PR #5, opened 2021-09-08
Escape < and > with &lt; and &gt;

westonpace:
Mkdocs uses python-markdown which allows inline HTML.  Some of the markdown (e.g. `TIMESTAMP<MILLISECOND>`) was being interpreted as HTML tags (and eventually rendering as `TIMESTAMP`).  The recommendation is to escape as `&lt;` and `&gt;`.  It appears that < and > in code tags is acceptable.

Another choice that might lead to more readable markdown is to just use ( and ) in the few places that < and > are currently used outside of code blocks.

One downside of either of the above approaches is that the simple logical types table is now quite wide (and requires scrolling on my viewport).

---
This change is [Reviewable](https://reviewable.io/reviews/substrait-io/substrait/5)

### DIFF
--- a/site/docs/relations/basics.md
+++ b/site/docs/relations/basics.md
@@ -14,7 +14,7 @@ Each relational operation is composed of several different properties. Common pr
 
 ## Relational Signatures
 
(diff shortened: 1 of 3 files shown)


jacques-n (approved):
Awesome, thanks @westonpace!

PR #19, opened 2021-09-09
Add Decimal(P, S) to types

emkornfield:
It is not clear if there is actually a lower bound restriction on scale.   It isn't mentioned in the arrow spec, I also couldn't find it in Trino, Iceberg or Spark docs.

---
This change is [Reviewable](https://reviewable.io/reviews/substrait-io/substrait/19)


jacques-n (approved):

jacques-n:
Thanks for the update @emkornfield !

PR #23, opened 2021-09-10
Remove physical types

jacques-n:
For people who need variations, they can use user defined extension types. (Simpler is better)

This is based on [this subthread](https://github.com/substrait-io/substrait/discussions/2#discussioncomment-1296988)

---
This change is [Reviewable](https://reviewable.io/reviews/substrait-io/substrait/23)


emkornfield:
I think the issue before is two items are conflated
1.  Memory Layout - how a particular type is laid out in memory.  I would guess plans can be agnostic to exact representation.
2. Logical Encodings - On the surface I think I agree with your assertion from the thread that having this modeled is likely important.

So I think this change is fine, but we might want to revisit point #2.

emkornfield (approved):

jacques-n:
Abandoning. Going with #25

PR #27, opened 2021-09-14
Type cleanups

jacques-n:
- remove timestamp_tz parameter
- add uuid (exists in iceberg & trino)
- add more range information for types.
- move simple, compound and variations to wip status

---
This change is [Reviewable](https://reviewable.io/reviews/substrait-io/substrait/27)


site/docs/types/simple_logical_types.md
```diff
@@ -6,22 +6,19 @@ To minimize type explosion, the project currently follows the guideline that a l
...
+| timestamp_tz | A timestamp with microseconds precision that is mapped to an instant in time. Similar to aware datetime in Python. Supports a range of [1000-01-01 00:00:00.000000..9999-12-31 23:59:59.999999] UTC. | timestamp&lt;micro;utc&gt; | timestamptz | - | timestamp(6) with time zone |
+| date            | A date. Range of [1000-01-01..9999-12-31]. | Date&lt;MILLISECOND&gt;      | date           | DateType      | Date                   |
+| time      | A time with microsecond precision since the beginning of any day. Range of [0..86,399,999,999] microseconds. | Time&lt;MICROSECOND;64&gt;   | time           | time(6)       | time(6)                |
+| interval_year   | Interval year to month. Supports a range of [0..120000] months. | INTERVAL&lt;YEAR_MONTH&gt;   | -              | -             | Interval year to month |
+| interval_day    | Interval day to second with microsecond precision. Supports a range of  [0..3,650,000] days and [0..86,399,999,999] microseconds. | INTERVAL&lt;DAY_TIME&gt;     | -              | -             | Interval day to second |
```
westonpace, line 23: Why can't intervals be negative?  I know Arrow & postgres support negative intervals.  I'm not sure about Trino but I would suspect it is allowed.

westonpace (commented):
Just the one question.  I don't have any strong feelings on UUID.  Arrow has stubbornly resisted them but semantically I'm not aware of any operations that can't be addressed by `fixed_size_binary(16)`.

site/docs/types/simple_logical_types.md
jacques-n: Good point on the interval ranges. I really struggled with those ranges. Frankly, I just really hate interval types :)

I've pushed an update. LMK if you think it makes sense.

jacques-n:
> Just the one question. I don't have any strong feelings on UUID. Arrow has stubbornly resisted them but semantically I'm not aware of any operations that can't be addressed by `fixed_size_binary(16)`.

Yeah, i'm in that camp. Unfortunately, it's first class in both Iceberg and Trino which made feel like we should include it to follow our own rules. It probably makes slightly more sense in the context of Substrait where we're more focused on logical concepts than physical (as opposed to Arrow).

rdblue:
@westonpace, there was a good discussion about UUID on the Iceberg list. For me, it came down to forcing users to choose between a bad representation for engines that is what they expect (string) and a bad representation for humans that has to be reformatted by hand. I'd rather keep it in the Iceberg spec and allow engines to fix this problem, like Trino does.

site/docs/types/simple_logical_types.md
```diff
@@ -6,22 +6,19 @@ To minimize type explosion, the project currently follows the guideline that a l
...
+| timestamp_tz | A timestamp with microseconds precision that is mapped to an instant in time. Similar to aware datetime in Python. Supports a range of [1000-01-01 00:00:00.000000..9999-12-31 23:59:59.999999] UTC. | timestamp&lt;micro;utc&gt; | timestamptz | - | timestamp(6) with time zone |
+| date            | A date. Range of [1000-01-01..9999-12-31]. | Date&lt;MILLISECOND&gt;      | date           | DateType      | Date                   |
+| time      | A time with microsecond precision since the beginning of any day. Range of [0..86,399,999,999] microseconds. | Time&lt;MICROSECOND;64&gt;   | time           | time(6)       | time(6)                |
+| interval_year   | Interval year to month. Supports a range of [-120,000..120,000] months. | INTERVAL&lt;YEAR_MONTH&gt;   | -              | -             | Interval year to month |
+| interval_day    | Interval day to second with microsecond precision. Supports a range of  [-3,650,000..3,650,000] days and [-86,399,999,999..86,399,999,999] microseconds. | INTERVAL&lt;DAY_TIME&gt;     | -              | -             | Interval day to second |
```
westonpace, line 23:
```suggestion
| interval_year   | Interval year to month. Supports a range of [-1,000..1,000] years and [-12,000..12,000] months. | INTERVAL&lt;YEAR_MONTH&gt;   | -              | -             | Interval year to month |
| interval_day    | Interval day to second with microsecond precision. Supports a range of  [-365,000..365,000] days and [-31,536,000,000,000,000..31,536,000,000,000,000] microseconds. | INTERVAL&lt;MONTH_DAY_NANO&gt;     | -              | -             | Interval day to second |
```
For intervals, the important thing is that `1 year 4 months` is not equal to `16 months` and `1 day 100 microseconds` is not equal to `8.64e+10 + 100 microseconds`.  The former is because 12 months is not 365 days and the latter is because that 1 day may or may not include a DST shift.

As a result, I think you need to specify the range for both units.  Otherwise, if I see `interval_year` with +/- 1000 years I might try and pack it into 16 bit data type using 4 bits for month and 12 for year.  This wouldn't interoperate with a system that supported a larger range of available months (e.g. Arrow)

It appears you have chosen 1000 years, which is fine, but very small.  Was it arbitrary or based on anything in particular?  The Arrow types can represent ~+/- 167 million years (signed 4 byte integer number of months).  Postgres supports +/- 178 million years.  If we're going to pick an arbitrary range I'd say we should pick `MAX(timestamp_tz) - MIN(timestamp_tz)` (~300,000 years I think) so that we can easily support time arithmetic (e.g. `timestamp_tz - timestamp_tz = interval`).

Regrettably, the Arrow DAY_TIME interval has millisecond resolution, so if we want to go for microseconds (and that makes good sense to me) I think we'd need to map to `MONTH_DAY_NANO` so I've updated that too.

My edit is assuming we want to stick to 1,000 years.

westonpace (changes requested):

site/docs/types/simple_logical_types.md
jacques-n: I was going for a 10,000 years range. Maybe I made a math mistake? I went for 10,000 years (0000-01-01 to 9999-12-31) as I believe that is the range sql defines for date times (and what I used for date and timestamp ranges above). Checking a few random systems, [datetime2 in sql server](https://docs.microsoft.com/en-us/sql/t-sql/data-types/datetime2-transact-sql?view=sql-server-ver15) and [datetime in mysql](https://dev.mysql.com/doc/refman/8.0/en/datetime.html) both match this range. I figured that the maximum amount an interval should be is basically the maximum distance between two timestamps (that was the rationale for the ranges).

>> the important thing is that 1 year 4 months is not equal to 16 months

Hmm. If I recall correctly most systems disagree with your statement and use these interchangeably because the conversion ratio is always guaranteed e.g. 1 year always equals 12 months. This is also why Arrow only uses a single 4 byte value for months for the INTERVAL_YEAR type. (I believe many systems just maintain a number of months for that measurement.)

For INTERVAL DAY TO SECOND, I agree that day and second are independent values (as opposed to year/month) as there is no guaranteed reliable conversion and thus must be stored as separate values. I think we should be find saying that 1,000,000 microseconds is equivalent to 1 second and only maintain microseconds.

westonpace:
> @westonpace, there was a good discussion about UUID on the Iceberg list. For me, it came down to forcing users to choose between a bad representation for engines that is what they expect (string) and a bad representation for humans that has to be reformatted by hand. I'd rather keep it in the Iceberg spec and allow engines to fix this problem, like Trino does.
@rdblue 

Thanks!  I'm happy leaving UUID in.  I think my original comment came off more negative than I intended.

site/docs/types/simple_logical_types.md
westonpace: > I was going for a 10,000 years range. Maybe I made a math mistake?

No, my math was wrong.  I agree with +/- 10k years given the range of dates defined.

> Hmm. If I recall correctly most systems disagree with your statement and use these interchangeably because the conversion ratio is always guaranteed e.g. 1 year always equals 12 months. This is also why Arrow only uses a single 4 byte value for months for the INTERVAL_YEAR type. (I believe many systems just maintain a number of months for that measurement.)

It appears you are correct.  Postgres defines 1 year as 12 months and if it needs to compare year/month intervals with day intervals it converts 1 year to 360 days.  The results making sorting intervals a little weird but I suppose that is unavoidable.  

I still think we should define the range for each time unit for each type.
I still think we need to use MONTH_DAY_NANO in Arrow

Do you think the microseconds range will need to be larger than 1 day to provide options for users that want to represent intervals of consistent length regardless of DST changes?  Unfortunately, 10k years is larger than what MONTH_DAY_NANO provides.

site/docs/types/simple_logical_types.md
```diff
@@ -6,22 +6,19 @@ To minimize type explosion, the project currently follows the guideline that a l
...
+| date            | A date. Range of [1000-01-01..9999-12-31]. | Date&lt;MILLISECOND&gt;      | date           | DateType      | Date                   |
+| time      | A time with microsecond precision since the beginning of any day. Range of [0..86,399,999,999] microseconds. | Time&lt;MICROSECOND;64&gt;   | time           | time(6)       | time(6)                |
+| interval_year   | Interval year to month. Supports a range of any combination of years and months that total less than or equal to 10,000 years. Each component can be specified as positive or negative. Examples maximums include: [10000y, -120000m, 1y119988m, 1000y108000m, etc]. Note that each component can never independentally specify more than 10,000 years, (even if the components have opposite signs e.g. -10000y200000m is **not** allowed). | INTERVAL&lt;YEAR_MONTH&gt;   | -              | -             | Interval year to month |
```
westonpace, line 22:
```suggestion
| interval_year   | Interval year to month. Supports a range of any combination of years and months that total less than or equal to 10,000 years. Each component can be specified as positive or negative. Examples minimums/maximums include: [10000y, -120000m, 1y119988m, 1000y108000m, etc]. Note that each component can never independently specify more than 10,000 years, (even if the components have opposite signs e.g. -10000y200000m is **not** allowed). | INTERVAL&lt;YEAR_MONTH&gt;   | -              | -             | Interval year to month |
```

westonpace (approved):
The latest commits have addressed my concerns about intervals.

PR #35, opened 2021-09-16
Added an example use case explaining consistent semantic operation on multiple query engines

westonpace:
This is related to the discussion #7


jacques-n (approved):
Nice addition!

PR #37, opened 2021-09-17
Minor protobuf fixes

jacques-n:
- Add boolean type and literals
- Add a typed null literal
- Fix map type definition
- Add java option to generate multiple files.

---
This change is [Reviewable](https://reviewable.io/reviews/substrait-io/substrait/37)

### DIFF
--- a/binary/type.proto
+++ b/binary/type.proto
@@ -1,10 +1,13 @@
 syntax = "proto3";
 package io.substrait;
 
+option java_multiple_files = true;
+
 message Type {
 
     oneof kind {
-        I8 i8 = 1;
+        Boolean bool = 1;
+        I8 i8 = 2;
         I16 i16 = 3;
         I32 i32 = 5;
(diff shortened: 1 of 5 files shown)


binary/type.proto
```diff
@@ -134,14 +140,9 @@ message Type {
...
-        repeated KeyValue key_values = 1;
-        Variation variation = 2;
-
-        message KeyValue {
-            Type key = 1;
-            Type value = 2;
-        }
+        Type key = 1;
+        Type value = 2;
+        Variation variation = 3;
```
rymurr, line 145: Not sure what these changes for the Map are? It looks like the KV is no longer repeated?

jacques-n: It was a mistake. This is for the type definition, not the literal value. The type definition just needs to list the two types.

jacques-n: To clarify, the original definition was a mistake, not the change.

PR #40, opened 2021-09-22
Propose update for simple output type expressions.

jacques-n:
Proposal for simple type expressions to support type parameterization for output type derivation.

---
This change is [Reviewable](https://reviewable.io/reviews/substrait-io/substrait/40)

### DIFF
--- a/binary/function.proto
+++ b/binary/function.proto
@@ -14,9 +14,27 @@ option java_multiple_files = true;
 message FunctionSignature {
 
     message FinalArgVariadic {
+        // the minimum number of arguments allowed for the list of final arguments (inclusive).
         int64 min_args = 1;
+
+        // the maxinum number of arguments allowed for the list of final arguments (exclusive)
         int64 max_args = 2;
+
+        // the type of parameterized type consistency
+        ParameterConsistency consistency = 3;
+
+        enum ParameterConsistency {
+            UNKNOWN = 0;
+
+            // All argument must be the same concrete type.
+            CONSISTENT = 1;
+
+            // Each argument can be any possible concrete type afforded by the bounds of any parameter defined in
+            // the arguments specification.
+            INCONSISTENT = 2;
+        }
     }
+
     message FinalArgNormal {}
 
     message Scalar {
@@ -96,10 +114,7 @@ message FunctionSignature {
     message Implementation {
 
         Type type = 1;
-        oneof implementation_type {
-            string uri = 2;
-            bytes binary = 3;
-        }
+        string uri = 2;
 
         enum Type {
             UNKNOWN = 0;
--- a/binary/function.proto
+++ b/binary/function.proto
@@ -3,6 +3,9 @@ syntax = "proto3";
 package io.substrait;
 
 import "type.proto";
+import "parameterized_types.proto";
+import "type_expressions.proto";
+import "extensions.proto";
 
 option java_multiple_files = true;
 
@@ -10,8 +13,14 @@ option java_multiple_files = true;
 // List of function signatures available.
 message FunctionSignature {
 
+    message FinalArgVariadic {
+        int64 min_args = 1;
+        int64 max_args = 2;
+    }
+    message FinalArgNormal {}
+
     message Scalar {
-        Identifier id = 1;
+        Extensions.FunctionId id = 1;
         repeated Argument arguments = 2;
         repeated string name = 3;
         Description description = 4;
@@ -19,13 +28,18 @@ message FunctionSignature {
         bool deterministic = 7;
         bool session_dependent = 8;
 
-        TypeExpression output_type = 9;
+        DerivationExpression output_type = 9;
 
-        repeated Implementation implementations = 10;
+        oneof final_variable_behavior {
+            FinalArgVariadic variadic = 10;
+            FinalArgNormal normal = 11;
+        }
+
+        repeated Implementation implementations = 12;
     }
 
     message Aggregate {
-        Identifier id = 1;
+        Extensions.FunctionId id = 1;
         repeated Argument arguments = 2;
         repeated string name = 3;
         Description description = 4;
@@ -33,17 +47,22 @@ message FunctionSignature {
         bool deterministic = 7;
         bool session_dependent = 8;
 
-        TypeExpression output_type = 9;
+        DerivationExpression output_type = 9;
 
-        bool ordered = 11;
+        oneof final_variable_behavior {
+            FinalArgVariadic variadic = 10;
+            FinalArgNormal normal = 11;
+        }
+
+        bool ordered = 14;
         uint64 max_set = 12;
         Type intermediate_type = 13;
 
-        repeated Implementation implementations = 14;
+        repeated Implementation implementations = 15;
     }
 
     message Window {
-        Identifier id = 1;
+        Extensions.FunctionId id = 1;
         repeated Argument arguments = 2;
         repeated string name = 3;
         Description description = 4;
@@ -51,11 +70,14 @@ message FunctionSignature {
         bool deterministic = 7;
         bool session_dependent = 8;
 
-        TypeExpression output_type = 9;
-
+        DerivationExpression intermediate_type = 9;
+        DerivationExpression output_type = 10;
+        oneof final_variable_behavior {
+            FinalArgVariadic variadic = 16;
+            FinalArgNormal normal = 17;
+        }
         bool ordered = 11;
         uint64 max_set = 12;
-        Type intermediate_type = 13;
         WindowType window_type = 14;
         repeated Implementation implementations = 15;
 
@@ -64,15 +86,6 @@ message FunctionSignature {
             STREAMING = 1;
             PARTITION = 2;
         }
-    }
-
-    message Identifier {
-        uint32 orgId = 1;
-        uint32 id = 2;
-    }
-
-    message CustomDerivationPolicy {
-        bytes policy = 11;
     }
 
     message Description {
@@ -96,19 +109,15 @@ message FunctionSignature {
     }
 
     message Argument {
+        ParameterizedType type = 1;
+        ArgumentKind kind = 2;
+        string name = 3;
 
-        message Plain {
-            Type type = 1;
-        }
-
-        message Constant {
-            Type type = 1;
-        }
-
-        message Variadic {
-            Type type = 1;
-            uint32 min = 2;
-            uint32 max = 3;
+        enum ArgumentKind {
+            UNKNOWN = 0;
+            VARIABLE_VALUE = 1;
+            CONSTANT_VALUE = 2;
+            TYPE = 3;
         }
 
     }
--- a/binary/function.proto
+++ b/binary/function.proto
@@ -19,12 +19,9 @@ message FunctionSignature {
         bool deterministic = 7;
         bool session_dependent = 8;
 
-        oneof derivation_type {
-            Type direct = 9;
-            CustomDerivationPolicy derivation_policy = 10;
-        }
+        TypeExpression output_type = 9;
 
-        repeated Implementation implementations = 11;
+        repeated Implementation implementations = 10;
     }
 
     message Aggregate {
@@ -36,10 +33,7 @@ message FunctionSignature {
         bool deterministic = 7;
         bool session_dependent = 8;
 
-        oneof derivation_type {
-            Type direct = 9;
-            CustomDerivationPolicy derivation_policy = 10;
-        }
+        TypeExpression output_type = 9;
 
         bool ordered = 11;
         uint64 max_set = 12;
@@ -57,10 +51,7 @@ message FunctionSignature {
         bool deterministic = 7;
         bool session_dependent = 8;
 
-        oneof derivation_type {
-            Type direct = 9;
-            CustomDerivationPolicy derivation_policy = 10;
-        }
+        TypeExpression output_type = 9;
 
         bool ordered = 11;
         uint64 max_set = 12;
@@ -121,4 +112,6 @@ message FunctionSignature {
         }
 
     }
+
 }
+
--- a/binary/parameterized_types.proto
+++ b/binary/parameterized_types.proto
@@ -1,0 +1,117 @@
+syntax = "proto3";
+package io.substrait;
+
+import "type.proto";
+import "extensions.proto";
+
+option java_multiple_files = true;
+
+message ParameterizedType {
+
+    oneof kind {
+        Type.Boolean bool = 1;
+        Type.I8 i8 = 2;
+        Type.I16 i16 = 3;
+        Type.I32 i32 = 5;
+        Type.I64 i64 = 7;
+        Type.FP32 fp32 = 10;
+        Type.FP64 fp64 = 11;
+        Type.String string = 12;
+        Type.Binary binary = 13;
+        Type.Timestamp timestamp = 14;
+        Type.Date date = 16;
+        Type.Time time = 17;
+        Type.IntervalYear interval_year = 19;
+        Type.IntervalDay interval_day = 20;
+        Type.TimestampTZ timestamp_tz = 29;
+        Type.UUID uuid = 32;
+
+        ParameterizedFixedChar fixed_char = 21;
+        ParameterizedVarChar varchar = 22;
+        ParameterizedFixedBinary fixed_binary = 23;
+        ParameterizedDecimal decimal = 24;
+
+        ParameterizedStruct struct = 25;
+        ParameterizedNamedStruct named_struct = 26;
+        ParameterizedList list = 27;
+        ParameterizedMap map = 28;
+
+        Extensions.TypeId user_defined = 31;
+
+        TypeParameter type_parameter = 33;
+
+    }
+
+    message TypeParameter {
+        string name = 1;
+        repeated ParameterizedType bounds = 2;
+    }
+
+    message IntegerParameter {
+        string name = 1;
+        NullableInteger range_start_inclusive = 2;
+        NullableInteger range_end_exclusive = 3;
+    }
+    
+    message NullableInteger {
+        int64 value = 1;
+    }
+    
+    message ParameterizedFixedChar {
+        IntegerOption length = 1;
+        Type.Variation variation = 2;
+        Type.Nullability nullability = 3;
+    }
+
+    message ParameterizedVarChar {
+        IntegerOption length = 1;
+        Type.Variation variation = 2;
+        Type.Nullability nullability = 3;
+    }
+
+    message ParameterizedFixedBinary {
+        IntegerOption length = 1;
+        Type.Variation variation = 2;
+        Type.Nullability nullability = 3;
+    }
+
+    message ParameterizedDecimal {
+        IntegerOption scale = 1;
+        IntegerOption precision = 2;
+        Type.Variation variation = 3;
+        Type.Nullability nullability = 4;
+    }
+
+    message ParameterizedStruct {
+        repeated ParameterizedType types = 1;
+        Type.Variation variation = 2;
+        Type.Nullability nullability = 3;
+    }
+
+    message ParameterizedNamedStruct {
+        // list of names in dfs order
+        repeated string names = 1;
+        ParameterizedStruct struct = 2;
+    }
+
+    message ParameterizedList {
+        ParameterizedType type = 1;
+        Type.Variation variation = 2;
+        Type.Nullability nullability = 3;
+    }
+
+    message ParameterizedMap {
+        ParameterizedType key = 1;
+        ParameterizedType value = 2;
+        Type.Variation variation = 3;
+        Type.Nullability nullability = 4;
+    }
+
+    message IntegerOption {
+        oneof integer_type {
+            int32 literal = 1;
+            IntegerParameter parameter = 2;
+        }
+    }
+}
+
--- a/binary/type.proto
+++ b/binary/type.proto
@@ -1,6 +1,8 @@
 syntax = "proto3";
 package io.substrait;
 
+import "extensions.proto";
+
 option java_multiple_files = true;
 
 message Type {
@@ -20,140 +22,155 @@ message Type {
         Time time = 17;
         IntervalYear interval_year = 19;
         IntervalDay interval_day = 20;
+        TimestampTZ timestamp_tz = 29;
+        UUID uuid = 32;
+
         FixedChar fixed_char = 21;
         VarChar varchar = 22;
         FixedBinary fixed_binary = 23;
         Decimal decimal = 24;
+
         Struct struct = 25;
         NamedStruct named_struct = 26;
         List list = 27;
         Map map = 28;
-        TimestampTZ timestamp_tz = 29;
-        UserDefined user_defined = 31;
-        UUID uuid = 32;
 
-        // A special type that can only be used for parameterized function arguments.
-        // This allows a user to express things like List<T>
-        Parameter parameter = 33;
-
-        // A special type that can only be used in function return type declarations.
-        TypeExpression expression = 34;
+        Extensions.TypeId user_defined = 31;
     }
 
-    message Parameter {
-        string name = 1;
+    enum Nullability {
+        NULLABLE = 0;
+        REQUIRED = 1;
     }
 
     message Boolean {
         Variation variation = 1;
+        Nullability nullability = 2;
     }
     message I8 {
         Variation variation = 1;
+        Nullability nullability = 2;
     }
 
     message I16 {
         Variation variation = 1;
+        Nullability nullability = 2;
     }
 
     message I32 {
         Variation variation = 1;
+        Nullability nullability = 2;
     }
 
     message I64 {
         Variation variation = 1;
+        Nullability nullability = 2;
     }
 
     message FP32 {
         Variation variation = 1;
+        Nullability nullability = 2;
     }
 
     message FP64 {
         Variation variation = 1;
+        Nullability nullability = 2;
     }
 
     message String {
         Variation variation = 1;
+        Nullability nullability = 2;
     }
 
     message Binary {
         Variation variation = 1;
+        Nullability nullability = 2;
     }
 
     message Timestamp {
         Variation variation = 1;
+        Nullability nullability = 2;
     }
 
     message Date {
         Variation variation = 1;
+        Nullability nullability = 2;
     }
 
     message Time {
         Variation variation = 1;
+        Nullability nullability = 2;
     }
 
     message TimestampTZ {
         Variation variation = 1;
+        Nullability nullability = 2;
     }
 
     message IntervalYear {
         Variation variation = 1;
+        Nullability nullability = 2;
     }
 
     message IntervalDay {
         Variation variation = 1;
+        Nullability nullability = 2;
     }
 
     message UUID {
         Variation variation = 1;
+        Nullability nullability = 2;
     }
 
     // Start compound types.
     message FixedChar {
-        IntegerParameter length = 1;
+        int32 length = 1;
         Variation variation = 2;
+        Nullability nullability = 3;
     }
 
     message VarChar {
-        IntegerParameter length = 1;
+        int32 length = 1;
         Variation variation = 2;
+        Nullability nullability = 3;
     }
 
     message FixedBinary {
-        IntegerParameter length = 1;
+        int32 length = 1;
         Variation variation = 2;
+        Nullability nullability = 3;
     }
 
     message Decimal {
-        IntegerParameter scale = 1;
-        IntegerParameter precision = 2;
+        int32 scale = 1;
+        int32 precision = 2;
         Variation variation = 3;
+        Nullability nullability = 4;
     }
 
     message Struct {
         repeated Type types = 1;
         Variation variation = 2;
+        Nullability nullability = 3;
     }
 
     message NamedStruct {
-
-        repeated Pair pairs = 1;
-        Variation variation = 2;
-
-        message Pair {
-            string name = 1;
-            Type type = 2;
-        }
+        // list of names in dfs order
+        repeated string names = 1;
+        Struct struct = 2;
     }
 
     message List {
         Type type = 1;
         Variation variation = 2;
+        Nullability nullability = 3;
     }
 
     message Map {
         Type key = 1;
         Type value = 2;
         Variation variation = 3;
+        Nullability nullability = 4;
     }
 
     message Variation {
@@ -161,70 +178,5 @@ message Type {
         string name = 2;
     }
 
-    message UserDefined {
-        int32 organization = 1;
-        string name = 2;
-    }
-
-    message IntegerParameter {
-        oneof integer_type {
-            int32 literal = 1;
-            String parameter_name = 2;
-        }
-    }
 }
 
-
-message TypeExpression {
-
-    oneof expression_type {
-        int32 integer_literal = 1;
-        UnaryOp unary_op = 2;
-        BinaryOp binary_op = 3;
-        Type.Parameter parameter = 4;
-        IfElse if_else = 5;
-        NotBindable not_bindable = 6;
-        Type type = 7;
-    }
-
-    message IfElse {
-        TypeExpression if_condition = 1;
-        TypeExpression if_return = 2;
-        TypeExpression else_return = 3;
-    }
-
-    message NotBindable {}
-
-    message UnaryOp {
-        OpType op_type = 1;
-        TypeExpression arg = 2;
-
-        enum OpType {
-            UNKNOWN = 0;
-            BOOLEAN_NOT = 1;
-        }
-    }
-
-    message BinaryOp {
-
-        OpType op_type = 1;
-        TypeExpression arg1 = 2;
-        TypeExpression arg2 = 3;
-
-        enum OpType {
-            UNKNOWN = 0;
-            PLUS = 1;
-            MINUS = 2;
-            MULTIPLY = 3;
-            DIVIDE = 4;
-            MIN = 5;
-            MAX = 6;
-            GREATER_THAN = 7;
-            LESS_THAN = 8;
-            AND = 9;
-            OR = 10;
-            EQUAL = 11;
-        }
-    }
-
-}
--- a/binary/type.proto
+++ b/binary/type.proto
@@ -31,6 +31,17 @@ message Type {
         TimestampTZ timestamp_tz = 29;
         UserDefined user_defined = 31;
         UUID uuid = 32;
+
+        // A special type that can only be used for parameterized function arguments.
+        // This allows a user to express things like List<T>
+        Parameter parameter = 33;
+
+        // A special type that can only be used in function return type declarations.
+        TypeExpression expression = 34;
+    }
+
+    message Parameter {
+        string name = 1;
     }
 
     message Boolean {
@@ -98,23 +109,23 @@ message Type {
 
     // Start compound types.
     message FixedChar {
-        int32 length = 1;
+        IntegerParameter length = 1;
         Variation variation = 2;
     }
 
     message VarChar {
-        int32 length = 1;
+        IntegerParameter length = 1;
         Variation variation = 2;
     }
 
     message FixedBinary {
-        int32 length = 1;
+        IntegerParameter length = 1;
         Variation variation = 2;
     }
 
     message Decimal {
-        int32 scale = 1;
-        int32 precision = 2;
+        IntegerParameter scale = 1;
+        IntegerParameter precision = 2;
         Variation variation = 3;
     }
 
@@ -154,4 +165,66 @@ message Type {
         int32 organization = 1;
         string name = 2;
     }
+
+    message IntegerParameter {
+        oneof integer_type {
+            int32 literal = 1;
+            String parameter_name = 2;
+        }
+    }
+}
+
+
+message TypeExpression {
+
+    oneof expression_type {
+        int32 integer_literal = 1;
+        UnaryOp unary_op = 2;
+        BinaryOp binary_op = 3;
+        Type.Parameter parameter = 4;
+        IfElse if_else = 5;
+        NotBindable not_bindable = 6;
+        Type type = 7;
+    }
+
+    message IfElse {
+        TypeExpression if_condition = 1;
+        TypeExpression if_return = 2;
+        TypeExpression else_return = 3;
+    }
+
+    message NotBindable {}
+
+    message UnaryOp {
+        OpType op_type = 1;
+        TypeExpression arg = 2;
+
+        enum OpType {
+            UNKNOWN = 0;
+            BOOLEAN_NOT = 1;
+        }
+    }
+
+    message BinaryOp {
+
+        OpType op_type = 1;
+        TypeExpression arg1 = 2;
+        TypeExpression arg2 = 3;
+
+        enum OpType {
+            UNKNOWN = 0;
+            PLUS = 1;
+            MINUS = 2;
+            MULTIPLY = 3;
+            DIVIDE = 4;
+            MIN = 5;
+            MAX = 6;
+            GREATER_THAN = 7;
+            LESS_THAN = 8;
+            AND = 9;
+            OR = 10;
+            EQUAL = 11;
+        }
+    }
+
 }
--- a/site/docs/expressions/scalar_functions.md
+++ b/site/docs/expressions/scalar_functions.md
@@ -65,7 +65,7 @@ Any function can declare a return type expression. A return type expression uses
 * Boolean: True and False
 * Type: A Substrait type (with possibly additional embedded expressions)
 
-These types are evaluated using a small set of operations to to support common scenarios. Since it is possible that output derivation ultimately determines that binding is not allowed, a `not_bindable` operation is provided that states that the particular binding is now sllowed. List of valid operations:
+These types are evaluated using a small set of operations to to support common scenarios. Since it is possible that output derivation ultimately determines that binding is not allowed, a `not_bindable` operation is provided that states that the particular binding is not allowed. List of valid operations:
 
 ```
 Math: +, -, *, /, min, max
@@ -100,7 +100,7 @@ For reference, here are are some common output type derivations and how they can
 
 | Operation                                     | Definition                                                   |
 | --------------------------------------------- | ------------------------------------------------------------ |
-| Add item to list                              | `add(<List<T>, T>) => List<T>)`                              |
+| Add item to list                              | `add(<List<T>, T>) => List<T>`                               |
 | Decimal Division                              | `divide(Decimal(P1,S1), Decimal(P2,S2)) => Decimal(P1 -S1 + S2 + MAX(6, S1 + P2 + 1), MAX(6, S1 + P2 + 1))` |
 | Do regex on only string maps to return values | `extract_values(Map<K,V>) => if(K==STRING) THEN LIST<V> ELSE !bindable` |
 
--- a/site/docs/expressions/scalar_functions.md
+++ b/site/docs/expressions/scalar_functions.md
@@ -65,7 +65,7 @@ Any function can declare a return type expression. A return type expression uses
 * Boolean: True and False
 * Type: A Substrait type (with possibly additional embedded expressions)
 
-These types are evaluated using a small set of operations to to support common scenarios. Since it is possible that output derivation ultimately determines that binding is not allowed, a `not_bindable` operation is provided that states that the particular binding is not allowed. List of valid operations:
+These types are evaluated using a small set of operations to support common scenarios. Since it is possible that output derivation ultimately determines that binding is not allowed, a `not_bindable` operation is provided that states that the particular binding is not allowed. List of valid operations:
 
 ```
 Math: +, -, *, /, min, max
--- a/site/docs/expressions/scalar_functions.md
+++ b/site/docs/expressions/scalar_functions.md
@@ -47,11 +47,10 @@ For scalar functions (one value produced for each record), a function binding de
 
 ### Input Argument Types
 
-Input arguments can be declared in one of three ways: complete, wildcarded, parameterized.
+Input arguments can be declared in one of two ways: materialized or parameterized.
 
-* Complete Type: Complete types are either simple types or compound types with all parameters fully defined. Examples include i32, fp32, VARCHAR(20), List&lt;fp32&gt;, etc.
-* Wildcard Types: Compound types can be wildcarded meaning that any valid value for the parameter can be accepted. Types are wildcarded by failing to express a particular value for the type. When serialized, this will typically mean using a null pointer or sentinel value. Examples would be  VARCHAR(*), List&lt;\*&gt;, DECIMAL(\*,10).
-* Parameterized types: Types can be parameterized such that the parameter can be used in a output derivation. Example parameterized types would be VARCHAR(T), List&lt;E&gt;, MAP&lt;K, fp32&gt;, etc. Parameters are simple names. Any parameters defined in an argument must be referenced at least once in an output expression. Parameterized argument types can only be used when a function is using expression based output type derivation.
+* Materialized Type: Materialized types are either simple types or compound types with all parameters fully defined (without any parameters). Examples include i32, fp32, VARCHAR(20), List&lt;fp32&gt;, etc.
+* Parameterized types: Types can be parameterized such that the parameter can be used in a output derivation e.g. `f(K) => K` or to guarantee consistency of type between input arguments `f(K,K) => boolean`. Example parameterized types would be VARCHAR(T), List&lt;E&gt;, MAP&lt;K, fp32&gt;, etc. A parameters is named with a simple UTF8 character or string.
 
 ### Direct Return Types
 
--- a/site/docs/expressions/scalar_functions.md
+++ b/site/docs/expressions/scalar_functions.md
@@ -1,16 +1,16 @@
 # Scalar Functions
 
-Scalar functions are a function that takes in values from a single record and produces an output value. To clearly specify the definition of functions, Substrait declares a extensible specification plus binding approach to function resolution. A scalar function signature includes the following properties:
+A function is a scalar function if that function takes in values from a single record and produces an output value. To clearly specify the definition of functions, Substrait declares a extensible specification plus binding approach to function resolution. A scalar function signature includes the following properties:
 
 | Property               | Description                                                  | Required                            |
 | ---------------------- | ------------------------------------------------------------ | ----------------------------------- |
 | Name                   | One or more user friendly utf8 strings that are used to reference this function in languages | At least one value is required.     |
-| List of arguments      | Argument properties are defined below. Arguments can be fully defined, parameterized or wildcarded. See further details below. | Optional, defaults to niladic.      |
+| List of arguments      | Argument properties are defined below. Arguments can be fully defined or calculated with a type expression. See further details below. | Optional, defaults to niladic.      |
 | Deterministic          | Whether this function is expected to reproduce the same output when it is invoked multiple times with the same input. This informs a plan consumer on whether it can constant reduce the defined function. An example would be a random() function, which is typically expected to be evaluated repeatedly despite having the same set of inputs. | Optional, defaults to true.         |
 | Session Dependent      | Whether this function is influenced by the session context it is invoked within. For example, a function may be influenced by a user who is invoking the function, the time zone of a session, or some other non-obvious parameter. This can inform caching systems on whether a particular function is cacheable. | Optional, defaults to false.        |
-| Variadic Behavior      | Whether the last argument of the function is variadic. Options include: [Single argument, M..N]. M must be less than or equal to N. N can be defined or left as "unlimited". | Optional, defaults to single value. |
+| Variadic Behavior      | Whether the last argument of the function is variadic or a single argument.  If variadic the argument can optionally have a lower bound (minimum number of instances) and upper bound (maximum number of instances) | Optional, defaults to single value. |
 | Description            | Additional description of function for implementers or users. Should be written human readable to allow exposure to end users. Presented as a map with language => description mappings. E.g. `{ "en": "This adds two numbers together.", "fr": "cela ajoute deux nombres"}`. | Optional                            |
-| Return Type Expression | The output type is expected to be a physical type of the expression. Return types can be one of a simple, fully-defined compound type or a type expression. See below for more on type expressions. | Required                            |
+| Return Value | The output type of the expression.  Return types can be expressed as a fully-defined type or a type expression. See below for more on type expressions. | Required                            |
 | Implementation Map     | A map of implementation locations for one or more implementations of the given function. Each key is a function implementation type. Implementation types include examples such as: AthenaArrowLambda, TrinoV361Jar, ArrowCppKernelEnum, GandivaEnum, LinkedIn Transport Jar, etc. [Definition TBD]. Implementation type has one or more properties associated with retrieval of that implementation. | Optional                            |
 
 
@@ -19,15 +19,15 @@ Scalar functions are a function that takes in values from a single record and pr
 
 There are two main types of arguments: Value Arguments and Type Arguments. 
 
-* Value Arguments: Arguments that where the function will refer to data value, could be a constant (literal-based expression) or variable (reference-based expression). This is the most common type of argument. Value arguments are not available in output derivation. Value arguments can be declared in one of two ways: concrete or parameterized. Concrete Types are either simple types or compound types with all parameters fully defined (without any parameters). Examples include i32, fp32, VARCHAR(20), List&lt;fp32&gt;, etc. Parameterized types are discussed further below.
-* Type Arguments: Type arguments are used to inform the evaluation and/or type derivation of the function. For example, you might have a function which is `truncate(<type> DECIMAL(P0,S0), <value> DECIMAL(P1, S1), <value> i32)`. This function declares two value arguments (arg1 and arg2) and type argument. The type argument's value can be used in output type determination. (The types of the values arguments can also be used for output type derivation.)
+* Value Arguments: An argument that refers to a data value.  This could be a constant (a literal expression defined in the plan) or variable (A reference expression that references data being processed by the plan). This is the most common type of argument. Value arguments are not available in output derivation. Value arguments can be declared in one of two ways: concrete or parameterized. Concrete Types are either simple types or compound types with all parameters fully defined (without referencing any type arguments). Examples include i32, fp32, VARCHAR(20), List&lt;fp32&gt;, etc. Parameterized types are discussed further below.
+* Type Arguments: Type arguments are used to inform the evaluation and/or type derivation of the function. For example, you might have a function which is `truncate(<type> DECIMAL(P0,S0), <value> DECIMAL(P1, S1), <value> i32)`. This function declares two value arguments and a type argument. The type argument's value can be used to determine the output type (The types of the values arguments can also be used for output type derivation.)
 
 #### Value Argument Properties
 
 | Property | Description                                                  | Required                                                   |
 | -------- | ------------------------------------------------------------ | ---------------------------------------------------------- |
-| Name     | A human name for this argument to help clarify use.          | Optional, defaults to a name based on position (e.g. arg0) |
-| Type     | A fully concrete or parameterized (bound or not).            | Required                                                   |
+| Name     | A human readable name for this argument to help clarify use.          | Optional, defaults to a name based on position (e.g. arg0) |
+| Type     | A fully defined type or a type expression.            | Required                                                   |
 | Constant | Whether this argument is required to be a constant for invocation. For example, in some system a regular expression pattern would only be accepted as a literal and not a column value reference. | Optional, defaults to false                                |
 
 #### Type Argument Properties
@@ -35,17 +35,17 @@ There are two main types of arguments: Value Arguments and Type Arguments.
 | Property | Description                                                  | Required                                                   |
 | -------- | ------------------------------------------------------------ | ---------------------------------------------------------- |
 | Type     | A partially or completely parameterized type. E.g. List&lt;K&gt; or K | Required                                                   |
-| Name     | A human name for this argument to help clarify use.          | Optional, defaults to a name based on position (e.g. arg0) |
+| Name     | A human readable name for this argument to help clarify use.          | Optional, defaults to a name based on position (e.g. arg0) |
 
 ### Parameterized Types
 
-Types are parametered by two types of values: by inner types (e.g. List&lt;K&gt;) and numeric values (e.g. VARCHAR(P,S)). Parameter names are simple strings (frequently a single character). There are two types of parameters: integer parameters and type parameters. 
+Types are parameterized by two types of values: by inner types (e.g. List&lt;K&gt;) and numeric values (e.g. VARCHAR(P,S)). Parameter names are simple strings (frequently a single character). There are two types of parameters: integer parameters and type parameters. 
 
-When the same parameter name is used multiple times in a function definition, the function can only bind if the exact same value is used for all parameters of that name. For example, if one had a function with a signature of `fn(VARCHAR(N), VARCHAR(N)), the function would be only be usable if both VARCHAR types had the same length value N. This necessitates that all instances of the same parameter name must be of the same parameter type (all instances are a type parameter or always intances are aninteger parameter).
+When the same parameter name is used multiple times in a function definition, the function can only bind if the exact same value is used for all parameters of that name. For example, if one had a function with a signature of `fn(VARCHAR(N), VARCHAR(N)), the function would be only be usable if both VARCHAR types had the same length value N. This necessitates that all instances of the same parameter name must be of the same parameter type (all instances are a type parameter or all instances are an integer parameter).
 
 #### Integer Parameter Bounds
 
-A type which declares one or more integer parameters can also declare a range of legal values for a function. For example, one could declare a function `fn(VARCHAR(N)) where N IN [0..20]` to describe a function which only takes in a varchar types that are less than or equal to 20 characters in length.
+A type which declares one or more integer parameters can also declare a range of legal values for a function. For example, one could declare a function `fn(VARCHAR(N)) where N IN [0..20]` to describe a function which only takes in varchar types that are less than or equal to 20 characters in length.
 
 #### Type Parameter Bounds
 
@@ -100,7 +100,7 @@ Fully defined with argument types:
 * `if(boolean a) then (integer) else (integer)`
 * `if(boolean a) then (type) else (type)`
 
-#### Example Type Expressions Uses
+#### Example Type Expressions
 
 For reference, here are are some common output type derivations and how they can be expressed with a return type expression:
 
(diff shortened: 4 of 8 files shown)


site/docs/expressions/scalar_functions.md
```diff
@@ -43,10 +43,68 @@ For scalar functions (one value produced for each record), a function binding de
...
 
 
+## Argument Types & Output Type Derivation
+
+### Input Argument Types
+
+Input arguments can be declared in one of three ways: complete, wildcarded, parameterized.
+
+* Complete Type: Complete types are either simple types or compound types with all parameters fully defined. Examples include i32, fp32, VARCHAR(20), List&lt;fp32&gt;, etc.
+* Wildcard Types: Compound types can be wildcarded meaning that any valid value for the parameter can be accepted. Types are wildcarded by failing to express a particular value for the type. When serialized, this will typically mean using a null pointer or sentinel value. Examples would be  VARCHAR(*), List&lt;\*&gt;, DECIMAL(\*,10).
```
cpcloud, line 53: What is the difference between a parameterized type and a wildcard type in any of your examples? Is there a use case where a parameterized type doesn't cover the behavior of wildcards?

How is `List<*>` distinct from `List<T>`, for example?

Is the wildcard terminology standard outside of the database community? I haven't encountered this language before, but maybe I just haven't done enough reading.

cpcloud: I would expect for example a function like `list_length`, that operates on every cell of a list-of-any-type column to have a signature of `list_length<K>(list: List<K>) -> u32`.

site/docs/expressions/scalar_functions.md
```diff
@@ -43,10 +43,68 @@ For scalar functions (one value produced for each record), a function binding de
...
+* Parameterized types: Types can be parameterized such that the parameter can be used in a output derivation. Example parameterized types would be VARCHAR(T), List&lt;E&gt;, MAP&lt;K, fp32&gt;, etc. Parameters are simple names. Any parameters defined in an argument must be referenced at least once in an output expression. Parameterized argument types can only be used when a function is using expression based output type derivation.
```
cpcloud, line 54: > Any parameters defined in an argument must be referenced at least once in an output expression.

What is the rationale here?

site/docs/expressions/scalar_functions.md
```diff
@@ -43,10 +43,68 @@ For scalar functions (one value produced for each record), a function binding de
...
+
+### Direct Return Types
+
+A direct return type is one that is fully known at function definition type. Example simple direct return types would be things such as i32, fp32. For compound types, a direct return type must be fully declared. Example fully defined types: VARCHAR(20), DECIMAL(25,5)
```
cpcloud, line 58: > A direct return type is one that is fully known at function definition type.

Should the last `type` be `time`?

Also, I find this language confusing. A function definition with a parameterized type is fully known at function definition time. It won't necessarily have been monomorphized at definition time, is that what you're referring to here? Maybe I'm being a bit pedantic but my thinking is that we shouldn't invent new words for things :)

site/docs/expressions/scalar_functions.md
```diff
@@ -43,10 +43,68 @@ For scalar functions (one value produced for each record), a function binding de
...
+
+### Return Type Expressions
+
+Any function can declare a return type expression. A return type expression uses a simplified set of expressions to describe how the return type should be returned. For example, a return expression could be as simple as the return of parameter declared in the arguments. For example `f(List<K>) =>K ` or can be a simple mathematical or conditional expression such as `add(decimal(a,b), decimal(c,d)) => decimal(a+c, b+d)`. For the simple expression language, there is a very narrow set of types:
+
+* Integer: 64 bit signed integer (can be a literal or a parameter value)
+* Boolean: True and False
+* Type: A Substrait type (with possibly additional embedded expressions)
+
+These types are evaluated using a small set of operations to to support common scenarios. Since it is possible that output derivation ultimately determines that binding is not allowed, a `not_bindable` operation is provided that states that the particular binding is now sllowed. List of valid operations:
```
cpcloud, line 68: > These types are evaluated using a small set of operations to to support common scenarios.

Double `to to`.

site/docs/expressions/scalar_functions.md
```diff
@@ -43,10 +43,68 @@ For scalar functions (one value produced for each record), a function binding de
...
+* `if(boolean a) then (type) else (type)`
+* `not_bindable()`
+
+#### Example Type Expressions Uses
+
+For reference, here are are some common output type derivations and how they can be expressed with a return type expression:
+
+| Operation                                     | Definition                                                   |
+| --------------------------------------------- | ------------------------------------------------------------ |
+| Add item to list                              | `add(<List<T>, T>) => List<T>)`                              |
```
cpcloud, line 103: You have an unmatched paren at the end here.

site/docs/expressions/scalar_functions.md
```diff
@@ -43,10 +43,68 @@ For scalar functions (one value produced for each record), a function binding de
...
+* Type: A Substrait type (with possibly additional embedded expressions)
+
+These types are evaluated using a small set of operations to to support common scenarios. Since it is possible that output derivation ultimately determines that binding is not allowed, a `not_bindable` operation is provided that states that the particular binding is now sllowed. List of valid operations:
+
+```
+Math: +, -, *, /, min, max
+Boolean: &&, ||, !, <, >, ==
+Input Parameters: type parameter, literal parameter
+Literals: integer, type
+Special: !bindable
```
cpcloud, line 75: Should `!bindable` have some way of customizing the error message? Maybe not, just wondering.

site/docs/expressions/scalar_functions.md
cpcloud: > binding is now sllowed

Should this be `binding is not allowed`?

site/docs/expressions/scalar_functions.md
```diff
@@ -43,10 +43,68 @@ For scalar functions (one value produced for each record), a function binding de
...
+
+#### Example Type Expressions Uses
+
+For reference, here are are some common output type derivations and how they can be expressed with a return type expression:
+
+| Operation                                     | Definition                                                   |
+| --------------------------------------------- | ------------------------------------------------------------ |
+| Add item to list                              | `add(<List<T>, T>) => List<T>)`                              |
+| Decimal Division                              | `divide(Decimal(P1,S1), Decimal(P2,S2)) => Decimal(P1 -S1 + S2 + MAX(6, S1 + P2 + 1), MAX(6, S1 + P2 + 1))` |
+| Do regex on only string maps to return values | `extract_values(Map<K,V>) => if(K==STRING) THEN LIST<V> ELSE !bindable` |
```
cpcloud, line 105: Why wouldn't this just be written as `extract_values(Map<String, V>)`?

site/docs/expressions/scalar_functions.md
jacques-n: I distinguished them here to specifically state that a defined parameter name must be used in an expression to be as strict/clear as possible.

site/docs/expressions/scalar_functions.md
jacques-n: Per my comment above. A wildcard could also be called an unnamed parameter. A parameter is something that must be used. It is only to be as intentional as possible.

site/docs/expressions/scalar_functions.md
jacques-n: Feel free to propose alternative language here.

site/docs/expressions/scalar_functions.md
jacques-n: Fair point, my original plan was to do multiple if statements. E.g. k==string or k==varchar. I just got lazy.

site/docs/expressions/scalar_functions.md
jacques-n: I wasn't thinking so. I wouldn't expect this information to be propogated to a user. I could imagine a situation where a user writes a query and 50 functions are initially evaluated but then found unbindable. Then on the 51, a binding works and is used. The fact that things can't bind feels like an internal plan producer concern and they should decide how to expose to users.

site/docs/expressions/scalar_functions.md
cpcloud: Hm, I guess I see this as analogous to an unused variable. It seems like introducing a new concept only to prevent people from not using type variables is pretty heavy weight since now someone has to build support for both, as well as the enforcement mechanism, when you can unify the notions of wildard and parameterized types under a single concept.

site/docs/expressions/scalar_functions.md
jacques-n: Updated

cpcloud: Maybe I missed it but is the plan to keep this as simple as possible? So, for example, if someone wanted to write a derivation function that corresponded to "any non-compound type", they're in for a giant `or` expression.

site/docs/expressions/scalar_functions.md
jacques-n: both fixed

site/docs/expressions/scalar_functions.md
jacques-n: Part of it was thinking about the serialization formats, which I guess we can ignore. When serializing something it seems weird to require a named unused parameter. If you say it is required at the spec level (e.g. you don't allow List<>, only List<K>) it seems like it would be confusing if the serialization then forced this on you. I suppose serialization could say either representation is fine and just leave it to implementors to have to tease out the distinction. In general, I'm trying to be more formal about these things as much as possible (and why I work on the protobuf sketch as I think about the problems).

I'm okay switching it but that was the reasoning :)

site/docs/expressions/scalar_functions.md
jacques-n: fixed

site/docs/expressions/scalar_functions.md
jacques-n: My goal is that the language should be as simple as reasonable. I'm open to adding a in-list type idea but I'm not entirely convinced it is necessary. All non-compound types doesn't actually seem like a real thing. Numeric types probably seems more real.

cpcloud: > All non-compound types doesn't actually seem like a real thing.

Well, I did base that example on extracting values from a map, and I think you'd probably at least want any non-compound type to be able to be part of a a map. :shrug: I don't feel too strongly about it, keeping it simple to start seems like a good idea.

site/docs/expressions/scalar_functions.md
westonpace: I agree with cpcloud here.  What about the case where it is unused in the output derivation, can be any type, but is not valid as a wildcard.  For example:

`EQUAL(FIXEDBINARY<*>, FIXEDBINARY<*>) -> boolean` is not valid because fixed binaries are only comparable if they have the same width.  However, the type parameters are not used in the output derivation.

site/docs/expressions/scalar_functions.md
jacques-n: Hmm... not super important at this moment but I wonder if we're using compound types to mean different things. I meant compound types as done in substrait where it is any type that is parameterized (including things like decimal, varchar, etc). Using that definition, I think most people would probably want to treat string and varchar much the the same and could imagine both being used as a key to map.

I better understand your comment about what maps would be restricted to if you limit "compound" to be types that contain other types such as struct, map and list. I've generally been referring to that set as "complex" as opposed to "compound". (Happy to revise terminology if something else is better for people.)

site/docs/expressions/scalar_functions.md
jacques-n: Good point on the equal. I've updated to remove the distinction between wildcard and parameterized.

site/docs/expressions/scalar_functions.md
jacques-n: I've updated to remove wildcards.

site/docs/expressions/scalar_functions.md
cpcloud: Ah yes! Sorry. I was definitely referring to what you've called _complex_ types, not _compound_ types.

Does that change your response at all? Fine if not, just wondering if you think "all non-complex types" is a predicate that will be commonly desired and used.

jacques-n: My general inclination would be to expose a generic OR_EQUALITY_LIST(value, type, type,...,type) operation as opposed to declare specific classes. Specific classes have the problem that if you then declared u8, for example, it wouldn't be in the "numeric" group. I'm torn on whether OR is enough or OR_EQUALITY_LIST is necessary. Love to hear your gut and others.

cpcloud: Before I answer that, are we expecting that every consumer/producer provides an output derivation where they are required, or will there be a default implementation? I could imagine not needing to worry about `OR` vs `IN` vs `OR_EQUALITY_LIST` (`OR` would probably suffice) if we have the right defaults to avoid having to repeat the same giant `OR` expression in every producer/consumer.

jacques-n: The derivation definition would be in the function signature yaml (or whatever format/name we give it), not the actual plan. In a function signature yaml, I'd be happy for that format to support named type expressions to normalize that kind of pattern (so it doesn't have be declared repetitively).

cpcloud: I see. In that case I suspect `OR` would be fine to start. Seems like it'd be very clear pretty quickly if an `IN` style operator/function was needed.

site/docs/expressions/scalar_functions.md
```diff
@@ -43,10 +43,67 @@ For scalar functions (one value produced for each record), a function binding de
...
+* `substract(integer a, integer b) => integer`
+* `min(integer a, integer b) => integer`
+* `max(integer a, integer b) => integer`
+* `equal(integer a, integer b) => boolean`
+* `greater_than(integer a, integer b) => boolean`
+* `less_than(integer a, integer b) => boolean`
+* `equal(Type a, Type b) => boolean`
+* `if(boolean a) then (integer) else (integer)`
+* `if(boolean a) then (type) else (type)`
+* `not_bindable()`
```
cpcloud, line 94: It looks like there's `!bindable` and `not_bindable()`. Are these the same thing?

cpcloud: Oh, is `not_bindable()` what is produced by the syntax `!bindable`?

site/docs/expressions/scalar_functions.md
```diff
@@ -43,10 +43,67 @@ For scalar functions (one value produced for each record), a function binding de
...
+
+Fully defined with argument types:
+
+* `not(boolean x) => boolean` 
+* `and(boolean a, boolean b) => boolean` 
+* `or(boolean a, boolean b) => boolean` 
+* `multiply(integer a, integer b) => integer`
+* `divide(integer a, integer b) => integer`
+* `add(integer a, integer b) => integer`
+* `substract(integer a, integer b) => integer`
```
cpcloud, line 85: I'm going to be that guy and suggest that we don't leave `integer` ambiguous here and give it an unambiguous integer type. I'd vote for `i64` since it's the widest type, supporting the most use cases. Hopefully this isn't an area of the spec where we need to be extremely concerned about integer sizes :)

cpcloud: Also, `substract` has an extra `s` in it.

jacques-n: I guess I could make some kind of witty joke that anything sub* in the substrait project must be subst* and that is why I mispelled... :D   .fixed now.

Wrt to i64, I agree and actually already specified that just above.

site/docs/expressions/scalar_functions.md
jacques-n: None of this is syntax, just concepts. Syntax would be delegated to the serialization formats. In the abbreviated section I try to make it easy to understand what the ops are. The section below is blown out to better describe the parameters, etc.

site/docs/expressions/scalar_functions.md
westonpace: What is function definition time?  The point when the function is written down in the spec?  Or the point when the query plan is created?

If the former, perhaps simply:

A direct return type is one that does not depend on any parameterized input type.

If the latter:

A direct return type is one that is not know until the query has begun executing.

I assume from the context that we are talking about the former.

westonpace (changes requested):
Partly I worry we are over-learning the decimal promotion case.  My gut tells me that there will be more complex logic than we are going to be able to express in a simple type expression.  That being said, I don't think they harm anything.

I'm not sure these changes are extensive enough to capture what I had in mind.  I've made a PR to try and express my thoughts but even that is just a rough start.  I've got a pretty good test case (I will describe more in discussions) for this language.  I might run through things this weekend to see if I can create something that satisfies my test case.

westonpace:
(all syntax in this comment is pseduocode / for example purposes only)

I think the biggest thing missing for me at the moment is the ability to specify a range of types for an input argument.  For a simple scalar function example consider `LIST_VALUE_LENGTH<T>(T)=>uint32` and returns the length of each element in the input array.  In this case `T` is only valid for "lists".

Initially my thinking was that I didn't need type constraints because I could simply list out all possible permutations.  This works well for the numeric kernels (e.g. ADD) because there are a finite number of numeric types.  Unfortunately, there are infinitely many list types (e.g. lists of structs).  Some possible ways to represent this:

 * Let an argument be a type expression
   * E.g. `LIST_VALUE_LENGTH<T>(List<T>)=>uint32`
 * Allow type paremeters to take a "type condition expression" which evaluates to a boolean
   * E.g. `LIST_VALUE_LENGTH<T>(T)=>uint32 : where T satisfies IS_LIST(T)`

binary/function.proto
```diff
@@ -19,12 +19,9 @@ message FunctionSignature {
...
         bool session_dependent = 8;
 
-        oneof derivation_type {
-            Type direct = 9;
-            CustomDerivationPolicy derivation_policy = 10;
-        }
+        TypeExpression output_type = 9;
 
-        repeated Implementation implementations = 11;
+        repeated Implementation implementations = 10;
```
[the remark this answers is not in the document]
jacques-n: I believe it will be. You would declare use of the same parameter name on each of the args and then a return type of boolean.

binary/function.proto
```diff
@@ -121,4 +112,6 @@ message FunctionSignature {
         }
 
     }
+
```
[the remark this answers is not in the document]
jacques-n: It means it is consistent for every row. Examples might be literals (e.g. 1) or functions of literals (e.g. 2+5).

site/docs/expressions/scalar_functions.md
jacques-n: Yes, the former. I'll update in next patch.

site/docs/expressions/scalar_functions.md
```diff
@@ -43,10 +43,67 @@ For scalar functions (one value produced for each record), a function binding de
...
+* Parameterized types: Types can be parameterized such that the parameter can be used in a output derivation e.g. `f(K) => K` or to guarantee consistency of type between input arguments `f(K,K) => boolean`. Example parameterized types would be VARCHAR(T), List&lt;E&gt;, MAP&lt;K, fp32&gt;, etc. A parameters is named with a simple UTF8 character or string.
+
+### Direct Return Types
+
+A direct return type is one that is fully known at function definition type. Example simple direct return types would be things such as i32, fp32. For compound types, a direct return type must be fully declared. Example fully defined types: VARCHAR(20), DECIMAL(25,5)
+
+### Return Type Expressions
+
+Any function can declare a return type expression. A return type expression uses a simplified set of expressions to describe how the return type should be returned. For example, a return expression could be as simple as the return of parameter declared in the arguments. For example `f(List<K>) =>K ` or can be a simple mathematical or conditional expression such as `add(decimal(a,b), decimal(c,d)) => decimal(a+c, b+d)`. For the simple expression language, there is a very narrow set of types:
```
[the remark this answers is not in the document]
jacques-n: In my model (probably not expressed well in the doc yet) there are two things: Type Parameters and Integer Parameters. Each have a name that can be referenced in output derivation if desired. I added TypeParameter to the list of available types in the type union to allow the expression List<K>. In the type definitions, integer parameters also replaced the direct integers in the types that take one or more integer parameters. As of right now, those are the only two types of parameters that types can be parameterized by.

site/docs/expressions/scalar_functions.md
```diff
@@ -43,10 +43,67 @@ For scalar functions (one value produced for each record), a function binding de
...
+* `not_bindable()`
+
+#### Example Type Expressions Uses
+
+For reference, here are are some common output type derivations and how they can be expressed with a return type expression:
+
+| Operation                                     | Definition                                                   |
+| --------------------------------------------- | ------------------------------------------------------------ |
+| Add item to list                              | `add(<List<T>, T>) => List<T>`                               |
+| Decimal Division                              | `divide(Decimal(P1,S1), Decimal(P2,S2)) => Decimal(P1 -S1 + S2 + MAX(6, S1 + P2 + 1), MAX(6, S1 + P2 + 1))` |
```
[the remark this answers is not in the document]
jacques-n: This is the common way these are expressed in SQL land. Brackets for type parameters. Parenthesis for integer parameters.

jacques-n (commented):
> (all syntax in this comment is pseduocode / for example purposes only)
> 
> I think the biggest thing missing for me at the moment is the ability to specify a range of types for an input argument. For a simple scalar function example consider LIST_VALUE_LENGTH<T>(T)=>uint32 and returns the length of each element in the input array. In this case T is only valid for "lists".
> 
> Initially my thinking was that I didn't need type constraints because I could simply list out all possible permutations. This works well for the numeric kernels (e.g. ADD) because there are a finite number of numeric types. Unfortunately, there are infinitely many list types (e.g. lists of structs). Some possible ways to represent this:
> 
> Let an argument be a type expression
> E.g. LIST_VALUE_LENGTH<T>(List<T>)=>uint32
> Allow type paremeters to take a "type condition expression" which evaluates to a boolean
> E.g. LIST_VALUE_LENGTH<T>(T)=>uint32 : where T satisfies IS_LIST(T)

In my proposal, we allow type parameters in arguments and expression for output derivation. So in your example `LIST_VALUE_LENGTH<T>(List<T>)=>uint32`.

cpcloud (commented):
> Reviewed 2 of 3 files at r1, 1 of 1 files at r6.
> *[Reviewable](https://reviewable.io/reviews/substrait-io/substrait/40)* status: 2 of 3 files reviewed, 18 unresolved discussions (waiting on @cpcloud, @emkornfield, @jacques-n, @rymurr, and @westonpace)

___
*[binary/type.proto, line 37 at r6](https://reviewable.io/reviews/substrait-io/substrait/40#-Mklz-UFDSxgo1VVlD51:-Mklz-UFDSxgo1VVlD52:b8vhdbk) ([raw file](https://github.com/substrait-io/substrait/blob/242e105a2af1fee6a311b5f5c091578398342804/binary/type.proto#L37)):*
> ```protobuf
>         // A special type that can only be used for parameterized function arguments.
>         // This allows a user to express things like List<T>
>         Parameter parameter = 33;
> ```

Putting `Parameter` and `TypeExpression` here seems like it would allow folks to generate invalid IR, such as UDF that returns a type parameter or one that returns a type expression. Is that actually intended?
___
*[binary/type.proto, line 172 at r6](https://reviewable.io/reviews/substrait-io/substrait/40#-Mkm-8y-23AVEbeUvjLi:-Mkm-8y-23AVEbeUvjLj:b-em055w) ([raw file](https://github.com/substrait-io/substrait/blob/242e105a2af1fee6a311b5f5c091578398342804/binary/type.proto#L172)):*
> ```protobuf
>         oneof integer_type {
>             int32 literal = 1;
>             String parameter_name = 2;
> ```

Curious why this isn't a plain `string`
___
*[binary/type.proto, line 215 at r6](https://reviewable.io/reviews/substrait-io/substrait/40#-Mkm-UE_3Nxtt7JwJ6Oi:-Mkm-UE_3Nxtt7JwJ6Oj:bemx7ia) ([raw file](https://github.com/substrait-io/substrait/blob/242e105a2af1fee6a311b5f5c091578398342804/binary/type.proto#L215)):*
> ```protobuf
> 
>         enum OpType {
>             UNKNOWN = 0;
> ```

Is `UNKNOWN` a useful variant? Isn't it already defined by simply being anything that isn't in the `OpType` enum?
___
*[site/docs/expressions/scalar_functions.md, line 58 at r1](https://reviewable.io/reviews/substrait-io/substrait/40#-MkE6CNari3x0OUTa10q-r1-58:-Mkm0zCX5raOtwjYxGnZ:bd1nkzz) ([raw file](https://github.com/substrait-io/substrait/blob/66c97718050b368c3aee7cd377b32e5ef0624c9e/site/docs/expressions/scalar_functions.md#L58)):*
<details><summary><i>Previously, jacques-n (Jacques Nadeau) wrote…</i></summary><blockquote>

Yes, the former. I'll update in next patch.
</blockquote></details>

Should we unify the language here and use "concrete type" for both input and output types?
___
*[site/docs/expressions/scalar_functions.md, line 10 at r6](https://reviewable.io/reviews/substrait-io/substrait/40#-Mkm-jOE-7FNI4OHQxMe:-Mkm-jOE-7FNI4OHQxMf:b-9bcif9) ([raw file](https://github.com/substrait-io/substrait/blob/242e105a2af1fee6a311b5f5c091578398342804/site/docs/expressions/scalar_functions.md#L10)):*
> ```Markdown
> | ------------------ | ------------------------------------------------------------ | ----------------------------------- |
> | Name               | One or more user friendly utf8 strings that are used to reference this function in languages | At least one value is required.     |
> | List of arguments  | Argument properties are defined below. Arguments can be fully defined, parameterized or wildcarded. See further details below. | Optional, defaults to niladic.      |
> ```

Should `wildcarded` still be here?
___
*[site/docs/expressions/scalar_functions.md, line 52 at r6](https://reviewable.io/reviews/substrait-io/substrait/40#-Mkm-phz9nw_moHvvYIS:-Mkm-phz9nw_moHvvYIT:b-zejgui) ([raw file](https://github.com/substrait-io/substrait/blob/242e105a2af1fee6a311b5f5c091578398342804/site/docs/expressions/scalar_functions.md#L52)):*
> ```Markdown
> Input arguments can be declared in one of two ways: materialized or parameterized.
> 
> * Materialized Type: Materialized types are either simple types or compound types with all parameters fully defined (without any parameters). Examples include i32, fp32, VARCHAR(20), List&lt;fp32&gt;, etc.
> ```

How about concrete type instead of materialized? How about "Concrete types are types that have no unbound type parameters. Examples include ..."

binary/function.proto
```diff
@@ -19,12 +19,9 @@ message FunctionSignature {
         bool deterministic = 7;
         bool session_dependent = 8;
 
-        oneof derivation_type {
-            Type direct = 9;
-            CustomDerivationPolicy derivation_policy = 10;
-        }
+        TypeExpression output_type = 9;
```
mbasmanova, line 22: It might be helpful to add a flag that says whether function has default null behavior, e.g. null in any input is guaranteed to produce a null result.

binary/function.proto
```diff
@@ -36,10 +33,7 @@ message FunctionSignature {
...
```
mbasmanova, line 36: Any particular reason intermediate_type is a concrete type rather than TypeExpression?

Also, what is max_set? In general, how do you feel about adding comments to describe various fields here to help the reader?

binary/type.proto
```diff
@@ -31,6 +31,17 @@ message Type {
         TimestampTZ timestamp_tz = 29;
         UserDefined user_defined = 31;
         UUID uuid = 32;
+
+        // A special type that can only be used for parameterized function arguments.
+        // This allows a user to express things like List<T>
+        Parameter parameter = 33;
+
+        // A special type that can only be used in function return type declarations.
```
mbasmanova, line 39: What makes it special? Can you give an example?

binary/type.proto
```diff
@@ -154,4 +165,66 @@ message Type {
...
+        BinaryOp binary_op = 3;
+        Type.Parameter parameter = 4;
+        IfElse if_else = 5;
+        NotBindable not_bindable = 6;
+        Type type = 7;
+    }
+
+    message IfElse {
+        TypeExpression if_condition = 1;
+        TypeExpression if_return = 2;
```
mbasmanova, line 192: Is this the type of "then" expressions? Why is it not called then_return? Shouldn't the type of "then" and "else" expressions be the same? If so, why do we need both if_return and else_return?

jacques-n (commented):
> Dismissed @cpcloud and @westonpace from 7 discussions.
> *[Reviewable](https://reviewable.io/reviews/substrait-io/substrait/40)* status: 2 of 3 files reviewed, 16 unresolved discussions (waiting on @cpcloud, @emkornfield, @jacques-n, @mbasmanova, @rymurr, and @westonpace)

___
*[binary/function.proto, line 22 at r7](https://reviewable.io/reviews/substrait-io/substrait/40#-MkE6CNari3x0OUTa10o-r7-22:-MknPTo-4y5jFJPjPMij:b-bkdiiu) ([raw file](https://github.com/substrait-io/substrait/blob/d72320c5497feff792b0d3bcc7e7c00a39510ee4/binary/function.proto#L22)):*
<details><summary><i>Previously, mbasmanova (Maria Basmanova) wrote…</i></summary><blockquote>

It might be helpful to add a flag that says whether function has default null behavior, e.g. null in any input is guaranteed to produce a null result.
</blockquote></details>

Great question. I'm torn on this one. I actually had that in an earlier version. I removed it as it felt like an function semantics detail as opposed to something that the plan itself needs to express. What do you think a substrait producer or consumer could do with this additional information that they wouldn't have already from another source? (For example, I'd expect that engines would already know this when they were doing code generation against the plan based on properties of their particular implementation).

My biggest argument for it is that probably makes sense for any kind of common structured description of a function semantic to be expressed just to make things easier to understand/validate. For example, I also had a error behavior property original that could be: [overflow, nullify, throw]. It also felt like something outside the realm of substrait and so I removed it but it probably should be added back if we add a null-if-null concept.
___
*[binary/type.proto, line 37 at r6](https://reviewable.io/reviews/substrait-io/substrait/40#-Mklz-UFDSxgo1VVlD51:-MknOXl85cNYBxTuo2M0:b-yae6ri) ([raw file](https://github.com/substrait-io/substrait/blob/242e105a2af1fee6a311b5f5c091578398342804/binary/type.proto#L37)):*
<details><summary><i>Previously, cpcloud (Phillip Cloud) wrote…</i></summary><blockquote>

Putting `Parameter` and `TypeExpression` here seems like it would allow folks to generate invalid IR, such as UDF that returns a type parameter or one that returns a type expression. Is that actually intended?
</blockquote></details>

Yeah, I'm torn on this. The other option is to create separate idl object hierachies for the two other categories of types.

Concrete Type
Parameterized Type
Type Expression

Since you can embed parameters in the pieces of a concrete type and the same for type expression in parameterized types, I took the shorter path of just having a single representation that will then be validated later. The problems is things like List<Struct<foo:K, bar:string>>. Either we implement three impls of list and struct: one that only accepts concrete, one that accepts concrete + parameter and one that accepts concrete + parameter + type expression or we do this pattern.

As I wrote more about it, maybe it isn't so bad to just have three separate unions. I was think all types would need to be duplicated but there are only a small number of types that actually need duplication (the ones with type parameters). Definitely makes the idl safer/reduces the needs for additional validation. Further thoughts?
___
*[binary/type.proto, line 172 at r6](https://reviewable.io/reviews/substrait-io/substrait/40#-Mkm-8y-23AVEbeUvjLi:-MknMxIb2NfGfmXVK8Xs:b-7yr5tv) ([raw file](https://github.com/substrait-io/substrait/blob/242e105a2af1fee6a311b5f5c091578398342804/binary/type.proto#L172)):*
<details><summary><i>Previously, cpcloud (Phillip Cloud) wrote…</i></summary><blockquote>

Curious why this isn't a plain `string`
</blockquote></details>

Yeah, weird. I'm confused on how it even compiled :/
___
*[binary/type.proto, line 215 at r6](https://reviewable.io/reviews/substrait-io/substrait/40#-Mkm-UE_3Nxtt7JwJ6Oi:-MknNCioEXCeD-Yuu2-9:bqg4kw) ([raw file](https://github.com/substrait-io/substrait/blob/242e105a2af1fee6a311b5f5c091578398342804/binary/type.proto#L215)):*
<details><summary><i>Previously, cpcloud (Phillip Cloud) wrote…</i></summary><blockquote>

Is `UNKNOWN` a useful variant? Isn't it already defined by simply being anything that isn't in the `OpType` enum?
</blockquote></details>

Protobuf3ism- unknown is the same as unset. Not useful directly but clarifies if the data was missing. If you use it for something proper you can't tell whether the field was forgotten or was set to that value.
___
*[binary/type.proto, line 39 at r7](https://reviewable.io/reviews/substrait-io/substrait/40#-MkE6CNari3x0OUTa10p-r7-39:-MknOCm-2NHb2U_U9eXO:b-4231db) ([raw file](https://github.com/substrait-io/substrait/blob/d72320c5497feff792b0d3bcc7e7c00a39510ee4/binary/type.proto#L39)):*
<details><summary><i>Previously, mbasmanova (Maria Basmanova) wrote…</i></summary><blockquote>

What makes it special? Can you give an example?
</blockquote></details>

This is used in return types to do things like 

fn(DECIMAL(P,S)) => DECIMAL(min(38,p+1))

Basically the output type is a calculation based on the input types.
___
*[binary/type.proto, line 192 at r7](https://reviewable.io/reviews/substrait-io/substrait/40#-MkE6CNari3x0OUTa10p-r7-192:-MknNIYp6D2MQimFYbq0:b-638qmo) ([raw file](https://github.com/substrait-io/substrait/blob/d72320c5497feff792b0d3bcc7e7c00a39510ee4/binary/type.proto#L192)):*
<details><summary><i>Previously, mbasmanova (Maria Basmanova) wrote…</i></summary><blockquote>

Is this the type of "then" expressions? Why is it not called then_return? Shouldn't the type of "then" and "else" expressions be the same? If so, why do we need both if_return and else_return?
</blockquote></details>

I'm not sure I'm understanding the question/thought. An example here might be:

Imagine fn(DECIMAL(P, S)) => IF P < 18 THEN DECIMAL(18,S) ELSE DECIMAL(38,S). 

Not a great example but the return type of the two branches are different. So in this case:

if_condition = P < 18
if_return = DECIMAL(18,S)
else_return = DECIMAL(38,S)
___
*[site/docs/expressions/scalar_functions.md, line 53 at r1](https://reviewable.io/reviews/substrait-io/substrait/40#-MkE6CNari3x0OUTa10q-r1-53:-MknRFz3DXVVKNY_8VF3:b6j2ken) ([raw file](https://github.com/substrait-io/substrait/blob/66c97718050b368c3aee7cd377b32e5ef0624c9e/site/docs/expressions/scalar_functions.md#L53)):*
<details><summary><i>Previously, jacques-n (Jacques Nadeau) wrote…</i></summary><blockquote>

I've updated to remove wildcards.
</blockquote></details>

addressed in patch.
___
*[site/docs/expressions/scalar_functions.md, line 54 at r1](https://reviewable.io/reviews/substrait-io/substrait/40#-MkE6CNari3x0OUTa10q-r1-54:-MknRH3k30P5Mg3h_u8d:b6j2ken) ([raw file](https://github.com/substrait-io/substrait/blob/66c97718050b368c3aee7cd377b32e5ef0624c9e/site/docs/expressions/scalar_functions.md#L54)):*
<details><summary><i>Previously, jacques-n (Jacques Nadeau) wrote…</i></summary><blockquote>

Good point on the equal. I've updated to remove the distinction between wildcard and parameterized.
</blockquote></details>

addressed in patch.
___
*[site/docs/expressions/scalar_functions.md, line 58 at r1](https://reviewable.io/reviews/substrait-io/substrait/40#-MkE6CNari3x0OUTa10q-r1-58:-MknRKPmD9n1ejTAkWrz:bhy9ot6) ([raw file](https://github.com/substrait-io/substrait/blob/66c97718050b368c3aee7cd377b32e5ef0624c9e/site/docs/expressions/scalar_functions.md#L58)):*
<details><summary><i>Previously, cpcloud (Phillip Cloud) wrote…</i></summary><blockquote>

Should we unify the language here and use "concrete type" for both input and output types?
</blockquote></details>

Will do.
___
*[site/docs/expressions/scalar_functions.md, line 10 at r6](https://reviewable.io/reviews/substrait-io/substrait/40#-Mkm-jOE-7FNI4OHQxMe:-MknRCCeCtYNfmsp18nq:bgvs6h8) ([raw file](https://github.com/substrait-io/substrait/blob/242e105a2af1fee6a311b5f5c091578398342804/site/docs/expressions/scalar_functions.md#L10)):*
<details><summary><i>Previously, cpcloud (Phillip Cloud) wrote…</i></summary><blockquote>

Should `wildcarded` still be here?
</blockquote></details>

will fix
___
*[site/docs/expressions/scalar_functions.md, line 52 at r6](https://reviewable.io/reviews/substrait-io/substrait/40#-Mkm-phz9nw_moHvvYIS:-MknRCztELXlfLSzAKXC:b-lhw40x) ([raw file](https://github.com/substrait-io/substrait/blob/242e105a2af1fee6a311b5f5c091578398342804/site/docs/expressions/scalar_functions.md#L52)):*
<details><summary><i>Previously, cpcloud (Phillip Cloud) wrote…</i></summary><blockquote>

How about concrete type instead of materialized? How about "Concrete types are types that have no unbound type parameters. Examples include ..."
</blockquote></details>

sounds good, will update.
___
*[site/docs/expressions/scalar_functions.md, line 103 at r6](https://reviewable.io/reviews/substrait-io/substrait/40#-MkE6CNari3x0OUTa10q-r6-103:-MknRSpCEc9qDkqpmSoL:b-r5o3je) ([raw file](https://github.com/substrait-io/substrait/blob/242e105a2af1fee6a311b5f5c091578398342804/site/docs/expressions/scalar_functions.md#L103)):*
<details><summary><i>Previously, jacques-n (Jacques Nadeau) wrote…</i></summary><blockquote>

This is the common way these are expressed in SQL land. Brackets for type parameters. Parenthesis for integer parameters. 
</blockquote></details>

any particular thoughts on the answer here @westonpace ?

jacques-n (commented):
> *[Reviewable](https://reviewable.io/reviews/substrait-io/substrait/40)* status: 2 of 3 files reviewed, 16 unresolved discussions (waiting on @cpcloud, @emkornfield, @jacques-n, @mbasmanova, @rymurr, and @westonpace)

___
*[binary/function.proto, line 36 at r7](https://reviewable.io/reviews/substrait-io/substrait/40#-MkE6CNari3x0OUTa10o-r7-36:-MknUc5j2JImIRQy7w7Z:b-hklhhk) ([raw file](https://github.com/substrait-io/substrait/blob/d72320c5497feff792b0d3bcc7e7c00a39510ee4/binary/function.proto#L36)):*
<details><summary><i>Previously, mbasmanova (Maria Basmanova) wrote…</i></summary><blockquote>

Any particular reason intermediate_type is a concrete type rather than TypeExpression?

Also, what is max_set? In general, how do you feel about adding comments to describe various fields here to help the reader?
</blockquote></details>

You're right, intermediate type should be a type expression. Will update. max_set is the maximum number of values this function will accept. E.g. if you have LIST_AGG and state that you only allow up to 1000 items in each list, you'd set this 1000. 

Yes, we should definitely add comments. It's currently pre-consensus so we've been focused more on the spec than the specific serialization (although I've been trying to keep the proto up to date so people can see how might be serialized). Once we get the core spec components better clarified, we plan on doing some consensus work around the serialization formats so I wasn't working super hard on making the proto consumable outside as a learning aid against the spec.

mbasmanova:
> > _[Reviewable](https://reviewable.io/reviews/substrait-io/substrait/40)_ status: 2 of 3 files reviewed, 16 unresolved discussions (waiting on @cpcloud, @emkornfield, @jacques-n, @mbasmanova, @rymurr, and @westonpace)
> 
> _[binary/function.proto, line 36 at r7](https://reviewable.io/reviews/substrait-io/substrait/40#-MkE6CNari3x0OUTa10o-r7-36:-MknUc5j2JImIRQy7w7Z:b-hklhhk) ([raw file](https://github.com/substrait-io/substrait/blob/d72320c5497feff792b0d3bcc7e7c00a39510ee4/binary/function.proto#L36)):_
> 
> Previously, mbasmanova (Maria Basmanova) wrote…
> You're right, intermediate type should be a type expression. Will update. max_set is the maximum number of values this function will accept. E.g. if you have LIST_AGG and state that you only allow up to 1000 items in each list, you'd set this 1000.
> 
> Yes, we should definitely add comments. It's currently pre-consensus so we've been focused more on the spec than the specific serialization (although I've been trying to keep the proto up to date so people can see how might be serialized). Once we get the core spec components better clarified, we plan on doing some consensus work around the serialization formats so I wasn't working super hard on making the proto consumable outside as a learning aid against the spec.

Thank you for a quick reply. I was hoping to compile this into C++ and try to map to [Velox's PlanNode](https://github.com/facebookincubator/velox/blob/main/velox/core/PlanNode.h) representation to see what's missing if anything. Would that be possible at all at this stage?

jacques-n:
> Thank you for a quick reply. I was hoping to compile this into C++ and try to map to [Velox's PlanNode](https://github.com/facebookincubator/velox/blob/main/velox/core/PlanNode.h) representation to see what's missing if anything. Would that be possible at all at this stage?

It should be possible. I'd expect most queries to be expressible. It would also be a lot of help to validate the design, find weaknesses, etc. I'm sure there are things missing and we're still changing things but for sure worth doing. FYI, I'm exploring doing the other side: using Calcite to produce a substrait plan. It really helps to exercise/better evaluate design decisions.

jacques-n:
> > Thank you for a quick reply. I was hoping to compile this into C++ and try to map to [Velox's PlanNode](https://github.com/facebookincubator/velox/blob/main/velox/core/PlanNode.h) representation to see what's missing if anything. Would that be possible at all at this stage?
> 
> It should be possible. I'd expect most queries to be expressible. It would also be a lot of help to validate the design, find weaknesses, etc. I'm sure there are things missing and we're still changing things but for sure worth doing. FYI, I'm exploring doing the other side: using Calcite to produce a substrait plan. It really helps to exercise/better evaluate design decisions.

I just remembered that the proto sketch for relations is pretty minimal right now. (The spec is more complete.) I'll add some more sketches there.

site/docs/expressions/scalar_functions.md
```diff
@@ -1,52 +1,118 @@
 # Scalar Functions
 
-Scalar functions are a function that takes in values from a single record and produces an output value. To clearly specify the definition of functions, Substrait declares a extensible specification plus binding approach to function resolution. This ensures a clear definition of function behavior separate from a particular use of each function while avoiding operations 
+Scalar functions are a function that takes in values from a single record and produces an output value. To clearly specify the definition of functions, Substrait declares a extensible specification plus binding approach to function resolution. A scalar function signature includes the following properties:
```
westonpace, line 3:
```suggestion
A function is a scalar function if that function takes in values from a single record and produces an output value. To clearly specify the definition of functions, Substrait declares a extensible specification plus binding approach to function resolution. A scalar function signature includes the following properties:
```
You could also maybe do "Scalar functions are functions that..."  but "Scalar functions are a function that" sounds weird to me.

This paragraph itself is rather intimidating and unclear to a neophyte but we tackle that more in the post-consensus phase.

site/docs/expressions/scalar_functions.md
```diff
@@ -1,52 +1,118 @@
...
 
-Substrait supports a number of functions. A scalar function signature includes the following properties:
+| Property               | Description                                                  | Required                            |
+| ---------------------- | ------------------------------------------------------------ | ----------------------------------- |
+| Name                   | One or more user friendly utf8 strings that are used to reference this function in languages | At least one value is required.     |
```
westonpace, line 7: What does it mean if there are multiple names?  Are they aliases of each other?

site/docs/expressions/scalar_functions.md
```diff
@@ -1,52 +1,118 @@
...
+| List of arguments      | Argument properties are defined below. Arguments can be fully defined, parameterized or wildcarded. See further details below. | Optional, defaults to niladic.      |
```
westonpace, line 8:
```suggestion
| List of arguments      | Argument properties are defined below. Arguments can be fully defined or calculated with a type expression. See further details below. | Optional, defaults to niladic.      |
```

site/docs/expressions/scalar_functions.md
```diff
@@ -1,52 +1,118 @@
...
+| Deterministic          | Whether this function is expected to reproduce the same output when it is invoked multiple times with the same input. This informs a plan consumer on whether it can constant reduce the defined function. An example would be a random() function, which is typically expected to be evaluated repeatedly despite having the same set of inputs. | Optional, defaults to true.         |
+| Session Dependent      | Whether this function is influenced by the session context it is invoked within. For example, a function may be influenced by a user who is invoking the function, the time zone of a session, or some other non-obvious parameter. This can inform caching systems on whether a particular function is cacheable. | Optional, defaults to false.        |
```
westonpace, line 10: Naive question.  Do we really need both session dependent and deterministic or can session dependent just be one particular case of non-deterministic?  Will a system be able to optimize more efficiently knowing a function is deterministic but not session dependent?

site/docs/expressions/scalar_functions.md
```diff
@@ -1,52 +1,118 @@
...
+| Variadic Behavior      | Whether the last argument of the function is variadic. Options include: [Single argument, M..N]. M must be less than or equal to N. N can be defined or left as "unlimited". | Optional, defaults to single value. |
```
westonpace, line 11:
```suggestion
| Variadic Behavior      | Whether the last argument of the function is variadic or a single argument.  If variadic the argument can optionally have a lower bound (minimum number of instances) and upper bound (maximum number of instances) | Optional, defaults to single value. |
```

`[Single argument, M..N]` <- I get what you are trying to express here but it doesn't come through to me except that I know the answer already.

site/docs/expressions/scalar_functions.md
```diff
@@ -1,52 +1,118 @@
...
+| Description            | Additional description of function for implementers or users. Should be written human readable to allow exposure to end users. Presented as a map with language => description mappings. E.g. `{ "en": "This adds two numbers together.", "fr": "cela ajoute deux nombres"}`. | Optional                            |
+| Return Type Expression | The output type is expected to be a physical type of the expression. Return types can be one of a simple, fully-defined compound type or a type expression. See below for more on type expressions. | Required                            |
+| Implementation Map     | A map of implementation locations for one or more implementations of the given function. Each key is a function implementation type. Implementation types include examples such as: AthenaArrowLambda, TrinoV361Jar, ArrowCppKernelEnum, GandivaEnum, LinkedIn Transport Jar, etc. [Definition TBD]. Implementation type has one or more properties associated with retrieval of that implementation. | Optional                            |
```
westonpace, line 14: Maybe a foolish ask but is there any chance we can get away with a URI here?

site/docs/expressions/scalar_functions.md
```diff
@@ -1,52 +1,118 @@
...
+| Variadic Behavior      | Whether the last argument of the function is variadic. Options include: [Single argument, M..N]. M must be less than or equal to N. N can be defined or left as "unlimited". | Optional, defaults to single value. |
+| Description            | Additional description of function for implementers or users. Should be written human readable to allow exposure to end users. Presented as a map with language => description mappings. E.g. `{ "en": "This adds two numbers together.", "fr": "cela ajoute deux nombres"}`. | Optional                            |
+| Return Type Expression | The output type is expected to be a physical type of the expression. Return types can be one of a simple, fully-defined compound type or a type expression. See below for more on type expressions. | Required                            |
```
westonpace, line 13:
```suggestion
| Return Value | The output type of the expression.  Return types can be expressed as a fully-defined type or a type expression. See below for more on type expressions. | Required                            |
```

I don't know what you meant by "physical type".  I don't see the merit in distinguishing between a simple type and a fully-defined compound type.  They are both "fully-defined types".

site/docs/expressions/scalar_functions.md
```diff
@@ -1,52 +1,118 @@
...
+* Value Arguments: Arguments that where the function will refer to data value, could be a constant (literal-based expression) or variable (reference-based expression). This is the most common type of argument. Value arguments are not available in output derivation. Value arguments can be declared in one of two ways: concrete or parameterized. Concrete Types are either simple types or compound types with all parameters fully defined (without any parameters). Examples include i32, fp32, VARCHAR(20), List&lt;fp32&gt;, etc. Parameterized types are discussed further below.
+* Type Arguments: Type arguments are used to inform the evaluation and/or type derivation of the function. For example, you might have a function which is `truncate(<type> DECIMAL(P0,S0), <value> DECIMAL(P1, S1), <value> i32)`. This function declares two value arguments (arg1 and arg2) and type argument. The type argument's value can be used in output type determination. (The types of the values arguments can also be used for output type derivation.)
```
westonpace, line 23:
```suggestion
* Value Arguments: An argument that refers to a data value.  This could be a constant (a literal expression defined in the plan) or variable (A reference expression that references data being processed by the plan). This is the most common type of argument. Value arguments are not available in output derivation. Value arguments can be declared in one of two ways: concrete or parameterized. Concrete Types are either simple types or compound types with all parameters fully defined (without referencing any type arguments). Examples include i32, fp32, VARCHAR(20), List&lt;fp32&gt;, etc. Parameterized types are discussed further below.
* Type Arguments: Type arguments are used to inform the evaluation and/or type derivation of the function. For example, you might have a function which is `truncate(<type> DECIMAL(P0,S0), <value> DECIMAL(P1, S1), <value> i32)`. This function declares two value arguments and a type argument. The type argument's value can be used to determine the output type (The types of the values arguments can also be used for output type derivation.)
```

I agree with what you are saying, I am just (hopefully) clearing up the wording.  `literal-based expression` and `reference-based expression` confused me for a while but I get what you are saying.  I removed `(arg1 and arg2)` because it isn't clear to me what you were trying to say.

site/docs/expressions/scalar_functions.md
```diff
@@ -1,52 +1,118 @@
...
 
+#### Value Argument Properties
 
+| Property | Description                                                  | Required                                                   |
+| -------- | ------------------------------------------------------------ | ---------------------------------------------------------- |
+| Name     | A human name for this argument to help clarify use.          | Optional, defaults to a name based on position (e.g. arg0) |
```
westonpace, line 29:
```suggestion
| Name     | A human readable name for this argument to help clarify use.          | Optional, defaults to a name based on position (e.g. arg0) |
```

site/docs/expressions/scalar_functions.md
```diff
@@ -1,52 +1,118 @@
...
+| Type     | A fully concrete or parameterized (bound or not).            | Required                                                   |
```
westonpace, line 30:
```suggestion
| Type     | A fully defined type or a type expression.            | Required                                                   |
```

site/docs/expressions/scalar_functions.md
```diff
@@ -1,52 +1,118 @@
...
+| Constant | Whether this argument is required to be a constant for invocation. For example, in some system a regular expression pattern would only be accepted as a literal and not a column value reference. | Optional, defaults to false                                |
 
-## Function Ids
+#### Type Argument Properties
 
-Each function signature is categorized based on a function signature id. The identifier includes two components:
+| Property | Description                                                  | Required                                                   |
+| -------- | ------------------------------------------------------------ | ---------------------------------------------------------- |
+| Type     | A partially or completely parameterized type. E.g. List&lt;K&gt; or K | Required                                                   |
+| Name     | A human name for this argument to help clarify use.          | Optional, defaults to a name based on position (e.g. arg0) |
```
westonpace, line 38:
```suggestion
| Name     | A human readable name for this argument to help clarify use.          | Optional, defaults to a name based on position (e.g. arg0) |
```

binary/parameterized_types.proto
```diff
@@ -0,0 +1,117 @@
+syntax = "proto3";
+package io.substrait;
+
+import "type.proto";
+import "extensions.proto";
+
+option java_multiple_files = true;
+
+message ParameterizedType {
```
westonpace, line 9: Is a type also a parameterized type?  In other words, in spots expecting a `ParameterizedType` can I just give a `Type`?  Is `VARCHAR(20)` a `ParameterizedType`?

binary/function.proto
```diff
@@ -105,20 +109,18 @@ message FunctionSignature {
     }
 
     message Argument {
+        ParameterizedType type = 1;
```
westonpace, line 112: Can `type` just be a `Type`? (I think I ask this same question a little further down where you define `ParameterizedType`)

site/docs/expressions/scalar_functions.md
```diff
@@ -1,52 +1,118 @@
...
 
-| Property        | Description                                                  |
-| --------------- | ------------------------------------------------------------ |
-| Organization Id | An unsigned 32 bit integer mapped to a list of known organizations listed in the [Substrait repository](https://github.com/substrait-io/substrait/blob/main/extensions/organizations.yaml). |
-| Function Id     | An unsigned 32 bit integer mapped to a list of known functions for the specific organization. For the Substrait organization, the function is listed in [here](https://github.com/substrait-io/substrait/blob/main/extensions/scalar_functions.yaml). |
+### Parameterized Types
 
-An organization is responsible for managing their own lists of functions. A function id should be used forever. If a function is deleted for any reason, the id should not be reused. Ideally each function signature will have one or more 
+Types are parametered by two types of values: by inner types (e.g. List&lt;K&gt;) and numeric values (e.g. VARCHAR(P,S)). Parameter names are simple strings (frequently a single character). There are two types of parameters: integer parameters and type parameters.
```
westonpace, line 42: What about the names in a struct parameterization?  That doesn't seem to fit the dichotomy of types/integers.

westonpace, line 42:
```suggestion
Types are parameterized by two types of values: by inner types (e.g. List&lt;K&gt;) and numeric values (e.g. VARCHAR(P,S)). Parameter names are simple strings (frequently a single character). There are two types of parameters: integer parameters and type parameters. 
```

site/docs/expressions/scalar_functions.md
```diff
@@ -1,52 +1,118 @@
...
+### Parameterized Types
 
-An organization is responsible for managing their own lists of functions. A function id should be used forever. If a function is deleted for any reason, the id should not be reused. Ideally each function signature will have one or more 
+Types are parametered by two types of values: by inner types (e.g. List&lt;K&gt;) and numeric values (e.g. VARCHAR(P,S)). Parameter names are simple strings (frequently a single character). There are two types of parameters: integer parameters and type parameters. 
 
-## Scalar Function Bindings
+When the same parameter name is used multiple times in a function definition, the function can only bind if the exact same value is used for all parameters of that name. For example, if one had a function with a signature of `fn(VARCHAR(N), VARCHAR(N)), the function would be only be usable if both VARCHAR types had the same length value N. This necessitates that all instances of the same parameter name must be of the same parameter type (all instances are a type parameter or always intances are aninteger parameter).
```
westonpace, line 44:
```suggestion
When the same parameter name is used multiple times in a function definition, the function can only bind if the exact same value is used for all parameters of that name. For example, if one had a function with a signature of `fn(VARCHAR(N), VARCHAR(N)), the function would be only be usable if both VARCHAR types had the same length value N. This necessitates that all instances of the same parameter name must be of the same parameter type (all instances are a type parameter or all instances are an integer parameter).
```

It also necessitates that all instances of the same parameter name are the exact same value which seems a stronger conclusion.

site/docs/expressions/scalar_functions.md
```diff
@@ -1,52 +1,118 @@
...
 
-For scalar functions (one value produced for each record), a function binding declares the function identifier to be used and the input arguments.
+#### Integer Parameter Bounds
 
+A type which declares one or more integer parameters can also declare a range of legal values for a function. For example, one could declare a function `fn(VARCHAR(N)) where N IN [0..20]` to describe a function which only takes in a varchar types that are less than or equal to 20 characters in length.
```
westonpace, line 48:
```suggestion
A type which declares one or more integer parameters can also declare a range of legal values for a function. For example, one could declare a function `fn(VARCHAR(N)) where N IN [0..20]` to describe a function which only takes in varchar types that are less than or equal to 20 characters in length.
```

site/docs/expressions/scalar_functions.md
```diff
@@ -1,52 +1,118 @@
...
+
+#### Type Parameter Bounds
+
+A parameterized type can either be unbounded (any type is allowed) or be bounded. When bounded, the function can only be bound to arguments of the types bound to. For example, one could declare a function `fn(Map<K,V> map) where K IN [STRING, VARCHAR(N), FIXEDCHAR(N)]` which would allow the function to accept a map argument as long as the key type is one of STRING, VARCHAR or FIXEDCHAR. Note, when duplicate parameter names are used in disjunct bounds options, the names will use the same bounds. For example, in the above map function, if one were to declare a bound such as where N IN [0..20], this limit would apply to either N parameter.
```
westonpace, line 52: If I have `fn(K, K) where K IN [STRING, VARCHAR(N), FIXEDCHAR(N)]` is `fn(STRING, VARCHAR(N))` a valid combination?

site/docs/expressions/scalar_functions.md
```diff
@@ -1,52 +1,118 @@
...
+* `add(integer a, integer b) => integer`
+* `subtract(integer a, integer b) => integer`
+* `min(integer a, integer b) => integer`
+* `max(integer a, integer b) => integer`
+* `equal(integer a, integer b) => boolean`
+* `greater_than(integer a, integer b) => boolean`
+* `less_than(integer a, integer b) => boolean`
+* `covers(Type a, Type b) => boolean`Covers means that type b matches type A for as much as type B is defined. For example, if type A is `VARCHAR(20)` and type B is `VARCHAR(N)`, type B would be considered covering. Similarlily if type A was `List<Struct<a:f32, b:f32>>`and type B was `List<Struct<>>`, it would be considered covering. Note that this is directional "as in B covers A" or "B can be further enhanced to match the definition A". 
+* `if(boolean a) then (integer) else (integer)`
+* `if(boolean a) then (type) else (type)`
```
westonpace, line 101: I don't understand the purpose of this section.

site/docs/expressions/scalar_functions.md
```diff
@@ -1,52 +1,118 @@
...
+
+#### Example Type Expressions Uses
```
westonpace, line 103:
```suggestion
#### Example Type Expressions
```

Or maybe "Example Uses for Type Expressions" or "Example Use Cases for Type Expressions"

site/docs/expressions/scalar_functions.md
```diff
@@ -1,52 +1,118 @@
...
+
+For reference, here are are some common output type derivations and how they can be expressed with a return type expression:
+
+| Operation                                                 | Definition                                                   |
+| --------------------------------------------------------- | ------------------------------------------------------------ |
+| Add item to list                                          | `add(<List<T>, T>) => List<T>`                               |
+| Decimal Division                                          | `divide(Decimal(P1,S1), Decimal(P2,S2)) => Decimal(P1 -S1 + S2 + MAX(6, S1 + P2 + 1), MAX(6, S1 + P2 + 1))` |
+| Do regex on only string maps to return values.            | `extract_values(Map<K,V>) => List<V> WHERE K IN [STRING, VARCHAR(N), FIXEDCHAR(N)]` |
```
westonpace, line 111:
```suggestion
| Select a subset of map keys based on a regular expression (requires stringlike keys)           | `extract_values(Map<K,V>) => List<V> WHERE K IN [STRING, VARCHAR(N), FIXEDCHAR(N)]` |
```

Where is the regular expression expressed?

site/docs/expressions/scalar_functions.md
```diff
@@ -1,52 +1,118 @@
...
+| Concatenate two fixed sized character strings             | `concat(FIXEDCHAR(A), FIXEDCHAR(B)) => CHAR(A+B)`            |
```
westonpace, line 112:
```suggestion
| Concatenate two fixed sized character strings             | `concat(FIXEDCHAR(A), FIXEDCHAR(B)) => FIXEDCHAR(A+B)`            |
```

site/docs/expressions/scalar_functions.md
```diff
@@ -1,52 +1,118 @@
...
+| Make a struct of a set of fields and a struct definition. | `make_struct(<type> T, K...) => T`                           |
```
westonpace, line 113: Where are the field names expressed?

westonpace (approved):
This is a clear improvement in my mind but I don't know that it solves everything just yet.  I didn't review the proto files very much as I'd rather review that with actual usage.

westonpace (commented):
> *[Reviewable](https://reviewable.io/reviews/substrait-io/substrait/40)* status: 0 of 8 files reviewed, 38 unresolved discussions (waiting on @cpcloud, @emkornfield, @jacques-n, @mbasmanova, @rymurr, and @westonpace)

___
*[binary/function.proto, line 22 at r7](https://reviewable.io/reviews/substrait-io/substrait/40#-MkE6CNari3x0OUTa10o-r7-22:-MkoCPoe8GPOVJ4JLHP2:bc9svlq) ([raw file](https://github.com/substrait-io/substrait/blob/d72320c5497feff792b0d3bcc7e7c00a39510ee4/binary/function.proto#L22)):*
<details><summary><i>Previously, jacques-n (Jacques Nadeau) wrote…</i></summary><blockquote>

Great question. I'm torn on this one. I actually had that in an earlier version. I removed it as it felt like an function semantics detail as opposed to something that the plan itself needs to express. What do you think a substrait producer or consumer could do with this additional information that they wouldn't have already from another source? (For example, I'd expect that engines would already know this when they were doing code generation against the plan based on properties of their particular implementation).

My biggest argument for it is that probably makes sense for any kind of common structured description of a function semantic to be expressed just to make things easier to understand/validate. For example, I also had a error behavior property original that could be: [overflow, nullify, throw]. It also felt like something outside the realm of substrait and so I removed it but it probably should be added back if we add a null-if-null concept.
</blockquote></details>

It would certainly be useful in automated property-based testing of query engines but I agree that may not fall into goals here.  I do worry a bit it would be a slippery slope.  For example, in addition to your "error behavior" there is "all nulls -> null" (e.g. maybe some nulls doesn't guarantee null but all nulls does) and "transitive", "never errors", etc.
___
*[site/docs/expressions/scalar_functions.md, line 103 at r6](https://reviewable.io/reviews/substrait-io/substrait/40#-MkE6CNari3x0OUTa10q-r6-103:-MkoDzPM0CuvRrCWcuRP:b-bbpuv8) ([raw file](https://github.com/substrait-io/substrait/blob/242e105a2af1fee6a311b5f5c091578398342804/site/docs/expressions/scalar_functions.md#L103)):*
<details><summary><i>Previously, jacques-n (Jacques Nadeau) wrote…</i></summary><blockquote>

any particular thoughts on the answer here @westonpace ?
</blockquote></details>

I don't like it but I won't fight it :)  I think I'd rather it be all angle brackets to avoid confusion with functions.

binary/function.proto
jacques-n: The idl for a parameterized type doesn't require any parameters to be included. So in that sense, logically yes. I guess I could make either an option here as a union but feels excessive.

binary/parameterized_types.proto
jacques-n: From a proto idl, you can represent VARCHAR(20) three different ways: with Type, ParameterizedType and DerivationExpression. They are very similar representations but to ensure that only the right kind of type information is input in each location, they are independent in the idl. I'm not sure of a better way to do this. The alternative is merging them all and having to write separate tools to validate semantics.

site/docs/expressions/scalar_functions.md
jacques-n: Yeah, they are aliases.

site/docs/expressions/scalar_functions.md
jacques-n: They play in different spaces/influence different optimization opportunities. 

Deterministic describes whether or not a value could be reduced to a constant in some expressions. If it is non-deterministic, you must execute for each record. Select rand(), a,b,c from foo is a good example. You wouldn't expect rand() to be the same value on any record. If you did the same thing as add(1,1), you could replace with the literal 2.

Session dependent describes whether or not a plan's interpretation is impacted by session properties and thus means you can't share something like a plan or data cache between different sessions (without also keying session properties). A function that returns username is a prime example.

site/docs/expressions/scalar_functions.md
jacques-n: My thinking was that providing the type of implementation in the map, it allows a system to only read manifests for implementations it can actually use. E.g. type => git-uri.

site/docs/expressions/scalar_functions.md
jacques-n: Good edit.

site/docs/expressions/scalar_functions.md
jacques-n: looks good

site/docs/expressions/scalar_functions.md
jacques-n: nice improvements, thanks

site/docs/expressions/scalar_functions.md
jacques-n: Yeah, there is no accomodation for that. For now, I would think that a plan builder would generate any required information using a type argument.

jacques-n: thanks

site/docs/expressions/scalar_functions.md
jacques-n: No. My goal with the wording previously was that K must the exact same value in all locations. If you want to support the pattern of mixed arguments, you'd need to use separate parameter names.

site/docs/expressions/scalar_functions.md
jacques-n: It declares the argument inputs/outputs for the available operations. The lines 77-80 were to make it easy to get a gist. The lines 83-101 were trying to be full definitions. (That was the intention, anyway.)

site/docs/expressions/scalar_functions.md
jacques-n: hehehe, good point. Will add post accepting your edits.

site/docs/expressions/scalar_functions.md
jacques-n: In the type T

cpcloud: What is the meaning of `K...` here? Does it mean every `K[0]` up to `K[n - 1]` are possibly distinct types? Or does it mean that every argument must be the same type `K`?

Also, what is the `<type> T` syntax? Isn't that covered by declaring type parameters in the function signature like `make_struct<T, K>(T, K...)`?

jacques-n: >What is the meaning of K... here?
I called this out elsewhere in the doc. In a variadic case, the last argument declaration can be either marked as consistent or inconsistent (I need to add this to the protobuf). If consistent, must resolve the same. If inconsistent, each instance can be a different value.

> Also, what is the <type> T syntax?

It's all pseudo syntax to help explain the concepts for the spec. We need to define an actual syntax.

PR #49, opened 2021-10-07
Proposed updates to support required and optional enumerations.

jacques-n:


site/docs/expressions/scalar_functions.md
```diff
@@ -22,6 +22,8 @@ There are two main types of arguments: Value Arguments and Type Arguments.
...
+* Optional Enumeration: Similar to required enumeration but more focused on supporting alternative behaviors. An optional enumeration always includes an "unspecified" default option that can be bound based on the capabilities of the plan consumer. When a plan states does not specify a behavior, the consumer is expected to resolve the option based on the first option the system can match. An example usecase might be `OVERFLOW_BEHAVIOR:[OVERFLOW, SATURATE, ERROR]` If unspecified, an engine would use the first of these that it implements. If specified, the engine would be expected to behave as specified or fail. Note, the value of an optional enumeration cannot be used in type derivation.
```
westonpace, line 26:
```suggestion
* Optional Enumeration: Similar to required enumeration but more focused on supporting alternative behaviors. An optional enumeration always includes an "unspecified" default option that can be bound based on the capabilities of the plan consumer. When a plan does not specify a behavior, the consumer is expected to resolve the option based on the first option the system can match. An example usecase might be `OVERFLOW_BEHAVIOR:[OVERFLOW, SATURATE, ERROR]` If unspecified, an engine would use the first of these that it implements. If specified, the engine would be expected to behave as specified or fail. Note, the value of an optional enumeration cannot be used in type derivation.
```

binary/expression.proto
```diff
@@ -18,6 +18,16 @@ message Expression {
...
+    }
+
+    message Enum {
+        oneof enum_kind {
+            string specified = 1;
+            Empty unspecified = 2;
+        }
+
+        message Empty {}
     }
```
westonpace, line 31:
```suggestion
    message Enum {
        optional string value = 1;
    }
```
Maybe I'm misunderstanding protobof?

binary/function.proto
```diff
@@ -124,15 +124,26 @@ message FunctionSignature {
     }
 
     message Argument {
-        ParameterizedType type = 1;
-        ArgumentKind kind = 2;
-        string name = 3;
+        string name = 1;
```
westonpace, line 127: In the YAML it looks like options/optional are part of the "parameters" section but here it looks like you are defining them for the "arguments" section?

extensions/aggregate_functions.yaml
```diff
@@ -2,49 +2,59 @@ functions:
...
-      overflow: [SILENT, SATURATE, ERROR]
     variants:
       - name: scalar
         decomposable: MANY
         parameters:
-          - K: [i8,i16,i32,i64,fp32,fp64]
+          - K: {types: [i8,i16,i32,i64,fp32,fp64]}
+          - overflow:
+            options: [SILENT, SATURATE, ERROR]
+            required: false
```
westonpace, line 12: Really minor nit: This is legal YAML but for consistency maybe pick one style and stick with it:

```
          - K:
            types: [i8,i16,i32,i64,fp32,fp64]
          - overflow:
            options: [SILENT, SATURATE, ERROR]
            required: false
```

...or...
```
          - K: {types: [i8,i16,i32,i64,fp32,fp64]}
          - overflow: {options: [SILENT, SATURATE, ERROR], required: false}
```

westonpace (commented):
I'll try and look at functions.yaml in more detail later (but don't wait for me)

extensions/aggregate_functions.yaml
jacques-n: this really made me laugh :)

it made more sense before I started introducing the templating thing...

binary/expression.proto
jacques-n: You could do this but it makes thing more implicit. The way I've done it makes it clearer (albeit slightly larger on the wire). Optional is implicit in pb3 so doesn't have to be stated. The union also allows us to introduce more than two options later.

The other nice thing is that the "enum_kind" becomes something to switch off of in several languages.

binary/function.proto
jacques-n: Yeah, i think this is clearer/more natural this way when using protobuf but I think it is easier to read/consume in yaml the other way (since there are no message types in Yaml). Basically, trying to be idiomatic to each form.

PR #54, opened 2021-10-18
Add field names to struct literals

cpcloud:
This change is [Reviewable](https://reviewable.io/reviews/substrait-io/substrait/54)

### DIFF
--- a/binary/expression.proto
+++ b/binary/expression.proto
@@ -70,15 +70,20 @@ message Expression {
             int32 years = 1;
             int32 months = 2;
         }
+
         message IntervalDayToSecond {
             int32 days = 1;
             int32 seconds = 2;
         }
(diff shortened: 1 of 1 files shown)


cpcloud:
I guess this really stems from the lack of `List` literals (it sounds unrelated but hear me out), so maybe it's premature since I don't have a problem to solve other than writing a producer that carries around struct field names :sweat_smile: .

The current `Struct` literal representation is the same as I think a `List` literal representation should be.

I think it's bit odd to shove a thing called `Struct` around for `List`s, but I also think it's odd to entirely duplicate messages by just slapping a different name on them.

On the other hand, plan clarity with respect to the type of a literal seems important so maybe duplicating the representation is fine.

PR #67, opened 2021-10-30
Rename sort type as sort direction

westonpace:
I can't find the term "sort type" in very many sources and when I do it is referring to something else (e.g. Redshift's interleaved vs compound).

On the other hand, "direction" is pretty common:

Iceberg and Spark have "SortOrder" which is a "SortDirection" and a "NullOrder".
Postgres calls it the "sort direction"

Splitting into Sort Order, Sort Direction, and Null Order to mirror Iceberg/Spark would be another option.

---
This change is [Reviewable](https://reviewable.io/reviews/substrait-io/substrait/67)

### DIFF
--- a/binary/expression.proto
+++ b/binary/expression.proto
@@ -150,10 +150,10 @@ message Expression {
         Expression expr = 1;
 
         oneof sort_kind {
-            SortType formal = 2;
+            SortDirection direction = 2;
             Extensions.FunctionId comparison_function = 3;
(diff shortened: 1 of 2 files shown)


jacques-n (approved):
Direction seems a little weird when thinking about clustered but that's probably not the first thing people think about and if you think "Direction" is more comprehensible for people, I'm all for that.

PR #70, opened 2021-11-09
Update functions to flatter format.

### LINKED ISSUE #65: Proposal: Make the system consumable function yaml easier to integrate
TLDR: @cpcloud and @jacques-n have both been working on building substrait producers and the current yaml functions are too difficult to work with. The yaml needs to be more discrete (which results in something less concise).

A key part of plan translation to/from substrait is the mapping of Substrait function definitions to producer or consumer function definitions. Currently, at least the following translators are underway:
* @jacques-n is working on converting a Calcite plan to Substrait. In Calcite, functions are mapped to very generic complex object hierarchies with embedded type resolution logic (e.g. `add(ANY, ANY)`).
* @cpcloud is working on converting Ibis plans to Substrait. In Ibis, functions are mapped to very specific function definitions `add(int32,int32)`.

This means the two efforts have touched different ends of the abstraction specification. Both efforts have found that this mapping process is overly complex using the current representation. In both cases, people have tried to create some kind additional abstractions to make the mapping process more straightforward. Since it is critical to facilitate ease of Substrait integration in many tools, this suggested that the level of abstraction/complexity is not right for easy integrations.

The current function signature yaml docs were primarily optimized to express things concisely. This includes the use of aliases, anchors and merges nodes as well as the capability to express type bounds. While the current form may be a relatively concise representation for human consumption, concision has come at the cost of integration of effort. Part of the integration complexity come from the fact that the functions are neither fully generic nor fully discrete and the current pattern for offset-based resolution. As we were discussing things, supporting fully discrete would make things easiest to reason about. 

Recommendation: Move to a representation that is much simpler. 
- Remove type bounds (only allow type concrete types or pure wildcard type)
- Remove yaml-only features

This simpler representation would require less sophistication of integrators comprehending the representation. Additionally, it also allows us to simplify the referencing scheme for binding to a specific function. Our current thoughts are that we should update the referencing scheme to express use a string pattern for referencing the specific discrete type we're bounding to. As an example, `add(i32,i32)` might become be referenced using one of `add<i32,i32>`, `add_ii`, `add:ii` or something similar to that. While a simple string representation might be somewhat, less fancy than other systems, it allows easier human consumption.

To make things less repetitive on the document side for things like description/name, we're currently thinking that we add an additional level of structure to the yaml document to share common attributes.

I'll look to post a patch proposing the changes.

jacques-n:
This closes #65.

---
This change is [Reviewable](https://reviewable.io/reviews/substrait-io/substrait/70)


extensions/functions_boolean.yaml
```diff
@@ -0,0 +1,35 @@
...
+  -
+    name: or
+    description: The boolean or of two values.
+    impls:
+      - args:
+          - value: boolean
+          - value: boolean
+        return: boolean
+  -
+    name: and
```
cpcloud, line 21: Should this have any discussion of TVL/Kleene? Or are you thinking that's a separate set of functions?

extensions/functions_arithmetic.yaml
```diff
@@ -0,0 +1,278 @@
...
+        intermediate: "STRUCT<fp64,i64>"
+        return: fp32?
+      - args:
+          - options: [ SILENT, SATURATE, ERROR ]
+            required: false
+          - value: fp64
+        nullability: DECLARED_OUTPUT
+        decomposable: MANY
+        intermediate: "STRUCT<fp64,i64>"
+        return: fp64?
```
cpcloud, line 278: newline

extensions/functions_arithmetic.yaml
```diff
@@ -0,0 +1,278 @@
...
+        intermediate: "STRUCT<i64,i64>"
+        return: i32?
+      - args:
+          - options: [ SILENT, SATURATE, ERROR ]
+            required: false
+          - value: i64
+        nullability: DECLARED_OUTPUT
+        decomposable: MANY
+        intermediate: "STRUCT<i64,i64>"
+        return: i64?
```
cpcloud, line 262: Are the semantics of this function a truncated integer average?

extensions/functions_aggregate_generic.yaml
```diff
@@ -0,0 +1,14 @@
...
+aggregate_functions:
+  - name: "count"
+    description: Count a set of values
+    impls:
+      - args:
+          - options: [ SILENT, SATURATE, ERROR ]
+            required: false
+          - value: any
+        nullability: DECLARED_OUTPUT
+        decomposable: MANY
```
cpcloud, line 12: Is the meaning of `decomposable` written down anywhere?

binary/extensions.proto
```diff
@@ -47,7 +47,6 @@ message Extensions {
             FunctionId function_id = 1;
             ExtensionId extension_id = 2;
             string name = 3;
-            uint32 index = 4;
```
cpcloud, line 50: Yay!

extensions/functions_arithmetic.yaml
```diff
@@ -0,0 +1,278 @@
+%YAML 1.2
+---
+scalar_functions:
+  -
+    name: "add"
+    description: "Add two values."
+    impls:
+      - args:
+          - options: [ SILENT, SATURATE, ERROR ]
```
cpcloud, line 9: This is repetitive, but I prefer the explictness.

cpcloud (commented):
LGTM, just a few minor questions/comments.

extensions/functions_aggregate_generic.yaml
jacques-n: Yes: https://github.com/substrait-io/substrait/blob/main/site/docs/expressions/aggregate_functions.md?plain=1#L12

extensions/functions_arithmetic.yaml
jacques-n: Good question. I figure we should go with whatever is the sql standard. Do you happen to know what it is? We can add to description here.

extensions/functions_arithmetic.yaml
jacques-n: done

extensions/functions_boolean.yaml
jacques-n: I was thinking this was three-valued. We should clarify in the descriptions. Were you thinking it shouldn't be?

(in general, I think in sql)

extensions/functions_boolean.yaml
```diff
@@ -0,0 +1,35 @@
...
+          - value: boolean
+        return: boolean
+  -
+    name: xor
+    description: The boolean xor of two values.
+    impls:
+      - args:
+          - value: boolean
+          - value: boolean
+        return: boolean
```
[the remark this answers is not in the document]
jacques-n: intellij apparently doesn't for yaml files...

fixed

extensions/functions_datetime.yaml
```diff
@@ -0,0 +1,95 @@
...
+          - value: interval_year
+        return: timestamp
+      - args:
+          - value: timestamp_tz
+          - value: interval_year
+        return: timestamp
+      - args:
+          - value: date
+          - value: interval_year
+        return: timestamp
```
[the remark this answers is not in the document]
jacques-n: nope. fixed for this and the others nearby.

extensions/functions_datetime.yaml
```diff
@@ -0,0 +1,95 @@
...
+          - value: interval_day
+        return: timestamp
+      - args:
+          - value: timestamp_tz
+          - value: interval_day
+        return: timestamp
+      - args:
+          - value: date
+          - value: interval_day
+        return: timestamp
```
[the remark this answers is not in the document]
jacques-n: yup, fixed.

extensions/functions_arithmetic.yaml
westonpace: SQL Server - Truncated integer
PostgreSQL / MySQL - Decimal

jacques-n: For the updates here I went with truncate. I figure we'll probably need to define a second set that return different data types/decimal promotion.

extensions/functions_boolean.yaml
cpcloud: TVL sounds right to me.

cpcloud (approved):

PR #75, opened 2021-11-15
Add SEMI and ANTI Join types

hannesmuehleisen:
This change is [Reviewable](https://reviewable.io/reviews/substrait-io/substrait/75)

### DIFF
--- a/binary/relations.proto
+++ b/binary/relations.proto
@@ -110,6 +110,8 @@ message JoinRel {
         OUTER = 2;
         LEFT = 3;
         RIGHT = 4;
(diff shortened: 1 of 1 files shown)


jacques-n (commented):
Do you think it makes sense to incorporate these into a standard join type? Do you want to share a couple of example SQL queries and how you would like to express the join condition? I just want to make sure we're comfortable with collapsing these into the same representation.

jacques-n (approved):

hannesmuehleisen:
Thanks!

PR #76, opened 2021-11-21
Various specification/documentation improvements

jacques-n:
- Clarify extensions including documentation, updated binary representation, distinction between simple and advanced.
- Provide documentation on serialization and more serialization examples, including details around surrogate keys and pointers
- Update all fields that are surrogate keys to use uint32 as opposed to complex structures. Update field names to better clarify behavior
- Add a capabilities proto message and documentation description
- Provide an example unknown.yaml file that shows how simpler functions can be declared
- Add extension capabilities to protobuf for advanced extensions (along with supporting documentation)
- Add a simple extension yaml schema and ensure all yaml files validate.
- Update type variations proto definition and extension definition pattern to be consistent with other extension types.
- Reorder binary serialization to be first in nav given it's greater maturity over text.
- Remove HintKeyValue and replace with AdvancedExtension use.

---
This change is [Reviewable](https://reviewable.io/reviews/substrait-io/substrait/76)


binary/capabilities.proto
```diff
@@ -0,0 +1,26 @@
+syntax = "proto3";
+
+package io.substrait;
+
+option java_multiple_files = true;
+option csharp_namespace = "Substrait.Protobuf";
+
+// Defines a set of Capabilities that a system (producer or consumer) supports.
```
cpcloud, line 8: Since the message is already capitalized and the point (I think) of this documentation is for humans to read it, the capitalization at best adds confusion.

> a system (producer or consumer)

It's fine to just say `producer or consumer`, the parenthetical isn't necessary.

binary/capabilities.proto
```diff
@@ -0,0 +1,26 @@
...
+message Capabilities {
+
+    // List of Substrait versions this system supports
```
cpcloud, line 11: Can you make the capitalization consistent?

binary/capabilities.proto
```diff
@@ -0,0 +1,26 @@
...
+    repeated string substrait_versions = 1;
+
+    // list of com.google.Any message types this system supports for advanced extensions.
```
cpcloud, line 14: What does `com.google.Any` have to do with this field?

binary/capabilities.proto
```diff
@@ -0,0 +1,26 @@
...
+    repeated string advanced_extension_type_urls = 2;
+
+    // list of simple extensions this system supports.
+    repeated SimpleExtension simple_extensions = 3;
+
+    message SimpleExtension {
+        string uri = 1;
+        repeated string function_keys = 2;
+        repeated string type_keys = 3;
+        repeated string type_variation_keys = 4;
```
cpcloud, line 24: I think these need documentation, and possibly examples.

binary/expression.proto
```diff
@@ -68,6 +77,18 @@ message Expression {
...
+        message Decimal {
+            // little-endian twos-complement integer representation of complete value (ignoring precision)
+            bytes value = 1;
+            
+            // declared scale of decimal literal
+            int32 scale = 2;
+            
+            // place to put decimal within twos-complement value.
+            int32 precision = 3;
+
```
cpcloud, line 89: Kill this extra newline. I'm going to open up a PR to wrangle all these random newlines everywhere and enforce a consistent style. It's starting to get a bit out of hand.

binary/expression.proto
```diff
@@ -68,6 +77,18 @@ message Expression {
...
 
+        message Decimal {
+            // little-endian twos-complement integer representation of complete value (ignoring precision)
+            bytes value = 1;
+            
+            // declared scale of decimal literal
+            int32 scale = 2;
+            
+            // place to put decimal within twos-complement value.
+            int32 precision = 3;
```
cpcloud, line 88: :sweat_smile: I hit this (missing field) recently, but forgot to open a PR to fix it.

binary/expression.proto
```diff
@@ -42,18 +42,27 @@ message Expression {
             double fp64 = 11;
             string string = 12;
             bytes binary = 13;
+            
+            // microseconds since epoch
             fixed64 timestamp = 14;
+            
+            // microseconds since epoch (at midnight)
```
cpcloud, line 49: Is this really microseconds? I thought it was milliseconds. Also, why is midnight mentioned? The UNIX epoch is already defined to be midnight.

binary/expression.proto
```diff
@@ -42,18 +42,27 @@ message Expression {
...
             fixed32 date = 16;
+            
+            // a time value in nanoseconds from midnight.
```
cpcloud, line 52: Nanos? I thought this was microseconds. In the systems I've worked in the most common representations for time were seconds and microseconds but I think only pandas represents times in nanoseconds. Other systems *can* represents nanos, but that's distinct from choosing to use nanos exclusively as pandas does.

binary/relations.proto
```diff
@@ -5,44 +5,47 @@ package io.substrait;
...
     }
 
     Hint hint = 3;
-    RuntimeConstraint constraint = 4;
+    io.substrait.extensions.AdvancedExtension advanced_extension = 4;
 
     message Direct {}
     message Emit {
-        repeated int32 output_mapping = 1;
+        repeated int32 output_mapping = 1 [packed = true];
```
cpcloud, line 26: This is already packed by default since we're using `proto3`: https://developers.google.com/protocol-buffers/docs/encoding#packed.

binary/capabilities.proto
```diff
@@ -0,0 +1,26 @@
+syntax = "proto3";
+
+package io.substrait;
+
+option java_multiple_files = true;
+option csharp_namespace = "Substrait.Protobuf";
+
+// Defines a set of Capabilities that a system (producer or consumer) supports.
+message Capabilities {
```
cpcloud, line 9: Can you split this into its own PR?

binary/capabilities.proto
```diff
@@ -0,0 +1,26 @@
...
+    // List of Substrait versions this system supports
+    repeated string substrait_versions = 1;
+
+    // list of com.google.Any message types this system supports for advanced extensions.
+    repeated string advanced_extension_type_urls = 2;
+
+    // list of simple extensions this system supports.
+    repeated SimpleExtension simple_extensions = 3;
+
+    message SimpleExtension {
```
cpcloud, line 20: Can you split the advanced/simple extension changes into a separate PR?

site/docs/expressions/specialized_record_expressions.md
```diff
@@ -4,12 +4,20 @@ While most all types of operations could be reduced to functions, in some cases
...
 For each data type, it is possible to create a literal value for that data type. The representation depends on the serialization format.
 
 
 
+## Cast Expression
+
+To convert a value from one type to another, Substrait defines a cast expression. Cast expression declare an expected type and an input argument.
+
+
+
```
cpcloud, line 20: I know there's preexisting extra newlines, but why all the extras?

cpcloud (changes requested):
Would like to see:

1. Capabilities in a separate PR
2. Extension changes in a separate PR

As is there's a lot of code/doc changes before those two changesets. It'll be easier to understand and faster to review if capabilities and extensions are each their own PR.

binary/expression.proto
jacques-n: I went with microseconds to be consistent with timestamp. Midnight is mentioned because this is a date. I would think that only certain values are valid (those at midnight)

FYI, I've abandoned these changes for now. Mostly extraneous from the more heavy lifting in this patch.

binary/expression.proto
jacques-n: Yeah, nanos doesn't make sense. Good point :)

binary/expression.proto
jacques-n: Yeah, someone else mentioned it. I've taken all literal changes out of this pr. We can capture separately.

binary/expression.proto
jacques-n: Don't be a newline hater!

binary/capabilities.proto
jacques-n: I have split it out but won't post until the other patches get merged since it depends on them (the docs part)

binary/capabilities.proto
jacques-n: com.google.Any defines [two fields](https://github.com/protocolbuffers/protobuf/blob/master/src/google/protobuf/any.proto#L154).

One is the type the any is holding (`type_label`) and the other is the binary data for that message type (`value`). This is a list of all `type_label`s that may be used in this plan. It allows a consuming system to understand whether there may be unexpected messages types without having to walk the entire tree.

binary/capabilities.proto
jacques-n: done

binary/capabilities.proto
jacques-n: agreed. baby steps, baby!

binary/relations.proto
jacques-n: fixed (in the separated out patch)

site/docs/expressions/specialized_record_expressions.md
jacques-n: beauty. I edit these primarily in a visual editor and probably did to make it look better. Removing for now (and extracting as a separate pr to be added after the core ones).

jacques-n (commented):
Comments are listed here but I am going to close this PR and instead use #79 and #80 (and additional once those are closed)

jacques-n:
Closing due to splitting. Now using #79 and #80

PR #77, opened 2021-11-21
Lift nullability, variation and NamedStruct

cpcloud:
This PR does a few things:

1. Moves the `Type.NamedStruct` out of the `Type` message namespace into the
   toplevel namespace of the `types.proto` file.
1. Similarly, `Type.Variation` is moved up.
1. DRYs up `Variation` and `Nullability` into the `Type` message, since each of
   those is repeated for every type. After moving `NamedStruct` and `Variation`
   out of the `Type` namespace this makes sense to deduplicate the message definition.
   This also simplifies producer code by allowing producers to handle variation and nullability
   *before* producing the specific type.

---
This change is [Reviewable](https://reviewable.io/reviews/substrait-io/substrait/77)

### DIFF
--- a/binary/type.proto
+++ b/binary/type.proto
@@ -53,10 +53,10 @@ message Type {
     }
 
     // The type's nullability
-    Nullability nullability = 1;
+    Nullability nullability = 32;
 
     // The type's variation. Optional.
-    Variation variation = 2;
+    Variation variation = 33;
 
     message Boolean {}
     message I8 {}
--- a/binary/type.proto
+++ b/binary/type.proto
@@ -6,8 +6,16 @@ import "extensions.proto";
 option java_multiple_files = true;
 option csharp_namespace = "Substrait.Protobuf";
 
-message Type {
+// A variation on an instance of a type
+//
+// Analogous to a typedef in C, a variation is useful for assigning a
+// different name to an existing type.
+message Variation {
+    int32 organization = 1;
+    string name = 2;
+}
 
+message Type {
     oneof kind {
         Boolean bool = 1;
         I8 i8 = 2;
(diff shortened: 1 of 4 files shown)


binary/type.proto
```diff
@@ -38,145 +46,68 @@ message Type {
...
     }
-
-    message Variation {
-        int32 organization = 1;
-        string name = 2;
-    }
-
 }
 
+message NamedStruct {
```
cpcloud, line 109: Since `NamedStruct` wasn't a type in the `oneof` (and wasn't nullable or variable), there's no utility in keeping it embedded in the `Type` message.

jacques-n (commented):
(reviewing on phone)

This change is incomplete since it removes the ability to construct many things such as nullability of i8 in a parameterized type. It also makes type construction inconsistent between parameterized types, type expressions and types. 

I considered doing this multiple times while working with types but kept coming back to consistency between the type classes being preferred (and the consistency ultimately kept code that worked with the types simpler than dry idl). I wonder if you've done this change having only been focused on type and not the other two type system variants?

I'm not saying I'm against the change but we should be convinced that the inconsistent type construction pattern is what we want.

cpcloud:
> nullability of i8 in a parameterized type.

Can you give an example of this?

> It also makes type construction inconsistent between parameterized types, type expressions and types.

That's fair. I was only looking at `Type`. I'll make them all match.

> and the consistency ultimately kept code that worked with the types simpler than dry idl)

This isn't my experience. Every location where I produce a type I have to call a function to translate the nullability. It would be much simpler if I could call that function one time, and _then_ recur into the specific `oneof` variant. I tried to make this clear in the last bullet:

> This also simplifies producer code by allowing producers to handle variation and nullability before producing the specific type.

jacques-n:
> > nullability of i8 in a parameterized type.
> 
> Can you give an example of this?
>

Parameterized type is a superset of type because it can be used in places where types can be parameterized or not. The difference is parametertized type allows us to not only express `i8?` but also express things like `List<T>` and `VARCHAR<L>` whereas type only allows literals like `List<i8>` and `VARCHAR<10>`.  The current pattern allows us to reuse the type definitions in the types.proto as opposed to restating them all. The only things that need to be implemented anew in parametertized_type.proto are new concepts and complex types since those now allow a union that covers a broader set of things.
 
> > It also makes type construction inconsistent between parameterized types, type expressions and types.
> 
> That's fair. I was only looking at `Type`. I'll make them all match.

I don't think you can, which is the complexity/problem.

> 
> > and the consistency ultimately kept code that worked with the types simpler than dry idl)
> 
> This isn't my experience...

You haven't worked with all three type systems. If you only work in a types.proto, I agree that abstracting the concepts out of the specific type instances make sense/clean things up. Once I started working in all three, moving between required a different pattern. If you're interested, you can see here how I dealt with this [here](https://github.com/jacques-n/substrait/blob/java/java/core/src/main/java/io/substrait/type/Types.java) and [here](https://github.com/jacques-n/substrait/blob/java/java/core/src/main/java/io/substrait/type/proto/ProtoTypes.java). [Note, this patch is still in progress.]

> Abstractly, nullability isn't a property of specific types, it's a property of any type. I'm not sure I understand the motivation to duplicate the nullability in every specific type.

This is only true for type, not for parameterized type nor type expression. Other members of the union.

I'd also note that in a perfect world, I don't think user-defined types would support variations. It feels like unnecessary level of composition since user-defined types are already arbitrarily defined.

As I've thought about this more, I've wondered whether the relationship/DRY for types between type, parameterized types and type expressions is important. I suspect that 99% of early implementations will only focus on type. The use of the other two concepts are only relevant if you want to start communicating function signatures between systems in binary form (as opposed to using the yaml uri reference pattern). However, before we commit to having inconsistent patterns between the two (and having no reuse), I thought it was worthwhile to make sure it is clear why it is structured the way it is before we change it.

jacques-n:
What happened? This PR brought up some good questions/ideas.

cpcloud:
Ah, I wasn't clear on whether this was viable or not as a PR. I haven't spent any time yet with types-outside-`type.proto` so I figured I should do that before revisiting. Happy to keep the discussion going though.

jacques-n:
To me, there are two options:

1. make type as easy to use as possible (the common entrypoint)
   1. This follows the changes here (roughly)
   1. Duplicates concepts in a different way in `ParameterizedType` and `DerivationExpression` since not all items in those union support these properties.
   1. Duplicates the simple type declarations (which are, admittedly, simple)
1. Keep the three consistent (what we have currently)
   1. Requires more complexity for type users
   1. Support reuse of simple types.

While I've built a bunch on the second option, the more I think about it, the more I'm inclined to the first option.

jacques-n:
@cpcloud thoughts on my comments here? I tried to reopen this but apparently the branch is gone :'(

PR #82, opened 2021-11-21
Add NamedRel

cpcloud:
This PR adds a `NamedRel` message for use at the root of an expression tree.

In particular, it's useful for being able to store a view without loss of information,
as well as being able to round trip a producer for testing.

### DIFF
--- a/binary/relations.proto
+++ b/binary/relations.proto
@@ -171,10 +171,10 @@ message SetRel {
 // This is for use at the root of a `Rel` tree.
 message RelRoot {
     // A relation
-    Rel input = 2;
+    Rel input = 1;
 
     // Field names in depth-first order
-    repeated string names = 3;
+    repeated string names = 2;
 }
 
 message Rel {
--- a/binary/relations.proto
+++ b/binary/relations.proto
@@ -172,7 +172,6 @@ message SetRel {
 message RelRoot {
     // A relation
     Rel input = 1;
-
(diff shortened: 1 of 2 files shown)


binary/relations.proto
```diff
@@ -176,5 +188,6 @@ message Rel {
       JoinRel join = 6;
       ProjectRel project = 7;
       SetRel set = 8;
+      NamedRel named = 9;
```
cpcloud, line 191: I guess one thing that is potentially awkward is the ability to compose `NamedRel`s, which seems valid, but not very useful.

jacques-n: That's why I don't think it belongs in the Rel union. I'm inclined to elevate, similar to NamedStruct. It also removes the inclusion of common, which doesn't really belong in this "marker".

jacques-n: One other idea. Maybe we call this RelRoot. That's what Calcite [calls it](https://github.com/apache/calcite/blob/master/core/src/main/java/org/apache/calcite/rel/RelRoot.java). It has some other cruft but it is basically a rel + field names (called "validated row type" in calcite).

cpcloud: `RelRoot` does seem like an apt name. Let's go with that. I'll move it out as well.

jacques-n (approved):
Lgtm, thanks!

PR #80, opened 2021-11-21
Clarify/solidify extensions

jacques-n:
- Clarify extensions including documentation, updated binary representation, distinction between simple and advanced.
- Provide documentation on serialization and more serialization examples, including details around surrogate keys and pointers
- Update all fields that are surrogate keys to use uint32 as opposed to complex structures. Update field names to better clarify behavior
- Add extension capabilities to protobuf for advanced extensions (along with supporting documentation)
- Update type variations proto definition and extension definition pattern to be consistent with other extension types.
- Reorder binary serialization to be first in nav given it's greater maturity over text.
- Remove HintKeyValue and replace with AdvancedExtension use.

PR #83, opened 2021-11-21
Format protobufs with clang-format and check in CI

cpcloud:
This PR reformats the protobuf files and adds a check in CI to catch any future
non-conforming formatting

### DIFF
--- a/binary/plan.proto
+++ b/binary/plan.proto
@@ -10,34 +10,34 @@ option csharp_namespace = "Substrait.Protobuf";
 
 // Either a relation or root relation
 message PlanRel {
-    oneof RelType {
-        // Any relation
-        Rel rel = 1;
-        // The root of a relation tree
-        RelRoot root = 2;
-    }
+  oneof RelType {
+    // Any relation
+    Rel rel = 1;
+    // The root of a relation tree
+    RelRoot root = 2;
+  }
 }
 
 // Describe a set of operations to complete.
 // For compactness sake, identifiers are normalized at the plan level.
 message Plan {
 
-    // a list of yaml specifications this plan may depend on
-    repeated io.substrait.extensions.SimpleExtensionURI extension_uris = 1;
-    
-    // a list of extensions this plan may depend on
-    repeated io.substrait.extensions.SimpleExtensionDeclaration extensions = 2;
-    
-    // one or more relation trees that are associated with this plan.
-    repeated PlanRel relations = 3;
+  // a list of yaml specifications this plan may depend on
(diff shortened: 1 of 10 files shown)


cpcloud:
Happy to split this up if that's better. All of the changes here except the CI bits were generated by running

```sh
$ clang-format -i binary/*proto
```

cpcloud:
I looked at Uber's [prototool](https://github.com/uber/prototool) but it pretty aggressively removes newlines altogether, which I wasn't too keen on.

binary/plan.proto
```diff
@@ -10,18 +10,18 @@ option csharp_namespace = "Substrait.Protobuf";
 
 // Either a relation or root relation
 message PlanRel {
-    oneof RelType {
-        // Any relation
-        Rel rel = 1;
-        // The root of a relation tree
-        RelRoot root = 2;
-    }
+  oneof RelType {
```
jacques-n, line 13: As part of this patch, can we make sure that we also check for other things:
* require proto3 syntax
* no use of excessive optional, packed references
* names follow standards:
  * snake_case for packages, files, fields, oneof kinds (this one was updated inconsistently and didn't notice until after merge)
  * CamelCase for messages, enum

cpcloud: I think we can do that, but I'd rather do it with a tool like [`buf`](https://docs.buf.build/introduction) (they provide [useful github actions](https://docs.buf.build/ci-cd/github-actions) to do this).

I think `prototool` is the only other alternative, and they recommend looking at `buf`!

cpcloud: I'd like to do that in a follow up because there are some changes that the buf linter recommends and I don't think those should be bundled in this PR.

jacques-n: How much hand work was this PR (versus automatic)? Asking as it may be painful to rebase #80 on top of these changes (I haven't tried, just guessing).

cpcloud: All of the work here except the CI bits was automatic, so virtually no hand work was done. Happy to wait for #80 before landing this.

mariusvniekerk: `buf lint` is quite nice.  Additionally `buf breaking` is awesome for drift detection

jacques-n:
I've merged #80, can you rebase this? Would be great to merge soon.

jacques-n:
I did the rebase and pushed here. Hopefully correctly!

jacques-n (approved):
Thanks for pulling this together @cpcloud

PR #85, opened 2021-11-22
Add precision and scale to decimal, clarify meaning of bytes

cpcloud:


cpcloud:
cc @bkietz

jacques-n (commented):
I wonder if we need to clarify the length of the bytes? Maybe we should state that is always 16 bytes?

cpcloud:
Yeah, let's clarify the byte length.

binary/expression.proto
```diff
@@ -59,6 +59,17 @@ message Expression {
             List list = 30;
         }
 
+        message Decimal {
+            // little-endian twos-complement integer representation of complete value (ignoring precision)
+            // Always 16 bytes in length
+            bytes value = 1;
+            // declared scale of decimal literal
+            int32 scale = 2;
```
jacques-n, line 67: Minor nit here: do you mind putting precision first? We typically list it that way (and it helps me remember which is which, which I always forget).

jacques-n (approved):
Much obliged on the reorder!

PR #87, opened 2021-11-25
Introduce Capabilities message type

jacques-n:
- Add a capabilities proto message and documentation description


cpcloud (approved):
I think this is a great start. I think we should explore a few more use cases here and see if there's additional information we can provide here to enable "consumer-driven" plans in cases where a producer may not have access to a type system.

PR #88, opened 2021-11-25
Minor expression cleanups

jacques-n:
- Add java package name for all proto files
- Add an explicit CAST expression
- Update the varchar literal to specify it's length
- Move AggregateFunction, AggregationPhase and SortField out of the Expression message (top-level in expression.proto).
- Remove AggregationPhase from aggregate rel since it is specified on the individual aggregation function bindings
- Add support for a filter as part of an aggregation measure (SQL2003)
- Cleanup some markdown linebreaks
- Add ReturnProgram to derivation expression to support SSA return definition (reduce repetition)

### DIFF
--- a/binary/expression.proto
+++ b/binary/expression.proto
@@ -6,6 +6,7 @@ import "type.proto";
 import "selection.proto";
 
 option java_multiple_files = true;
+option java_package = "io.substrait.proto";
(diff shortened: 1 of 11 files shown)


cpcloud (approved):
LGTM

PR #95, opened 2021-11-29
Move NamedStruct out of the Type message

### LINKED ISSUE #90: Style: rename Type.NamedStruct to toplevel Schema?
Currently `Type.NamedStruct` is not part of type system and is only being used to represent schemas in read relations. I think we should move this out and rename it to `Schema` until there's another concrete, documented use case.

cpcloud:
Closes #90

### DIFF
--- a/binary/relations.proto
+++ b/binary/relations.proto
@@ -48,7 +48,7 @@ message RelCommon {
 
 message ReadRel {
   RelCommon common = 1;
-  Type.NamedStruct base_schema = 2;
+  NamedStruct base_schema = 2;
   Expression filter = 3;
   MaskExpression projection = 4;
   io.substrait.extensions.AdvancedExtension advanced_extension = 10;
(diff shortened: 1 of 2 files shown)


jacques-n:
The more I think about this, the more I'm torn about whether it makes sense for us to use the term schema. What do you think about NamedSchema or something like that? Saying that a struct type isn't a schema (which is what naming this implies) feels a bit confusing/misleading. People will then start complaining that expressions don't have output schema when they do, they just have positional schema as opposed to name based schema...

cpcloud:
Yeah, I think am anchoring on "schema" because it's what's use to describe the column name/type pairs of a table in SQL.

I'm happy to keep `NamedStruct` for now in light of your thoughts there.

jacques-n (approved):

gabry-lab:
hey, @cpcloud ,how to represent the type of an expressions? if it's really output a schema.
For example, an UDF [from_json](https://spark.apache.org/docs/latest/sql-ref-functions-builtin.html#json-functions), it output an named struct.

PR #97, opened 2021-11-29
Document date literal units and move to int64/millis since epoch

### LINKED ISSUE #93: Format: clarify the units of date literals
`date` literals [are defined as `fixed32`](https://github.com/substrait-io/substrait/blob/main/binary/expression.proto#L47) values, but no information about what the units of those values are is documented.

A 32-bit integer doesn't allow for more than about 9 decimal digits, so seconds0since-epoch seems like the only possible interpretation of units here if my arithmetic is correct.

I guess this raises another question of whether we should move to `int64` and use milliseconds-since-epoch?

cpcloud:
Closes #93


binary/expression.proto
```diff
@@ -44,7 +44,8 @@ message Expression {
       string string = 12;
       bytes binary = 13;
       fixed64 timestamp = 14;
-      fixed32 date = 16;
+      // Date in units of milliseconds since the UNIX epoch.
+      int64 date = 16;
```
jacques-n, line 48: if we use microseconds everywhere else (i believe we do), i think we should use them here as well.

Do you want to take a pass on all of these for comments to make sure we're consistent?

Also, we should rationalize whether we should use fixed64 for this as well (or switch timestamp other to int64)

cpcloud: > if we use microseconds everywhere else (i believe we do), i think we should use them here as well.

Makes sense.

> Do you want to take a pass on all of these for comments to make sure we're consistent?

Yup, will do.

> Also, we should rationalize whether we should use fixed64 for this as well (or switch timestamp other to int64)

Will comment on this in a bit.

cpcloud: @jacques-n Are these literals allowed to be negative? Currently they aren't by virtue of `fixed32`/`fixed64` being unsigned.

jacques-n: Oversight. I guess I wanted `sfixed64`

jacques-n: But it may be just a stupid optimization that confuses people. Probably makes sense to just go with int64 even if it will typically be a large number (thus the inclination towards a fixed width value)

cpcloud: Ok, I'll revamp the temporal things. Stand by.

cpcloud: Revamped.

jacques-n:
Can you update timestamp_tz literal as well?

cpcloud:
@jacques-n Done

jacques-n (approved):

PR #98, opened 2021-11-29
Deduplicate Switch/If expressions according to the intended use case

### LINKED ISSUE #91: Question: why are IfThen and SwitchExpression messages near exact copies of each other?
What is the use case for having `IfThen` and `SwitchExpression` messages that are effectively exact copies of each other?

```protobuf
  message IfThen {
    repeated IfClause ifs = 1;
    Expression else = 2;

    message IfClause {
      Expression if = 1;
      Expression then = 2;
    }
  }

  message SwitchExpression {
    repeated IfValue ifs = 1;
    Expression else = 2;

    message IfValue {
      Expression if = 1;
      Expression then = 2;
    }
  }
```

I understand the difference between the two different types of `CASE` expressions in SQL.

Can't we represent `SwitchExpression` in terms of an `IfThen` with an equality condition?

If we still need to distinguish between the two ... cases (pun intended), can we refactor this to use an `enum` for differentiating instead of two effectively identical constructs?

cpcloud:
Closes #91

### DIFF
--- a/binary/expression.proto
+++ b/binary/expression.proto
@@ -15,8 +15,8 @@ message Expression {
     FieldReference selection = 2;
     ScalarFunction scalar_function = 3;
     WindowFunction window_function = 5;
-    CasePredicateExpression case_predicate = 6;
-    CaseValueExpression case_value = 7;
+    IfThen if_then = 6;
+    SwitchExpression switch_expression = 7;
     SingularOrList singular_or_list = 8;
     MultiOrList multi_or_list = 9;
     Enum enum = 10;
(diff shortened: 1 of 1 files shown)


binary/expression.proto
```diff
@@ -148,32 +148,31 @@ message Expression {
     }
   }
 
-  message IfThen {
-
-    repeated IfClause ifs = 1;
+  message CasePredicateExpression {
```
jacques-n, line 151: Can we keep the IfThen names as opposed to changing this? I think it does a better job of clarifying distinction from the other type. It also avoids being overly sql centric for a pretty standard programming concept.

cpcloud: That's fair. I can keep the original names.

jacques-n: I think that CaseValue (or SwitchValue) are an improvement over simply "switch". Just felt that ifthen was probably clearer to most people.

cpcloud: Done

jacques-n (approved):

Issue #101, opened 2021-12-03
Save the Partition Infos in ReadRel


rui-mo:
It would be very useful to save the Partition Infos (including the partition index, file path, start and length) in ReadRel.

PR #102, opened 2021-12-03
Add Partition Infos in ReadRel

rui-mo:
This pr adds the Partition Infos in ReadRel.
Closes https://github.com/substrait-io/substrait/issues/101.

### DIFF
--- a/proto/substrait/relations.proto
+++ b/proto/substrait/relations.proto
@@ -86,6 +86,15 @@ message ReadRel {
 
       FileFormat format = 5;
 
+      // the index of the partition this item belongs to
+      uint64 partition_index = 6;
+
+      // the start position in byte to read from this item
+      uint64 start = 7;
+      
+      // the length in byte to read from this item
+      uint64 length = 8;
+
       enum FileFormat {
         FILE_FORMAT_UNSPECIFIED = 0;
         FILE_FORMAT_PARQUET = 1;


binary/relations.proto
```diff
@@ -86,6 +86,15 @@ message ReadRel {
 
       Format format = 5;
 
+      // the index of the partition this item belongs to
+      uint64 partition_index = 6;
+
+      // the start position in byte to read from this item
+      uint64 start = 7;
```
westonpace, line 93: Can you give a bit more detail what file format you are thinking of?  For example, given a parquet file, I don't think I can apply an arbitrary byte offset.  Instead I would normally expect to see a row group number of some kind.

rui-mo: Thanks for your comment, and we are thinking of Parquet file. Actually to read which row group can be determined by the internal logic according the start and length. For example, in Spark, if the midpoint of a row group locates in the region [start, start + length), then this row group will be be read. In Arrow, we also added that kind of logic. Therefore, the start and length would be very useful for determining the row groups to read.

jacques-n:
Let's merge this with the expectation that we're probably going to need to continue iterate over time as more people are working with readers.

jacques-n (approved):

binary/relations.proto
westonpace: If the start position was in the middle of a row group then would that row group be returned?

If the row group is returned should it be sliced as closely as possible to the start position?

If it is sliced what happens if the start position is in the middle of a value?  Or in the middle of a compressed buffer?

rui-mo: - For Q1: In current implementation, if the midpoint of the current rowgroup is not in the range of [start, start + length), this rowgroup will not be read by this partition. But because all the partitions must include the midpoints of all the rowgroups, this rowgroup will be read by other partition, and will not be missed.
- For Q2&3: It will not be sliced, because even if it is not read by current partition, it will be read by other partition.

westonpace: Ok, that makes sense.  The reader will still read entire row groups but will only read those row groups whose midpoints are contained in the range.  Thank you for the explanation.

jacques-n: Can one of you propose some additional comments and/or information in the spec docs to make sure this is clear to future users?

westonpace: I took a stab at it: #109

proto/substrait/relations.proto
```diff
@@ -86,6 +86,15 @@ message ReadRel {
...
       FileFormat format = 5;
 
+      // the index of the partition this item belongs to
+      uint64 partition_index = 6;
+
+      // the start position in byte to read from this item
+      uint64 start = 7;
+      
+      // the length in byte to read from this item
+      uint64 length = 8;
```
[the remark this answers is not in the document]
rui-mo: Could we set these fields to be optional to avoid this issue?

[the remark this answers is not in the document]
cpcloud: Proto fields are already optional. `proto3` support for explicit optional is pretty nascent and it's not likely that many systems are prepared to handle `optional` in `proto3` yet. There are some very important details around the meaning of field presence here https://github.com/protocolbuffers/protobuf/blob/v3.12.0/docs/field_presence.md that we'd need to be comfortable with before introducing this IMO.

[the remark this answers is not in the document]
westonpace: If the user doesn't specify slicing parameters then we assume they are reading the whole file.  The defaults match up with this nicely.

`start` - A default of 0 is what you would put in if you were explicitly reading the whole file anyways.
`length` - This is a more interesting case but we are in luck because 0 is an invalid length so we just need to document that 0 means "read the whole file"
`partition_index` - If they are going to read the whole file then calling it partition index 0 of 0 seems fine to me.  Nothing needed here.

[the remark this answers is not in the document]
cpcloud: Sounds great.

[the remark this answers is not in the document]
rui-mo: Thanks, LGTM!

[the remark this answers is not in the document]
westonpace: By the way, I did add default behavior explanation for `length` to the proto in https://github.com/substrait-io/substrait/pull/109/files#diff-cc42d548188fb7975a0eae621be239983723002ff219559a7876112d4fa78894R101 (though it seems that PR needs a rebase)

PR #103, opened 2021-12-04
Move to buf and address buf lints; add ci checks with buf

cpcloud:
This PR moves validation to `buf` in CI and addresses buf's lint complaints

Major changes:

- Add enum value prefixes: https://docs.buf.build/configuration/v1beta1/lint-rules#enum_value_prefix
- Make the package and directory structures match: https://docs.buf.build/configuration/v1beta1/lint-rules#file_layout


jacques-n:
Haven't looked through all of this but I  don't think you should have renamed the root directory to substrait. If you want to change binary to proto or something I think that would be okay. If that means pushing things another directory down to behave as buf wants wrt to directory to package equivalence, lets do that.

cpcloud:
What's wrong with the directory name being substrait?

jacques-n:
> What's wrong with the directory name being substrait?

A few reasons:
* It isn't descriptive (not easy to guess where the binary or protobuf representation of substrait is?) 
* It implies that the rest of the repository isn't substrait. (e.g., the schema for the yaml file)
* It is hard to imagine where we would put the similar types of artifacts for the human readable representation (once we complete it). 

But mostly, I just find it confusing.

jacques-n:
That enum prefixing is disgusting. Sigh. Guess we just have to live with it to be good cross-language lovers.

Other than the root directory name, this looks great. Thanks for working through the change @cpcloud

cpcloud:
@jacques-n I've moved files under `proto`. Let me know if any other changes are needed.

I also don't really like the enum prefixing, but C++ scoping rules are what we're stuck with unfortunately :(

jacques-n (approved):
LGTM, thanks!

jacques-n:
Random question @cpcloud , does buf require a network to run? I hope not.

I ask because I'm trying to understand exactly what the buf.lock file uses as reference and how it is maintained.

cpcloud:
> Random question @cpcloud , does buf require a network to run? I hope not.

AFAIK most operations do not require network access to run.

The primary subcommands that do are:

1. `buf push`: push a buf module to a registry
2. `buf registry`: logging into/out of a buf registry
3. `buf mod update`: update the `buf.lock` manifest

> I ask because I'm trying to understand exactly what the buf.lock file uses as reference and how it is maintained. I guess it could use git (which would be cool).

`buf.lock` records specific information about dependencies (https://docs.buf.build/configuration/v1/buf-lock).

We currently don't have any dependencies in substrait (the Google protos might be considered one, I'm not yet sure if that should be explicit) so the file is empty except for the version number of the substrait module

cpcloud:
Lock file updates contacting the internet is pretty standard behavior for any dependency management tool these days. There's no requirement for anything in the substrait package to use the network at all right now.

jacques-n:
Use for dependencies doesn't worry me. I was trying to understand how fingerprinting and checking semantic versioning works.

cpcloud:
If by checking semantic versioning you mean `buf breaking`, then I don't know. However, the documentation [here](https://docs.buf.build/breaking/overview) suggests that it is apply a large set of rules to detect the breakages.

Fingerprinting I'm not sure about.

Is there a specific situation in which you're worried about network usage? Might help to assist in targeting searching/code spelunking.

jacques-n:
Sorry I'm not be clearing. It looks like I'm referencing `buf breaking`. 

My questions were how it was trying to determine the basis to compare the version against. Initially I was concerned it was storing the proto in a registry and comparing it there. I now have actually gone and looked at the docs briefly and it seems like you give it a url or git location for the basis of determining breaking changes. That makes sense. I had assumed this patch also did this behavior and was trying to understand how it implicitly knew the basis (and that was the concern I had around remote calls). I now realize that this patch doesn't actually do this. :D

https://docs.buf.build/tour/detect-breaking-changes#run-buf-breaking

cpcloud:
IMO I don't think we should start using `buf breaking` in CI without a larger community discussion around what 1.0 means for the project. Looking forward to that day!

PR #104, opened 2021-12-06
refactor: change field name for List value type

gforsyth:
Proposing either the changes in this PR or alternatively changing
`List.values` to `List.elements` so that the `*_type` argument is
consistent with the field name.


jacques-n (approved):
Makes sense, thanks!

PR #105, opened 2021-12-06
Propose alternative for empty map/list literals to reduce comment reliance

jacques-n:


jacques-n:
@cpcloud and @gforsyth , would love your thoughts on this alternative for empty map/list. It feels like it relies less on comments to interpret things (and semantic validation beyond proto)

cpcloud (approved):
This much better than what I did.

gforsyth (approved):
yep, this looks good to me

PR #106, opened 2021-12-07
Date literal typo

bkietz:


jacques-n:
@cpcloud and I had discussed making this microseconds instead of milliseconds since it seemed best to keep consistent with everything else. It may be that we need to update the spec docs?

bkietz:
As a note, arrow has only the Date64 type which is millis since epoch. If we're keeping micros and updating the spec docs instead, we should also remove Date64 from the "equivalent arrow type" column there.

Related: the interval types don't correspond to any arrow type either

bkietz:
FWIW, I'm not aware of any systems which identify dates with microseconds. AFAICT:

Milliseconds are used by arrow::date64 and [java.sql.Date](https://docs.oracle.com/javase/7/docs/api/java/sql/Date.html) (which I _think_ was the inspiration for arrow::date64)

Whole days are used by arrow::date32, [velox](https://github.com/facebookincubator/velox/blob/main/velox/type/Date.h#L77), [impala](https://github.com/apache/impala/blob/b28da054f3595bb92873433211438306fc22fbc7/be/src/runtime/date-value.h#L38) and [iceberg](https://github.com/Netflix/iceberg/blob/5684ac1b77851737ed316e7f9d302f902b3e1e67/api/src/main/java/com/netflix/iceberg/transforms/Dates.java#L46-L52)

cpcloud:
Parquet also uses whole days https://github.com/apache/parquet-format/blob/master/LogicalTypes.md#date

bkietz:
I'll refactor this PR to shift to whole days for substrait::date as well, to see what that would look like

proto/substrait/expression.proto
```diff
@@ -18,7 +18,8 @@ message Expression {
     SwitchExpression switch_expression = 7;
     SingularOrList singular_or_list = 8;
     MultiOrList multi_or_list = 9;
-    Enum enum = 10;
+    Enum enum
+    = 10;
```
cpcloud, line 22: This seems ... off.

cpcloud (changes requested):
Can you split the formatting changes out into a separate PR?

jacques-n:
@bkietz , do you think you'll be able to split out the literal change from formatting changes?

What do people think about the update to int32 days?

proto/substrait/expression.proto
bkietz: Whoops, I left clang-format on write. I do think it'd be beneficial to check in a clang-format configuration to substrait, though

jacques-n:
@cpcloud, you happy with the switch to days for literal?

cpcloud:
Fine by me

jacques-n (approved):
Thanks @bkietz !

PR #107, opened 2021-12-07
Name _reference fields consistently

bkietz:
(instead of using _pointer sometimes)


bkietz:
related: Type.user_defined_type_reference is slightly inconsistent with similar fields

jacques-n (approved):
Thanks for catching and fixing @bkietz

PR #112, opened 2021-12-08
Enforce editorconfig settings in CI

cpcloud:


jacques-n (approved):
LGTM. We should probably apply to site but that can come later :)

Issue #115, opened 2021-12-20
Move the Partition Index out of the message 'FileOrFiles'


zzcclp:
It's better to move the Partition Index out of the message 'FileOrFiles', in general, one partition will include some files to be operated. For example, in Spark, one 'FilePartition' will include one partition index and some 'PartitionedFile', one 'PartitionedFile' indicates one file info to be scanned.

![image](https://user-images.githubusercontent.com/9430290/146727670-d21aad03-a239-44e1-8f43-129bc1c31c83.png)

westonpace:
Can this issue be closed (since the associated PR was closed)?

zzcclp:
> Can this issue be closed (since the associated PR was closed)?

Closed.

PR #116, opened 2021-12-20
Move the Partition Index out of the message 'FileOrFiles'

zzcclp:
Move the Partition Index out of the message 'FileOrFiles'.
Refer to #115 .


zzcclp:
@jacques-n @rui-mo please help to review this pr, thanks.

rui-mo:
@zzcclp According to our previous communication, the case of files belonging to different Partitions should also be considered. So each FileOrFiles has a partition_index here.

PR #120, opened 2022-01-06
Introduce a Java module

### LINKED ISSUE #21: Implement Calcite native library that parses sql and returns a binary substrait representation
The idea here is to provide a reasonable way for people to give users immediate access to the high quality SQL parsing of Calcite with minimal effort. We'd use GraalVM for AOT compilation and start with a fairly simple function similar to `substrait parse(string)`

It would be nice if part of this effort was to expose this library with a command line tool that could be piped to other future tools. (For example, create an additional cli that will take a plan and return the results with Datafusion.)

A big question is what catalog to expose in an example cli. Some ideas:
- Require a user to provide --table <alias>=<path> for Parquet files (similar to how local paths are explored in Docker)
- Pass in table declarations --table t1=(int c1, int c2) and treat the read objects as opaque (for later binding use by an execution system)?
- Other ideas?

A second fun thing to add would be a separate library that is plan in and out and applies a list of optimization rules using one of the existing Calcite optimizers. Lower priority than the sql parser initially but could be intersting to evaluate different optimization patterns and start exposing nice Calcite interfaces for things like python/rust/etc.

jacques-n:
- Create a new set of java modules
- Add a core module that provides java idomatic apis for rels, types, expressions, extensions, etc.
- Add a antlr grammar that parses types expressions
- Create a new isthmus module that converts sql queries into substrait plans.
- Add a isthmus command line application that enables substrait conversion.
- Add PR builds around java build and native library builds across windows, mac and linux.

This closes #21


jacques-n:
Bugs reported:

- Grouping expressions seem to be duplicated.
- Decimal value is empty when it shouldn't be.

yma11:
Hi @jacques-n, as synced, I met 2 issues about json file and deserialization when using this tool.
1) Duplicated groupingExpression.
`          // select l_orderkey, sum(l_orderkey+l_partkey) from lineitem where l_orderkey >= 100 group by l_orderkey
          "groupings": [{
            "groupingExpressions": [{
              "selection": {
                "directReference": {
                  "structField": {
                    "field": 0
                  }
                },
                "rootReference": {
                }
              }
            }]
          }, {
            "groupingExpressions": [{
              "selection": {
                "directReference": {
                  "structField": {
                    "field": 0
                  }
                },
                "rootReference": {
                }
              }
            }]
          }],`
2) Decimal value is empty after deserialization. I am trying to get the value using `s_literal_expr.decimal().value()`.

java/isthmus/build.gradle.kts
```diff
@@ -0,0 +1,53 @@
+plugins {
+    id("java")
+    id("idea")
+    id("com.palantir.graal") version "0.10.0"
+}
+
+java {
+    toolchain {
+        languageVersion.set(JavaLanguageVersion.of(17))
```
adamkennedy, line 9: This not only requires a non-LTS version of Java, but also requires the use of preview features from Java 17 (IDEA throws an error by default unless you turn on preview features).

Should this be moved to a LTS version? Either 11 for broad compatibility or (just released this week) 18?

adamkennedy (changes requested):
Noted the challenging Java version this uses

java/isthmus/build.gradle.kts
jacques-n: FYI, Java 17 is a lts release.

I agree that 17 is an aggressive choice (but it's also a new project).

We are actually using some of the pattern matching that is preview in 17. We could take it out.

jacques-n:
Hey @yma11 
> 1. Duplicated groupingExpression.
>    `          // select l_orderkey, sum(l_orderkey+l_partkey) from lineitem where l_orderkey >= 100 group by l_orderkey "groupings": [{ "groupingExpressions": [{ "selection": { "directReference": { "structField": { "field": 0 } }, "rootReference": { } } }] }, { "groupingExpressions": [{ "selection": { "directReference": { "structField": { "field": 0 } }, "rootReference": { } } }] }],`

I believe I have fixed this in the update I just committed.

> 2. Decimal value is empty after deserialization. I am trying to get the value using `s_literal_expr.decimal().value()`.

I'm not seeing this problem. I have these two tests. Which one fails to uncover what you're seeing?

1. [Proto/Pojo Conversion](https://github.com/substrait-io/substrait/blob/d6e40d2641571cf28b5ceda7bd6091c1ba850281/java/core/src/test/java/io/substrait/type/proto/LiteralRoundtripTest.java#L16)
1. [Calcite/Pojo Conversion](https://github.com/substrait-io/substrait/blob/d6e40d2641571cf28b5ceda7bd6091c1ba850281/java/isthmus/src/test/java/io/substrait/isthmus/CalciteLiteralTest.java#L123)

jacques-n:
Resolved as part of #147

yma11:
Thanks, @jacques-n.
> I'm not seeing this problem. I have these two tests. Which one fails to uncover what you're seeing?
> 
> 1. [Proto/Pojo Conversion](https://github.com/substrait-io/substrait/blob/d6e40d2641571cf28b5ceda7bd6091c1ba850281/java/core/src/test/java/io/substrait/type/proto/LiteralRoundtripTest.java#L16)
> 2. [Calcite/Pojo Conversion](https://github.com/substrait-io/substrait/blob/d6e40d2641571cf28b5ceda7bd6091c1ba850281/java/isthmus/src/test/java/io/substrait/isthmus/CalciteLiteralTest.java#L123)

I looked at the ProtoExpressionConverter. We have no big difference when trying to get value of decimal literal: `literal.getDecimal().getValue()` vs `s_literal_expr.decimal().value()`.  What's different is, in my use case, I firstly serialize  substrait plan to json file and then deserialize the json file to substrait plan, the literal expression in json format is:
`
                        "literal": {
                          "decimal": {
                            "value": "AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAg=",
                            "precision": 2,
                            "scale": 1
                          },
                          "nullable": false
                        }
`
and `s_literal_expr.decimal().value()` returns a string instead of decimal value.
`inline const ::std::string& Expression_Literal_Decimal::value() const {
  // @@protoc_insertion_point(field_get:substrait.Expression.Literal.Decimal.value)
  return value_.GetNoArena();
}`
Could this be the cause of empty value?

jacques-n:
Closing this as it has been moved to https://github.com/substrait-io/substrait-java

PR #130, opened 2022-01-24
Fix typo in github actions workflow

jvanstraten:
That path doesn't look right to me. Though it looks like the checker was just falling back to checking all files before, so I guess it was doing its job despite the typo.


cpcloud (approved):

cpcloud:
Thanks @jvanstraten!

PR #132, opened 2022-01-28
Add CrossProduct Relation and Single-join type

pdet:
These are necessary to run some of the flattened subqueries.

### DIFF
--- a/proto/substrait/relations.proto
+++ b/proto/substrait/relations.proto
@@ -127,7 +127,16 @@ message JoinRel {
     JOIN_TYPE_RIGHT = 4;
     JOIN_TYPE_SEMI = 5;
     JOIN_TYPE_ANTI = 6;
+    JOIN_TYPE_SINGLE=7;
   }
+
+  substrait.extensions.AdvancedExtension advanced_extension = 10;
+}
+
+message CrossRel {
+  RelCommon common = 1;
+  Rel left = 2;
+  Rel right = 3;
 
   substrait.extensions.AdvancedExtension advanced_extension = 10;
 }
@@ -234,5 +243,6 @@ message Rel {
     ExtensionSingleRel extension_single = 9;
     ExtensionMultiRel extension_multi = 10;
     ExtensionLeafRel extension_leaf = 11;
+    CrossRel cross = 12;
   }
 }
--- a/site/docs/relations/logical_relations.md
+++ b/site/docs/relations/logical_relations.md
@@ -156,7 +156,7 @@ The join operation will combine two separate inputs into a single output, based 
 | Right | Return all records from the right input. For each cross input match, return a record including the data from both sides. For any remaining non-matching records from the right input, return the left record along with nulls for the right input. |
 | Semi | Returns records from the left input. These are returned only if the records have a join partner on the right side. |
 | Anti  | Return records from the left input. These are returned only if the records do not have a join partner on the right side. |
-| Single | Returns at most one join partner per entry on the left input. In case no partner is found, NULL is returned. |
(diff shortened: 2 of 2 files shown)


jacques-n:
Can you propose updates to the spec docs as well to match these changes?

Also, what is a single-join?

pdet:
Yes, I'll update the spec docs in a bit

A single join is like a LEFT OUTER JOIN, but returns at most one join partner per entry on the left side (and NULL if no partner is found).
It is created as a result of the subquery flattening and can either be the join type of a regular join or the join type of a Delim Join.

I think after this I only need to add support to the Mark and Delim Join to get subqueries working :-).

proto/substrait/relations.proto
```diff
@@ -127,11 +127,20 @@ message JoinRel {
     JOIN_TYPE_RIGHT = 4;
     JOIN_TYPE_SEMI = 5;
     JOIN_TYPE_ANTI = 6;
+    JOIN_TYPE_SINGLE=7;
   }
 
   substrait.extensions.AdvancedExtension advanced_extension = 10;
 }
 
+message CrossRel {
```
cpcloud, line 136: Does this need to be a separate relation from `Join` or can it be another join variant? No strong opinion on that from me, just curious if there's a reason for it to be a distinct relation.

pdet: If there is no requirement of a join expression, this could be a cross_join type. I'm also fine with that.

cpcloud: I see, maybe it makes sense to keep this as you have it then, so that a plan must explicitly be encoded with cross join and we don't have to deal with cross joins with filters.

site/docs/relations/logical_relations.md
```diff
@@ -136,8 +154,9 @@ The join operation will combine two separate inputs into a single output, based
 | Outer | Return all records from both the left and right inputs. For each cross input match, return a record including the data from both sides. For any remaining non-match records, return the record from the corresponding input along with nulls for the opposite input. |
 | Left  | Return all records from the left input. For each cross input match, return a record including the data from both sides. For any remaining non-matching records from the left input, return the left record along with nulls for the right input. |
 | Right | Return all records from the right input. For each cross input match, return a record including the data from both sides. For any remaining non-matching records from the right input, return the left record along with nulls for the right input. |
-
-
+| Semi | Returns records from the left input. These are returned only if the records have a join partner on the right side. |
+| Anti  | Return records from the left input. These are returned only if the records do not have a join partner on the right side. |
+| Single | Returns at most one join partner per entry on the left input. In case no partner is found, NULL is returned. |
```
jacques-n, line 159: Can you add more insight into what happens if there is more than one partner (e.g. operation can use arbitrarily among partners). Also, can you clarify the null statement. I assume you always return the left side as in a left outer join. I believe you are saying here that null is returned for the right fields when no partner is identified.

proto/substrait/relations.proto
jacques-n: I'm fine either way on this. I think there are some nice properties to having crossrel separate but it will probably look very similar to join ultimately.

site/docs/relations/logical_relations.md
pdet: Hey Jacques, I've updated the text a bit.

Just as an extra clarification: 
The single join is defined in this paper http://btw2017.informatik.uni-stuttgart.de/slidesandpapers/F1-10-37/paper_web.pdf
In section 3.2.

One thing that might differ from system to system, is the execution on a single join when more than a join partner is found per tuple.

DuckDB follows the SQLite semantics for subqueries, which means that in the case of >1 result, we just return the first result we find, rather than throwing an error. But as you can see in sec 3.2, an error can be thrown as well. Hence I've also decided to add it there.

jacques-n (approved):
Looks good, thanks for the updates!

PR #133, opened 2022-02-01
Added some comments to the .proto file explaining the different path types in a LocalFiles message

westonpace:


proto/substrait/algebra.proto
```diff
@@ -77,9 +77,16 @@ message ReadRel {
 
     message FileOrFiles {
       oneof path_type {
+        // A URI that can refer to either a single folder or a single file
         string uri_path = 1;
+        // A URI where the path portion is a glob expression that can
+        // identify zero or more paths.
+        // Consumers should support the POSIX syntax.  The recursive
+        // globstar (**) may not be supported.
```
cpcloud, line 85: Is there any reason why recursive globs wouldn't be supported? Globs are usually implemented far below substrait (in glibc for example) so why would substrait have anything to say about specific glob constructs?

westonpace: This came from Slack where @jacques-n suggested keeping it simple.  Glob support would require additional work in C++ but not much.  Mainly I want to call out specific syntax to make it clear what the semantics of the operation would be.  Regardless of what tools consumers use to implement this feature its important they all agree on the meaning.  There are many tools that support globs but do not support recursive globs and so it helps to be specific.

jacques-n (approved):
LGTM. Merging

PR #134, opened 2022-02-02
feat: add subquery representation

### LINKED ISSUE #119: Introduce the ability to express a correlated subquery
We should introduce a way to express a correlated subquery. We need to come up with the best way to represent this. Some prior art:

- [Calcite](https://github.com/apache/calcite/blob/master/core/src/main/java/org/apache/calcite/rex/RexSubQuery.java) - A filter relation that defines a dynamic variable of it's data that is used in a new expression type that contains a subtree. The subquery subtree references the fields of this dynamic variable to connect the sub and main trees. (`[dynamic variable].[field name]` e.g. `$cor0.P_PARTKEY`)
- [Trino](https://github.com/trinodb/trino/blob/master/core/trino-main/src/main/java/io/trino/sql/planner/plan/ApplyNode.java) - A relational node type that has a special set symbols from outer tree that are then referenced via special expression assignments from the inner tree. (I think, just a quick review of the code.)
- [Spark](https://github.com/apache/spark/blob/0a6be8cf59d5f9978337ec39953f0f7a88aec636/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/subquery.scala#L72) Looks like an expression that contains a subtree, a set of outer expressions and an exprId (that I believe is used inside the inner subtree). This looks/feels a bit like Calcite although Calcite also has the agg functions & groupings.

Other examples people think that should provide inspiration?

cpcloud:
This PR adds support for correlated subqueries to substrait.

Notes:

1. It'll be easier to review commit by commit, since I had to combine
   expressions and relations together to get the protos to compile.
2. While there are actually only three kinds of subqueries (scalar, `EXISTS` and `ANY`),
   we chose to keep a few of the higher level representations so that information about
   the original intention is preserved.

Closes #119.

cc @pdet @Mytherin, would love your thoughts here!

### DIFF
--- a/proto/substrait/algebra.proto
+++ b/proto/substrait/algebra.proto
@@ -595,24 +595,23 @@ message Expression {
       SetComparison set_comparison = 4;
     }
 
-    // A subquery with one row and one column, often an aggregate though not
-    // required to be.
+    // A subquery with one row and one column. This is often an aggregate
+    // though not required to be.
     message Scalar {
       Rel input = 1;
     }
 
-    enum SetPredicateOp {
-      SET_PREDICATE_OP_UNSPECIFIED = 0;
-      SET_PREDICATE_OP_EXISTS = 1;
-      SET_PREDICATE_OP_UNIQUE = 2;
-    }
-
     // A predicate over a set of rows in the form of a subquery
-    /// EXISTS and UNIQUE are common SQL forms of this operation
+    // EXISTS and UNIQUE are common SQL forms of this operation.
     message SetPredicate {
+      enum PredicateOp {
+        PREDICATE_OP_UNSPECIFIED = 0;
+        PREDICATE_OP_EXISTS = 1;
+        PREDICATE_OP_UNIQUE = 2;
+      }
       // TODO: should allow expressions
-      SetPredicateOp set_predicate_op = 1;
-      Rel tuple_set = 2;
+      PredicateOp predicate_op = 1;
+      Rel tuples = 2;
     }
 
     // Predicate checking that the left expression is contained in the right subquery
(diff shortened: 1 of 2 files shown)


proto/substrait/relations.proto
```diff
@@ -568,11 +568,86 @@ message Expression {
...
+      // This value must be >= 1
+      uint32 steps_out = 1;
+    }
+  }
 
+  message Subquery {
+    // Subquery relation expression
+    oneof subquery_type {
+      Scalar scalar = 1;
+      SetPredicate exists_predicate = 2;
```
jacques-n, line 591: the name of this union item doesn't match the type.

proto/substrait/relations.proto
```diff
@@ -246,3 +246,366 @@ message Rel {
     CrossRel cross = 12;
   }
 }
+
```
jacques-n, line 248: it probably makes sense to rename this file to something more apt to the content inside.

jacques-n (commented):
Generally looks good. Interested to see if others agree.

proto/substrait/relations.proto
cpcloud: Fixed.

proto/substrait/relations.proto
cpcloud: The best I came up with was `algebra.proto` 🤷🏻.

jacques-n:
This looks good to me but I'd like to get some more eyes on it.

jacques-n:
I lied... we need to also add the docs part of this.

Once people agree with the shape of this, let's try to update the appropriate parts of the expressions/etc sections.

Mytherin (dismissed):
Looks good to me as a way of representing unflattened subqueries. The only potential missing feature I see here is support for the `ARRAY`subquery, e.g. in Postgres:

```sql
select array(select * from generate_series(0,5,1));
--  {0,1,2,3,4,5}
```

cpcloud:
> Looks good to me as a way of representing unflattened subqueries. The only potential missing feature I see here is support for the `ARRAY`subquery, e.g. in Postgres:
> 
> ```sql
> select array(select * from generate_series(0,5,1));
> --  {0,1,2,3,4,5}
> ```

Thanks!

Yeah, functions that accept subqueries (or sets of rows) would be an interesting follow up.

This would likely also need work in the type system.

cpcloud:
I will add some documentation to the spec later today about subqueries.

jacques-n (approved):
Thanks for working through this!

PR #135, opened 2022-02-04
ci: proto-format-check CI job not working

jvanstraten:
I don't think the proto-format-check a.k.a. "Check Protobuf Style" CI task was doing anything, because jidicula/clang-format-action@v4.3.1 does [not](https://github.com/jidicula/clang-format-action/blob/v4.3.1/README.md?plain=1#L36-L52) check .proto files [yet](https://github.com/jidicula/clang-format-action/blob/v4.5.0/README.md?plain=1#L59-L60), and thus it simply wasn't matching any files. If I'm right CI should fail for the first commit here, since clang-format complains about lots of things for me locally.

### DIFF
--- a/.github/workflows/pr.yml
+++ b/.github/workflows/pr.yml
@@ -27,7 +27,7 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v2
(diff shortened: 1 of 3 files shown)


cpcloud (approved):
Yes! Thank you.

PR #139, opened 2022-02-07
Adding Alias to Projected Expression

pdet:
Extending the Projection relation to hold an alias for an expression.

I guess the alias can also be directly added to the expression message if desired.

### DIFF
--- a/proto/substrait/algebra.proto
+++ b/proto/substrait/algebra.proto
@@ -107,10 +107,15 @@ message ReadRel {
   }
 }
 
+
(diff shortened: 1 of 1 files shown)


jacques-n:
Hey @pdet, Substrait only uses field names at the edges of the plan (read, write, and plan root). Everything internal to the plan is ordinal based. So this doesn't seem like a change we should include. 

Maybe give an example of what you're trying accomplish that you don't think you can currently accomplish.

PR #121, opened 2022-01-06
Add DistributeRel Message

ZJie1:

PR #142, opened 2022-02-21
An updated Exchange rel.

jacques-n:
This work is an update based on discussion on #121 (I was unable to update to push these changes to that branch.)


jacques-n:
This is superseding #121

EpsilonPrime:
Given that ExchangeRel (and DdlRel and WriteRel) aren't part of Rel's oneof is it possible to use them in a plan or are they intended for use outside of a plan?

Issue #131, opened 2022-01-28
Plan validator/consumer reference implementation


jvanstraten:
tl;dr: requesting a tool that can give an authoritative answer to the question "does this plan conform to the Substrait spec?"

I've been studying the Substrait specification and protobuf files for a few days now, with the purpose of contributing more extensively to the consumer for [Arrow's query engine](https://github.com/apache/arrow/pull/12279). While I think I understand the basic ideas for most of it, I'm finding that the specification is not very precise about some of the nitty-gritty details. For example, is the `type_variation_reference` field of the various builtin type messages mandatory? If yes, what's the type variation of, for example, a literal (which doesn't define a reference)? Based on that I figure the answer is "no," but that means that `type_variation_reference == 0` (or possibly some other reserved value?) must mean "no variation," as protobuf just substitutes a default value for primitive types when a field isn't specified. Does that then also mean that anchor 0 is reserved for other such links? And so on. The most annoying part of these types of questions is that I don't know if I just glossed over some part of the spec. I don't like asking pedantic questions when I'm not 100% sure I'm not the problem myself.

The core issue here, in my opinion, is that there is no objective, authoritative way to determine for some given Substrait plan whether A) it is valid and B) what it means. The only way to do that right now is to interpret a plan by hand, which, well... is open to interpretation. Even if a spec is completely precise about everything, it's still very easy for a human to make mistakes here. If however there *was* a way to do that, I could just feed it some corner cases to resolve above question, or look through its source code.

[...]

PR #146, opened 2022-03-05
Add Rust bindings

andygrove:
This PR adds Rust bindings. The build script (`build.rs`) copies the `.proto` files into a local directory and then uses [prost](https://crates.io/crates/prost) to compile the `.proto` files and generate Rust source code.

I tested with `cargo publish --dry-run --allow-dirty` and it completed without error.


rust/README.md
```diff
@@ -0,0 +1,24 @@
...
+
+## Generate Documentation
+
+``` bash
+cargo doc --no-deps
+```
+
+## Publishing the crate
+
+We need to specify `--allow-dirty` when publishing because we copy the `.proto` files into the local directory but
```
jacques-n, line 19: Is there way we can do this without the copy? E.g. just give the relative path to the files (or symlink the files?)

jacques-n:
Can you add a github action that builds?

rust/README.md
andygrove: No, at least not that I know of.

Rust "crates" that get published to crates.io are essentially just tarballs with the file extension `.crate` that contains everything needed to compile from source, so we must include the `.proto` files in this crate tarball.

If we reference the proto files using a relative path that is outside of the `rust` directory then it will not make sense in the context of running against the extracted tarball.

One workaround would be to put the `Cargo.toml` in the root of the repository but that does not seem ideal.

If we had these bindings in a separate repo then we could use a git submodule to pull the main repo with the proto files. That is perhaps the cleanest option and we can use version control to make sure we are referencing a specific version of the proto files.

andygrove: I also tried with symlinks but it leads to the same issue where Cargo sees local files that are not checked in:

```
$ cargo publish --dry-run
    Updating crates.io index
error: 9 files in the working directory contain changes that were not yet committed into git:

build.rs
proto/substrait/function.proto
proto/substrait/capabilities.proto
proto/substrait/type.proto
proto/substrait/type_expressions.proto
proto/substrait/parameterized_types.proto
proto/substrait/algebra.proto
proto/substrait/plan.proto
proto/substrait/extensions/extensions.proto

to proceed despite this and include the uncommitted changes, pass the `--allow-dirty` flag
```

andygrove: I updated https://github.com/andygrove/substrait-rs to demonstrate the git submodule approach, which works well. I also added the GitHub action there.

jacques-n: Got it, thanks

jacques-n:
@cpcloud , can you do a review of this since you work in rust more than me?

@andygrove , I think you need to update to spdx asl2 license headers same as the proto files (not the big ones you're using).

houqp (approved):

jvanstraten:
Worth noting here that I already made Rust bindings a while ago for the validator that I've been working on for the past few weeks: https://github.com/jvanstraten/substrait/tree/validator (see also #131). It looks like our approach is pretty much the same, so hopefully I'll be able to use these bindings for the validator as well... but I'm not yet sure if that's possible, since I had to resort to some procedural macro magic on top of the files generated by prost for some validator-specific introspection logic for some self-checking features of the validator, which would probably be a bit ugly to expose to users as well.

Nevertheless, there are some quality of life things that I already did but I don't think are in this PR yet:

 - my [build.rs](https://github.com/jvanstraten/substrait/blob/2142bd6d48de5398ca6586df7b0ae6e9faa18588/validator/core/build.rs) automatically searches for the .proto files, so the list doesn't need to be hardcoded;
 - cargo clippy and cargo fmt checks in CI and pre-commit; and
 - CI on different operating systems (for what it's worth for what is mostly generated code, though I remember having some issues with prost/protoc's path handling on Windows specifically... can't hurt to check);

so it might be worth merging this functionality into this PR already, but I can also just do it later as part of my eventual PR for the validator.

rust/README.md
jvanstraten: @jacques-n So, what's the plan for this? Multiple repositories + submodules (presumably with all repositories moved to the substrait-io organization) or merging this PR?

I'm asking because I'm running into the same problems with the validator and its various language bindings now. My workarounds don't work 100% yet, and I've already had to resort to some *ugly* stuff.

Also, while trying to get things working right for the validator, I realized that the approach in this PR doesn't work 100% either:

 - `build.rs` is not run locally by `cargo package`/`cargo publish`, at least not until it tries to verify that the .crate file it built will compile out of context. For the crate to be packaged correctly, it's essential that `cargo build` is run first, or it will fail validation because the files don't exist at best (just run `cargo package` in a clean clone of the repo; it will fail), or publish with out-of-date .proto files that were copied earlier at worst.
 - I don't think you'd be able to use these bindings by specifying the git repository as the dependency, since `Cargo.toml` is not in the repository root. (I've already run into similar problems with other tools that assume that it lives there, so I actually already had to put a workspace `Cargo.toml` and the associated `Cargo.lock` file in the root. As more and more bindings will likely be added over time, I suspect this is going to create a lot of clutter.)
 - Highly unlikely, but if the `../proto` directory would ever happen to exist in a user's build environment when the .crate file is built as a dependency, `build.rs` might behave erratically. The published crate should in general never be probing outside of its build directory (except to look for system dependencies like protoc, I guess).

I don't think there *is* a good way to do this without resorting to submodules. The "best" solution I suppose would be to use `cargo xtask codegen` to do the copy rather than `build.rs`, but `cargo-xtask` is not a standard thing, and `cargo package` would still complain without `--allow-dirty`. `cargo` just doesn't like crates relying on context outside of the manifest directory, and with good reason IMO. Likewise for the Python bindings I have for the validator, and it probably applies to many other languages as well.

If we *do* go the multiple-repository route, I would suggest to move the proto and yaml files into a separate repository as well. That way, the main repository can have submodules for it *and* the language bindings (and validator), so a user can just get everything in one go if they want, and so we can have some centralized CI and release logic to help ensure everything stays in sync.

andygrove: > * I don't think you'd be able to use these bindings by specifying the git repository as the dependency, since `Cargo.toml` is not in the repository root.

That part actually does work. Cargo is smart enough to search the repo for the appropriate `Cargo.toml`. We are depending on this PR branch from [datafusion-substrait](https://github.com/datafusion-contrib/datafusion-substrait) for example. 

However, I agree with your other points and would also prefer so see these bindings in their own repo with a submodule to pull in the proto files.

jvanstraten: Since it doesn't look like there's going to be a resolution on this anytime soon, I had another look at the build logic for the validator. I ended up doing pretty much what you're doing now, also copying over the files in build.rs. I did figure out a way to resolve the third point I made, namely by adding a sort of marker file in the crate directory that's excluded in the package, in order to determine whether build.rs is being run from the git repository or from a downloaded crate file. If the file doesn't exist, it knows not to probe anything outside of its manifest directory. [Here](https://github.com/jvanstraten/substrait/blob/2e1c1172c3ff3d327ac255db56176989f6fa8deb/validator/rs/build.rs)'s my file for reference.

Good to know that Cargo itself doesn't require the manifest file to be at the root, by the way. It's smarter than I thought, I guess :) Unfortunately some of the third-party stuff I'm using isn't, I think, so I don't think I can easily move my workspace file to a nicer location in my soon-to-be PR.

jacques-n:
Hey @jvanstraten and @andygrove, i've created a [substrait-rs](https://github.com/substrait-io/substrait-rs) repository. Can you create a PR for that instead (using git submodule to point to the main repo)?

andygrove:
> Hey @jvanstraten and @andygrove, i've created a [substrait-rs](https://github.com/substrait-io/substrait-rs) repository. Can you create a PR for that instead (using git submodule to point to the main repo)?

Thanks @jacques-n I will create a PR in the next few hours

andygrove:
> > Hey @jvanstraten and @andygrove, i've created a [substrait-rs](https://github.com/substrait-io/substrait-rs) repository. Can you create a PR for that instead (using git submodule to point to the main repo)?
> 
> Thanks @jacques-n I will create a PR in the next few hours

Actually, @jvanstraten may have something further along. Let's wait to hear from them first.

jvanstraten:
@andygrove Are you still planning to move this PR to substrait-rs sometime? I can also do it, if you prefer.

andygrove:
@jvanstraten I have created a PR against the substrait-rs repo now - https://github.com/substrait-io/substrait-rs/pull/1 so will close the one

PR #147, opened 2022-03-07
Add some needed functions into yaml specification

rui-mo:
This pr added some needed functions into yaml specification.


jacques-n:
Resolved by #152

PR #149, opened 2022-03-10
Add Input Rel

rui-mo:
This pr added input rel to specify the action of fetching data from specific iterator generated by last computing.


jacques-n:
I don't have a clear enough sense of this operator to give feedback. It looks like it is a readrel with a specialized input type (an in memory iterator of some sort). In order for this to be useful, I'd expect us to also define writerel with a similar capability. This would allow us to stitch together discrete plan trees in a single plan message.

jacques-n:
(The close above was accidental)

jacques-n:
Two options here. Add another read_type or use the ExtensionTable for your own arbitrary table type.

jacques-n:
Closing this wil expectation that ReadRel will be used instead.

PR #150, opened 2022-03-10
Add output format in RelRoot

rui-mo:
This pr added output format in RelRoot. With this information, the computing's result can be converted into certain format according to the requirement.
One use case would be to specify which columnar format is expected. If conversion is needed, the format conversion will be conducted after the whole computing.


jacques-n:
This seems unnecessary. We already have a [type variation](https://substrait.io/types/type_variations/) capability for data types. Can't that be used instead?

rui-mo:
Thanks for the comment, will firstly try to use the existing specifications.

PR #153, opened 2022-03-27
Upgrade mkdocs so that jinja2 bug doesn't cause a problem.

jacques-n:
See: https://github.com/mkdocs/mkdocs/issues/2799

### DIFF
--- a/site/requirements.txt
+++ b/site/requirements.txt
@@ -1,4 +1,4 @@
-mkdocs==1.2.4
+mkdocs==1.2.2
 mkdocs-material==7.2.6
 mkdocs-minify-plugin==0.4.0
 mkdocs-redirects==1.0.3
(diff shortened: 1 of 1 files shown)


jacques-n:
@rymurr, it looks like your mkdocs-proto plugin is locked to a single mkdocs version :(

```
ERROR: Cannot install -r ./site/requirements.txt (line 2), -r ./site/requirements.txt (line 3), -r ./site/requirements.txt (line 4), -r ./site/requirements.txt (line 6), -r ./site/requirements.txt (line 7), -r ./site/requirements.txt (line 8) and mkdocs==1.2.4 because these package versions have conflicting dependencies.

The conflict is caused by:
    The user requested mkdocs==1.2.4
    ...
    mkdocs-protobuf 0.0.8 depends on mkdocs==1.2.2
```

PR #154, opened 2022-03-30
Update mkdocs-protobuf plugin

rymurr:
previously mkdocs-protobuf had pinned mkdocs. This patch updates the
mkdocs-protobuf plugin so that mkdocs is no longer pinned and removes
the workaround from #153


rymurr:
`mkdocs-awesome-pages` needed an update too to fix compatibility issues

jacques-n (approved):
LGTM, thanks!

PR #155, opened 2022-03-30
Plan validator/consumer reference implementation

### LINKED ISSUE #131: Plan validator/consumer reference implementation

jvanstraten:
Closes #131; see issue for details on what this is and why.

It's probably about time I start throwing this out here for some more public review; by most definitions this is probably well past the MVP stage. For a long time already I've been making sure that my commits pass CI, and CI is pretty extensive, so really, you should already be able to use the validator by just cloning my branch and following the build instructions. I'm kind of on the fence about whether to make this a draft PR and continuing development in this branch or whether I should just do a feature freeze in this branch and continue development in another; let me know what would make more sense. This PR is already pretty massive, after all.

~~I'll still be adding tests (and fixes for bugs uncovered by those tests), but~~ I don't intend to add any more functionality or do any more refactoring in this PR, so I've now marked this as ready for review.

A semi-structured braindump of what I've done so far and what is still to do follows.

Validation stuff supported thus far:
 - Errors for missing required fields, warnings for fields unknown to the validator, etc.
 - Linking up references and anchors.
 - Syntax-checking and optionally resolving YAML URIs.
 - Type propagation and checking.
 - Extension file schema validation via jsonschema.
 - Probably lots more I'm not thinking of right now; I pretty much just implemented validation for everything I came across, which is everything that's currently reachable from `substrait.Plan`.

Validation stuff NOT supported yet:
 - Anything to do with YAML extension file parsing beyond just schema validation. Most of the resolution logic for extensions is there; the extension definitions just aren't parsed yet. This involves stuff like type expressions, which is a whole new can of worms that I've yet to open. I'm planning to add support for that in a later PR.

Input:
 - The Rust crate only accepts binary protobuf serializations as input for the plan, because this is all that `prost` supports.
 - The Python library can also work with JSON, YAML, and [JSOM](https://github.com/saulpw/jsom) (the latter being a bit unstable because JSOM is still rather unstable, but it's being worked on).
 - Beyond the plan, the validator can be configured:
    - severity of certain classes of diagnostics can be clamped, so you can for example disable diagnostics you don't care about;
    - URIs and the resolution thereof can be overridden, for example when a YAML file isn't reachable or you want to cache them locally;
    - some misc. options related to dealing with unknown fields.

Output:
 - Validity verdict: valid, invalid, or unknown. The latter is automatically used for example when the validator encounters a field it doesn't know about yet, so the validator can lag behind the protobuf definitions to the extent that rustc allows it to (adding oneof or enum variants will probably fail it due to match statements missing branches, but you can usually just return `Ok(())` for new branches if you don't want to implement them yet, and the boilerplate logic will take care of the rest).
 - List of diagnostics: the validator doesn't just fail immediately when it sees a problem; it will keep going. Almost all problems beyond protobuf not being able to deserialize the input are recovered from intelligently.
 - The validator's interpretation of the plan as either an HTML page that you can click through or a corresponding protobuf meta-representation of the plan. The former is extremely useful for debugging not only the validator itself, but also Substrait plans in general, regardless of their validity. It'll give you a nice relation graph, for instance. In many cases it also annotates nodes with human-readable descriptions of what they represent, in addition to the raw protobuf messages and primitives.

Supported usage methods:
 - As a pure-Rust crate.
 - As a Rust crate with native dependency on libcurl for YAML resolution.
 - As a Python module using urllib for YAML resolution.
 - As a pure-C library for any language with C FFI capability, built using CMake.
 - As a Python-based command-line tool.

Supported distribution methods:
 - Properly self-contained Python wheels for Windows, MacOS, and manylinux.
 - Python source distributions.
 - Rust crates (already made a [test 0.0.1 release](https://crates.io/search?q=substrait-validator) on crates.io; contact me via slack or something for handing over project ownership).

Documentation:
 - User-facing documentation is mostly via readme files in the repository right now.
 - `cargo doc` for the Rust crate. Almost everything has at least *a* docstring. Most of the validation code should also be easy to understand, with the complexity mostly hidden by the boilerplate tree traversal logic and data type modules. I've tried, at least, to keep those bits approachable even for people new to Rust (though an IDE that tells you what the types of things are is pretty much a must).
 - The Python module has docstrings as well, but it's unfortunately comparatively messy, in no small part due to the ugliness of the classes generated by protobuf. It's easy to use, though.
 - The C module is pretty much just documented via the generated header file.
 - I also documented the release process for the Python module and Rust crates, which is rather nonstandard here and there because of the "code generation" phase that copies the protobuf and schema files into the appropriate working directories (neither the Rust or Python toolchains I was using have nice hooks for this).

Devops stuff and testing:
 - Github workflows build and test everything on the major operating systems (windows-latest, macos-latest, ubuntu-latest).
 - ~~I'm assuming that CI runs will have to be approved each time, so I also have a PR open [on my own repo](https://github.com/jvanstraten/substrait/pull/2) to test them.~~
 - Added some pre-commit hooks, most for Rust but also some general ones for pre-existing CI checks.
 - There are (generated) workflows for releasing the Python module (notably manylinux). The tests are run by fully building the wheels/sdists and installing them, so the whole release process should be covered, but I haven't tested actually pushing to PyPI.
 - Unit testing is admittedly somewhere between scarce and nonexistent, the biggest reasons being that lots of things in Substrait plans are highly context-sensitive and that I can't define plans in a reasonable format from within Rust directly.
 - There is however a validation test runner which makes it pretty easy to test the actual validation logic. I've already added a small majority of TPC-H queries as integration tests as well as a large amount of positive and negative corner tests.
 - Rust code is free of clippy and format violations via CI checks and pre-commit hooks.
 - Likewise for Python code via black and flake8.

Specification FIXMEs:
 - While implementing the validation logic, I came across a number of things in the spec that either don't make sense to me or are just wrong. Rather than updating the spec and significantly polluting this PR in the process, I used FIXME tags throughout the Rust sources to keep track. A grep for FIXME in validator/rs should list them and only them. There's one thing that I did fix immediately because I couldn't do anything sensible with it otherwise, being that switch expression messages lacked a field for the expression to be matched.

There is also https://github.com/jvanstraten/substrait/pull/3: a combination of this PR and Isthmus #120. Isthmus does a few things that I'm pretty sure are wrong according to the spec (my interpretation thereof anyway) or IMO a very bad idea (allowing anchor/reference 0, so you can't ever tell the difference between a reference that someone forgot to add and a reference to anchor 0), the latter of which I've also actually changed in Isthmus in that branch because it was easy to do. There's also a [demo notebook](https://github.com/jvanstraten/substrait/tree/validator%2Bisthmus/validator/demo) in there that you can use to easily play around with Isthmus and the validator (you'll have to actually clone the repo and run it though; the HTML output is loaded in IPython using an iframe, so it's not included in the notebook file).


jvanstraten:
CI failure is due to https://github.com/rust-lang/rust-clippy/issues/8658, not sure what's going on with the new clippy version or how to work around it yet.

jvanstraten:
@jacques-n Since the Rust bindings got their own repo, I guess it might make sense to move this to (a) separate repo(s) as well. I'm not sure what the best approach would be, though, since both the Python and Rust packages of the validator have these packaging issues. Also, the validator has its own protobuf message tree to describe its output systematically, which would either need to be where it is in the PR right now (`proto/substrait/validator/*.proto`) or in yet another repository. I guess the most technically-correct submodule structure would be this monster...

```mermaid
  graph TD;
      validator-rs-->substrait;
      validator-rs-->validator-proto;
      validator-c-->validator-rs;
      validator-py-->substrait;
      validator-py-->validator-proto;
      validator-py-->validator-rs;
```

but I think a single repo for the validator (or still just merging it into the main repo) is better at this point, despite the packaging process still being a bit awkward. I did solve all the issues in the end, and I'd be inclined to tolerate a bit of ugliness in the release process in order to not have to deal with keeping all the submodules in that graph in sync.

jacques-n:
> ...repo

Holy crap, this thing is massive! I guess that is what I get when I look away for a little bit. Nice work @jvanstraten, very exciting.

Some thoughts:
* I'm inclined to do this as a single new repository (substrait-validator). 
  * It keeps the core repo build quick.
  * If it isn't, it means that any time we update the spec, whomever is updating the spec will also need to update this code. That's feels like a lot for someone to understand. (or the pre-commit tests would need to be able to have this run against an older version of the protos than once in the same repo).
   * It seems like we'll start supporting multiple plan versions with this, something that makes it feel like it should be independent.
* As an independent repo, maybe we should import the git log as is to start. Seems more trackable than squash merging.
* Any patch that is this large tends to get poor reviews. Any thoughts on how we can address? Does it make sense to refactor this pr into several patches that are more bite-size so we can get a higher quality review?

What do you think about each of the above?

jvanstraten:
Some quick responses only because I'm currently on vacation (will be back in two weeks);

> I'm inclined to do this as a single new repository

I'd have to do some work splitting it up properly when I'm back from vacation, but sure.

> If it isn't, it means that any time we update the spec, whomever is updating the spec will also need to update this code.

This was kind of by design though, since it then forces the validator to stay exactly in sync with the spec. If it *isn't* in sync, it's not nearly as useful, because then what are you really validating?

I spent quite some effort to make sure that the validator does the right thing even if some new feature isn't implemented. Everything has sane defaults, such as automatically generating not-yet-implemented warning diagnostics when a specified field isn't explicitly traversed by the validation function. You'd typically only have to add a oneof match case that doesn't do anything when adding something to the spec.

> It seems like we'll start supporting multiple plan versions with this, something that makes it feel like it should be independent.

I guess. I'm struggling with the idea of versioning in the usual way with git though, because without extra effort an application would never be able to use two versions of substrait at once using the official protobuf library; there can only be one `substrait.Plan` message tree linked to a binary at once. prost doesn't have this restriction, but any multi-version consumer/producer will have this problem. The website should also be version-aware IMO. This is a complex problem that requires more thought than just what the validator requires.

> As an independent repo, maybe we should import the git log as is to start. Seems more trackable than squash merging.

I could probably squash some of the earlier commits together to clean out all the WIP commits, but otherwise I agree that merging this is better than squashing everything into one big commit.

> Any patch that is this large tends to get poor reviews. Any thoughts on how we can address? Does it make sense to refactor this pr into several patches that are more bite-size so we can get a higher quality review?

I'd call this more of an initial commit for a (sub)project than a patch, but I'll be honest: I'm very bad at splitting things into bitesize pieces. In general I feel like splitting (the initial impl of) a big project into features and implementing them one by one is a recipe for spaghetti code and inefficiency because of how difficult it becomes to thoroughly refactor or redo something after it's landed; like, the first ~5 versions of the boilerplate traversal code I made turned out to be terrible to work with, which I just didn't realize until I started using them. It also doesn't help that I needed things to work all the way up to rhe Python bindings to properly test the Rust code, because prost doesn't support JSON(-like) protobuf deserialization (completely) and hardcoding test plans using handwritten Rust syntax seemed like a laborious task and impossible to maintain. All in all, this was about as "core features only" as I could manage without having to take a shot in the dark.

Rather than line-by-line reviewing everything here I'd suggest just playing around with it and glossing over the code, just to convince yourself that it's not completely broken, to check that I interpreted the spec at least more or less correctly, and to see that it is (already) useful. It makes more sense to me to deal with the inevitable bug squashing and improvements of what's basically an initial commit via patches as they come up, than to try to get it exactly right via lenghty reviews the first time.

jacques-n:
My inclination is to import this as a string of commits into the new repo https://github.com/substrait-io/substrait-validator. @cpcloud , @westonpace , @rymurr or @andygrove, any interest in reviewing all/part of this?

jvanstraten:
Okay, I'll work on converting all the build logic to the submodule approach. When done, I'll open a PR into the new repo and close this one.

Issue #141, opened 2022-02-20
Setup automated releases


jacques-n:
@cpcloud and I have been discussing introducing automated releases/release notes and starting to version things. We should get this going since other projects would be much better off depending on a released set of substrait proto definitions, etc as opposed to using a git hash.

cpcloud:
Yep, I can look into this.

winningsix:
Any more details for this? One quick question around this, for a ```Substrait``` dependent project, will Substrait be linked as a Git submodule or include protobuf there?

jacques-n:
I think this will depend on the language. For things like Javascript, Java and, python I think we should publish compiled files. For c and c++ I think we should probably publish a tgz with the proto files. What do others think?

FelixYBW:
Do we have plan to create and release the helper tools for each language as well? Like substrait_plan_builder etc.? The builder function will be useful to create and test substriat plans consumers. If so for c/c++ we may need to publish a library as well in future.

jacques-n:
> Do we have plan to create and release the helper tools for each language as well? Like substrait_plan_builder etc.?

Yes, I think we should do that. I'm definitely trying to provide that as part of the [java patch](https://github.com/substrait-io/substrait/pull/120) I'm working on.

PR #156, opened 2022-04-04
ci: set up automated releases

cpcloud:
This PR implements:

1. Automated GitHub source release of substrait including tagging, updating
   release notes, and publishing to GitHub Releases.
1. Automated release of the substrait protos to the [`buf` Schema Registry](https://docs.buf.build/bsr)

The automation is done using [`semantic-release`](https://github.com/semantic-release/semantic-release) and commit structure is verified using `commitlint` in CI.


jacques-n:
@rymurr , might you be willing to take a look at this and give your feedback?

My initial thought is: any chance we can package this up as an action as opposed to a bunch of code in our repo? (Or do as a submodule?) As we have more repos, I'd love to use a consistent pattern and not have to replicate things multiple times. 

Additionally, it would be good if we added a site page to this set of changes that explained the release process, message requirements, etc.

cpcloud:
> 



> @rymurr , might you be willing to take a look at this and give your feedback?
> 
> My initial thought is: any chance we can package this up as an action as opposed to a bunch of code in our repo? (Or do as a submodule?) As we have more repos, I'd love to use a consistent pattern and not have to replicate things multiple times.

Yep, makes sense.

> 
> Additionally, it would be good if we added a site page to this set of changes that explained the release process, message requirements, etc.

I am happy to write down a brief summary of what semantic-release does in our docs, but I'd rather not spell out the details until the dust has settled, we've done a few releases and we're happy with the process.

Happy to put some links to the message requirements, they're spelled out in detail elsewhere.

.commitlintrc.json
```diff
@@ -0,0 +1,126 @@
+{
```
rymurr, line 1: Hey @cpcloud I was speaking to @jacques-n about the need to add the commit lint to this PR. My understanding is that its required to help semver decide if its a major or minor release. Is that correct?

Would you mind just saying a few words so that future us knows why we did this?

.github/workflows/pr.yml
```diff
@@ -64,3 +63,26 @@ jobs:
             ajv validate -s ../text/simple_extensions_schema.yaml --strict=true --spec=draft2020 -d "$i"
           done
         working-directory: ./extensions
+  commitlint:
```
rymurr, line 66: it would be great to document the rules for future commiters and add a pre commit hook (so they aren't surprised by failed ci and have to re-write their commit). I am happy to take that as a follow up if you want.

buf.gen.yaml
```diff
@@ -1,10 +1,10 @@
 version: v1
 plugins:
-  - name: cpp
+  - remote: buf.build/protocolbuffers/plugins/cpp:v3.20.0-1
```
rymurr, line 3: Does this generate the proto files for each language? Do they get uploaded anywhere? 

eg how would i import this into a java project?

or do we add publish steps to maven central, pypi, etc in a follow up?

ci/release/run.sh
```diff
@@ -0,0 +1,14 @@
+#!/usr/bin/env bash
+# shellcheck shell=bash
+
+set -euo pipefail
+
+npx --yes \
```
rymurr, line 6: what are the rules for how a release version is decided?

rymurr (commented):
Looks good @cpcloud . A few minor comments added.

I agree with @jacques-n that moving the bash scripts into the actions would be better. Or at the least combining them into a single bash script w/ flags and functions to cut down on code dup (this is especially true of `run.sh` and `dry_run.sh` imho).

ci/release/run.sh
cpcloud: https://github.com/semantic-release/semantic-release#how-does-it-work

rymurr: cool, thanks. So I guess we have to be careful to add the `BREAKING CHANGE:` footer to a commit?

I see that `buf` has a `breaking` command, perhaps we can run that as part of CI and fail if the correct footer isn't present?

cpcloud: That's a great idea!

.commitlintrc.json
cpcloud: > Hey @cpcloud I was speaking to @jacques-n about the need to add the commit lint to this PR. My understanding is that its required to help semver decide if its a major or minor release. Is that correct?

`commitlint` per se isn't necessary for release, but it will prevent commits that don't conform to semantic-release and therefore couldn't be used to compute a new release version.

jacques-n:
This ticket addresses #141. Linking here.

jacques-n:
@cpcloud and @rymurr, we're stalled out here. What do we need to do to wrap this up?

cpcloud:
Not stalled, I've just been in Europe for the past ten days. Let's wrap this up early next week. I will address the remaining feedback this week.

jacques-n:
Vive la CI!

buf.gen.yaml
cpcloud: Publishing to the GitHub release and BSR for now.

cpcloud:
@rymurr @jacques-n This is ready for another round of review.

jacques-n (approved):
LGTM. Let's merge and we'll see how things go :D

PR #157, opened 2022-04-04
style: format protos using buf format

cpcloud:
This PR reformats protos using `buf format` and adds a CI check for it as well.


rymurr (approved):
LGTM

PR #159, opened 2022-04-05
Added a slight clarification to logical_relations.md

### LINKED ISSUE #158: Does ReadRel::base_schema include columns hidden by projection?
The `ReadRel` relation states:

Direct Schema | Defines the schema of the output of the read (before any emit remapping/hiding).
Projection | A masked complex expression describing the portions of the content that should be read

Does the `Direct Schema` include fields hidden by `Projection` (which is not technically the same thing as emit remapping / hiding)?

westonpace:
Closes #158 

This clarifies that the direct schema of a read relation should report the schema before projection (i.e. it will include fields that are removed by the projection)


jacques-n (approved):
LGTM +1

PR #166, opened 2022-04-12
fix function description for gt and gte

gforsyth:
not a functional change, just updating the descriptions to match


jacques-n (approved):
Good catch, thanks @gforsyth !

PR #168, opened 2022-04-18
build(go): set Go package names and enable buf generate for Go

zeroshade:
This PR adds a Go module for the substrait protobuf files. There's a gen-proto.sh file which uses `git rev-parse` to find the top of the cloned git tree and then uses that to run `protoc` against the `.proto` files to generate the go code into the appropriate directory. This can be easily replicated by running `go generate ./substrait` from the `go` directory (or `go generate .` from the `go/substrait` directory). 

For a Go module to be importable, the generated code needs to be committed to the repository so it can be pulled down and compiled by the `go` command since it will not run `go generate` as part of building at all. 

One thing I was unsure of was whether or not it should be kept as a manual process that someone needs to run `go generate` and commit the updated `.pb.go` files or if a Github action should be created to automatically do so whenever changes are merged to the `.proto` files in master. But for now this will at least provide an initial Go module to build off of.


go/README.md
```diff
@@ -0,0 +1,18 @@
+# substrait-go
+
+Experimental Go bindings for [substrait](https://substrait.io).
+
+## Generate from proto files
+
+```bash
+go generate ./substrait
```
cpcloud, line 8: Can you instead use `buf gen` for this?

You'll need to add these lines to `buf.gen.yaml`:

```yaml
  - name: go
    out: gen/proto/go
```

go/substrait/gen-proto.sh
```diff
@@ -0,0 +1,5 @@
+#!/bin/sh
```
cpcloud, line 1: I don't think this script is necessary if you use `buf gen`.

cpcloud (changes requested):

go/substrait/gen-proto.sh
zeroshade: Updated `buf.gen.yaml` and removed this script.

go/README.md
zeroshade: for ease of import paths, is it okay that my change to buf.gen.yaml uses `out: go` so that the import path for Go would be: `github.com/substrait-io/substrait/go/substrait` rather than `github.com/substrait-io/substrait/gen/proto/go/substrait` ?

jacques-n:
We had talked about putting different language bindings in different repos. @cpcloud , you good with proposing that we put this in a substrait-go repo?

jacques-n:
I've created https://github.com/substrait-io/substrait-go to store go bindings related to substrait.

cpcloud:
@jacques-n Sounds good to me. Happy to review this PR over in `substrait-go`.

zeroshade:
@jacques-n @cpcloud I've modified this PR to only update the buf.gen.yaml and the .proto files with the `option go_package` line pointing at the `substrait-go` repository. 

I've created https://github.com/substrait-io/substrait-go/pull/1 over at the new repository.

jacques-n:
This looks reasonable to me. Since this has a circular reference, what order should we merge this and https://github.com/substrait-io/substrait-go/pull/1 in?

jacques-n:
> This looks reasonable to me. Since this has a circular reference, what order should we merge this and [substrait-io/substrait-go#1](https://github.com/substrait-io/substrait-go/pull/1) in?

My assumption is this should be merged first. Just want to confirm.

zeroshade:
@jacques-n Yep, this should get merged first and then i'll update the submodule in the substrait-go PR

buf.gen.yaml
```diff
@@ -8,3 +8,8 @@ plugins:
     out: gen/proto/java
   - name: python
     out: gen/proto/python
+  - name: go
+    out: gen/proto/go
+    opt:
+      - paths=source_relative
+
```
cpcloud, line 15: Seems like there's a few extra newlines here that should be removed.

cpcloud (approved):
LGTM except for the extra newlines in `buf.gen.yaml`!

buf.gen.yaml
cpcloud: Actually, I can just clean this up and merge it!

cpcloud: Done.

cpcloud (approved):

cpcloud:
Thanks @zeroshade! Keep it coming!

PR #178, opened 2022-04-28
docs: correct masked expression example

### LINKED ISSUE #173: "Masked Complex Expression" example looks wrong
In https://substrait.io/expressions/field_references/#masked-complex-expression it seems the example is not consistent. The mask, textual description, and given result all diverge with each other.

If we keep the textual description, I would for example expect this mask:
```
0:[0,1:[..5:[0,2]]],2,3

OR

0:
  - 0
  - 1:
    ..5:
      - 0
      - 2
2
3
```
(also note typographical issues: "O" vs. "0", spacing after YAML hyphenation)

to produce this result:
```
struct:
  - struct:
    - integer
    - list:
      struct: 
        - i32
        - string
  - i32
  - i64
```

Am I misunderstanding something?

pitrou:


cpcloud (approved):
LGTM, thanks!

PR #180, opened 2022-04-28
fix(ci): cd into buf-configured proto directory

cpcloud:


cpcloud:
Typically you would use `fix` for CI things, but we need to trigger a release to test the publishing step so that's why I'm doing it here.

jacques-n (approved):

PR #181, opened 2022-04-28
fix(ci): archive the generated code

cpcloud:


ci/release/prepare.sh
```diff
@@ -6,3 +6,7 @@ set -euo pipefail
 # build artifacts
 buf build
 buf generate
+
+for dir in gen/proto/*; do
+  tar -czf "$(dirname "${dir}")"/"${dir//\//.}.tar.gz" "${dir}"
```
cpcloud, line 11: The resulting tarballs are named like `gen.proto.${language}.tar.gz`.

jacques-n: Which languages is this useful for? I assume go would just use buf. For java, really need a maven coordinate jar file, etc.

cpcloud: Ah, so should we just remove it then?

cpcloud: All languages should be able to use buf if there's a protoc plugin, that said I don't know much about Java package management.

jacques-n: Yeah, I think we should remove this step. Java has something similar to npm or pip which is a separate process we'll need to undertake. Ideally buf would do the work but it doesn't look like they have focused on that ecosystem.

cpcloud: Roger that, I'll remove this and keep the conventionalcommits changes. That style is a bit better IMO. Here's the recent ibis release with that style: https://github.com/ibis-project/ibis/releases/tag/3.0.0

jacques-n (approved):
Sounds good. LGTM

PR #182, opened 2022-04-28
fix: remove not function that expects two arguments

sanjibansg:
This PR removes the `not` function in `functions_boolean.yaml` that expects two boolean arguments.


extensions/functions_boolean.yaml
```diff
@@ -1,14 +1,6 @@
 %YAML 1.2
 ---
 scalar_functions:
-  -
-    name: not
-    description: The boolean not of a provided value.
-    impls:
-      - args:
-          - value: boolean
```
jacques-n, line 9: I think we should just remove the second arg as opposed to removing the entire function. (It should be a single boolean argument input as opposed to two.)

extensions/functions_boolean.yaml
```diff
@@ -1,14 +1,6 @@
 %YAML 1.2
 ---
 scalar_functions:
-  -
-    name: not
```
cpcloud, line 5: Should we keep the function but correct it to have a single argument?

extensions/functions_boolean.yaml
westonpace: The unary not kernel was already at the end of the list

jacques-n: lol

extensions/functions_boolean.yaml
jacques-n: as pointed out to me, apparently we already had the correct one elsewhere in the file.

extensions/functions_boolean.yaml
sanjibansg: Yes, the unary not function is already present in the end
https://github.com/substrait-io/substrait/blob/c62c1492dd0ecea04f3d053e93eedf1b79877ab2/extensions/functions_boolean.yaml#L31

cpcloud (dismissed):
LGTM

jacques-n:
I think we need this commit to be marked `BREAKING CHANGE` according to the semantic release plugin since removal of a function is a breaking change. See [here](https://www.npmjs.com/package/semantic-release).

@cpcloud , any chance you can point to a good example of what a commit message should look like?

cpcloud:
Here's an example from ibis: https://github.com/ibis-project/ibis/commit/3f48cb8b71ee30e8e8cdc3b13a50939dc51bfaa0. The commit message is:

```
refactor(api): remove distinct ColumnExpr API

Remove `ColumnExpr.distinct()` because it's a confusing API.

BREAKING CHANGE: Replace `t["a"].distinct()` with `t[["a"]].distinct()`.
```

cpcloud:
@jacques-n How do you want to proceed here?

jacques-n:
> @jacques-n How do you want to proceed here?

Isn't the commit message still missing the breaking change notice? Frankly, I don't think anyone is depending on this change so I'm not stuck on calling it a breaking change other than for rigor.

sanjibansg:
> 

I have updated the commit message to include `BREAKING CHANGE`, does this look good now?

westonpace (approved):
The commit message looks correct to me now.

cpcloud (approved):

cpcloud (dismissed):
Sorry, I meant to block accidental merging until #210 is merged :)

jacques-n (approved):
Now that #210 is merged, merging this.

PR #186, opened 2022-05-09
feat: add flag FailureBehavior in Cast expression

ZJie1:


site/docs/expressions/specialized_record_expressions.md
```diff
@@ -11,7 +11,7 @@ For each data type, it is possible to create a literal value for that data type.
 
 
 ## Cast Expression
-To convert a value from one type to another, Substrait defines a cast expression. Cast expression declare an expected type and an input argument.
+To convert a value from one type to another, Substrait defines a cast expression. Cast expression declare an expected type, an input argument and an enumeration flag that contains enhancement conditions, such as `CONDITION_FLAG_NULL_ON_FAILURE` which  prevents throws and instead returns null on cast failure.
```
mbasmanova, line 14: typo: "Cast expression declare" -> "Cast expression declares"

proto/substrait/algebra.proto
```diff
@@ -482,6 +482,13 @@ message Expression {
   message Cast {
     Type type = 1;
     Expression input = 2;
+    ConditionFlag condition_flag = 3;
+
+    enum ConditionFlag {
+      CONDITION_FLAG_UNSPECIFIED = 0;
+      CONDITION_FLAG_NULL_ON_FAILURE = 1;
+      CONDITION_FLAG_THROW_ON_FAILURE = 2;
```
mbasmanova, line 490: Curious why do we need both NULL_ON_FAILURE and THROW_ON_FAILURE. Since these are mutually exclusive, I expect that only one would be sufficient.

mbasmanova (commented):
A question.

proto/substrait/algebra.proto
ZJie1: When do conversion, if the `nullOnFailures= true` in velox, we can use `substraitCastExpr.set_condition_flag(::substrait::CONDITION_FLAG_NULL_ON_FAILURE)`.
if the `nullOnFailures = false`, we can use `substraitCastExpr.set_condition_flag(::substrait::CONDITION_FLAG_THROW_ON_FAILURE)`.
If we only have this one variation, we can use `CONDITION_FLAG_UNSPECIFIED` for the case `nullOnFailures = false`, but I am afraid this will be confused when there have other variations. Your thought?

site/docs/expressions/specialized_record_expressions.md
jvanstraten: It should be either "**A** cast expression declare**s**" or "Cast expression**s** declare", actually. As you corrected it and as it was originally written it's missing an article.

I would also rewrite the addition to something like "and an argument specifying whether the operation should throw or return null on failure."

proto/substrait/algebra.proto
cpcloud: It's probably best to keep it explicit for clarity and stipulate the default of `THROW` in the prose in the spec.

proto/substrait/algebra.proto
```diff
@@ -482,6 +482,13 @@ message Expression {
   message Cast {
     Type type = 1;
     Expression input = 2;
+    ConditionFlag condition_flag = 3;
```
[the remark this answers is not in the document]
cpcloud: Yeah, let's change this to `FailureBehavior failure_behavior`.

site/docs/expressions/specialized_record_expressions.md
```diff
@@ -11,7 +11,7 @@ For each data type, it is possible to create a literal value for that data type.
 
 
 ## Cast Expression
-To convert a value from one type to another, Substrait defines a cast expression. Cast expression declare an expected type and an input argument.
+To convert a value from one type to another, Substrait defines a cast expression. Cast expression declares an expected type, an input argument and an enumeration flag that contains enhancement conditions, such as `CONDITION_FLAG_NULL_ON_FAILURE` which  prevents throws and instead returns null on cast failure.
```
cpcloud, line 14:
```suggestion
To convert a value from one type to another, Substrait defines a cast expression. Cast expressions declare an expected type, an input argument and an enumeration specifying failure behavior, indicating whether cast should return null on failure or throw an exception.
```

cpcloud (changes requested):

proto/substrait/algebra.proto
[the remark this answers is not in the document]
ZJie1: change it.

proto/substrait/algebra.proto
ZJie1: Change it like below.

``` 
   enum FailureBehavior {
      // throw an exception as default failure behavior
      FAILURE_BEHAVIOR_UNSPECIFIED = 0;
      FAILURE_BEHAVIOR_RETURN_NULL = 1;
    }
```
 Here we just use the default `FAILURE_BEHAVIOR_UNSPECIFIED = 0` represents the `THROW` cause the default(the first) in `enum` must end with `_UNSPECIFIED`. would this be ok?
Or we must use like below?
``` 
   enum FailureBehavior {
      FAILURE_BEHAVIOR_UNSPECIFIED = 0;
      FAILURE_BEHAVIOR_THROW_EXCEPTION = 1;
      FAILURE_BEHAVIOR_RETURN_NULL = 2;
    }
```

cpcloud: Let's use

```protobuf
   enum FailureBehavior {
      FAILURE_BEHAVIOR_UNSPECIFIED = 0;
      FAILURE_BEHAVIOR_THROW_EXCEPTION = 1;
      FAILURE_BEHAVIOR_RETURN_NULL = 2;
   }
```

The `_UNSPECIFIED` suffixes are exactly that, they indicate that there is no error behavior specified. The default behavior should be one of the enum variants that is not `_UNSPECIFIED`.

proto/substrait/algebra.proto
```diff
@@ -482,6 +482,13 @@ message Expression {
   message Cast {
     Type type = 1;
     Expression input = 2;
+    FailureBehavior failure_behavior = 3;
+
+    enum FailureBehavior {
+      // throw an exception as default failure behavior
+      FAILURE_BEHAVIOR_UNSPECIFIED = 0;
+      FAILURE_BEHAVIOR_RETURN_NULL = 1;
+    }
```
cpcloud, line 491:
```suggestion
    enum FailureBehavior {
      // throw an exception as default failure behavior
      FAILURE_BEHAVIOR_UNSPECIFIED = 0;
      FAILURE_BEHAVIOR_RETURN_NULL = 1;
      FAILURE_BEHAVIOR_THROW_EXCEPTION = 2;
    }
```

cpcloud (changes requested):

cpcloud:
@ZJie1 Will merge after the additional enum variant is added!

proto/substrait/algebra.proto
ZJie1: Got it. Thanks!
Changed this in the latest commit.

cpcloud (approved):

PR #191, opened 2022-05-10
Add invocation property to AggregateFunction message for specifying distinct vs all

JamesRTaylor:


proto/substrait/function.proto
```diff
@@ -78,6 +78,18 @@ message FunctionSignature {
...
+
+    AggregateInvocationType invocation = 16;
+
+    enum AggregateInvocationType {
+      AGGREGATE_INVOCATION_TYPE_UNSPECIFIED = 0;
+
+      // Aggregate function invocation returns all values
+      AGGREGATE_INVOCATION_TYPE_ALL = 1;
+
+      // Aggregate function invocation only returns distinct values
```
jacques-n, line 90: I feel like the "only returns distinct values" here and below are misleading. The function doesn't return distinct values, it returns a calculation that is based on only distinct values.

JamesRTaylor:
I've updated the comment, @jacques-n. I also moved the new invocation property to the right proto message.

jacques-n:
@cpcloud and @jvanstraten , you good with this change? It's not breaking but it is an introduction of a new feature in the spec.

cpcloud (approved):
LGTM, this will get picked up in the next release on Sunday!

jacques-n:
Thanks @JamesRTaylor !

jvanstraten:
Looks like I'm too late, but looks good to me for what it's worth :)

PR #199, opened 2022-05-16
Support fractional seconds for interval_day literals

jvanstraten:
While updating the type system documentation, I realized that `interval_day` literals don't currently support the advertised microsecond precision. I added support via a third microseconds component in order to be completely backward-compatible.


jacques-n (approved):
Doh, wish this had been right to begin with rather than having a third field. Agree that third field is cleaner for compatibility. Thanks @jvanstraten

PR #200, opened 2022-05-16
Type system documentation rewrite

### LINKED ISSUE #195: Confusion about interval types (and type compliance in general)
The way the documentation describes `interval_year` and `interval_day` has always confused me. What I intuitively feel like it *should* say is that `interval_year` describes a timestamp difference with month granularity (with the purpose of being able to say "next month" or "next year" despite the actual timestamp difference implied with that being context-sensitive), `interval_day` describes an exact timestamp difference with at least microsecond granularity, and both types must support a range of at least +/-10000 years. If I were to start from scratch with a data storage format for these types, I would probably use a single i16 signifying months for `interval_year`, and distribute `interval_day` over i64 seconds and i32 microseconds (or nanoseconds) because that's the distribution I'm used to from Linux and C.

Following [the relevant part of the spec](https://github.com/substrait-io/substrait/blob/720dbeac31f11655514cd737815d0bac0ad47fc8/site/docs/types/simple_logical_types.md?plain=1#L22-L23) to the letter however, my takeaways are:

 - `interval_year` *must* be stored as two separate integers, one for years and one for months, because for example 1y0m is defined to be different from 0y12m. Otherwise, I don't understand what all the examples are for; a range [-12000, 12000] months would have sufficed, and the rest would just be describing the semantics of a string to `interval_year` conversion.
 - Likewise, `interval_day` *must* be stored as two separate integers, one for days and one for at least microseconds; I couldn't, for example, reuse `DECIMAL<18, 6>` (signifying seconds) or (equivalently) i128 microseconds, or use seconds/microseconds as I probably would have done otherwise.
 - The microsecond portion of `interval_day` must be stored as at least an i64, either with microsecond or nanosecond granularity (it took me a while to figure out that those numbers are based on 2^63/1000).
 - Unlike `interval_year`, `interval_day` can legally assume values beyond +/-10000 years, for example by setting both day and micros to the maximum or minimum value.

Notably, the Arrow data type marked as equivalent for `interval_year` wouldn't be compliant; Arrow stores `INTERVAL<YEAR_MONTH>` as a single i32 specifying months. `INTERVAL<MONTH_DAY_NANO>` (i32 months, i32 days, i64 nanos) is also not an exact match with `interval_day`, but AFAICT it's at least a superset thereof.

I would propose to rewrite/clarify the specification of the data types after we answer the following questions:

 1) Would it be sensible to rewrite the docs for `interval_year` and `interval_day` based on the first paragraph, or is my lack of a background in data science showing?
    - If yes, should we change the names to `interval_month` and `interval_micros`, or something else that better captures what they represent?
 2) Should types in general be specified by means of some canonical in-memory format, or merely represent abstract sets of possible values that are hopefully convenient?
    - Integers and arguably floating points are currently specified as the former, intervals strongly imply a particular physical representation, and most of the other types are specified abstractly.
    - AFAICT [the answer to this one the latter](https://github.com/substrait-io/substrait/blob/720dbeac31f11655514cd737815d0bac0ad47fc8/site/docs/types/type_variations.md?plain=1#L3), i.e. type variations should be used to specify the physical storage type when needed, so the base types can be defined abstractly.
 3) Should it be legal for implementations to internally operate on values outside the Substrait-specified range, if they use a different physical representation internally?
    - If yes, this would allow Arrow to use `INTERVAL<MONTH_DAY_NANO>` without additional checks, but it might allow for a lot of subtle undefined behavior in practice. For example, 1 microsecond divided by 3 could then be rounded to 0 microseconds by one engine, and to 333 nanoseconds by another. Not sure what the best approach would be here, or if these corner cases are even something we need to care deeply about.

jvanstraten:
Closes #195

This is an almost complete rewrite/reformulation of the type system specification per my current understanding to make intentions a bit more explicit. This isn't supposed to change any specified behavior, so LMK if I still misunderstood something...


site/docs/types/simple_logical_types.md
```diff
@@ -1,24 +1,22 @@
-# Simple Types
+# Simple Logical Types
 
-Substrait tries to cover the most common types used in data manipulation. Simple types are those that don't support any form of configuration. For simplicity, any generic type that has only a small number of discrete implementations is declared directly, as opposed to via configuration.
+Simple types are those that don't support any form of configuration. For simplicity, any generic type that has only a small number of discrete implementations is declared directly, as opposed to via configuration.
```
jacques-n, line 3: As I've thought more about this, I think we should just remove the distinction between simple and compound types. Technically simple types also have properties: they are all configurable with nullability and type variation. Additionally, the number of times people have said "why don't you have x" when they were looking at one table and not the other suggests that the page separation is also confusing. Can you just collapse to a single page? (please include a forward from the old page to the new consolidated page for SEO purposes).

site/docs/types/simple_logical_types.md
```diff
@@ -1,24 +1,22 @@
...
-| uuid            | A universally-unique identifier composed of 128 bits. Typically presented to users in the following hexadecimal format: `c48ffa9e-64f4-44cb-ae47-152b4e60e77b`. Any 128-bit value is allowed, without specific adherence to RFC4122.                                                                                                                                                                                                                |                                | uuid           |               | UUID                        |
+| Type Name       | Description                                                  | Canonical physical representation (protobuf)
```
jacques-n, line 5: As above, please remove canonical.

site/docs/types/type_system.md
```diff
@@ -0,0 +1,36 @@
+# Type System
+
+Substrait tries to cover the most common types used in data manipulation. To minimize type explosion, the project currently follows the guideline that a logical type should probably only be included in the specification if it is included in at least two of the following open source Projects: Apache Arrow, Apache Iceberg, Apache Spark and Trino. Types beyond this common core may be represented using [simple extensions](../extensions/index.md#simple-extensions).
+
+## Logical vs. Physical Types
```
jacques-n, line 5: As above, I think you're using a specific version of these terms which doesn't map well to relational algebra land. As mentioned elsewhere, all types in substrait carry type variation. It was an early concrete change in Substrait to avoid a separation between logical and physical type definitions. This section is useful but I'd title this section "Type Interpretation Guidelines" or something similar.

site/docs/types/type_system.md
```diff
@@ -0,0 +1,36 @@
...
+
+All types in Substrait must have a declared nullability, either `REQUIRED` to specify non-nullable or `NULLABLE` to specify nullable. In the latter case, values of the type may assume the special null value, otherwise they may not. When represented as text, a type is `NULLABLE` if and only if the logical type name is suffixed with a `?`, for example `i8?` or `VARCHAR?<10>` (see [type syntax parsing](type_parsing.md)).
+
+[Unless otherwise specified](../expressions/scalar_functions.md#nullability-handling), operations that take multiple arguments support any mix of nullability at their input. They yield a nullable type if any of the arguments are nullable, and at runtime yield null if any argument is null.
+
+Null literals in Substrait are typed: for example, a null for an `i8?` is different from a null for a `VARCHAR?<10>`. There is no separate null type.
+
+## Coercion and Casting
+
+The only context in which Substrait may be argued to perform type coercion is for nullability, as specified in the previous section. In all other contexts, explicit [cast expressions](../expressions/casts.md) are needed.
```
jacques-n, line 36: This isn't a property of type, it is a property of the function definition system. As such, it is a little weird it is noted here.

site/docs/types/user_defined_types.md
```diff
@@ -1,4 +1,4 @@
-# User-Defined Types
+# User-Defined Logical Types
```
jacques-n, line 1: please remove logical.

site/docs/types/user_defined_types.md
```diff
@@ -12,3 +12,5 @@ A YAML example of an extension type is below:
 ```
 
 This declares a new type (namespaced to the associated YAML file) called "point". This type is composed of two `i32` values named longitude and latitude. Once a type has been declared, it can be used in function declarations.  [TBD: should field references be allowed to dereference the components of a user defined type?]
+
+The canonical physical representation of user-defined types is a protobuf [Any](https://developers.google.com/protocol-buffers/docs/proto3#any) message.
```
jacques-n, line 16: Please reword this. There is no canonical physical representation. There is a protobuf representation of literals only.

jacques-n (commented):
Thanks @jvanstraten, great to get more eyes on this stuff!

I did a first review pass. Haven't yet gone through the intricacies of all your type definition changes. My big feedback is you're introducing a new set of concepts of logical and physical types which we started with and then specifically removed as it doesn't map well to the protobuf representation and also has conflicts with the definition of logical and physical when talking about relational algebra operations.

site/docs/types/compound_logical_types.md
```diff
@@ -1,15 +1,15 @@
...
-| MAP&lt;K, V&gt;                   | An unordered list of type K keys with type V values.         | map&lt;k,v&gt;            | map&lt;k,v&gt;       | -              | map&lt;k,v&gt;                    |
+| Type Name                    | Description                                                  | Canonical physical representation (protobuf)
+| ---------------------------- | ------------------------------------------------------------ | ------------------------------------------------
+| FIXEDCHAR&lt;L&gt;           | A fixed-length unicode string of L characters. L must be within [1..2,147,483,647]. When casting, strings shorter than L are padded with spaces, and strings longer than L are right-trimmed. | L-character `string`
+| VARCHAR&lt;L&gt;             | A unicode string of at most L characters.L must be within [1..2,147,483,647]. When casting, strings shorter than L are not padded, and strings larger than L are right-trimmed. | `string` with at most L characters
+| FIXEDBINARY&lt;L&gt;         | A binary string of L bytes. When casting, values shorter than L are padded with zeros, and values longer than L are right-trimmed. | L-byte `bytes`
```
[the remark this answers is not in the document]
jvanstraten: I agree, but the original text does define what I can only describe as casting behavior there. I left it there since casts are not otherwise defined yet.

site/docs/types/compound_logical_types.md
```diff
@@ -1,15 +1,15 @@
-# Compound Types
+# Compound Logical Types
```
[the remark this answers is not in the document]
jvanstraten: Okay, that explains a lot (like the filename).

I do think there should be a word to distinguish "type" from the "~~logical~~ type", though. One of the things that went very wrong in the Arrow Substrait consumer is the lack of an understanding that a Substrait type is actually a four-tuple of the ~~logical~~ type, nullability, an optional variation, and a list of parameters (I know the latter doesn't actually exist for simple types, but in my head and in the validator it's just an empty set of parameters). I used "base type" in the validator sources, would naming them that way make more sense to you?

site/docs/types/compound_logical_types.md
```diff
@@ -1,15 +1,15 @@
...
+| ---------------------------- | ------------------------------------------------------------ | ------------------------------------------------
+| FIXEDCHAR&lt;L&gt;           | A fixed-length unicode string of L characters. L must be within [1..2,147,483,647]. When casting, strings shorter than L are padded with spaces, and strings longer than L are right-trimmed. | L-character `string`
+| VARCHAR&lt;L&gt;             | A unicode string of at most L characters.L must be within [1..2,147,483,647]. When casting, strings shorter than L are not padded, and strings larger than L are right-trimmed. | `string` with at most L characters
+| FIXEDBINARY&lt;L&gt;         | A binary string of L bytes. When casting, values shorter than L are padded with zeros, and values longer than L are right-trimmed. | L-byte `bytes`
+| DECIMAL&lt;P, S&gt;          | A fixed-precision decimal value having precision (P, number of digits) <= 38 and scale (S, number of fractional digits) 0 <= S <= P. | 16-byte `bytes`
+| STRUCT&lt;T1,...,Tn&gt;      | A list of types in a defined order. | `repeated Literal`, types matching T1..Tn
```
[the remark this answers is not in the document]
jvanstraten: I was pretty much just copying the protobuf representation of literals, since that's the only place where what I've been calling the canonical physical representation exists. But confusion noted, I'll give it another shot.

site/docs/types/compound_logical_types.md
```diff
@@ -1,15 +1,15 @@
...
-| STRUCT&lt;T1,...,T2&gt; | A list of types in a defined order. | struct_&lt;*&gt;                  | struct&lt;*&gt; | struct&lt;*&gt; | row&lt;*&gt;   |
-| NSTRUCT&lt;N:T1,...,N:T2&gt; | **Pseudo-type**: A struct that maps unique names to value types. Each name is a UTF-8-encoded string. Each value can have a distinct type. Note that NSTRUCT is actually a pseudo-type. This is because Substrait's core type system is based entirely on ordinal positions, not named fields. Nonetheless, when working with systems outside Substrait, names are important. | struct_&lt;*&gt; | struct&lt;*&gt; | struct&lt;*&gt; | row&lt;*&gt; |
-| LIST&lt;T&gt;               | A list of values of type T. The list can be between [0..2,147,483,647] values in length. | list                | list           | list           | array                       |
-| MAP&lt;K, V&gt;                   | An unordered list of type K keys with type V values.         | map&lt;k,v&gt;            | map&lt;k,v&gt;       | -              | map&lt;k,v&gt;                    |
+| Type Name                    | Description                                                  | Canonical physical representation (protobuf)
```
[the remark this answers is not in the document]
jvanstraten: I named it that way because:

 - type variation specified: storage format should be whatever corresponds to the variation;
 - type variation not specified: storage format is the "canonical" format.

My reasoning is that once we elaborate on a communication protocol, there should be some canonical data format, even if it's only useful for testing because it isn't very efficient. By precedence, the most logical choice for this format seems to me to be the virtual table format. Writing that format down was my intention for the column.

Where exactly am I going wrong with that line of reasoning?

site/docs/types/simple_logical_types.md
jvanstraten: I'm fine with collapsing the pages and tables into one, but I kind of like the distinction between simple and compound personally. To me, `i8` is a concrete type, while `struct` is not; the concrete form would be `struct<>`, but `i8<>` makes no sense. It's just a way to talk about parameterizable and non-parameterizable types that doesn't take 7-8 syllables, though.

site/docs/types/compound_logical_types.md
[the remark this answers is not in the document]
jacques-n: I think we should talk about things like: 
* All types have certain properties
    * Can have multiple physical variations
    * Can be nullable or non-nullable.
* Some types:
    * have integer parameters
    * have parameters composed of other types.

I understand the concern and confusion. I'm not sure it warrants two terms because there aren't really two separate type concepts. 

If we are looking for a comparison, a more apt comparison to code may be that we have type classes and type instances. For example, I can have I8?[0] and I8[1]. They are both instances of the I8 type class. However, that feels like it is talking about implementation of the spec, not the spec itself. I8 does not exist independent of its definition that includes nullability and variation within the spec.

site/docs/types/simple_logical_types.md
jacques-n: I think this is the root of the disconnect. I wouldn't agree that I8 is a concrete type. I8?[0] or I8[0] are concrete types. **All types are all parameterized**. Some of those parameters are common and some are unique. I think the confusion elsewhere stems from the fact that we don't say "I8 is a parameterized type" when it is (nullability and physical variation).

site/docs/types/type_system.md
jvanstraten: Terminology aside, I'm getting really confused now. If variations don't relate to different ~~physical~~ storage methods of a ~~logical~~ type, then what are they? Just some kind of inheritance-like thing or something? Or is the word "physical" merely overloaded differently in my mind than it is in yours due to relational algebra stuff?

Also, does

> all types in substrait carry type variation

imply that variations aren't optional...? Because then every single plan I've ever seen is invalid (so needless to say I think that would be a bad idea), and the validator is wrong about a lot of things.

site/docs/types/type_system.md
jvanstraten: I disagree strongly here. Yes, it's not a property of *a* type, but coercion (or less generally promotion) is one of the most important properties of a type *system*. It's one of the first questions I ask myself when I encounter a new one.

Even in Substrait, functions merely exist in the context of a type system. Some counterexamples to this purely being a function definition thing:

 - If-else expressions. One of my FIXMEs in the validator is due to nullability behavior there not being specified. I was strict about requiring equal nullability for the branches and non-nullable conditions initially, but IIRC I got tired of manually correcting Isthmus-generated plans when making the test cases.
 - Filter conditions. If the condition is nullable, do all columns in the schema become nullable? Is it illegal to do that? Does it throw an error at runtime if a null comes by? Or does null just imply false? Again, I ended up being lenient here in the validator.
 - Equality functions, which are currently assumed to exist as part of joins, aggregations, set stuff, etc. I don't remember off the top of my head what I did here in the validator.

Whenever this behavior is unspecified, a producer can't accurately deduce the return type of things without making assumptions. Overload resolution depends on nullability, so it needs to be able to do this deduction for the same reasons a read relation has to specify a schema; as soon as a type isn't 100% concrete, it can't know exactly which functions can operate on it. Exhaustively specifying this behavior for everything that takes an expression what the type must be, all the way down to nullability, will only lead to confusion and inconsistency. These two sentences would solve all these problems at once with what seems to me to be reasonably sane default behavior.

I'd also be fine with stating that coercion or promotion doesn't exist, and that types always need to match exactly (unless handled via function nullability handling, I guess). Personally I'd even prefer that. But my gut says it would break a lot of existing stuff.

site/docs/types/type_system.md
jacques-n: Yeah, I'd say the docs aren't super clear on this and it is good that you're trying to clean things up. Let's start with the simple case:

Many systems have a single canonical representation of each type of data. In those systems, people can use the default type variation of 0 meaning "system default".  

In some systems, a system could have multiple variations of representation for the same data. For example, I may have a normal i64 and a rle-encoded i64. In this case, the system would decide which of the type variations each of those things are mapped to. E.g. `i64[0] => normal` and `i64[1] => rle-encoded`.

> that variations aren't optional...?

i64 is the same as i64[0]. So they are optional to specify, they just default 0. The property always exists and has a value. This is the "default" variation for the simple case where systems only have one variation. Protobuf hides default values by default and thus 0 is implied rather than being concrete when sending on the wire. It's absence is the same as it's presence with value 0. In the simple case, the default equals "there is no variation". One thing that is missing is that we should formally specify the properties of the 0 type variation. Specifically it has `function behavior == inherits`. This was implied but isn't written anywhere.

Put another way, types always have a variation property. If not expressly stated, they default to 0, which means "system default".

site/docs/types/type_system.md
jacques-n: > coercion (or less generally promotion) is one of the most important properties of a type system

If your statement is that "a non-nullable type instance can be promoted to a nullable type instance without loss of information", I agree with that statement. However, the specific behavior of things like you list need to specifically defined. Wrt to your specific statements:

> If-else expressions. One of my FIXMEs in the validator is due to nullability behavior there not being specified. I was strict about requiring equal nullability for the branches and non-nullable conditions initially, but IIRC I got tired of manually correcting Isthmus-generated plans when making the test cases.

Agree that we need to better specify. This was my intention: 
* All branches must return the same type of data ignoring nullability. 
* All branches must return the same type variations.
* If all of the branches are required, the outcome is required. If any one of the branches is nullable, the output is nullable.

> Filter conditions. If the condition is nullable, do all columns in the schema become nullable? Is it illegal to do that? Does it throw an error at runtime if a null comes by? Or does null just imply false? Again, I ended up being lenient here in the validator.

I don't understand this question. A filter relation has no impact on the schema of the data. If the condition returns null, we should behave the same as databases do which I believe is to exclude the given record (we can't confirm that the filter condition passes).

> Equality functions, which are currently assumed to exist as part of joins, aggregations, set stuff, etc. I don't remember off the top of my head what I did here in the validator.

Agree this needs to be better defined.

site/docs/types/compound_logical_types.md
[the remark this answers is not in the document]
jvanstraten: Just to be clear, I was just looking for a term to use for the first part of a type, as in:

```
        data type
   _________|________
   STRUCT ? [0] <...>
   ‾‾|‾‾‾ | ‾|‾ ‾‾|‾‾
    ???   |  |  parameter pack
          |  |
nullability  variation
```

(using your syntax; I'm still confused about the [0] part, but that's more suited for another thread in this list)

Type class is fine; incidentally that's also what I used internally in the validator. I'm fine with whatever. I simply used "logical type" for it in the proposal without any further intended implications, so s/logical type/type class/g (I'll get around to rewriting it based on the feedback here when I feel like I'm clear on everything again).

site/docs/types/simple_logical_types.md
jvanstraten: I would just say `i8` is syntactic sugar for `i8[0]`. When someone is talking about `i8` and they're not specifically talking about type classes (which is almost never in practice, and quite clear from context normally), I assume they mean `i8[0]`. I *think* the only disconnect after that is that I would never refer to variations using integers. I can only assume you're referring to the protobuf value of `type_variation_reference` (otherwise I have no idea where those numbers are coming from), but AFAICT the only context in which type parsing is done is in YAML extensions, where anchor/reference pairs don't exist yet. In extensions they have names, just like functions and user-defined type classes. Talking about any number except 0 would also be wildly ambiguous, since plans can use any set of anchor numbers they want.

Assuming we can agree that `i8[0]` is represented using type variation reference 0, and that there is no type variation anchor defined for variation 0 in the extension set because it's the default/base/0/whatever variation, expanding the name would syntactically make it either `[]` or ` `, so the latter is what I've been doing to cut back on noise. Hence, to me, `i8` can both be used to refer to the type class `i8`, and to the concrete type non-nullable `i8` with the default variation. A type with a non-default variation could for example be `string[arrow]`.

site/docs/types/type_system.md
jvanstraten: Okay, so (the integer vs name thing from the other thread aside) I *think* this just boils down to us misunderstanding each other, but both understanding the actual spec the same way. Phew :)

site/docs/types/compound_logical_types.md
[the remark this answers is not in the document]
jacques-n: I'd use "name" for the first part I guess.(?)

site/docs/types/type_system.md
jvanstraten: > If your statement is that "a non-nullable type instance can be promoted to a nullable type instance without loss of information", I agree with that statement. However, the specific behavior of things like you list need to specifically defined.

I'm using the word "promotion" to mean a something that happens automatically, i.e. a coercion, that is guaranteed to not lose information. For example, if I have a function that only takes nullable arguments, and i have a non-nullable type, I can still pass it to that function without an explicit cast, due to this coercion rule.

From this discussion I think your intentions however are that coercions should never happen. Which is fine, even preferable IMO. However, by the current specification if-else branches must then have exactly equal types, including nullability:

> When an if expression is declared, all return expressions must be the same identical type.

and since the spec says nothing about the expected type for the various conditions, a strict validator should probably reject anything that isn't a non-nullable boolean with the default variation. The "implicit equality functions" should then also return that type rather than being viral about nulls, i.e. null == 1 equals false instead of null. Again, that's perfectly consistent, I was just worried it might be *too* strict.

> I don't understand this question. A filter relation has no impact on the schema of the data. If the condition returns null, we should behave the same as databases do which I believe is to exclude the given record (we can't confirm that the filter condition passes).

If that's what databases do, then fine. I honestly wouldn't know, so I don't really have an opinion about what the behavior should be, other than that it needs to be specified. By the coercion logic I proposed, one might argue that a nullable filter condition makes the columns in the schema nullable and yields a row with all nulls if the condition yields null, which seems to be in line with the viral nature of null I see everywhere else. When the default logic is that conditions must be non-nullable unless otherwise specified there would also be no confusion, it'd just need an extra note mentioning that null means false in the context of a condition to get the behavior you describe.

site/docs/types/compound_logical_types.md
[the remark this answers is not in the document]
jvanstraten: Meh... The `STRUCT` part is more than just a name; it also encapsulates the semantics of the type and what the parameter pack (if any) means. "User-defined type name" also feels overly broad to me.

Do you have anything against type class? I suppose it's a bit of an overloaded term from object-oriented type systems, but at least Substrait's type system isn't using the word "class" for anything else that I'm aware of.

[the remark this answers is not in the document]
jacques-n: I'm ok with class.

site/docs/types/compound_logical_types.md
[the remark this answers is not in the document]
jacques-n: Since literals only contain literals, the wording makes more sense in that context than it does here. (For example, you can only put literals inside a map literal.)

site/docs/types/compound_logical_types.md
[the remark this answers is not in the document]
jacques-n: I think the issue here is canonical. Substrait has not opinions about the default or best representation of data. Using the word canonical sounds too much like "recommended" or "preferred" or ideal. Substrait doesn't portend to have any opinion about such things. So in most cases, 99.999% of all data related to a substrait plan has no default or standard representation. The only place where Substrait has to have an opinion is when it stores literals. This is unfortunate but unavoidable. But this is probably 0.001% or less of all data relating to a Substrait plan. Saying that the 0.001% part of things is "canonical" is misleading.

For reference, even in protobuf, the physical representation of int32 isn't actually what most people would think of as a int32 values. Instead, it's a Google varint encoded integer. The only thing int32 is the name of the type in the context of the protobuf idl and how the varint encoded value is exposed at the api level. So calling that a physical representation is also weird.

site/docs/types/simple_logical_types.md
jacques-n: I agree that we should codify 0 if not defined is "system preferred" or similar.

site/docs/types/type_system.md
jacques-n: >  However, by the current specification if-else branches must then have exactly equal types, including nullability:

I think we need to allow coercion specifically in those locations. 

> I don't really have an opinion about what the behavior should be

It some ways, this is great. It's a huge test of the specification. It's very rigorous. This is also a highly inefficient way of trying to formalize the specification. It'd be great if more people helped who had a common ground of the SQL specification and the great amount of commonalities the vast majority of these systems carry (and Substrait is just trying to reflect).

site/docs/types/compound_logical_types.md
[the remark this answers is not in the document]
jacques-n: By the way, nice little ascii graphic above :D

site/docs/expressions/casts.md
```diff
@@ -0,0 +1,3 @@
+# Casts
+
+Cast expressions may be used to convert between types. Behavior is currently unspecified (TBD).
```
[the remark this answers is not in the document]
jvanstraten: Also, actually, I've since realized since that casts were already listed in the "Specialized Record Expressions" page.

jacques-n:
It is very difficult to review this patch as the file renaming makes diff not be helpful for the largest set of changes. Can you please do the file rename as a separate commit so we can look at the diff to the files independent of the rename. (I tried locally and my local diff tool also can't automatically recognize that this is a rename + modify.)

jvanstraten:
I'm not sure how much it helped (the diff tools seem to just give up on the tables anyway), but here you go.

site/docs/types/type_system.md
```diff
@@ -0,0 +1,16 @@
+# Type System
+
+Substrait tries to cover the most common types used in data manipulation. To minimize type explosion, the project currently follows the guideline that a logical type should probably only be included in the specification if it is included in at least two of the following open source Projects: Apache Arrow, Apache Iceberg, Apache Spark and Trino. Types beyond this common core may be represented using [simple extensions](../extensions/index.md#simple-extensions).
```
jacques-n, line 3: Let's remove the "to minimize type explosion...trino" sentence.

site/docs/types/type_system.md
```diff
@@ -0,0 +1,16 @@
...
+| ------------------------------- | ------------------- | --------------------------------------------------------- | ----------------------------------------------------------------------------------------------------------------------------------------------------------------------------
+| [Class](type_classes.md)        | Always              | `i8`, `string`, `STRUCT`, extensions                      | Together with the parameter pack, describes the set of non-null values supported by the type. Subdivided into simple and compound type classes.
+| Nullability                     | Always              | Either `NULLABLE` (`?` suffix) or `REQUIRED` (no suffix)  | Describes whether values of this type can be null. Note that null is considered to be a special value of a nullable type, rather than the only value of a special null type.
+| [Variation](type_variations.md) | Always              | Either `[0]` (system-preferred) or an extension           | Allows different variations of the same type class to exist in a system at a time, usually distinguished by in-memory format.
```
jacques-n, line 11: It would be good to note that i8 we frequently skip expressing the variation when writing about the default variation.

site/docs/types/type_system.md
```diff
@@ -0,0 +1,16 @@
+# Type System
+
+Substrait tries to cover the most common types used in data manipulation. To minimize type explosion, the project currently follows the guideline that a logical type should probably only be included in the specification if it is included in at least two of the following open source Projects: Apache Arrow, Apache Iceberg, Apache Spark and Trino. Types beyond this common core may be represented using [simple extensions](../extensions/index.md#simple-extensions).
+
+Substrait types fundamentally consist of four components:
```
jacques-n, line 5: nice addition.

jacques-n (commented):
Mostly looks good. Thanks for splitting this out. It *did* make it easier to review.

jacques-n (approved):
LGTM. Thanks for the improvements @jvanstraten !

site/docs/types/type_system.md
jvanstraten: Added "no suffix" to the example list, does that capture it? b70f3d9

PR #202, opened 2022-05-18
Add orc file format

JkSelf:
Add orc file format.

### DIFF
--- a/proto/substrait/algebra.proto
+++ b/proto/substrait/algebra.proto
@@ -115,7 +115,7 @@ message ReadRel {
       enum FileFormat {
         FILE_FORMAT_UNSPECIFIED = 0;
         FILE_FORMAT_PARQUET = 1;
-        FILE_FORMAT_PARQUET = 2;
+        FILE_FORMAT_ORC = 2;
       }
     }
   }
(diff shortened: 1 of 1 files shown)


proto/substrait/algebra.proto
```diff
@@ -115,6 +115,7 @@ message ReadRel {
       enum FileFormat {
         FILE_FORMAT_UNSPECIFIED = 0;
         FILE_FORMAT_PARQUET = 1;
+        FILE_FORMAT_PARQUET = 2;
```
westonpace, line 118:
```suggestion
        FILE_FORMAT_ORC = 2;
```

westonpace (changes requested):

jacques-n:
I was thinking that we would change this structure per the comments on #169 and #183. @westonpace do you think you can push #169 forward? Then this could be layered in with a similar pattern.

westonpace:
I should be able to find some time to work on this today.

proto/substrait/algebra.proto
JkSelf: Thanks for your review. I have updated it to `FILE_FORMAT_ORC`.

jacques-n:
Hi @JkSelf , this basic change was included using a new pattern in #169.

PR #203, opened 2022-05-18
ci: remove breaking change detection

cpcloud:
This PR removes the buf breaking change detection until we hit 1.0.

### DIFF
--- a/.github/workflows/pr.yml
+++ b/.github/workflows/pr.yml
@@ -36,6 +36,25 @@ jobs:
       - uses: actions/checkout@v2
       - uses: bufbuild/buf-setup-action@v0.7.0
       - uses: bufbuild/buf-lint-action@v1
+      - name: check for breaking changes
+        id: check-breaking
+        run: |
+          set -euo pipefail
+
+          if ! buf breaking --against 'https://github.com/substrait-io/substrait.git#branch=main'; then
+            breaking="true"
+          else
+            breaking="false"
+          fi
+
+          echo "::set-output name=breaking::${breaking}"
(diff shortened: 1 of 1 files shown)


cpcloud:
cc @westonpace @jvanstraten @jacques-n

Was thinking we should disable this until 1.0, so that we don't have to trigger a major release while we're still iterating towards 1.0.

westonpace (dismissed):
Agreed.  Though I think, instead of disabling breaking change, we change it to require "feature" classification (e.g. bump the minor version).  This matches the semantic versioning scheme for 0.x releases.

cpcloud:
That makes sense.

jacques-n:
I'm -1 on disabling the breaking commit comment check. I think we should be deliberate about overriding. Having a failing precommit that we decide to override is better than accidentally introducing breaking changes when we didn't mean to. That being said... 

I think I'll start a thread on the mailing list. I don't think we should make 1.0 a thing. It's just a number and I don't see a great reason to have a different system before vs after. 

And if we do want to make 1.0 a thing, I think we should still document breaking changes as breaking even if we don't bump major version until later... Introducing a breaking change but not telling people in the release notes is a bad outcome from my pov.

westonpace:
+1 for hashing it out on the ML.

jvanstraten:
+1 for ML thread, but pending that, I guess +/-0 for this PR. I always thought that for 0.x.y x++ means breaking and y++ means anything else, but apparently semver [makes no assertions about that](https://semver.org/spec/v2.0.0.html#spec-item-4). TIL.

cpcloud:
Superseded by #210.

PR #209, opened 2022-05-25
Document version behavior

jvanstraten:
As discussed during the sync meeting, here's my view of versioning combined with what I think is the current consensus, as a starting point for further discussion. RFC


jacques-n (approved):
Thanks @jvanstraten

PR #212, opened 2022-05-25
docs: correct markdown formatting

westonpace:


cpcloud (dismissed):
Thank you!

cpcloud:
Is this intended to be `fix:`? Maybe better as `docs:`?

westonpace:
Changed to `docs:`.  I guess my brain was thinking "it fixes a bug in the docs" but I agree that `docs:` for corrections as well as new docs features makes more sense.

jvanstraten:
Just pointing out that the title was changed, not the commit message. Presumably conventional commits will end up looking at the latter instead of the former, considering the PR CI also lints the commits rather than the PR title?

cpcloud:
Since we have squash turned on and as the only option, the final commit that goes in is the one that GitHub generates from the PR title just before you click the "yes I definitely want to merge" button.

It's not clear to me which approach is better:

### Squash Model

#### Pros

1. Contributor commits can be whatever they want (we'd turn off the commitlint CI)

#### Cons

1. No intermediate commits enter into the semantic-release version computation.
2. Onus is on maintainers to make sure the final PR title conforms (may be automatable in a similar way to what we're doing with commitlint)

### Rebase and Merge Model

#### Pros

1. All commits enter into semantic release. If they conform, they're in. No risk of losing information.
2. Maintainers aren't responsible for making sure the final commit conforms.

#### Cons

1. All commits enter into semantic release. Risk of irrelevant changes causing a version bump. It's up to contributors to conform to the commit conventions.

I'm sure there are other pros and cons for each.

As a related aside, I personally do not consider a 1:1 commit to PR relationship all that useful. There are plenty of concise ways to get the commits associated with a given PR using `gh` or the web UI so I don't consider that a pro of the squash model or a con of the rebase and merge model.

jacques-n:
The bigger downside to me of rebase & merge is two parts: 
- No guarantee that all commits pass CI.
- Makes it more complex to revert a patch.

> I personally do not consider a 1:1 commit to PR relationship all that useful. There are plenty of concise ways to get the commits associated with a given PR using gh or the web UI so I don't consider that a pro of the squash model or a con of the rebase and merge model.

I prefer not to depend on Github UIs/CLIs to be able to understand the composition of a changeset. I mostly use non-Github owned tools to interact with Git and will often merge patches without using the Github UI.

Isn't there a way for the release notes tool to incorporate the information from the consolidated commit message?

cpcloud:
> I prefer not to depend on Github UIs/CLIs to be able to understand the composition of a changeset.

I personally have not found this to be a problem in practice. If it is, then it might be worth considering splitting up the PRs into smaller changesets. I realize this isn't always possible.

> Isn't there a way for the release notes tool to incorporate the information from the consolidated commit message?

I don't think this makes sense. Consider all the problems that arise when mixing commits:

What would the squashed commit message be if there were a fix and feature in a single PR? What if one of them is breaking and one of them isn't? What they are mixed together with non-version-bumping changes? What if there's a revert of a previous commit with a fix or feature (this is handled automatically by semantic release without squashing)

Semantic release is designed around the philosophy that a commit's effect on the version is not multitudinous.

I think I'm now pretty -1 on squashing.

jacques-n:
Lol, I'm pretty -1 on not squashing. Every project I work on squashes. Calcite, Arrow, Iceberg, internal projects now and in the past... 

I'm running around today so can't really provide the time against this until at least tomorrow. My high level is if squash loses meaning, it should me multiple prs.

westonpace:
> What would the squashed commit message be if there were a fix and feature in a single PR?
> What if one of them is breaking and one of them isn't? What if they are mixed together with non-version-bumping changes?

Don't do that.  That should be two PRs.

> What if there's a revert of a previous commit with a fix or feature (this is handled automatically by semantic release without squashing)

What does semantic release do differently in this case?  Also, see my above point :)

I'm +1 on squashing because not squashing makes it difficult to do reviews because the PR is constantly rebasing.  For example, consider the workflow:

 * User makes a big PR
 * Reviewer asks for some changes
 * User addresses changes
 * Reviewer checks to see if the changes were properly addressed
 
 In that final step it is often very valuable to see just the changes that happened since the last review.  If we are forcing users to constantly do their own squashing then that will be impossible to see.

cpcloud:
> What does semantic release do differently in this case?

It'll introduce a patch version: https://semantic-release.gitbook.io/semantic-release/support/faq#how-can-i-revert-a-release

> Don't do that. That should be two PRs.

Isn't that effectively two commits then? What's the difference? All the non-semantic commits get squashed?

> User makes a big PR

Hm, I'm tempted to say "don't do that" 😄.

> If we are forcing users to constantly do their own squashing then that will be impossible to see.

No one would be required to squash (except possibly just before merging if desired). I realize that sounds heretical, but in my experience it hasn't be a problem in practice if people are making reasonably sized PRs.

cpcloud:
Maybe a GitHub issue isn't the best place to have this discussion.

I personally would rather not squash, but if we're not going to it's up to people merging PRs to make sure that the meaningful commits in a PR are reflected in the final title.

I suppose you could build something (or maybe someone already has!) to handle squash merges, but you'd probably be rebuilding most of semantic release and having to solve the problems I listed above to do so.

jvanstraten:
Weak +1 on squash-merging. I personally actually prefer everything being a proper merge so no history rewriting/amending/rebasing is necessary, but I also don't think it's all that reasonable for a first-time contributor to understand what our commit message conventions are. Squash-merging puts the onus on only those who have write privileges.

If we do only squash-merge, CI should IMO be modified to not nag about commit messages in PRs, since then it doesn't really matter.

> User makes big PR

😶

westonpace:
I have fixed the commit message (wasn't sure if that was needed) so we can merge this.  Feel free to continue the discussion on the ML.  As long as what we pick is clear I think we can move forwards.

I do think a squash approach is simpler for new contributors.  I am not volunteering to rewrite all of semantic release though :laughing: so I'm content with a non-squash approach.  I'll update my vote to +0.

cpcloud (approved):

cpcloud:
I'm not content with squash, but it's not a hill I'm willing to die on 😂. Let's just continue with squash for now. We can always reevaluate after more contributions start rolling in.

westonpace:
@cpcloud GH should use the issue description as the commit message footer.

PR #217, opened 2022-06-01
feat: support nullable and non-default variation user-defined types

jvanstraten:
As mentioned on slack (and encountered in the validator, though I'd forgotten about it apparently), it's currently impossible to specify nullability for user-defined types. It was also impossible to specify variations, which *I* see no reason to disallow, but there was no conclusion about that yet. If we do think that should be illegal, I'll also update the type docs in this PR to mention this (and why).

### DIFF
--- a/proto/substrait/parameterized_types.proto
+++ b/proto/substrait/parameterized_types.proto
@@ -38,7 +38,12 @@ message ParameterizedType {
     ParameterizedList list = 27;
     ParameterizedMap map = 28;
 
-    uint32 user_defined_pointer = 31;
+    ParameterizedUserDefined user_defined = 30;
+
+    // Deprecated in favor of user_defined, which allows nullability and
+    // variations to be specified. If user_defined_pointer is encountered,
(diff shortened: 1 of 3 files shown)


jacques-n:
I'm good with this but let's leave open for a couple days to see if anyone has a disagreement with it.

jacques-n:
@cpcloud @westonpace are you good with this change happening?

westonpace (approved):
Makes sense to me.

jvanstraten:
I guess it's bad form to push to a PR that's already under review and more-or-less accepted, but I had some realizations of my own:

 - I named the user-defined type reference field `user_defined_type_reference` without really thinking about it, while in `algebra.proto` and `extensions.proto` the field is juse named `type_[reference|anchor]`. Better to just keep things as consistent as possible.
 - Updated the type expression and parameterized type protos accordingly, which I didn't think to do until I started looking at them more thoroughly today.

jacques-n (approved):

PR #219, opened 2022-06-02
feat: add aggregate function min/max support

weijietong:


extensions/functions_arithmetic.yaml
```diff
@@ -276,3 +276,105 @@ aggregate_functions:
...
+  - name: "min"
+    description: Min a set of values.
+    impls:
+      - args:
+          - options: [ SILENT, SATURATE, ERROR ]
+            required: false
+          - value: i8
+        nullability: DECLARED_OUTPUT
+        decomposable: MANY
+        intermediate: i8
```
jinfengni, line 288: Seems both the `intermediate` and `return` type should be `i8?`, indicating that a nullable i8.  

Same applied to other impls.

extensions/functions_arithmetic.yaml
```diff
@@ -276,3 +276,105 @@ aggregate_functions:
         decomposable: MANY
         intermediate: "STRUCT<fp64,i64>"
         return: fp64?
+  - name: "min"
+    description: Min a set of values.
+    impls:
+      - args:
+          - options: [ SILENT, SATURATE, ERROR ]
```
jacques-n, line 283: I don't see the purpose of multiple overflow behaviors for min. How would min ever overflow?

extensions/functions_arithmetic.yaml
weijietong: Got it. BTW do we need to define another function definication that the `return` type is not nullable ?

weijietong: Got it. BTW do we need to define another function definication that the `return` type is not nullable ? The [specification](https://substrait.io/extensions/) shows that the functoin signature (which is the key in the yaml file) doesn't include the return type.

extensions/functions_arithmetic.yaml
weijietong: I just got the meaning of the `options` which indicates the behavior of a function overflow. So will remove it.

extensions/functions_arithmetic.yaml
weijietong: done

extensions/functions_arithmetic.yaml
weijietong: done

jacques-n:
It looks like your commit doesn't match the [expected commit pattern](https://www.conventionalcommits.org/en/v1.0.0-beta.2/). This is a new feature.

jacques-n (approved):
LGTM. Thanks @weijietong !

PR #220, opened 2022-06-02
Adding Join Suffix

vibhatha:
This PR is intended to include the suffixes for join operations.

### DIFF
--- a/proto/substrait/algebra.proto
+++ b/proto/substrait/algebra.proto
@@ -148,6 +148,9 @@ message JoinRel {
     JOIN_TYPE_SINGLE = 7;
   }
 
+  string left_suffix = 8;
+  string right_suffix = 9;  
+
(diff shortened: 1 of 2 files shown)


cpcloud:
This seems like a UX concern for producers to handle rather than something that Substrait needs to model. What's the motivation for including it in Substrait?

jacques-n:
I don't know what a suffix is. The docs update doesn't do a good job of explaining the purpose. 

Searching around it seems like this has to do with field naming(?). Substrait plan nodes don't deal with names, only ordinal positions so if that is the purpose of this patch, it seems like there is a misunderstanding of how substrait works.

vibhatha:
@jacques-n this is a misunderstanding of the usage. Thanks a lot for pointing out. I closing this one.

PR #221, opened 2022-06-07
docs: add terminology section to website

jvanstraten:


jvanstraten:
Context:

I've (somewhat reluctantly) started looking into tackling the documentation for function definitions and type expressions today. My intention was to open a draft PR with comments added to all fields and messages in [function|parameterized_types|type_expressions].proto (after adding missing fields) as a basis for discussion, because I see so many problems with it that any discussion would instantly become a mess without threads. But then I got stuck doing even that, because of inconsistent and missing terminology. For example, the protos use "variable" and "argument" interchangeably, [when they actually mean "parameter"](https://stackoverflow.com/questions/156767/whats-the-difference-between-an-argument-and-a-parameter). Meanwhile, the word "parameter" is used both for named bindings in derivation expressions and the parameters of a type. I know, argument and parameter are usually used interchangeably, but you just can't do that in a context like this without things instantly becoming confusing.

I also wasn't sure how to put this in an issue in a way that facilitates discussion, so here it is via PR instead. Likewise, I ended up defining the terms in a YAML file and compiling that to a mkdocs page to sort the terms link everything up (the generated markdown is just checked in for now).

Consider the terms I used to just be a proposal; I'm fine with swapping things around, but I do really want unambiguous terms for all of them, so I don't have to roll my own terminology in the validator code and have it be inconsistent all over the place. Once we settle on something, I'll be happy to go over all the existing files in the repo to refactor as needed.

site/docs/spec/terminology.yaml
```diff
@@ -0,0 +1,230 @@
...
+
+type variation: >-
+  a specific variation of a {type class}, allowing different variations of the
+  same {type class} to exist in a system at a time. Usually this is set to the
+  {system-preferred variation} for the {type class}.
+system-preferred variation: >-
+  a special {type variation} that is implicitly defined for each {type class},
+  serving as a default value for when {type variation}s are not in use.
+
+type parameter pack: >- # wasn't really used
```
jacques-n, line 142: Couple of initial thoughts here on this and much below this:
- Trying to finalize things we agree on and don't know if agree on in the same patch might slow down the merging of this patch.
- You started using parameter when talking about functions. I basically said I'm only going to use parameter for type related stuff and arguments for function related stuff. This is a slight manipulation of the terms but avoids the need here and above of constantly having "function type parameter". By using both, it actually makes things much more confusing to me. Especially when you didn't use "type type" below. So now we have type argument here but we also have type arguments in functions.
- I think that you've collapsed data types and parameterized types together which I think is suboptimal. There are three independent type abstractions, one building on the next. Type literals, parameterized types and derivation expressions. They all deal with types, they are at just different levels of capability (expressibility). Each can only be used at a certain location within the simple extensions.
- One of the things that is especially confusing when reading your decomposition of this is what appears to be a mixing of compound types and parameterized types. Compound types are type literals that express no additional abstract concepts. For example, `VARCHAR<40>`. This is discrete from a parameterized type such as `VARCHAR<T>` or a type class e.g. `VARCHAR`. In a function declaration, I cannot use compound types for bindings (e.g. `VARCHAR<40>`). I can only parameterized types (e.g. `VARCHAR<T>`). The parameter can be used or ignored. In parameterized types I can also use the `anyX` pseudo-type parameter (X is an optional numeric identifier after any) which allows is a wildcard which is reusable. This is useful when you want to say something like `add(any1, any1) => any1`. This is a function which says I can "add" two values of any type to get a new version of that type. The any is a wildcard but it is the same wildcard everytime it is referenced.

site/docs/spec/terminology.yaml
```diff
@@ -0,0 +1,230 @@
...
+  [here](../expressions/scalar_functions.md).
+aggregate function: >-
+  a {function} that computes a single {datum} for a complete {dataset}, as
+  defined [here](../expressions/aggregate_functions.md).
+window function: >-
+  a {function} that computes a {datum} for each row of a {dataset}, making use
+  of some window of surrounding rows in addition to the row it is computed for,
+  as defined [here](../expressions/window_functions.md).
+
+function declaration: >-
```
jacques-n, line 70: I've been calling the whole function function declaration and the specific implementation function signature. It seems like you're missing the latter concept?

site/docs/spec/terminology.yaml
jacques-n: I will take a more thorough pass to your concepts in a bit.

jacques-n (commented):
@jvanstraten , I agree that your pattern is nicer. However I want to avoid creating technical debt or custom systems around managing the website. I'm also not supportive of committing generated website content in the core repo. Can you figure out a way to make this work within the existing paradigm as opposed to tacking an additional manual process on? 

I'm assuming that mkdocs has all sorts of plugin apis that could enable this functionality if you think it is important. To me, if it is important to have the functionality, let's make sure we do it in a way that matches the build, serving, editing, etc system that we use for the rest of the website. (For example, I use `mkdocs serve` right now to peruse the changes as I make them to the website. Let's make sure this works within that construct.)

Alternatively, we separate this into two steps: one is about content (just a simple markdown doc) and then separately is a fancier mkdocs plugin you add that makes it look nicer/smarter.

Additionally, as I reviewed the doc, I think it is important to actually make this readable from a user's perspective. For example, the information around function argument, enum function argument, value function argument, etc should probably be shown hierarchially (and there seem to be many places where showing the information hierarchically would be a vast improvement over a simple list. So I wonder if what we should be supporting is something much simpler than you're formatting tool. E.g. something along the lines of "hey, link this marked term to the part of the doc that has a matching id" (which I assume we can get mkdocs to generate if doesn't already).

site/docs/spec/terminology.yaml
```diff
@@ -0,0 +1,230 @@
...
+type function parameter: >- # changed from "type argument"
+  a {function parameter} to which a {type function argument} can be bound.
+  The {data type} passed to the argument must match the {data type pattern}
+  defined along with the parameter.
+enum function parameter: >- # changed from "enumeration"
+  a {function parameter} to which an {enum function argument} can be bound.
+  The allowed values are defined along with the parameter by means of a set
+  of string options and an optional default option.
+
+function binding: >- # didn't really exist
```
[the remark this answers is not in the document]
jvanstraten: Those are fundamentally different things. Binding happens during compilation, and involves connecting the parameters in the function definition (or actually declaration/prototype in this case, since Substrait doesn't know the function definition) to argument expressions to build a complete expression tree. Invocation (aka calling a function) is the process of actually executing the body of a function to evaluate its return value, once all its dependencies (argument expressions and/or program state for impure functions) have been evaluated, either at runtime or during constant propagation. After binding, an expression tree with a function binding in it may be modified during optimization, but the bound function generally can't be invoked yet. The relationships are `definition 1-->* binding 1-->* invocation`.

FWIW, in the context you linked the difference between "binding" and "invocation" is like "addition" and "add" (if anything), so I don't think the distinction is important enough to require a refactor there, in the same way that you can get away with mixing "parameter" and "argument" in 99% of cases. It's the 1% that I'm trying to cover.

jvanstraten:
I hate to say it, but I think I'm just going to have to give up on type derivation and everything related to it.

 - It's been abundantly clear for months now that I don't understand type derivation as it's currently formulated.
 - I don't know how to even begin to properly understand it if I can't rely on basic compiler/DSL design terminology to be used correctly.
 - I can't personally define a terminology section to try to remedy this, because I don't understand the existing terminology to begin with, and replacing the current terminology with something that's consistent with literature is apparently not desirable.
 - I can't ask questions about type expressions and get answers I'd understand if I don't even understand the applicable terminology, though in general my current understanding is insufficient for me to formulate sane questions that can be efficiently answered.
 - I can't implement or improve the documentation of something I don't innately understand, at least not without continuing to resort to this extremely time-consuming and demoralizing trial-and-error process of submitting something and hoping I got it right. It works -- sort of -- for small things, but, when properly defined and sufficiently descriptive, I seriously expect type derivations to be about 1/3rd of the spec.
 - It very much seems like I'm the only one in the community who has a problem with type derivations and the terminology it (and the rest of the spec) relies on, so I'm forced to conclude that the problem is with me to begin with, and not with the specification.

Unfortunately that means that the validator won't get support for them for the foreseeable future (for what that's worth).

This PR specifically seems pretty hopeless to me since we fundamentally disagree on both the terminology itself and the approach for representing that terminology on the website. It'd be a lot more efficient if someone with opinions closer to yours would draft this up instead -- assuming I'm not the only one who thinks this is important to have in the first place.

site/docs/spec/terminology.yaml
cpcloud: "Compound" as a descriptor for `VARCHAR<40>` seems like non-standard type system terminology.

`VARCHAR<40>` is as fully specified a type as `int32`. Calling `VARCHAR<40>` "compound" would be like saying `int32` is a compound version of `int<N>` (assume that exists for argument's sake).

> Compound types are type literals that express no additional abstract concepts.

By this definition every monomorphized type including any type that doesn't have a parameter (`int32`, `float64`, etc) is a compound type. If everything is a compound type it doesn't really make sense to keep it around as a distinct concept or term.

jvanstraten: FWIW, Substrait has used the term "simple" to refer to all types that don't need a `<...>` pack to be concrete, and "compound" to refer to all types that do, [since its conception](https://github.com/substrait-io/substrait/blob/c7f7ef80e70f5b758dfab604ee06dbfb4ddfe020/doc/types_logical_compound.md). But I agree. "Compound" feels synonymous to "nested" to me.

PR #222, opened 2022-06-08
feat: adding temporal functions

rok:
This adds new scalar functions to work with timestamp/date/time types.

### DIFF
--- a/extensions/functions_datetime.yaml
+++ b/extensions/functions_datetime.yaml
@@ -1,11 +1,13 @@
 %YAML 1.2
 ---
 scalar_functions:
-  - name: extract
+  -
+    name: extract
     description: Extract portion of a date/time value.
     impls:
       - args:
-          - options: [ YEAR, QUARTER, MONTH, DAY, DAY_OF_WEEK, DAY_OF_YEAR, ISO_WEEK, IS_LEAP_YEAR, HOUR, SECOND, SUBSECOND, MILLISECOND, MICROSECOND, NANOSECOND ]
+          - options: [ YEAR, QUARTER, MONTH, DAY, DAY_OF_WEEK, DAY_OF_YEAR, ISO_WEEK, IS_LEAP_YEAR, HOUR, SECOND,
+                       SUBSECOND, MILLISECOND, MICROSECOND, NANOSECOND ]
             name: The part of the value to extract.
             required: true
           - value: timestamp
@@ -17,7 +19,8 @@ scalar_functions:
           - value: timestamp
         return: boolean
       - args:
-          - options: [ YEAR, QUARTER, MONTH, DAY, DAY_OF_WEEK, DAY_OF_YEAR, ISO_WEEK, IS_LEAP_YEAR, HOUR, SECOND, SUBSECOND, MILLISECOND, MICROSECOND, NANOSECOND, IS_DST ]
+          - options: [ YEAR, QUARTER, MONTH, DAY, DAY_OF_WEEK, DAY_OF_YEAR, ISO_WEEK, IS_LEAP_YEAR, HOUR, SECOND,
+                       SUBSECOND, MILLISECOND, MICROSECOND, NANOSECOND, IS_DST ]
             name: The part of the value to extract.
             required: true
           - value: timestamp_tz
@@ -46,7 +49,8 @@ scalar_functions:
             required: true
           - value: time
         return: i64
-  - name: "add"
+  -
+    name: "add"
     description: Add an interval to a date/time type.
     impls:
       - args:
@@ -73,7 +77,8 @@ scalar_functions:
           - value: date
           - value: interval_day
         return: timestamp
-  - name: "add_intervals"
+  -
+    name: "add_intervals"
     description: Add two intervals together.
     impls:
       - args:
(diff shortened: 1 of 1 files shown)


extensions/functions_datetime.yaml
```diff
@@ -1,36 +1,52 @@
...
         return: i64
       - args:
-          - options: [ YEAR, MONTH, DAY, SECOND ]
+          - options: [ IS_LEAP_YEAR ]
+            name: The part of the value to extract.
+            required: true
+          - value: timestamp
+        return: boolean
+      - args:
+          - options: [ YEAR, QUARTER, MONTH, DAY, DAY_OF_WEEK, DAY_OF_YEAR, ISO_WEEK, IS_LEAP_YEAR, HOUR, SECOND, SUBSECOND, MILLISECOND, MICROSECOND, NANOSECOND, IS_DST ]
```
jacques-n, line 20: Can you add definitions and what common data tools use each to a comment somewhere? That will make it easier to evaluate whether things should be included.

extensions/functions_datetime.yaml
```diff
@@ -1,36 +1,52 @@
...
             name: The part of the value to extract.
             required: true
           - value: timestamp_tz
         return: i64
       - args:
-          - options: [ YEAR, MONTH, DAY ]
+          - options: [ IS_LEAP_YEAR, IS_DST ]
```
jacques-n, line 26: I don't know what IS_DST means in the context of a instance in time.

extensions/functions_datetime.yaml
```diff
@@ -57,8 +73,7 @@ scalar_functions:
           - value: date
           - value: interval_day
         return: timestamp
-  -
-    name: "add_intervals"
+  - name: "add_intervals"
```
jacques-n, line 76: Why did you reformat all of these? Let's not do reformatting changes like this. To be consistent, if we think they are important let's add to a linter check as part of precommit.

extensions/functions_datetime.yaml
```diff
@@ -193,3 +203,66 @@ scalar_functions:
...
+          - options: [ TIMEZONE ]
+            name: Timezone of provided local time.
+            required: true
+          - value: timestamp
+        return: timestamp_tz
+  - name: "strptime"
+    description: Parse string into timestamp/date/time using provided format.
+    impls:
+      - args:
+          - options: [ FORMAT ]
```
jacques-n, line 219: Same as above, this seems to be a misunderstanding of options.

extensions/functions_datetime.yaml
```diff
@@ -193,3 +203,66 @@ scalar_functions:
...
+    description: Convert local time to UTC using given local time's timezone.
+    impls:
+      - args:
+          - options: [ TIMEZONE ]
+            name: Timezone of provided local time.
+            required: true
+          - value: timestamp
+        return: timestamp_tz
+  - name: "strptime"
+    description: Parse string into timestamp/date/time using provided format.
```
jacques-n, line 216: We need to declare the format rules this function uses.

extensions/functions_datetime.yaml
```diff
@@ -193,3 +203,66 @@ scalar_functions:
...
+    impls:
+      - args:
+          - options: [ FORMAT ]
+            name: Expected format of the input timestamp/date/time.
+            required: true
+          - value: string
+        return: timestamp
+      - args:
+          - options: [ FORMAT ]
```
jacques-n, line 225: This isn't allowed: two variations of the same function with the same input argument types. Both above and below are option + string.

jacques-n (changes requested):
Several items need to be addressed.

rok:
@jacques-n apologies I got distracted. Will address Monday.

extensions/functions_datetime.yaml
rok: Most likely. If options are enums, how do I bring in a string parameter?

extensions/functions_datetime.yaml
rok: Would that go into description here? I would start by adopting https://man7.org/linux/man-pages/man3/strptime.3.html

extensions/functions_datetime.yaml
rok: Removing.

extensions/functions_datetime.yaml
```diff
@@ -193,3 +203,66 @@ scalar_functions:
...
+            name: Expected format of the input timestamp/date/time.
+            required: true
+          - value: string
+        return: time
+  - name: "strftime"
+    description: Convert timestamp/date/time to string using provided format
+    impls:
+      - args:
+          - options: [ FORMAT ]
```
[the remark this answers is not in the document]
rok: Agreed. Will do.

jacques-n:
Closing this due to inactivity. Please reopen if you want to proceed.

rok:
@jacques-n I have pushed some changes, but I can't reopen the PR (permissions). Can you please reopen and review?

jacques-n:
It looks like I can't reopen after you force pushed. 
image

I suggest just opening a new pr and referring to this one (or you could try force-pushing your old changes back, reopening, then overpushing). I'm not sure why Github makes this complicated...

rok:
@jacques-n switching to #272

PR #223, opened 2022-06-09
fix: add overflow behavior to integer division

westonpace:
When working on https://github.com/apache/arrow/pull/13285 we noticed that there is no overflow behavior for division.  My first thought was that overflow is not possible on division but it turns out that there is one special case of overflow when the data type is signed integers.

For example, consider 8 bit signed integers.  The range is -128,127.  However, `-128 / -1 => 128` which is outside the range.  I'm not entirely sure what `SATURATE` would mean in this case but it seems it could mean 127 for the simplicity of having one set of overflow handling options for all the arithmetic.

### DIFF
--- a/extensions/functions_arithmetic.yaml
+++ b/extensions/functions_arithmetic.yaml
@@ -126,18 +126,26 @@ scalar_functions:
     description: "Divide one value by another. Partial values are truncated."
     impls:
       - args:
+          - options: [ SILENT, SATURATE, ERROR ]
+            required: false
           - value: i8
           - value: i8
         return: i8
       - args:
+          - options: [ SILENT, SATURATE, ERROR ]
+            required: false
           - value: i16
           - value: i16
         return: i16
       - args:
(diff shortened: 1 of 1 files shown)


westonpace:
CC @sanjibansg who discovered this discrepency

jacques-n:
Good catch. Do we need to add to decimal as well?

As a note, I don't see the the introduction of an optional argument to a function as a breaking change so consider this fix identification as a valid non-breaking change. (I think you went through this as well but want to call out explicitly.)

jacques-n (changes requested):
Looking for same changes wrt to Decimal.

westonpace:
@jacques-n I'm not sure I understand.  Division for decimal already specifies overflow I think: https://github.com/substrait-io/substrait/blob/a79eb07a15cfa157e795f028a83f746967c98805/extensions/functions_arithmetic_decimal.yaml#L57

PR #225, opened 2022-06-15
feat: support placeholders for prepared statements

jvanstraten:
Adds support for placeholders/plan parameters akin to SQL prepared statements. See also [this](https://substrait.slack.com/archives/C02D7CTQXHD/p1655280726253919) thread on the Substrait slack.

### DIFF
--- a/site/docs/expressions/specialized_record_expressions.md
+++ b/site/docs/expressions/specialized_record_expressions.md
@@ -8,7 +8,8 @@ These constructs should be focused on different expression types as opposed to s
 ## Literal Expressions
 For each data type, it is possible to create a literal value for that data type. The representation depends on the serialization format. Literal expressions include both a type literal and a possibly null value.
 
-
+## Placeholder Expressions
+Placeholders allow plans to be parameterized akin to prepared statements in SQL. A placeholder expression behaves like a literal, but rather than encoding the value of the literal, it just encodes a reference to a plan parameter. The parameter in turn encodes type information and a reference to where the data should come from. The parameter may also be given a literal value, either to serve as a default value, or to make a parameterized plan executable without actually replacing the placeholders with literals.
 
 ## Cast Expression
 To convert a value from one type to another, Substrait defines a cast expression. Cast expressions declare an expected type, an input argument and an enumeration specifying failure behavior, indicating whether cast should return null on failure or throw an exception.
--- a/proto/substrait/algebra.proto
+++ b/proto/substrait/algebra.proto
@@ -357,6 +357,7 @@ message Expression {
     MultiOrList multi_or_list = 9;
     Cast cast = 11;
     Subquery subquery = 12;
+    Placeholder placeholder = 13;
 
     // deprecated: enum literals are only sensible in the context of
     // function arguments, for which FunctionArgument should now be
@@ -806,6 +807,13 @@ message Expression {
       // right side of the expression
       Rel right = 4;
     }
+  }
+
+  // A placeholder expression acts like a parameter passed to the plan, to
+  // support things like prepared statements in SQL.
+  message Placeholder {
+    // Points to a plan_parameter_anchor defined in this plan.
+    uint32 plan_parameter_reference = 1;
   }
 }
 
--- a/proto/substrait/plan.proto
+++ b/proto/substrait/plan.proto
@@ -5,6 +5,7 @@ package substrait;
 
 import "substrait/algebra.proto";
 import "substrait/extensions/extensions.proto";
+import "substrait/type.proto";
 
 option csharp_namespace = "Substrait.Protobuf";
 option go_package = "github.com/substrait-io/substrait-go/proto";
@@ -18,6 +19,28 @@ message PlanRel {
     Rel rel = 1;
     // The root of a relation tree
     RelRoot root = 2;
+  }
+}
+
+// A parameter for a plan, to support things like prepared statements in SQL.
+message PlanParameter {
+  // An arbitrary integer used to refer to this parameter in the plan.
+  // Required.
+  uint32 plan_parameter_anchor = 1;
+
+  // The reference used to refer to this parameter outside of the plan.
+  // Required.
+  oneof reference {
+    string named = 2;
+    int32 indexed = 3;
+  }
+
+  // The data type or value of the parameter. Required. A value may be passed
+  // either to serve as a default or to make the plan executable without losing
+  // the distinction between parameter and literal.
+  oneof type_or_value {
+    Type type = 4;
+    Expression.Literal value = 5;
   }
 }
 
@@ -42,4 +65,11 @@ message Plan {
   // unused. In many cases, a consumer may be able to work with a plan even if
   // one or more message types defined here are unknown.
   repeated string expected_type_urls = 5;
+
+  // An optional list of parameters used in the plan, to support things like
+  // prepared statements in SQL. When nonempty, the plan can only be executed
+  // when all parameters have a value (either passed along with the parameter
+  // or via another channel), but it *can* be optimized or otherwise
+  // manipulated.
+  repeated PlanParameter parameters = 6;
 }


ashvina:
LGTM 👍. Thanks !
I think it would be helpful to add an example/test case but I am not sure what would be the best way to do that. For e.g. a test case with a parameterized query which validates the generated plan would make sense to me.
I don't intend to add a task on your plate. I am just trying to think what should future PRs contribute and document w.r.t. schema change contributions.

site/docs/expressions/specialized_record_expressions.md
```diff
@@ -8,7 +8,8 @@ These constructs should be focused on different expression types as opposed to s
 ## Literal Expressions
 For each data type, it is possible to create a literal value for that data type. The representation depends on the serialization format. Literal expressions include both a type literal and a possibly null value.
 
-
+## Placeholder Expressions
```
jacques-n, line 11: Can you add some information about what you think the cycle will look like here? It sounds like you're going to:

- submit a plan with placeholders where the placeholder is populated with types instead of values
- resubmit the same plan with the same plaeholders but they are instead populated by values, mapping using the reference oneof with a name or index.

This feels awkward. It seems like we need a new message which is "here are the parameters for a previously defined plan". Since it is directly related to the previous plan, it seems like it'd just be something like:
```
message PlanBind {
  repeated ParameterBind = 1;
}

message ParameterBind {
  uint32 plan_parameter_reference = 1;
  Expression.Literal value = 2;
}
```

It also seems like there is no need for any of the reference oneof within `PlanParameter` nor the value branch of the type_or_value oneof.

jvanstraten: My intention with this was to be flexible in order to make it easy for existing systems to use Substrait, rather than to restrict things to the bare minimum needed to make something work. That being said, here's some use cases for these parameters:

 - A sends parameterized plan to B. B stores and perhaps pre-optimizes the plan akin to a SQL prepared statement. Later, A sends arguments to B to execute the plan. The latter would (still) be out of band for Substrait, or might perhaps use a relation extension that references a previously submitted/optimized plan and binds the parameters; I don't think this is really worth thinking about until we work on a protocol of some kind in general (notably, your suggestion lacks any way to refer to the stored plan, so it'd be unusable without out-of-band information anyway). The reference field allows for named and/or positional arguments, depending on how B natively stores and binds arguments to parameters; again (still) out of the scope of Substrait. It feels weird to me to use anchors for this, extend their use beyond a single message.
 - As above but with predefined default values bound to the parameters.
 - Different flow: A sends parameterized plan to B, B has the information needed to fill out the arguments (previously submitted, computed, stored as secrets of some kind, whatever), and then forwards the plan to C for execution. By allowing values to bind to the parameters directly, B doesn't need to concern itself with rewriting the whole relation tree to do the binding, which would otherwise require knowledge of all possible expression and relatiom types in Substrait, including any advanced extensions that may be in use.
 - Tracking information related to where literals came from by using parameters with values, perhaps for debugging purposes or to rebind later.
 - Deduplicating large literal values used in the plan several times. In fact, it might be worth adding something more general for this that includes subexpressions or relation subtrees.

None of these things are possible without the reference oneof and/or the value variant, except maybe the first, if a preexisting consumer is expected to just do away with named arguments. We support column names at the periphery of the plan despite not using them internally, so it seemed fitting to allow for the same thing here.

The relevance of the numeric reference is kinda disputable since we already have anchor numbers and an index into the repeated parameter field. I suggested it this way as a separation of concerns typs of thing: the reference interfaces with the outside world, the anchor interfaces with the relation tree, and (same as for the extension declarations) the index into the outer repeated field is of no significance.

jacques-n: I think that we're overloading this to try to solve a bunch of things and should narrow this down and deliberately include functionality for each particular thing. My interpretation of this particular thing is that it is similar to a JDBC/ODBC prepared statements statements. Is this the main use case or is it something else?

> A sends parameterized plan to B. B stores and perhaps pre-optimizes the plan akin to a SQL prepared statement. Later, A sends arguments to B to execute the plan. The latter would (still) be out of band for Substrait, or might perhaps use a relation extension that references a previously submitted/optimized plan and binds the parameters; I don't think this is really worth thinking about until we work on a protocol of some kind in general (notably, your suggestion lacks any way to refer to the stored plan, so it'd be unusable without out-of-band information anyway). The reference field allows for named and/or positional arguments, depending on how B natively stores and binds arguments to parameters; again (still) out of the scope of Substrait. It feels weird to me to use anchors for this, extend their use beyond a single message.

I definitely think that converting a parameterized plan to a fully resolved plan should be a concern of Substrait. Defining the concept of `Placeholdered Plan + PlaceHolder values => Fully Resolved Plan` feels like something we should define formally in Substrait as opposed to elsewhere. I'm fine if the envelope is defined externally (for thing like which plan are we talking about). 

> It feels weird to me to use anchors for this, extend their use beyond a single message.

I agree a little bit. On the other hand, I feel like introducing a new concept here feels like it just makes the plan consumption side more complex.

> As above but with predefined default values bound to the parameters.

Do you have an end-user scenario for default values? This definitely doesn't exist in the places where I've seen this pattern (SQL systems). (If this is important, I'd be more inclined that you should just use coalesce rather than introduce a new property of a placeholder.)

> Different flow: A sends parameterized plan to B, B has the information needed to fill out the arguments (previously submitted, computed, stored as secrets of some kind, whatever), and then forwards the plan to C for execution. By allowing values to bind to the parameters directly, B doesn't need to concern itself with rewriting the whole relation tree to do the binding, which would otherwise require knowledge of all possible expression and relatiom types in Substrait, including any advanced extensions that may be in use.

You have two different concepts here: easy replacement and secrets.
* For easy replacement, I'd rather offer this up as a library function people can use than having to introduce a special structure just for this purpose.
* For secrets, I'm be inclined to model those as functions just like something such as "give me the current user". This is how most systems model these kinds of things. At some iteration of plan processing, the function call has constant reduction applied to it and turns into the resolved value.  

> Tracking information related to where literals came from by using parameters with values, perhaps for debugging purposes or to rebind later.

I don't think this makes sense as a first class concern. If we agree it is important, I'd rather just introduce a new annotation expression type. Additionally, trying to track provenance of parts of expression trees is a complex and painful thing due to rewrites, etc. I suggest we avoid making promises around provenance tracking/knowledge at the spec level.


> Deduplicating large literal values used in the plan several times. In fact, it might be worth adding something more general for this that includes subexpressions or relation subtrees.

This feels like a completely separate idea. Let's not overload this with that. (Especially since we'd want to have that be done consistently.) FWIW, I think we already have this ability mostly with relational subtree (we're missing a subtree-reference relation). We don't have it for expressions but I could imagine where it would probably belong. That being said, I'd question whether expression collapsing like that is over-engineering versus just relying on a simple wrapper compression codec...

jvanstraten: > I think that we're overloading this to try to solve a bunch of things and should narrow this down and deliberately include functionality for each particular thing. My interpretation of this particular thing is that it is similar to a JDBC/ODBC prepared statements statements. Is this the main use case or is it something else?

I guess so. I submitted this in response to [Max Goisser's question on Slack](https://substrait.slack.com/archives/C02D7CTQXHD/p1655280726253919) because it seemed like low-hanging fruit to fix the problem rather than telling them to awkwardly work around it with niladic functions. Note also that they're asking for string-based parameters, while pointing out that some systems use integers as well. I then intentionally generalized the idea to make it more universally applicable, because in my experience, that leads to much better specifications or implementations than focusing on one perceived use case at a time. However, I don't think it's really worth discussing this at this point (and certainly not buried in a PR review thread), and I should just yield, so I won't respond to the rest of what you wrote (I read it, though).

However, when distilling this to the absolute bare minimum Substrait would need to support this without sideband information (the latter referring to your mention of this during the latest sync meeting), supporting this is not *at all* low-hanging fruit, because Substrait plans lack any method to be referred to later on. How would a bound plan refer to a previously prepared plan? An integer seems way too naive. A string seems difficult -- how would a producer generate it? A GUID relies on randomness to actually be unique, and thus isn't perfect. In fact, as long as Substrait is only defined as a unidirectional and stateless producer to consumer link (if that), this can never be guaranteed to work; it's always possible (and trivial) to find a collision. Unless, of course, you use the plan itself as the reference (which the consumer could then hash to do a cache lookup, if it wants), with values bound.

Alternatively, we could specify that the unique identifiers produced by the producer only need to be unique within the context of a given producer/consumer pair, which would give the producer the information it needs to generate sufficiently unique identifiers. However, this would force the consumer to key the prepared statements by means of (producer, identifier) pairs, which seems nontrivial, and also prevents a producer from using a statement prepared by a different producer.

The more *correct* method is for the consumer to generate some unique token, and respond to the producer with a unique ID that it can use to reference the plan later. That might end up being something simple like a sequence number, or might be something much more involved; I don't know. My point is that it can't currently be specified. At best, I could use an `Any` for the plan reference. Or I guess we could do what SQL implementations probably tend to do and abuse the regular tabular result, in which case I guess the backreference would confusingly be a `substrait.Expression.Literal`.

So I see four things that I can/could have do here:

1) ~~Best-effort specification: define something that will work today, using a mechanic that will (as you rightfully point out) will become pretty niche once a protocol is defined, but at least is dead simple to implement for a consumer.~~
2) Least-effort specification: define something that will be incomplete and therefore unusable for the foreseeable future (i.e. until someone proposes a bidirectional Substrait protocol), using an `Any` to refer back to a prepared plan since we don't know yet what this reference will look like.
3) (Hypothetically) a mega PR that specifies this along with a protocol (but I have no intention of doing this).
4) Give up and close this PR, and
   a) recommend using niladic functions after all, which will obviously not ever be widely supported; or
   b) simply say Substrait doesn't support this at this time and they should either wait a few years or cook up their own protocol.

None of the remaining things seem particularly appealing to me, so I guess I'm open to suggestions.

jvanstraten:
Closing this due to inactivity. If someone else has a need for this feature and holds strong opinions about how they should be implemented, please open an issue or new PR.

PR #226, opened 2022-06-16
docs: fix spec components links

guykhazma:
Minor fix to the links to point to the right sections.
I kept the existing [redirect map](https://github.com/substrait-io/substrait/blob/fdf7c19b58c60a13157983e00c5964625206b077/site/mkdocs.yml#L49) since I am not sure whether it is in use somewhere. 
If it's not used I think it would be better to avoid the redirect map and point directly to the right subsection.   
For example, we could use directly `types/type_classes/#compound-types` instead of `/types/compound_logical_types#compound-types`

Note that the redirect plugin doesn't support redirecting to a specific subsection (see [here](https://github.com/mkdocs/mkdocs-redirects/issues/16))

### DIFF
--- a/site/docs/spec/specification.md
+++ b/site/docs/spec/specification.md
@@ -19,10 +19,10 @@ Once all portions of the specification have been moved to commit (or eliminated)
 
 | Priority | Status | Section                                                      | Description                                                  |
 | -------- | ------ | ------------------------------------------------------------ | ------------------------------------------------------------ |
-| 1        | wip    | [Simple Types](/types/simple_logical_types#simple-types)                  | A way to describe the set of basic types that will be operated on within a plan. Only includes simple types such as integers and doubles (nothing configurable or compound). |
-|          | wip    | [Compound Types](/types/compound_logical_types#compound-types)              | Expression of types that go beyond simple scalar values. Key concepts here include: configurable types such as fixed length and numeric types as well as compound types such as structs, maps, lists, etc. |
+| 1        | wip    | [Simple Types](/types/type_classes#simple-types)                  | A way to describe the set of basic types that will be operated on within a plan. Only includes simple types such as integers and doubles (nothing configurable or compound). |
+|          | wip    | [Compound Types](/types/type_classes#compound-types)              | Expression of types that go beyond simple scalar values. Key concepts here include: configurable types such as fixed length and numeric types as well as compound types such as structs, maps, lists, etc. |
 |          | wip    | [Type Variations](/types/type_variations)                    | Physical variations to base types.                           |
-|          | sketch | [User Defined Types](/types/user_defined_types#user-defined-types)              | Extensions that can be defined for specific IR producers/consumers. |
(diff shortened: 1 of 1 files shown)


site/docs/spec/specification.md
```diff
@@ -19,10 +19,10 @@ Once all portions of the specification have been moved to commit (or eliminated)
...
-| 1        | wip    | [Simple Types](/types/simple_logical_types)                  | A way to describe the set of basic types that will be operated on within a plan. Only includes simple types such as integers and doubles (nothing configurable or compound). |
-|          | wip    | [Compound Types](/types/compound_logical_types)              | Expression of types that go beyond simple scalar values. Key concepts here include: configurable types such as fixed length and numeric types as well as compound types such as structs, maps, lists, etc. |
+| 1        | wip    | [Simple Types](/types/simple_logical_types#simple-types)                  | A way to describe the set of basic types that will be operated on within a plan. Only includes simple types such as integers and doubles (nothing configurable or compound). |
+|          | wip    | [Compound Types](/types/compound_logical_types#compound-types)              | Expression of types that go beyond simple scalar values. Key concepts here include: configurable types such as fixed length and numeric types as well as compound types such as structs, maps, lists, etc. |
```
jvanstraten, line 23:
```suggestion
| 1        | wip    | [Simple Types](/types/type_classes#simple-types)                  | A way to describe the set of basic types that will be operated on within a plan. Only includes simple types such as integers and doubles (nothing configurable or compound). |
|          | wip    | [Compound Types](/types/type_classes#compound-types)              | Expression of types that go beyond simple scalar values. Key concepts here include: configurable types such as fixed length and numeric types as well as compound types such as structs, maps, lists, etc. |
```

site/docs/spec/specification.md
```diff
@@ -19,10 +19,10 @@ Once all portions of the specification have been moved to commit (or eliminated)
...
 |          | wip    | [Type Variations](/types/type_variations)                    | Physical variations to base types.                           |
-|          | sketch | [User Defined Types](/types/user_defined_types)              | Extensions that can be defined for specific IR producers/consumers. |
+|          | sketch | [User Defined Types](/types/user_defined_types#user-defined-types)              | Extensions that can be defined for specific IR producers/consumers. |
```
jvanstraten, line 25:
```suggestion
|          | sketch | [User Defined Types](/types/type_classes#user-defined-types)              | Extensions that can be defined for specific IR producers/consumers. |
```

jvanstraten (changes requested):
Ah right, I just forgot to update those links in #200. The redirect map is there to make sure that if a third party has a link to one of the old pages it will at least redirect to something related rather than 404'ing, but internally we should just use the right links.

That whole page is kinda outdated, by the way. Take it with a big grain of salt.

guykhazma:
Thanks for the heads up @jvanstraten, I am going over the documentation first and will then will dive deeper into the code.
I have fixed the links.

guykhazma:
@jvanstraten FYI in case you expect me to merge this, I don't have permissions to do so

jvanstraten:
I don't either. I believe @jacques-n or @cpcloud have to do that.

jacques-n (approved):

PR #227, opened 2022-06-16
feat: add modulo operation

guykhazma:
This PR adds the modulo operation to the list of possible binary operations.

In general, the list seems to be lacking multiple operators like bitwise shift operations (see the list for [postgres](https://www.postgresql.org/docs/7.4/functions-math.html) and [calcite](https://github.com/apache/calcite/blob/b9c2099ea92a575084b55a206efc5dd341c0df62/linq4j/src/main/java/org/apache/calcite/linq4j/tree/ExpressionType.java#L28)).
I have started with the modulo operation as it seem to be the most basic one which is missing but I can add the rest as well.

### DIFF
--- a/proto/substrait/type_expressions.proto
+++ b/proto/substrait/type_expressions.proto
@@ -145,6 +145,7 @@ message DerivationExpression {
       BINARY_OP_TYPE_OR = 10;
       BINARY_OP_TYPE_EQUALS = 11;
       BINARY_OP_TYPE_COVERS = 12;
+      BINARY_OP_TYPE_MODULO = 13;
     }
   }
 
--- a/site/docs/expressions/scalar_functions.md
+++ b/site/docs/expressions/scalar_functions.md
@@ -95,7 +95,7 @@ Any function can declare a return type expression. A return type expression uses
 These types are evaluated using a small set of operations to support common scenarios. List of valid operations:
 
 ```
-Math: +, -, *, /, min, max
+Math: +, -, *, /, % (modulo), min, max
 Boolean: &&, ||, !, <, >, ==
 Parameters: type, integer
 Literals: type, integer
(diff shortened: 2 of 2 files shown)


guykhazma:
I see that the modulo function is defined in the extensions module (see [here](https://github.com/substrait-io/substrait/blob/fdf7c19b58c60a13157983e00c5964625206b077/extensions/functions_arithmetic.yaml#L157)).
So there seems to be a need for a mechanism that would make sure that the functions defined in the extensions can be expressed with the proto definitions.

guykhazma:
I understand that the proto is used to deduce the return type of a function, but once you decide to define different types of binary operations as different proto messages it makes sense to include also to include modulo.
Otherwise, why not use a generic BinaryOpType without the enum? 

With regards to the extension, I may be missing something, let me try and explain what i meant.
I didn't meant using extensions to be used for return type computation, what I meant is that for the operators that have been identified as worthy enough to be included in the enum of `BinaryOpType` it makes sense to expect the serialization of to use the corresponding `BinaryOpType`. If nothing enforces that then similar to my question above why have this enum in the first place?

jacques-n:
Do you have a concrete use case for modulo in type derivation? The goal was to keep the capabilities of type derivation expressions as narrow as possible to minimize development burden. If you do have a case, can you share the yaml definition of the function and output derivation function that uses such a function. As @jvanstraten mentioned, just because we can come up with an example doesn't mean it is actually necessary to introduce it. 

> So there seems to be a need for a mechanism that would make sure that the functions defined in the extensions can be expressed with the proto definitions.

I'm sorry, I don't understand this sentence. 

> for the operators that have been identified as worthy enough to be included in the enum of BinaryOpType it makes sense to expect the serialization of to use the corresponding BinaryOpType. 

I'm sorry, I don't understand this sentence.

guykhazma:
I understand my confusion now, thanks, I mixed expressions with return type derivation as @jvanstraten pointed out.
I don't have a concrete use case which seems reasonable in order to introduce the modulo operation in type derivation so I'll close this PR.

PR #228, opened 2022-06-16
feat: add functions to yaml

rui-mo:
This PR added starts_with, ends_with, contains functions to yaml files.

### DIFF
--- a/extensions/functions_string.yaml
+++ b/extensions/functions_string.yaml
@@ -44,3 +44,122 @@ scalar_functions:
           - value: i32
           - value: i32
         return: "string"
+  - name: starts_with
+    description: Whether this string starts with another string.
+    impls:
+      - args:
+          - value: "varchar<L1>"
+          - value: "varchar<L2>"
+        return: "BOOLEAN"
+      - args:
+          - value: "varchar<L1>"
+          - value: "string"
+        return: "BOOLEAN"
+      - args:
+          - value: "varchar<L1>"
+          - value: "fixedchar<L2>"
+        return: "BOOLEAN"
+      - args:
+          - value: "string"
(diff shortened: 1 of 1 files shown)


extensions/functions_comparison.yaml
```diff
@@ -73,4 +73,12 @@ scalar_functions:
           - value: any1
         return: BOOLEAN
         nullability: DECLARED_OUTPUT
-
+  -
+    name: "in"
+    description: Whether a value is in a list of values.
+    impls:
+      - args:
+          - value: any1
```
jvanstraten, line 81: I think you meant this?

```suggestion
          - value: T
```

jvanstraten (changes requested):

jacques-n:
I think you're looking for the orlist expression (as opposed to introducing in).

For the character matching functions, I suggest you also include all permutations of fixed character, string and varchar combinations.

extensions/functions_comparison.yaml
rui-mo: The function we need is already covered by OrList expression. Removed this.

rui-mo:
> I think you're looking for the orlist expression (as opposed to introducing in).
> 
> For the character matching functions, I suggest you also include all permutations of fixed character, string and varchar combinations.

Thanks for your comment. Removed in function and added the permutations for starts_with, ends_with, contains.

jacques-n:
Please also update commit title. Should be feat and more specific

rui-mo:
> Please also update commit title. Should be feat and more specific

Updated. Thank you.

jacques-n (approved):
Looks mostly good. I will update the title to be more explicit when I merge.

PR #230, opened 2022-06-22
feat: add functions for arithmetic, rounding, logarithmic, and string transformations

sanjibansg:
This PR adds definitions of extension functions in Arithmetic, Rounding, Logarithmic & String Transformations.

### DIFF
--- a/extensions/functions_arithmetic.yaml
+++ b/extensions/functions_arithmetic.yaml
@@ -253,13 +253,13 @@ scalar_functions:
     impls:
       - args:
           - value: i8
-        return: fp32
+        return: fp64
       - args:
           - value: i16
-        return: fp32
+        return: fp64
       - args:
           - value: i32
-        return: fp32
+        return: fp64
       - args:
           - value: i64
         return: fp64
--- a/extensions/functions_arithmetic.yaml
+++ b/extensions/functions_arithmetic.yaml
@@ -154,6 +154,40 @@ scalar_functions:
           - value: fp64
         return: fp64
   -
+    name: "negate"
+    description: "Negation of the value"
+    impls:
+      - args:
+          - options: [ SILENT, SATURATE, ERROR ]
+            required: false
+          - value: i8
+        return: i8
+      - args:
+          - options: [ SILENT, SATURATE, ERROR ]
+            required: false
+          - value: i16
+        return: i16
+      - args:
+          - options: [ SILENT, SATURATE, ERROR ]
+            required: false
+          - value: i32
+        return: i32
+      - args:
+          - options: [ SILENT, SATURATE, ERROR ]
+            required: false
+          - value: i64
+        return: i64
+      - args:
+          - options: [ SILENT, SATURATE, ERROR ]
+            required: false
+          - value: fp32
+        return: fp32
+      - args:
+          - options: [ SILENT, SATURATE, ERROR ]
+            required: false
+          - value: fp64
+        return: fp64
+  -
     name: "modulus"
     description: "Get the remainder when dividing one value by another."
     impls:
@@ -173,6 +207,68 @@ scalar_functions:
           - value: i64
           - value: i64
         return: i64
+  -
+    name: "power"
+    description: "Take the power with the first value as the base and second as exponent"
+    impls:
+      - args:
+          - options: [ SILENT, SATURATE, ERROR ]
+            required: false
+          - value: i8
+          - value: i8
+        return: i8
+      - args:
+          - options: [ SILENT, SATURATE, ERROR ]
+            required: false
+          - value: i16
+          - value: i16
+        return: i16
+      - args:
+          - options: [ SILENT, SATURATE, ERROR ]
+            required: false
+          - value: i32
+          - value: i32
+        return: i32
+      - args:
+          - options: [ SILENT, SATURATE, ERROR ]
+            required: false
+          - value: i64
+          - value: i64
+        return: i64
+      - args:
+          - options: [ SILENT, SATURATE, ERROR ]
+            required: false
+          - value: fp32
+          - value: fp32
+        return: fp32
+      - args:
+          - options: [ SILENT, SATURATE, ERROR ]
+            required: false
+          - value: fp64
+          - value: fp64
+        return: fp64
+  -
+    name: "sqrt"
+    description: "Square root of the value"
+    impls:
+      - args:
+          - value: i8
+        return: fp32
+      - args:
+          - value: i16
+        return: fp32
+      - args:
+          - value: i32
+        return: fp32
+      - args:
+          - value: i64
+        return: fp64
+      - args:
+          - value: fp32
+        return: fp32
+      - args:
+          - value: fp64
+        return: fp64
 aggregate_functions:
   - name: "sum"
     description: Sum a set of values.
--- a/extensions/functions_arithmetic.yaml
+++ b/extensions/functions_arithmetic.yaml
@@ -178,13 +178,9 @@ scalar_functions:
           - value: i64
         return: i64
       - args:
-          - options: [ SILENT, SATURATE, ERROR ]
-            required: false
           - value: fp32
         return: fp32
       - args:
-          - options: [ SILENT, SATURATE, ERROR ]
-            required: false
           - value: fp64
         return: fp64
   -
--- a/extensions/functions_comparison.yaml
+++ b/extensions/functions_comparison.yaml
@@ -78,7 +78,9 @@ scalar_functions:
     description: Whether a value is not a number.
     impls:
       - args:
-          - value: any1
+          - value: fp32
         return: BOOLEAN
-        nullability: DECLARED_OUTPUT
+      - args:
+          - value: fp64
+        return: BOOLEAN
 
--- a/extensions/functions_comparison.yaml
+++ b/extensions/functions_comparison.yaml
@@ -73,4 +73,12 @@ scalar_functions:
           - value: any1
         return: BOOLEAN
         nullability: DECLARED_OUTPUT
+  -
+    name: "is_nan"
+    description: Whether a value is not a number.
+    impls:
+      - args:
+          - value: any1
+        return: BOOLEAN
+        nullability: DECLARED_OUTPUT
 
--- a/extensions/functions_logarithmic.yaml
+++ b/extensions/functions_logarithmic.yaml
@@ -1,0 +1,63 @@
+%YAML 1.2
+---
+scalar_functions:
+  -
+    name: "ln"
+    description: "Natural logarithm of the value"
+    impls:
+      - args:
+          - options: [ SILENT, SATURATE, ERROR ]
+            required: false
+          - value: fp32
+        return: fp32
+      - args:
+          - options: [ SILENT, SATURATE, ERROR ]
+            required: false
+          - value: fp64
+        return: fp64
+  -
+    name: "log10"
+    description: "Logarithm to base 10 of the value"
+    impls:
+      - args:
+          - options: [ SILENT, SATURATE, ERROR ]
+            required: false
+          - value: fp32
+        return: fp32
+      - args:
+          - options: [ SILENT, SATURATE, ERROR ]
+            required: false
+          - value: fp64
+        return: fp64
+  -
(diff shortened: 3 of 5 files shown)


extensions/functions_arithmetic.yaml
```diff
@@ -153,6 +153,40 @@ scalar_functions:
           - value: fp64
           - value: fp64
         return: fp64
+  -
+    name: "negate"
+    description: "Negation of the value"
+    impls:
+      - args:
+          - options: [ SILENT, SATURATE, ERROR ]
```
jacques-n, line 161: Ugh... I suppose this exists because of the uneveness of twos-compliment but ugh...

extensions/functions_arithmetic.yaml
```diff
@@ -173,6 +207,68 @@ scalar_functions:
           - value: i64
           - value: i64
         return: i64
+  -
+    name: "power"
+    description: "Take the power with the first value as the base and second as exponent"
+    impls:
+      - args:
```
jacques-n, line 214: Where did you come up with these signatures from? For example, why does i8^i8 return i8? It seems like we'll run out of space very quickly. I feel like many systems support something like `power(i64, fp64) => fp64` and `power(fp64,fp64) => fp64` and that is enough.

extensions/functions_arithmetic.yaml
```diff
@@ -173,6 +207,68 @@ scalar_functions:
...
+          - options: [ SILENT, SATURATE, ERROR ]
+            required: false
+          - value: fp64
+          - value: fp64
+        return: fp64
+  -
+    name: "sqrt"
+    description: "Square root of the value"
+    impls:
+      - args:
```
jacques-n, line 254: Is there a particular reason that smaller numbers use fp32 as an output? It seems like maybe they should all be fp64. Are using a particular system's definition here?

extensions/functions_comparison.yaml
```diff
@@ -73,4 +73,12 @@ scalar_functions:
           - value: any1
         return: BOOLEAN
         nullability: DECLARED_OUTPUT
+  -
+    name: "is_nan"
+    description: Whether a value is not a number.
+    impls:
+      - args:
+          - value: any1
```
jacques-n, line 81: This doesn't seem like the right argument type. Should we just have two of these: one for fp32 and one for fp64?

extensions/functions_comparison.yaml
```diff
@@ -73,4 +73,12 @@ scalar_functions:
...
+        return: BOOLEAN
+        nullability: DECLARED_OUTPUT
```
jacques-n, line 83: Why is this declared output? Shouldn't nan be null if the input is null?

extensions/functions_rounding.yaml
```diff
@@ -0,0 +1,106 @@
...
+---
+scalar_functions:
+  -
+    name: "ceil"
+    description: "Rounding to the ceiling of the value"
+    impls:
+      - args:
+          - options: [ SILENT, SATURATE, ERROR ]
+            required: false
+          - value: i8
```
jacques-n, line 11: I don't know what rounding an i8 (or any of the i* values below) means.

extensions/functions_rounding.yaml
```diff
@@ -0,0 +1,106 @@
...
+        return: i64
+  -
+    name: "round"
+    description: "Rounding off the value"
+    impls:
+      - args:
+          - options: [ SILENT, SATURATE, ERROR ]
+            required: false
+          - value: i8
+        return: i8
```
jacques-n, line 80: Same as above, what does this mean?

extensions/functions_rounding.yaml
```diff
@@ -0,0 +1,106 @@
...
+    name: "ceil"
+    description: "Rounding to the ceiling of the value"
+    impls:
+      - args:
+          - options: [ SILENT, SATURATE, ERROR ]
+            required: false
+          - value: i8
+        return: i8
+      - args:
+          - options: [ SILENT, SATURATE, ERROR ]
```
jacques-n, line 14: Why would I have overflow behaviors for ceil?

extensions/functions_rounding.yaml
```diff
@@ -0,0 +1,106 @@
...
+            required: false
+          - value: fp64
+        return: i64
+  -
+    name: "floor"
+    description: "Rounding to the floor of the value"
+    impls:
+      - args:
+          - options: [ SILENT, SATURATE, ERROR ]
```
jacques-n, line 43: Why would I have overflow behaviors for floor?

extensions/functions_rounding.yaml
```diff
@@ -0,0 +1,106 @@
...
+            required: false
+          - value: i8
```
jacques-n, line 45: What is the floor of an i8? For example, what is the floor of 7?

extensions/functions_string.yaml
```diff
@@ -163,3 +164,23 @@ scalar_functions:
           - value: "fixedchar<L1>"
           - value: "varchar<L2>"
         return: "BOOLEAN"
+  -
+    name: lower
+    description: Transforms the string to lower case characters
```
jacques-n, line 169: Let's add some additional definition here of how lower case is defined. I assume there is a utf definition?

extensions/functions_string.yaml
```diff
@@ -163,3 +164,23 @@ scalar_functions:
...
+    impls:
+      - args:
+          - value: "varchar<L1>"
```
jacques-n, line 172: Shall we add fixedchar?

extensions/functions_string.yaml
```diff
@@ -163,3 +164,23 @@ scalar_functions:
...
+        return: "varchar<L1>"
+      - args:
+          - value: "string"
+        return: "string"
+  -
+    name: upper
```
jacques-n, line 178: same comments as on lower.

jacques-n (changes requested):
Please address comments provided.

extensions/functions_arithmetic.yaml
westonpace: I agree.  Arrow also upcasts all to fp64.

extensions/functions_arithmetic.yaml
```diff
@@ -173,6 +207,68 @@ scalar_functions:
...
+        return: fp32
+      - args:
+          - options: [ SILENT, SATURATE, ERROR ]
+            required: false
+          - value: fp64
+          - value: fp64
+        return: fp64
+  -
+    name: "sqrt"
+    description: "Square root of the value"
```
westonpace, line 252: What happens if the value is negative?

extensions/functions_comparison.yaml
westonpace: It appears Arrow supports all numeric types but for anything other than float or double the return is hard-coded to false.  I'd be onboard with only supporting floating point types.

extensions/functions_comparison.yaml
westonpace: I don't have strong feelings (null is not nan) but Arrow seems to agree with @jacques-n:

```
>>> pc.is_nan(None)
<pyarrow.BooleanScalar: None>
```

extensions/functions_logarithmic.yaml
```diff
@@ -0,0 +1,63 @@
+%YAML 1.2
+---
+scalar_functions:
+  -
+    name: "ln"
```
westonpace, line 5: It would be good to document what these options apply to (and their meaning) somewhere.  In arrow, arithmetic functions overflow by default.  However, logrithmic functions go to `-inf` or `NaN`.  I'm not sure if this latter behavior is "saturate" or "silent".

extensions/functions_rounding.yaml
westonpace: Arrow does not support integer round explicitly.  However, we will auto-cast an integer column to a double column and then call round on that.  I think this has led to some confusion in the docs.  I agree with @jacques-n that these kernels should not exist.

extensions/functions_string.yaml
westonpace: Arrow defers to utf8proc which is somewhat sparse on details.  However, you are correct, there is a standard UTF-8 way of lowercasing, though it [sometimes does the wrong thing semantically](https://github.com/JuliaLang/julia/issues/35615).

extensions/functions_arithmetic.yaml
westonpace: @jacques-n do you have a strong preference which direction we go here?  Should we have multiple `power` functions?

westonpace:
@jacques-n curious on your opinion here.  A number of these functions are floating point functions and in Arrow we have two variants.  A checked version which returns an error and an unchecked version which returns NaN.  For example, dealing with negative numbers in `sqrt`.  Do we consider returning `NaN` to be `SATURATE` or `SILENT`?  Or should we call it something else?  Does both `SATURATE` and `SILENT` apply here?

jvanstraten:
No one asked me, but signalling NaNs by definition are error markers, so ignoring the insertion of one should hopefully be "silent", not "saturate". Also, if anything, all floating point operations that don't have domain issues saturate to +/- infinity on overflow or +/- 0 on underflow; put differently, overflow does not exist in the first place, because positive and negative infinities can be represented. So basically, I don't think trying to coerce the semantics of unsigned and two's complement integer arithmetic into IEEE 754 arithmetic is very sensible. What *might* be sensible though is to add the rounding options that IEEE 754 itself defines. I'd propose something like

```yaml
- name: rounding
  options: [ TIE_TO_EVEN, TIE_AWAY_FROM_ZERO, TRUNCATE, CEILING, FLOOR ]
  required: false
- name: on_domain_error
  options: [ NAN, ERROR ]
  required: false
```

if we want to be thorough, with the latter only defined for operations that are actually affected by domain errors.

cpcloud:
> @jacques-n curious on your opinion here. A number of these functions are floating point functions and in Arrow we have two variants. A checked version which returns an error and an unchecked version which returns NaN. For example, dealing with negative numbers in `sqrt`. Do we consider returning `NaN` to be `SATURATE` or `SILENT`? Or should we call it something else? Does both `SATURATE` and `SILENT` apply here?

I agree with @jvanstraten's assessment here. It's not really possible to say whether an operation should be saturating or silent given only the knowledge that it returns NaN. Even if we could, as Jeroen said saturation doesn't make a whole lot of sense for IEEE floating point numbers with valid domains.

westonpace:
+1 from me as well for that answer.  Thanks @jvanstraten .  I have one question regarding rounding.  Is that needed for the case where the inputs/outputs are integers?  Or is this some kind of floating point rounding rule for the case where the exact answer falls between two valid floating point numbers due to limits in precision?

jvanstraten:
> Or is this some kind of floating point rounding rule for the case where the exact answer falls between two valid floating point numbers due to limits in precision?

It's that; this isn't about rounding to some nearby integer but about rounding to one of the two nearest possible representations.

I suppose the same options could be used for float -> int, though the `CEILING` and `FLOOR` options would need some additional specification for the boundary conditions (e.g. does 127.5 with `CEILING` round to i8 127 or does it trigger overflow behavior?). Note that i32/i64 -> fp32 and i64 to fp64 also need this rule because the mantissa isn't large enough to represent all possible values precisely, but i8/i16 -> fp32 and i8/i16/i32 -> fp64 never need to round.

Also, in case anyone is unsure:

 - `TIE_TO_EVEN`: choose the nearest possible representation. If the desired value is exactly halfway between the representations, tie toward the nearest even integer.
 - `TIE_AWAY_FROM_ZERO`: choose the nearest possible representation. If the desired value is exactly halfway between the representations, tie toward the one with the larger magnitude.
 - `TRUNCATE`: if the desired value is not exactly representable, choose the one that is closest to zero.
 - `CEILING`: if the desired value is not exactly representable, choose the one that is closest to positive infinity.
 - `FLOOR`: if the desired value is not exactly representable, choose the one that is closest to negative infinity.

ianmcook:
This PR should be closed. It is superseded by #245 and #322.

sanjibansg:
> This PR should be closed. It is superseded by other PRs including #245, #267, #322.

Sure, agreed.

PR #235, opened 2022-06-27
Initial DDL implementation

curino:
This work was done in collaboration with @jcamachor, and it is an initial proposal to be discussed, evolved and documented. This code adds `protobuf` specification for:

1. `CREATE TABLE [AS]`
2.  `CREATE VIEW`
3. `INSERT [OR REPLACE]`
4. `DELETE`
5. `UPDATE`

### DIFF
--- a/proto/substrait/algebra.proto
+++ b/proto/substrait/algebra.proto
@@ -902,28 +902,34 @@ message AggregateFunction {
     AGGREGATION_INVOCATION_DISTINCT = 2;
   }
 
-  // A CREATE TABLE operation
+  // A CREATE TABLE [AS] operation
   message CreateTable {
     // the table to insert into
     NamedTable table = 1;
+    // the schema of the table (with types, nullability,
+    // defaults for each column)
     NamedStruct base_schema = 2;
-    // A query (for the CTAS semantics)
+    // An optional query (for the CTAS semantics)
     // with named fields matching all columns that
-    // do not have detaults in the table
+    // do not have defaults in the table
     optional RelRoot query = 3;
+
+    //TODO: add constraints
   }
 
   // A CREATE VIEW AS operation
   message CreateView {
     // the table to insert into
     NamedView table = 1;
-    NamedStruct base_schema = 2;
-    // The query defining the body of the view
-    // all fields types in the query must match the base_schema
+    // an optional schema (if different than the query output)
+    optional NamedStruct base_schema = 2;
+    // The query defining the body of the view all fields
+    // types in the query must match the base_schema if present
     RelRoot query = 3;
   }
 
-  // An Insert operation
+
+  // An INSERT [OR REPLACE] operation
   message Insert {
     // the table to insert into
     NamedTable table = 1;
@@ -935,7 +941,7 @@ message AggregateFunction {
     RelRoot query = 3;
   }
 
-  // An Delete operation
+  // An DELETE operation
   //NOTE: we are not supporting multi-table deletes (only available in MySQL)
   message Delete {
     // the table to insert into
@@ -944,7 +950,7 @@ message AggregateFunction {
     Rel tup_delete = 2;
   }
 
-  // An Update operation
+  // An UPDATE operation
   message Update {
     // the table to update
     NamedTable table = 1;
--- a/proto/substrait/algebra.proto
+++ b/proto/substrait/algebra.proto
@@ -914,7 +914,7 @@ message AggregateFunction {
     // do not have defaults in the table
     optional RelRoot query = 3;
 
-    //TODO: add constraints
+    //TODO: add constraints, indexes, etc.?
   }
 
   // A CREATE VIEW AS operation
@@ -926,8 +926,9 @@ message AggregateFunction {
     // The query defining the body of the view all fields
     // types in the query must match the base_schema if present
     RelRoot query = 3;
-  }
 
+    //TODO: add materialization and other options
+  }
 
   // An INSERT [OR REPLACE] operation
   message Insert {
@@ -948,6 +949,8 @@ message AggregateFunction {
     NamedTable table = 1;
     // query that returns the tuples to delete (that must be part of table)
     Rel tup_delete = 2;
+
+    //TODO: add cascading semantics
   }
 
   // An UPDATE operation
@@ -955,21 +958,20 @@ message AggregateFunction {
     // the table to update
     NamedTable table = 1;
     // Expression to assign new value to one (or more) columns
-    repeated Assignment assignments = 1;
+    repeated Assignment assignments = 2;
     // query that returns the tuples to be modified (that must be part of table)
-    ReadRel tup_delete = 2;
+    Rel tup_update = 3;
   }
-
 
   // A modifying operation
   message ModOp {
     oneof operation {
-      CreateTable crtab = 1;
-      CreateView crview = 2;
-      Insert ins = 3;
-      Delete del = 4;
-      Update upd = 5;
-      // consider adding UPSERT semantics
+      CreateTable create_table = 1;
+      CreateView create_view = 2;
+      Insert insert = 3;
+      Delete delete = 4;
+      Update update = 5;
+      //TODO: consider adding UPSERT/MERGE semantics
     }
   }
 }
--- a/proto/substrait/algebra.proto
+++ b/proto/substrait/algebra.proto
@@ -55,6 +55,21 @@ message RelCommon {
   }
 }
 
+// A base table. The list of string is used to represent namespacing (e.g., mydb.mytable).
+// This assumes shared catalog between systems exchanging a message.
+message NamedTable {
+  repeated string names = 1;
+  substrait.extensions.AdvancedExtension advanced_extension = 10;
+}
+
+// A view (materialized or virtual). The list of string is used to represent namespacing (e.g., mydb.myview).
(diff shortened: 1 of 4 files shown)


curino:
There is a cyclic dependency between `type.proto` and `algebra.proto`. It is somewhat easily resolvable by moving the `Literal` definition into `type.proto`, except for 3 cross-references to `Type.List empty_list`, `Type.Map empty_map`, and `Type null`. With @jcamachor we think those could (maybe should?) be replaced with a `Literal.EmptyList`, `Literal.EmptyMap`, and `Literal.Null` dedicated objects.  Opinions?

jacques-n:
Several thoughts here.

1) I don't think it makes sense to introduce these operations at the root level. I think in all cases we can use a new relation.
2) For CTAS, INSERT and MERGE, I think we should use the preexisting [Write](https://substrait.io/relations/logical_relations/#write-operator) relation and add an enumeration: CREATE, CREATE_OR_APPEND, APPEND, APPEND_OR_UPDATE.
3) For Update, I suggest we introduce a new relation. We can Use an array of expressions for the update clause. (The assignment message is from a totally different system and is something to avoid.)
4) For the delete case, again, think this should just be a new relation. The question I have here is what is the semantic? Do we return key fields, full records?
4) I suggest that we have a default value that can optionally specified at the write relation level as a struct literal. If specified, the expectation would be the values in the struct that are not populated from the output would be populated with the values specified. Rather than introduce some kind of mapping system, we could simply say that the relation that is the input must match the default value schema. This feels like it would keep things simple.

curino:
@jacques-n,  Damn you are fast! Thanks for the quick feedback!   

1. Sounds good, happy to go with that.

2. Happy to try to implement the version you have (`Write` is not in the `protobuf`, which we discussed as the issue of spec vs `protobuf` issue before, but we can add it first and then specialize it).  So if we understand you correctly you would see `INSERT/DELETE/UPDATE` returning a single-row, single-column table containing the number of affected rows (common) and `CREATE TABLE` returning something like a single-row, single-column table with a boolean if the operation succeeded? (the latter is not very common/standardized at least as far as few min googling suggests). 
I don't love enums semantic switches, but we can do that if you feel strongly. In general, @jcamachor and I find it more tidy to have a message type per primary operation.  

3. I think an array of expression might not be enough, they need to also have a paired column we are assigning (hence extracting `Assignment`, if that doesn't work for you we can create a similar but dedicated message for that). Some expression have `a = expr` type structure, but I think it is harder to rely on that, then assign explicitly. 

4. For delete again I would return "tuple affected", the use of Rel inside it is to allow the full richness of that to "select" the tuples we operate on.

5. (I am not 100% sure I understand what you are proposing, but I will answer based on my current best guess) Is this a  performance consideration? I find having the quantities (names, types, defaults) co-located makes it easier. So unless there is some performance consideration I am not finding it simpler/easier to use.

jacques-n:
> So if we understand you correctly you would see `INSERT/DELETE/UPDATE` returning a single-row, single-column table containing the number of affected rows (common) and `CREATE TABLE` returning something like a single-row, single-column table with a boolean if the operation succeeded?

The way I've patterned this previously is that it returns a stream of records that could be some metadata about the write operation. Imagine you write many files, you might return the name of each file as well as the records in that file. This could apply even without files: imagine a partioned output. Maybe we can return a name of each partition along with records in each. I'd suggest an output schema that is maybe specific to each type of operation (insert versus update, for example). It would include appropriate descriptions for those things plus all of the stream rotation description column values (if any are any defined). This is obviously a bit hand-wavey. Would be great if you proposed something more concrete on this front. 

>    I don't love enums semantic switches, but we can do that if you feel strongly. In general, @jcamachor and I find it more tidy to have a message type per primary operation.

I'm okay with either. I guess the main concern I would have is related to extensibility of table types (similar to read rel table types). Right now we can add new write options and could support insert + update + create, etc. I don't want people to have extend many different things to add write capabilities. How do we solve for that.

> 3. I think an array of expression might not be enough, they need to also have a paired column we are assigning (hence extracting `Assignment`, if that doesn't work for you we can create a similar but dedicated message for that). Some expression have `a = expr` type structure, but I think it is harder to rely on that, then assign explicitly.

I suggest we propose an inverse set of fields to the read rel. In read rel, we have full schema + projection. I'd do the same here but the projection would be inverted (this is where the input schema maps to the output schema's full detail). With the added additional field of optional values (as noted below).

> 4. For delete again I would return "tuple affected", the use of Rel inside it is to allow the full richness of that to "select" the tuples we operate on.

"Tuple affected" is a little vague. In some systems (say deltalake), this could be the whole record. In other systems (say sql server with primary key), this could be just the primary key.As above, I'd suggest using a mapping here. For many systems, I think we'd only want primary key as opposed to multiple

> 5. (I am not 100% sure I understand what you are proposing, but I will answer based on my current best guess) Is this a  performance consideration? I find having the quantities (names, types, defaults) co-located makes it easier. So unless there is some performance consideration I am not finding it simpler/easier to use.

One of the key goals I have is trying to avoid requiring semantic validation on protobuf as much as possible (and instead use protobufs default structures to provide correctness guarantees). This means avoiding introducing something to a general location that is meaningful only in very rare situations (and would be disallowed 99% of the time). Introducing default value at the type level means we have it everywhere, not just in the one place where it is actually relevant. 

Conceptually speaking, what you're actually doing is applying a mask on write that does a coalesce on certain fields. In that situation, it feels better to constrain the behavior to that location rather than overload types generally. Once in that place, it feels like the existing model of a literal struct is sufficient (and we could define the coalesce behavior locally). We could also introduce a entirely new concept when combined with the item 3 above's "reverse projection". That also seems like it would be fine. It just seems like it might dupe a lot of existing literal definitions.

westonpace:
Do we want to draw any distinction between a top-level relation that could potentially modify the underlying data and a top-level relation that is purely read-only?  I'm not too concerned from an Arrow perspective as Acero is query-only and there are no plans to support DDL.  However, from a user perspective, I wonder if it can possible be difficult at times for a producer to know if the plan it is producing could potentially have side effects.  I think it would generally be straightforward for the consumer to know if a query will have side effects.  So having some ability for the producer to state that a query should be "side-effect free" might be nice.

jacques-n:
> Do we want to draw any distinction between a top-level relation that could potentially modify the underlying data and a top-level relation that is purely read-only? I'm not too concerned from an Arrow perspective as Acero is query-only and there are no plans to support DDL. However, from a user perspective, I wonder if it can possible be difficult at times for a producer to know if the plan it is producing could potentially have side effects. I think it would generally be straightforward for the consumer to know if a query will have side effects. So having some ability for the producer to state that a query should be "side-effect free" might be nice.

I'm more inclined to provide standard libraries that answer these questions then add extra duplicative summary information to a plan. Curious what others think.

jacques-n:
> I think this is might be closer to what you have in mind (much more compact), I am still fuzzy about it though, and we need to run through with @jcamachor all the corner cases for semantics check, but wanted to check what you think:
> 
> ```
> // The operator that modifies the schema/content of a database
> message WriteRel {
>   // The table to be modified
>   NamedTable table = 1;
>   // The columns that will be modified (representing after-image of a schema change)
>   NamedStruct base_schema = 2;
>   // The default values for the columns (representing after-image of a schema change)
>   repeated Literal defaults = 3;
> 
>   //TODO add constraints/indexes/etc..?
> 
>   // The type of operation to perform
>   WriteOp op = 4;
> 
>   // The relation that determines the tuples to add/remove/modify
>   // remaining columns are left as-is or to default (for INSERT).
>   Rel query = 5;
> 
>   // The rel that specifies the output of the operator. It allows references
>   // to DELETED.<col_name> or INSERTED.<col_name>. It can also compute simply count of
>   // affected tuples (per common behavior of most RDBMS). Defaults to no output.
>   Rel output = 6;
> 
>   enum WriteOp {
>     WRITE_OP_UNSPECIFIED = 0;
>     WRITE_OP_CREATE_TABLE = 1;
>     WRITE_OP_DROP_TABLE = 2;
>     WRITE_OP_ALTER_TABLE = 3;
>     WRITE_OP_TRUNCATE_TABLE = 4;
> 
>     WRITE_OP_CREATE_VIEW = 5;
>     WRITE_OP_CREATE_OR_REPLACE = 6;
>     WRITE_OP_DROP_VIEW = 7;
> 
>     WRITE_OP_INSERT = 8;
>     WRITE_OP_INSERT_OR_REPLACE = 9;
>     WRITE_OP_DELETE = 10;
>     WRITE_OP_UPDATE = 11;
>   }
> }
> ```
> 
> Changes posted in PR #427

Generally. Yes. I wouldn't include pure DDL operations in the WriteRel, they feel different (create/drop view, create table, drop table, alter table). Basically, there seem to be two main types of writes: writes that take a set of records as input (e.g. CTAS) and writes that don't (CREATE TABLE). Keeping the second separate feels better than combining here. In general, I have a little question of whether we need to even support DDL at the Substrait level. It feels very database specific (and not requiring any of the other parts of Substrait).

I also don't really understand what the output Rel is. I'd expect something more like: 

```
OutputPattern {

oneof pattern {
  None none = 1;
  RecordCount record_count = 2;
  PartitionList partition_list = 3;
  StreamList stream_list = 4;
}

  message None {}

  message RecordCount{}

  message PartitionList {
    Type.Struct structure = 1;
  }

  message StreamList {
    Type.Struct structure = 1;
  }

}
```
That's just a rought sketch

PR #237, opened 2022-06-30
Add the key-value pair map to store the runtime configurations in substrait plan

JkSelf:
Add the key-value pair to store the configurations in substrait plan.


JkSelf:
@jacques-n Can you help to review? Thanks for your help!

jacques-n:
We already have advanced extensions to solve this need. You can create whatever any type you want and then slot it in the optimization or enhancement slot as makes sense. We chose structured extensibility as opposed to unstructured such as key/value pairs. You can see more on our extension thinking [here](https://substrait.io/serialization/binary_serialization/#advanced-extensions).

jacques-n:
Closing this since it appears that we already have equivalent functionality with a plan's advanced extensions.

Issue #238, opened 2022-07-03
feat(proto): Add Write relation to support multiple outputs


rtpsw:
Currently, a Substrait plan produces a single output corresponding to its root relation. In some cases, multiple outputs are needed, e.g., for multi-table computation as well as for check-point and debug output. The solution proposed here is to add a Write relation, which is a pass-through relation that does not modify the table yet writes it out to some external sink.

westonpace:
I'm going to leave this open as a "generally good idea" but I'm adding the help `wanted label` as I don't think you're actively working on this?

rtpsw:
AFAIK, we ended up solving this using a [`NamedTapRel` extension](https://github.com/apache/arrow/blob/45918a90a6ca1cf3fd67c256a7d6a240249e555a/cpp/proto/substrait/extension_rels.proto#L48-L60), i.e., outside the scope of Substrait itself. I think this can be closed. @icexelloss?

PR #239, opened 2022-07-03
feat(proto): add write relation to support multiple outputs

rtpsw:
See https://github.com/substrait-io/substrait/issues/238

### DIFF
--- a/proto/substrait/algebra.proto
+++ b/proto/substrait/algebra.proto
@@ -142,6 +142,49 @@ message ReadRel {
   }
 }
 
+// This operator passes-through while writing to an external sink
+message WriteRel {
+  RelCommon common = 1;
+  Rel input = 2;
+  substrait.extensions.AdvancedExtension advanced_extension = 10;
+
+  oneof write_type {
+    LocalFiles local_files = 3;
+  }
+
+  message LocalFiles {
+
+    // The consumer should avoid data duplication across items
+    repeated FileOrFiles items = 1;
+    substrait.extensions.AdvancedExtension advanced_extension = 10;
+
+    message FileOrFiles {
+      oneof path_type {
+        string uri_path = 1;
+        string uri_path_glob = 2;
+        string uri_file = 3;
+        string uri_folder = 4;
+      }
+
+      FileFormat format = 5;
+
+      // the index of the partition this item belongs to
+      uint64 partition_index = 6;
+
+      // the project-expressions to apply when writing this item
+      repeated Expression project_expressions;
+
+      // the filter-condition to apply when writing this item
+      Expression filter_condition;
+
+      enum FileFormat {
+        FILE_FORMAT_UNSPECIFIED = 0;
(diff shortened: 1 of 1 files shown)


rtpsw:
cc @icexelloss

jacques-n:
#236 is focused on write rel. Let's get that merged first. Would be good to get your feedback/thoughts on the direction there.

rtpsw:
> #236 is focused on write rel. Let's get that merged first. Would be good to get your feedback/thoughts on the direction there.

@jacques-n, thanks for pointing me to this PR. I'll post there.

rtpsw:
> @rtpsw Can you remind me why we need this again?

See [issue description](https://github.com/substrait-io/substrait/issues/238#issue-1292305059). Though it's not on our critical path, it has been useful in my local work, and I expect it would be useful more generally. It's possible this issue would get merged into https://github.com/substrait-io/substrait/pull/236 .

cpcloud:
@rtpsw Is this superseded by https://github.com/substrait-io/substrait/pull/252?

jacques-n:
This is superseded by #284. Closing. Let us know if we missed anything.

Thanks!

jvanstraten:
I'm not sure, but isn't the current idea that a plan should only have one root relation? Nothing in the protos prevents multiple root relations and I don't think the website says anything about it, but if that's the case then that technically disallows the use case you put forward, and a minor spec change would still be required. Personally I don't see why this should be disallowed; if a consumer doesn't want to support multiple roots then it can just reject a plan with multiple in it.

Otherwise, a WriteRel that passes all data through is not necessary, because you could write the hypothetical `sink(WriteRel(source))` like this:

 - relation 0: source
 - relation 1 (root?): WriteRel(ReferenceRel(0))
 - relation 2 (root): sink(ReferenceRel(0))

The actual new relations are not part of the oneof in Rel yet, that's a bug that's being addressed in #288. Also, WriteRel as currently defined isn't actually the complementary operation of ReadRel. You mention "check-point and debug output," so I imagine you're also interested in file sinks, which currently fundamentally don't work because WriteRel is more of a ReadModifyWriteRel or UpdateRel than a strict write relation, and thus file output was conveniently left out.

rtpsw:
We should distinguish between the use case, which is additional outputs for check-point and debug purposes, from the draft proposal in this PR. While I'm no longer requesting the current PR, due to the other PRs described above, I'm still interested in the use case (though not at a high priority). I think the use case is not handled until a way to get check-point and debug output is spelled out. So, [my question](https://github.com/substrait-io/substrait/pull/239#issuecomment-1245040106) stands and #238 should be left open.

> I imagine you're also interested in file sinks, which currently fundamentally don't work because WriteRel is more of a ReadModifyWriteRel or UpdateRel than a strict write relation, and thus file output was conveniently left out.

I agree a solution for check-point and debug purposes could be based on files. However, I don't think the file has to be a sink. The update or read-modify-write operation fits a random-access/re-writable (relational) file.

PR #240, opened 2022-07-07
feat: add grouping functions

guykhazma:
This PR adds support for the grouping function which is needed in order to support query 27 in TPCDS (see this [issue](https://github.com/substrait-io/substrait-java/issues/71))

Once this PR will get merged I'll open a PR to add the support for substrait-java (see draft branch [here](https://github.com/substrait-io/substrait-java/compare/main...guykhazma:substrait-java:issue71)).

In addition, do we want to add support for the [`group_id`](https://github.com/apache/calcite/blob/b9c2099ea92a575084b55a206efc5dd341c0df62/core/src/main/java/org/apache/calcite/sql/fun/SqlGroupIdFunction.java) function as well? It is not part of the SQL standard but is supported by some engines.

### DIFF
--- a/extensions/functions_aggregate_generic.yaml
+++ b/extensions/functions_aggregate_generic.yaml
@@ -24,8 +24,9 @@ aggregate_functions:
         return: i64
   - name: "grouping"
     description: >-
-      Indicates whether a specified column in a GROUP BY is aggregated or not,
-      returns 1 for aggregated or 0 for not aggregated in the result set
+      Indicates whether the specified column(s) in a GROUP BY are aggregated or not.
+      The result is a bitmap which indicates whether an argument is aggregated or grouped.
+      If the N-th argument is aggregated then the N-th bit is set to 1.
     impls:
       - args:
           - options: [ SILENT, ERROR ]
--- a/extensions/functions_aggregate_generic.yaml
+++ b/extensions/functions_aggregate_generic.yaml
@@ -22,3 +22,25 @@ aggregate_functions:
         decomposable: MANY
         intermediate: i64
         return: i64
+  - name: "grouping"
+    description: >-
+      Indicates whether a specified column in a GROUP BY is aggregated or not,
+      returns 1 for aggregated or 0 for not aggregated in the result set
+    impls:
+      - args:
+          - options: [ SILENT, ERROR ]
+            required: false
+          - value: any
+        nullability: DECLARED_OUTPUT
+        return: boolean
+  - name: "grouping_id"
+    description: "Return the level of grouping for a set of columns"
+    impls:
+      - args:
+          - options: [ SILENT, ERROR ]
+            required: false
+          - value: any
+        variadic:
+          min: 0
+        nullability: DECLARED_OUTPUT
+        return: i64
--- a/extensions/functions_aggregate_generic.yaml
+++ b/extensions/functions_aggregate_generic.yaml
@@ -31,8 +31,10 @@ aggregate_functions:
           - options: [ SILENT, ERROR ]
             required: false
           - value: any
+        variadic:
+          min: 1
         nullability: DECLARED_OUTPUT
-        return: boolean
+        return: i64
   - name: "grouping_id"
     description: "Return the level of grouping for a set of columns"
     impls:
--- a/text/simple_extensions_schema.yaml
+++ b/text/simple_extensions_schema.yaml
@@ -196,7 +196,7 @@ $defs:
         items:
           type: object
           additionalProperties: false
-          required: [ intermediate, return ]
+          required: [ return ]
           properties:
             args:
               $ref: "#/$defs/arguments"


emkornfield:
not an expert her but it seems the reasons for them not to be functions, is that there is there implied input has special metadata associated with it.  I've been meaning to try to catch up with substrait, but I can't say if this distinction is meaningful enough to make them not a function.

extensions/functions_aggregate_generic.yaml
```diff
@@ -22,3 +22,23 @@ aggregate_functions:
...
         return: i64
+  - name: "grouping"
+    description: "Indicates whether the column specified in a GROUP BY clause aggregated or not"
+    impls:
+      - args:
+          - options: [ SILENT, ERROR ]
+            required: false
+          - value: any
+        nullability: DECLARED_OUTPUT
+        return: i64
```
ianmcook, line 33:
```suggestion
        return: boolean
```

guykhazma: I chose to use `i64` because calcite is `BIGINT`, see [here](https://github.com/apache/calcite/blob/b9c2099ea92a575084b55a206efc5dd341c0df62/core/src/main/java/org/apache/calcite/sql/fun/SqlGroupingFunction.java#L43).

extensions/functions_aggregate_generic.yaml
```diff
@@ -22,3 +22,23 @@ aggregate_functions:
         decomposable: MANY
         intermediate: i64
         return: i64
+  - name: "grouping"
+    description: "Indicates whether the column specified in a GROUP BY clause aggregated or not"
```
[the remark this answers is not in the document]
guykhazma: I have updated the description to include information about the expected output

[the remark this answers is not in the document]
guykhazma: I have updated the description to include information about the expected output

hannes:
DuckDB supports `GROUPING` according to the SQL standard. From the description in the change here it's not clear to me that those are the semantics intended but maybe they are. 
Screenshot 2022-07-18 at 06 58 11

guykhazma:
@hannes yes this is the semantics of the `grouping` function I am referring to.
The description I am using is also used in Spark (see [here](https://spark.apache.org/docs/latest/api/sql/index.html#grouping)). I can change it to the above description if we prefer that.

extensions/functions_aggregate_generic.yaml
jvanstraten: +1 for boolean. Judging by the docs you linked, it only does that because it's variadic and uses one bit for each argument. The way you defined the function here it only takes a single argument, and thus only needs to output a single bit, which is what booleans are meant for. I guess you could also make this function variadic, but without a bigint type, the number of arguments would be artificially limited to 64 to avoid overflow.

extensions/functions_aggregate_generic.yaml
```diff
@@ -22,3 +22,25 @@ aggregate_functions:
         decomposable: MANY
         intermediate: i64
         return: i64
+  - name: "grouping"
+    description: >-
+      Indicates whether a specified column in a GROUP BY is aggregated or not,
+      returns 1 for aggregated or 0 for not aggregated in the result set"
```
jvanstraten, line 28: There's a stray `"` at the end there.

guykhazma: thanks, fixed

extensions/functions_aggregate_generic.yaml
guykhazma: ok, agree, changed it to `boolean` and kept it with single argument.

extensions/functions_aggregate_generic.yaml
```diff
@@ -22,3 +22,25 @@ aggregate_functions:
...
+      returns 1 for aggregated or 0 for not aggregated in the result set
+    impls:
+      - args:
+          - options: [ SILENT, ERROR ]
+            required: false
+          - value: any
+        nullability: DECLARED_OUTPUT
+        return: boolean
+  - name: "grouping_id"
+    description: "Return the level of grouping for a set of columns"
```
mbasmanova, line 37: What does "level of grouping" means here?

mbasmanova:
In Presto, grouping sets are implemented using GroupId operator which duplicates input for each grouping set and produces an extra "group_id" column that identifies a grouping set. The aggregation operator then groups by all columns mentioned in the grouping sets + group_id column. In this design "grouping" expression becomes a scalar expression over group_id column. 

GroupId operator implementation in Velox + examples of query plans: 

https://github.com/facebookincubator/velox/pull/1793/files#diff-58637420124ffe154b91ddfe97854192bc53cca18543a4046db8ebcb04edb7fdR813

"Grouping" expression rewrite in Presto:

https://github.com/prestodb/presto/blob/master/presto-main/src/main/java/com/facebook/presto/sql/planner/GroupingOperationRewriter.java#L41

Given this implementation, it is not obvious to me that it makes sense to define "grouping" aggregate function in Substrait. It would be helpful to see some examples of queries and corresponding query plans that would use "grouping" aggregate function.

extensions/functions_aggregate_generic.yaml
```diff
@@ -22,3 +22,25 @@ aggregate_functions:
...
         intermediate: i64
         return: i64
+  - name: "grouping"
+    description: >-
+      Indicates whether a specified column in a GROUP BY is aggregated or not,
+      returns 1 for aggregated or 0 for not aggregated in the result set
+    impls:
+      - args:
+          - options: [ SILENT, ERROR ]
+            required: false
```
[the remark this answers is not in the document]
jvanstraten: The `required` tag is for the error handling option, which just means it defaults to the first value. I'm not entirely sure what the option is for, though.

mbasmanova:
A follow-up question is whether Substrait supports GroupId plan node: https://github.com/prestodb/presto/blob/master/presto-main/src/main/java/com/facebook/presto/sql/planner/plan/GroupIdNode.java#L60

mbasmanova:
@jvanstraten 

> please note that Substrait aggregate relations always emit a grouping set index column

I didn't realize that. Thank you for pointing out. I assume this is specified by the "followed by an integer" clause below.

https://substrait.io/relations/logical_relations/#aggregate-operation 

> The list of distinct columns from each grouping set (ordered by their first appearance) followed by the list of measures in declaration order, followed by an integer describing the associated particular grouping set the value is derived from.

I wonder if there is any particular reason for this design. It looks like neither Presto nor Trino can represent aggregations this way.

jvanstraten:
All engines are inevitably going to do things differently and internally represent things in different ways, for performance in particular cases or otherwise. If everyone were already doing the same thing, Substrait wouldn't be necessary.

An interchange format like Substrait should ideally use a minimal, high-level format with only one way to do something. That one way necessarily probably won't map one-to-one to your favorite engine, again because they all do things differently. It might not even be performant, in this case probably mapping to some combination of aggregate and projection nodes to implement Substrait's idea of what an aggregate fundamentally is. But the idea is that a plan written using only Substrait core stuff should be highly compatible among consumers.

In order to also be able to represent plans that are optimized for a particular engine, Substrait supports extensions; you're free to define whatever aggregate node semantics you want as an extension relation or (as in this PR) as a function. But of course the catch is that no other consumer will understand it.

If Substrait would try to support everything everyone does in its core, either everyone would have to implement everyone else's semantics and idiosynchrasies on top of their own (unlikely to happen...) or there would be no such thing as a cross-consumer plan in practice.

mbasmanova:
@jvanstraten Thank you for explaining. Curious, which of the engines do implement grouping sets the way these are defined in Substrait.

extensions/functions_aggregate_generic.yaml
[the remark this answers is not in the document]
guykhazma: I am not sure why, but I was under the impression that when you don't provide any parameter to the grouping function it should default to the list of columns/expression used by the groupby.
I don't find any link that mention that so I will change it to support multiple arguments and require at least one

jacques-n:
As I've reviewed this more, I think that the existing representation of grouping sets in an aggregate operation accomplish this behavior. Closing with the recommendation of using that functionality.

PR #243, opened 2022-07-13
feat: Enable arena option

ZJie1:


ZJie1:
> This change probably isn't needed. On the other hand, it is harmless so I have no problem with it in case someone really does need to use the older protoc.
> 
> [protocolbuffers/protobuf#8119](https://github.com/protocolbuffers/protobuf/issues/8119)

Thanks for the reply, will close this since we could update protobuf version and it works for us.

amaira2011 (approved):

PR #244, opened 2022-07-14
feat(arithmetic): add abs and sign to scalar function extensions

gforsyth:
The overflow behavior for `abs` is annoying but probably necessary (unless we want to aggressively upcast).

Following Arrow's lead on return types for `sign`

### DIFF
--- a/extensions/functions_arithmetic.yaml
+++ b/extensions/functions_arithmetic.yaml
@@ -392,6 +392,70 @@ scalar_functions:
           - value: fp64
           - value: fp64
         return: fp64
+  - 
+    name: "abs"
+    description: >
+      Calculate the absolute value of the argument.
+
+      Integer values allow the specification of overflow behavior to handle the
+      unevenness of the twos complement, e.g. Int8 range [-128 : 127].
+    impls:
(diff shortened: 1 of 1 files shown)


extensions/functions_arithmetic.yaml
```diff
@@ -181,6 +181,67 @@ scalar_functions:
...
+      - args:
+          - value: fp64
+        return: fp64
+  -
+    name: "sign"
+    description: >
+      Return the signedness of the argument.
+
+      Integer values return signedness as Int8.
+      Floating point values return signedness with the same type as the input.
```
jvanstraten, line 224: I guess for ints the return value is one of [-1, 0, 1]. But what about floats? Would it be [-1.0, -0.0, 0.0, 1.0, NaN] or something?

Also, why do the integer variants return all `i8` but the floating point variants don't all return `fp32`?

jvanstraten: Yep.

jvanstraten (approved):

jacques-n (approved):
Thanks @gforsyth !

PR #245, opened 2022-07-14
feat: add functions for arithmetic, rounding, logarithmic, and string transformations

gforsyth:
Picking up on @sanjibansg 's work in #230 -- I've fixed up what I believe were the remaining comments and I've removed the `round` operation so that can be handled and discussed separately (although there seems to be a reasonable consensus around https://github.com/substrait-io/substrait/pull/230#issuecomment-1170230630


richtia:
@gforsyth Is there a convention to follow for the name/description?  Some files have both in double quotes, some only have one, some neither.  Seems like it's mismatched.

gforsyth:
> @gforsyth Is there a convention to follow for the name/description? Some files have both in double quotes, some only have one, some neither. Seems like it's mismatched.

I went with double-quotes for a single line description but no quotes for a multiline string.  But that's arbitrary

extensions/functions_arithmetic.yaml
```diff
@@ -181,6 +211,72 @@ scalar_functions:
           - value: i64
           - value: i64
         return: i64
+  -
+    name: "power"
+    description: >
+      Take the power with the first value as the base and second as exponent.
+
+      All integer arguments are promoted to int64.
```
jacques-n, line 219: Given that this is the behavior, let's not include the variations that take anything else. Let's just expect engines to do upromotions.

extensions/functions_string.yaml
```diff
@@ -163,3 +164,29 @@ scalar_functions:
           - value: "fixedchar<L1>"
           - value: "varchar<L2>"
         return: "BOOLEAN"
+  -
+    name: lower
+    description: Transforms the string to lower case characters
```
jacques-n, line 169: Can we specify using what rules? I assume there are specific utf8 rules around this that we should reference?

extensions/functions_string.yaml
```diff
@@ -163,3 +164,29 @@ scalar_functions:
...
+      - args:
+          - value: "fixedchar<L1>"
+        return: "fixedchar<L1>"
+  -
+    name: upper
+    description: Transforms the string to upper case characters
+    impls:
+      - args:
+          - value: "varchar<L1>"
+        return: "varchar<L1>"
```
jacques-n, line 186: Same comment as lower.

extensions/functions_arithmetic.yaml
```diff
@@ -161,6 +161,36 @@ scalar_functions:
...
+        return: i64
+      - args:
+          - value: fp32
+        return: fp32
+      - args:
+          - value: fp64
+        return: fp64
   -
     name: "modulus"
     description: "Get the remainder when dividing one value by another."
```
jvanstraten, line 196: Modulus and remainder are two different things, and even then everyone seems to have their own ideas about them; please specify the behavior for negative numbers. The same thing actually applies for integer division: does it round toward negative infinity or zero, and does this depend on the sign of the divisor? Remainder is then fully defined. I personally much prefer the behavior of Python over what C typically does (and I say "typically" because C shies away from actually specifying it and leaves it to the processor/compiler).

extensions/functions_arithmetic.yaml
```diff
@@ -181,6 +211,72 @@ scalar_functions:
...
+            required: false
+          - value: i64
+          - value: i64
+        return: i64
+      - args:
+          - options: [ SILENT, SATURATE, ERROR ]
+            required: false
+          - value: fp32
+          - value: fp32
+        return: fp64
```
jvanstraten, line 251: I don't really see why fp32 ^ fp32 should promote to fp64. Both have "similar" ranges in that they include infinity, and if you really need double accuracy you should just promote to double first. If we take the C math library as precedent, we have `double pow(double, double)` and float `powf(float, float)`.

I feel the same about the integer power functions, though a bit less strongly I guess. Why should power be different from multiplication and addition? Both can overflow, and the former requires double the bits. All languages I know (except VHDL/Verilog, but those are kinda special) require promotion first if you need a larger output type. I guess I can see where it's coming from, but I don't think this sort of smartness belongs in the Substrait spec.

IMO: just let the producer decide if they want to upcast or not, and if a compound operation that does the upcast and the operation simultaneously exists in the consumer it is up to them to optimize to that operation when they encounter the `pow(cast<i64>(x), cast<i64>(x))` pattern.

extensions/functions_rounding.yaml
```diff
@@ -0,0 +1,27 @@
...
+scalar_functions:
+  -
+    name: "ceil"
+    description: "Rounding to the ceiling of the value"
+    impls:
+      - args:
+          - options: [ SILENT, SATURATE, ERROR ]
+            required: false
+          - value: fp32
+        return: i64
```
jvanstraten, line 12: The range of even an fp32 is MUCH larger than the range of i64. ceil and floor normally return floating point numbers for a reason. If you want an i64, the pattern is `cast<i64>(floor(x))`.

extensions/functions_arithmetic.yaml
```diff
@@ -181,6 +211,72 @@ scalar_functions:
...
+      - args:
+          - options: [ SILENT, SATURATE, ERROR ]
+            required: false
+          - value: fp64
+          - value: fp64
+        return: fp64
+  -
+    name: "sqrt"
```
jvanstraten, line 259: Same as power, sqrt is normally defined as `double sqrt(double)` and `float sqrtf(float)`. I suppose you could add `sqrt(i64) -> fp64` to those two because an `i64` can't be cast to fp64 without losing information, but the other ones are redundant.

jvanstraten (changes requested):
In addition, I think you forgot to implement this comment: https://github.com/substrait-io/substrait/pull/230#issuecomment-1170230630. It seems to me like we had reached consensus about that? sqrt and logarithms need domain error handling, all floating point operations need rounding mode, and SILENT/SATURATE/ERROR makes no sense for functions that return floats.

extensions/functions_string.yaml
gforsyth: I'm going to punt on the string transforms to a separate PR and try to scope this one down a bit, so I've removed these changes for the moment.

extensions/functions_arithmetic.yaml
jvanstraten: Uh, oops.

gforsyth: heh, all good. I didn't know that Github was _finally_ allowing comments on lines that hadn't been touched (which I've been waiting for for ages)

extensions/functions_arithmetic.yaml
```diff
@@ -181,6 +211,41 @@ scalar_functions:
...
+    description: "Take the power with the first value as the base and second as exponent."
+    impls:
+      - args:
+          - options: [ SILENT, SATURATE, ERROR ]
+            required: false
+          - value: i64
+          - value: i64
+        return: i64
+      - args:
+          - options: [ SILENT, SATURATE, ERROR ]
```
jvanstraten, line 225: Float pow still has the integer options instead of float options.

extensions/functions_arithmetic.yaml
```diff
@@ -181,6 +211,41 @@ scalar_functions:
...
+        return: fp32
+      - args:
+          - options: [ SILENT, SATURATE, ERROR ]
+            required: false
+          - value: fp64
+          - value: fp64
+        return: fp64
+  -
+    name: "sqrt"
+    description: "Square root of the value"
```
jvanstraten, line 238: sqrt still lacks rounding mode and domain error options.

extensions/functions_rounding.yaml
```diff
@@ -0,0 +1,27 @@
+%YAML 1.2
+---
+scalar_functions:
+  -
+    name: "ceil"
+    description: "Rounding to the ceiling of the value"
+    impls:
+      - args:
+          - options: [ SILENT, SATURATE, ERROR ]
```
jvanstraten, line 9: ceil still has overflow options (like floor, it doesn't need any options AFAICT).

jvanstraten (changes requested):

jvanstraten (changes requested):
The logarithms still lack domain error handling (undefined for <= 0). LGTM now otherwise.

jvanstraten (approved):

extensions/functions_logarithmic.yaml
```diff
@@ -0,0 +1,95 @@
...
+    description: "Logarithm of the value with the given base"
+    impls:
+      - args:
+          - name: rounding
+            options: [ TIE_TO_EVEN, TIE_AWAY_FROM_ZERO, TRUNCATE, CEILING, FLOOR ]
+            required: false
+          - name: on_domain_error
+            options: [ NAN, ERROR ]
+            required: false
+          - value: fp32
```
jacques-n, line 81: Please provide names for these arguments to clarify their purpose.

jacques-n (approved):
Looks good. Thanks @gforsyth for working through the feedback!

PR #246, opened 2022-07-15
feat: add random function

ianmcook:
Adds function `random`. Includes a nullary implementation and unary implementation that takes a constant seed value.

I put this in a new YAML file because AFICT most engines consider random generation functions to be in category of their own. I can move it to `functions_arithmetic.yaml` if y'all prefer that.

In most engines, the random generation function returns a value of type double in 0 <= x < 1, and this is consistent with that. ~~But the type of the seed value varies across engines, so I set that to `any`.~~ Seed values of type `i64` and `fp64` are allowed.

For the seed argument, I attempted to follow the [Value Argument Properties](https://substrait.io/expressions/scalar_functions/#value-argument-properties) format, but I could not find any other examples of that in the existing YAML files, so I'm not sure whether I did it correctly.

### DIFF
--- a/extensions/functions_random.yaml
+++ b/extensions/functions_random.yaml
@@ -3,7 +3,10 @@
 scalar_functions:
   -
     name: random
-    description: "Return a value between 0 (inclusive) and 1 (exclusive)"
+    description: >
+      Return a value between 0 (inclusive) and 1 (exclusive).
+
+      Not suitable for cryptographic applications.
     impls:
       - deterministic: false
         return: fp64
--- a/extensions/functions_random.yaml
+++ b/extensions/functions_random.yaml
@@ -13,7 +13,14 @@ scalar_functions:
       - args:
           - value:
               - name: Seed
-                type: any1
+                type: i64
+                constant: true
+        deterministic: false
+        return: fp64
+      - args:
+          - value:
+              - name: Seed
+                type: fp64
                 constant: true
         deterministic: false
         return: fp64
--- a/extensions/functions_random.yaml
+++ b/extensions/functions_random.yaml
@@ -1,0 +1,14 @@
+%YAML 1.2
+---
+scalar_functions:
+  -
+    name: random
+    description: "Return a value between 0 (inclusive) and 1 (exclusive)"
+    impls:
+      - args:
+          - value:
+              - name: Seed
+                type: any1
+                required: false
+        deterministic: false
+        return: fp64
--- a/extensions/functions_random.yaml
+++ b/extensions/functions_random.yaml
@@ -5,10 +5,12 @@ scalar_functions:
     name: random
     description: "Return a value between 0 (inclusive) and 1 (exclusive)"
     impls:
+      - deterministic: false
+        return: fp64
       - args:
           - value:
               - name: Seed
                 type: any1
-                required: false
+                constant: true
         deterministic: false
         return: fp64


extensions/functions_random.yaml
```diff
@@ -0,0 +1,16 @@
...
+  -
+    name: random
+    description: "Return a value between 0 (inclusive) and 1 (exclusive)"
+    impls:
+      - deterministic: false
+        return: fp64
+      - args:
+          - value:
+              - name: Seed
+                type: any1
```
jacques-n, line 13: I don't think the any type makes sense here. Any is really focused on a situation where engines can support any type. Most engines would only support one discrete type (say i64 or fp64). In those cases, we should have discrete implementations.

extensions/functions_random.yaml
```diff
@@ -0,0 +1,16 @@
+%YAML 1.2
+---
+scalar_functions:
+  -
```
jacques-n, line 4: Does it make sense to include both a random and a secure random function (and declare the security/quality of the random function)?

jacques-n (commented):
Thanks @ianmcook , separately file makes sense to me. A couple of more specific comments inline.

extensions/functions_random.yaml
ianmcook: Got it. I'll look at what seed types are commonly supported and I'll add separate implementations for each.

extensions/functions_random.yaml
ianmcook: I'm not sure. Can you give an example or two of the functions in different engines that this might map to? Thanks

ianmcook:
I considered adding a `uniform` function here also, but that is not nearly as common as `random` so I omitted it. If anyone feels that that should be included (👍) or omitted (👎), please react to this comment accordingly.

extensions/functions_random.yaml
westonpace: postgres has [`get_random_bytes`](https://www.postgresql.org/docs/current/pgcrypto.html#id-1.11.7.35.10) but they also have a whole set of crypto-style functions (encrypt, hash, etc.) and so maybe we can wait and add all those later.  Either way, we should probably add a note like this one from postgres:

> The random() function uses a simple linear congruential algorithm. It is fast but not suitable for cryptographic applications

westonpace:
-1 on uniform because then the next question is "why not normal? exponential? etc."  Those would be cool features to have (and it seems like one of the reasons I'm always reaching for numpy) but we can maybe add those in their own dedicated section.

jvanstraten:
-1 on uniform et al because they're not primitives. You can make all of them by just doing arithmetic on the result of `random`.

extensions/functions_random.yaml
jvanstraten: Fast vs cryptographic could also be an option argument. Same as for the arithmetic overflow and rounding behaviors, if an engine doesn't support one or more of the options it should just reject selectively.

ianmcook: I added
>Not suitable for cryptographic applications

to the description. I would prefer to save cryptographic random generation functions for a separate PR.

ianmcook:
Side note: after this is merged, I intend to start a discussion with the DuckDB folks about it, because DuckDB (following the example of PostgreSQL) does not accept seed values to `random()` but instead provides a separate function `setseed()` that the user can use to set a seed value. It will be interesting to see if we can have all consumers including DuckDB conform to one standard for how to set seeds (ideally the one implemented here which I believe is more common) or whether that proves to be infeasible.

jvanstraten:
Looking at this again, I'm wondering under what circumstances two invocations of `random` on a column should yield the same list of values. The seed argument implies to me that if I would make a ProjectRel that adds two columns using the expression `random(42)` for both, I should get two identical columns, but is that actually what it means or is it just *additional* entropy? If the former, what about different consumers or different sessions to the same consumer? Also, should there not also be a `random()` function that doesn't take a seed argument? Or should the seed maybe be associated with a complete plan rather than a function? Random seeds are weird.

westonpace:
Hmm, @jvanstraten makes a good point.  Traditionally a scalar function is "stateless".  The value of row X is independent of the value of row X-1.  For a seedless random you could argue that this is true (even though it technically isn't) by pretending the random function is a true random number generator and ignoring the fact that there is a state machine inside.

For a seeded random this seems pretty much impossible.  This means, for example, that a seeded random will get different results depending on the batch size of your data (not to mention threading), which seems to defeat the purpose.

extensions/functions_random.yaml
jacques-n: > > Fast vs cryptographic could also be an option argument.
> 
> I am -1 on this because it does not match the semantics of any engine or client AFAIK. I think it's better to have separate functions.

This makes an assumption that the systems would expose this as an option to users. There is no requirement for systems to expose functions that have options to end users with the functions. A system may just map random() to the fast one and secure_random() or something similar to the safe one.

ianmcook: > > > Fast vs cryptographic could also be an option argument.
> > 
> > 
> > I am -1 on this because it does not match the semantics of any engine or client AFAIK. I think it's better to have separate functions.
> 
> This makes an assumption that the systems would expose this as an option to users. There is no requirement for systems to expose functions that have options to end users with the functions. A system may just map random() to the fast one and secure_random() or something similar to the safe one.

Yes, I recognize that producers and consumers are under no requirement to map functions directly to/from Substrait functions and function options directly to/from Substrait function options. However, when there is a question like this about whether we should represent two capabilities as two separate functions or as one function that accepts two option values, in general it would seem wise to follow the examples of various existing engines and clients.

jacques-n:
I second @westonpace's comments. I suggest we remove the seed as an argument. @jvanstraten, do you concur?

jvanstraten:
Yup, that would solve my issues with it :+1:

ianmcook:
Sounds good. I will remove the implementation with the seed argument.

Presumably we will eventually want to define a seeded pseudorandom function in Substrait, no? @jacques-n @westonpace @jvanstraten do you think this would best be defined as a separate function with a different name (e.g. `seeded_random`)?

jvanstraten:
Seeded random feels like a snowflake too special for a normal function to me. I'd personally just add an optional random seed field at the plan level and call it a day. But that might be too simplistic too.

julianhyde:
I don't buy @westonpace's argument about "stateless functions". Yes, most regular functions in SQL are [pure](https://en.wikipedia.org/wiki/Pure_function): given the same arguments, they return the same results.

But `random` is not pure. (A pure random function would always return the same value, so would be utterly useless.) And so we need a different model for what is 'reasonable' behavior.

A seed argument doesn't ensure complete determinacy (e.g. if you change the plan to push down filters) but, for a given platform and execution plan gives a pretty reasonable guarantee. Good enough to be able to write unit tests. Without a seed argument all bets are off.

jvanstraten:
Note that Substrait already conceptually defines impure functions as either "session-dependent but otherwise pure" or "non-deterministic" depending on options. Unseeded `random()`s impurity, and that of any other stateful function, is sufficiently covered by the latter. It's just the specification of the initial state for random that's the issue. You normally seed an RNG only once, not for every binding of a `random()` function and obviously not for every call (though the `constant` flag skirts around that issue).

The most important things I'm wondering about right now, especially for unit tests, is whether the seed should be conveyed by the plan or be set by the test runner, and whether there are cases where you might *want* a somehow more granular seed configuration.

julianhyde:
If you have a seed parameter then that gives you (as the test writer) the flexibility to pass in some kind of parameter as the argument. That seems more straightforward than creating plumbing to provide the seed via some back door.

westonpace:
> A seed argument doesn't ensure complete determinacy (e.g. if you change the plan to push down filters) but, for a given platform and execution plan gives a pretty reasonable guarantee. Good enough to be able to write unit tests. Without a seed argument all bets are off.

I would not be surprised if other engines work differently than Acero.  However, in Acero, this statement isn't really true.  For example, consider a plan:

`SCAN -> FILTER -> PROJECT -> SINK`

If the filter node is selective enough then an engine may choose to wait and batch results (instead of sending empty or single row batches through the rest of the pipeline).  If the engine is scanning in parallel then this means that the batches coming out of the filter node are completely unpredictable.  Even if the project node is applying a seeded random then the output will be non-deterministic.

> That seems more straightforward than creating plumbing to provide the seed via some back door.

I wouldn't argue for that either.  If a user needs this deterministic behavior in order to write tests then it could still be implemented as a window function where the window is the entire dataset.  This would treat seeded random the same as other stateful functions like cumulative sum.

julianhyde:
> If the engine is scanning in parallel then this means that the batches coming out of the filter node are completely unpredictable

It's a soluble problem. For instance, if `FILTER` is operating over N partitions then `random` in partition 0 uses seed, `random` in partition 1 uses `seed + 1`, etc.

> If a user needs this deterministic behavior in order to write tests then it could still be implemented as a window function where the window is the entire dataset

You're saying if you want the query to be deterministic then write a different query. We can do better than that. (Postgres does, for example.)

The deterministic mode is more for the benefit of the people writing the DBMS than the users. The users may or may not need determinism, but the people writing the DBMS sure as hell do, because you can't write or test a component unless the components under it have a deterministic mode.

westonpace:
> It's a soluble problem. For instance, if FILTER is operating over N partitions then random in partition 0 uses seed, random in partition 1 uses seed + 1, etc.

This won't work but I'll save it for the bottom since this is getting into the weeds.

> You're saying if you want the query to be deterministic then write a different query.

A window function is currently defined by the spec as such:

```
Window functions are functions which consume values from
multiple records to produce a single output. They are similar
to aggregate functions, but also have a focused window of
analysis to compare to their partition window. Window functions
are similar to scalar values to an end user, producing a single
value for each input record. However, the consumption visibility
for the production of each single record can be many records.
```

This sounds like we are talking about but the language is a bit unclear and I will admit I am not super confident here.  What I was trying to say was "If you have a nail (seeded random) and a hammer (window functions) then use the hammer.  You can maybe use the backend of a screwdriver (pseudo-scalar function) but we probably shouldn't put that in the instruction manual.

> The deterministic mode is more for the benefit of the people writing the DBMS than the users.

Then why put it in the spec?  If an engine can use a scalar seeded random then they should.  Acero cannot.  For unit testing we instead rely on a deterministic scheduler.

> It's a soluble problem. For instance, if FILTER is operating over N partitions then random in partition 0 uses seed, random in partition 1 uses seed + 1, etc.

Imagine we have three batches, A B and C.  Our filter is x < 0.  Our filter is configured to wait until at least 2 rows have accumulated before emitting a batch.

A:
| x |
| --- |
| -1 |
| 1 |

B:
| x |
| --- |
| -2 |
| 2 |

C:
| x |
| --- |
| -3 |
| 3 |

If the data arrives in ABC order then the first batch the project node will see is:

| x |
| --- |
| 1 |
| 2 |

So our seeded random will yield...

| x | RAND(X, 0)
| --- | --- |
| 1 | 87 |
| 2 | 123 |

If the data arrives in BAC order then the first batch the project node will see is:

| x |
| --- |
| 2 |
| 1 |

And the seeded random will yield...

| x | RAND(X, 0) |
| --- | --- |
| 2 | 87 |
| 1 | 123 |

westonpace:
In restrospect, such a filtering node would be in violation of Substrait, since it cannot maintain ordering guarantees.  However, even if we can rely on row order, I think there are examples where we cannot rely on batch size.

 * If the input dataset is a parquet dataset and the exact same dataset is saved with a different row group size then this would modify the output.
 * If there is an order destroying node in place, such as a hash-join, then a window function would either impose an order or reject the plan.  In the meantime, a projection would just silently return non-deterministic results.
 * If there is an order preserving join (asof, merge) one could imagine it batching data, in order, until either sufficient time has passed, or some threshold has hit.  This "sufficient time has passed" criteria will be non-deterministic.

ianmcook:
I worry that this discussion has drifted away from practical considerations. In practice, when users set random seeds, I think they are not expecting results with reproducible row order. I don't think we should get hung up about that.

ianmcook:
On the question of where seeds should be set: I believe there are three options:

1. As an argument to the seeded random generation function, as in MySQL (this is what is implemented here)
2. With a separate set seed function that returns void and is called only for its side effect, as in PostgreSQL (I'm not sure that is even possible in Substrait; I think we can rule out this option)
3. As an optional argument at the plan level (this is what @jvanstraten suggests above)

Are we confident that option 3 is sufficiently flexible to meet user needs? Some engines (MySQL) allow users to call `rand()` with a seed multiple times within a single query, and each time. For example (using the console at https://tour.tidb.io):

```
select rand(42) x, rand(42) y from mysql.tidb;
```

| x  | y  |
|---|---|
| 0.6629  | 0.6629  | 
| 0.1516 | 0.1516 | 
| 0.7694  | 0.7694  | 
| 0.3925  | 0.3925  | 

Notice the values are the same in both columns. Option 3 does not allow that. Option 1 does.

jvanstraten:
Ultimately I'm fine with either 1 or 3, as long as we clearly specify when two equally-seeded random functions are expected to yield the same values.

extensions/functions_random.yaml
```diff
@@ -0,0 +1,26 @@
...
+    name: random
+    description: >
+      Return a value between 0 (inclusive) and 1 (exclusive).
```
cpcloud, line 7:
```suggestion
      Return a random value between 0 (inclusive) and 1 (exclusive).
```

extensions/functions_random.yaml
```diff
@@ -0,0 +1,26 @@
...
+
+      Not suitable for cryptographic applications.
+    impls:
+      - deterministic: false
+        return: fp64
+      - args:
+          - value:
+              - name: Seed
```
cpcloud, line 15: What are the semantics of seed? For example, would `random(5)` return the same random number for every row of the source relation? Or is it designed to be the seed for some process that creates an RNG before executing the "inner loop" of the source data.

PR #247, opened 2022-07-15
feat: add least and greatest functions to functions_comparison.yml

richtia:
PR to add functions for `least` and `greatest`.

### DIFF
--- a/extensions/functions_comparison.yaml
+++ b/extensions/functions_comparison.yaml
@@ -218,11 +218,8 @@ scalar_functions:
   -
     name: "least"
     description: >-
-      Returns the smallest value.  The function will return null if any argument evaluates
+      Evaluates each argument and returns the smallest one.  The function will return null if any argument evaluates
       to null.
-
-      String comparison is done in lexicographical ordering, one character at a time, from
-      left to right. Uppercase letters are less than lowercase letters.
     impls:
       - args:
           - value: T
@@ -232,11 +229,8 @@ scalar_functions:
   -
     name: "least_skip_null"
     description: >-
-      Returns the smallest value. The function will return null only if all arguments evaluate
+      Evaluates each argument and returns the smallest one. The function will return null only if all arguments evaluate
       to null.
-
-      String comparison is done in lexicographical ordering, one character at a time, from
-      left to right. Uppercase letters are less than lowercase letters.
     impls:
       - args:
           - value: T
--- a/extensions/functions_comparison.yaml
+++ b/extensions/functions_comparison.yaml
@@ -218,8 +218,8 @@ scalar_functions:
   -
     name: "least"
     description: >-
-      Evaluates each argument and returns the smallest one.  The function will return null if any argument evaluates
-      to null.
+      Evaluates each argument and returns the smallest one.  The function will return
+      null if any argument evaluates to null.
     impls:
       - args:
           - value: T
@@ -229,8 +229,8 @@ scalar_functions:
   -
     name: "least_skip_null"
     description: >-
-      Evaluates each argument and returns the smallest one. The function will return null only if all arguments evaluate
-      to null.
+      Evaluates each argument and returns the smallest one. The function will return
+      null only if all arguments evaluate to null.
     impls:
       - args:
           - value: T
@@ -240,15 +240,22 @@ scalar_functions:
   -
     name: "greatest"
     description: >-
-      Returns the largest value. Only return null if 'all' arguments evaluate to null.
-
-      String comparison is done in lexicographical ordering, one character at a time, from left to right.
-      Uppercase letters are less than lowercase letters.
-
-      There is no greatest_skip_null function because it behaves the same as greatest.
+      Evaluates each argument and returns the largest one.  The function will return
+      null if any argument evaluates to null.
     impls:
       - args:
           - value: T
         variadic:
           min: 2
-        return: T
+        return: T|?
+  -
+    name: "greatest_skip_null"
+    description: >-
+      Evaluates each argument and returns the largest one. The function will return
+      null only if all arguments evaluate to null.
+    impls:
+      - args:
+          - value: T
+        variadic:
+          min: 2
+        return: T&?
--- a/extensions/functions_comparison.yaml
+++ b/extensions/functions_comparison.yaml
@@ -218,27 +218,22 @@ scalar_functions:
   -
     name: "least"
     description: >-
-      Returns the smallest value.
-
-      The `null_handling` option specifies how to handle null arguments.  If `ANY_NULL`
-      is specified, the function will return null if any argument evaluate to null.
-      `ALL_NULLS` means the function will return null only if all arguments evaluate to null.
+      Returns the smallest value.  The function will return null if any argument evaluates
+      to null.
 
       String comparison is done in lexicographical ordering, one character at a time, from
       left to right. Uppercase letters are less than lowercase letters.
     impls:
       - args:
           - value: T
-        options:
-          null_handling:
-            values: [ ANY_NULL, ALL_NULLS ]
         variadic:
           min: 2
         return: T|?
   -
     name: "least_skip_null"
     description: >-
-      Returns the smallest value. Skip any null arguments.
+      Returns the smallest value. The function will return null only if all arguments evaluate
+      to null.
 
       String comparison is done in lexicographical ordering, one character at a time, from
       left to right. Uppercase letters are less than lowercase letters.
@@ -247,7 +242,7 @@ scalar_functions:
           - value: T
         variadic:
           min: 2
-        return: T
+        return: T&?
   -
     name: "greatest"
     description: >-
@@ -255,6 +250,8 @@ scalar_functions:
 
       String comparison is done in lexicographical ordering, one character at a time, from left to right.
       Uppercase letters are less than lowercase letters.
+
+      There is no greatest_skip_null function because it behaves the same as greatest.
     impls:
       - args:
           - value: T
--- a/extensions/functions_comparison.yaml
+++ b/extensions/functions_comparison.yaml
@@ -215,3 +215,30 @@ scalar_functions:
         variadic:
           min: 2
         return: any1
+  -
+    name: "least"
+    description: >-
+      Returns the smallest value. Only return null if 'all' arguments evaluate to null.
+
+      String comparison is done in lexicographical ordering, one character at a time, from left to right.
+      Uppercase letters are less than lowercase letters.
+    impls:
+      - args:
+          - value: T
+        variadic:
+          min: 2
+        return: T
+
+  -
+    name: "greatest"
+    description: >-
+      Returns the largest value. Only return null if 'all' arguments evaluate to null.
+
+      String comparison is done in lexicographical ordering, one character at a time, from left to right.
+      Uppercase letters are less than lowercase letters.
+    impls:
+      - args:
+          - value: T
+        variadic:
+          min: 2
+        return: T
--- a/extensions/functions_comparison.yaml
+++ b/extensions/functions_comparison.yaml
@@ -218,17 +218,36 @@ scalar_functions:
   -
     name: "least"
     description: >-
-      Returns the smallest value. Only return null if 'all' arguments evaluate to null.
+      Returns the smallest value.
 
-      String comparison is done in lexicographical ordering, one character at a time, from left to right.
-      Uppercase letters are less than lowercase letters.
+      The `null_handling` option specifies how to handle null arguments.  If `ANY_NULL`
+      is specified, the function will return null if any argument evaluate to null.
+      `ALL_NULLS` means the function will return null only if all arguments evaluate to null.
+
+      String comparison is done in lexicographical ordering, one character at a time, from
+      left to right. Uppercase letters are less than lowercase letters.
+    impls:
+      - args:
+          - value: T
+        options:
+          null_handling:
+            values: [ ANY_NULL, ALL_NULLS ]
+        variadic:
+          min: 2
+        return: T|?
+  -
+    name: "least_skip_null"
+    description: >-
+      Returns the smallest value. Skip any null arguments.
+
+      String comparison is done in lexicographical ordering, one character at a time, from
+      left to right. Uppercase letters are less than lowercase letters.
     impls:
       - args:
           - value: T
         variadic:
           min: 2
         return: T
-
   -
     name: "greatest"
     description: >-


jvanstraten:
I feel like string could use some more details in the description. Case sensitivity? Lexicographic vs natural? etc. Float too maybe for NaN behavior.

westonpace:
Hmm...the entire topic of "comparison" probably deserves a nice hefty block of prose somewhere (perhaps on the site itself).  Otherwise we are at risk of repeating ourselves all over the YAML.  The same goes for overflow, overflow vs NaN, etc.

I wonder if we want a section on the website for "functions" where some of this text can live.

ianmcook:
In PostgreSQL, `least` and `greatest` return null only if _all_ the arguments evaluate to null. I think that's the behavior we want consumers to implement, so we should say something about that in the descriptions. (Because I believe in some other databases/engines, `least` and `greatest` return null if _any_ argument is null.)

richtia:
> I feel like string could use some more details in the description. Case sensitivity? Lexicographic vs natural? etc. Float too maybe for NaN behavior.

Do you or @westonpace have a suggestion for how to handle the NaN behavior?

jvanstraten:
> Hmm...the entire topic of "comparison" probably deserves a nice hefty block of prose somewhere (perhaps on the site itself).

Isn't this kind of behavior up to the extension that specifies the function, though? I could imagine different engines having subtly different native ways of doing comparisons, in which case conforming with Substrait's "defaults" might cost performance. You'd then want to have the option to override Substrait's defaults by just using a different function.

The alternative is associating ordering information with types instead, or at least default ordering information. SortRel actually already requires this for the default ascending/descending sorts, but beyond [how nulls are to be ordered](https://substrait.io/relations/basics/#orderedness) it leaves ordering up to the imagination of the user. Also, for the sort-by-function method it leaves the function signature (return type?) and the behavior for return values outside of [-1, 0, 1] unspecified.

Digression: personally I don't like having these SQL-esque default "ascending/descending" sorts at all; the implication that all types should have exactly one default ordering method seems odd to me. There is no logical way to order a 2D coordinate, for instance: you could just order by X first and then by Y, but that's as meaningless as any other sort order (Y first, by polar coordinates, by Hilbert curve, whatever). Instead, I'd much prefer having only the "custom function identifier" and "clustered" methods. If it were up to me I'd deprecate/remove the default orderings and define something like this instead:

 - By function: `compare(T, T) -> i8` or `less_than(T, T) -> boolean`
 - By function, nulls first: `compare(T!, T!) -> i8` or `less_than(T!, T!) -> boolean`
 - By function, nulls last: `compare(T!, T!) -> i8` or `less_than(T!, T!) -> boolean`
 - Clustered by implicit identity function (I don't really like it, but implicit identity is used all over the place already anyway, for example in set relations)
 - Clustered using equality or comparison function: `compare(T, T) -> i8` or `equals(T, T) -> boolean`

where by `T!` I mean the non-nullable version of the nullable type that's being sorted, so you only need to define ordering functions for non-nullable types without loss of generality.

If I'm not the only one who feels this way I can escalate this to an issue or PR.

> Otherwise we are at risk of repeating ourselves all over the YAML.

Personally I'd much rather documentation be repeated ad nauseam than only be specified in one place where someone might not find it. It requires more maintenance, but no one wants to or can be expected to scour the complete documentation for clues when they need one specific piece of information, especially when (at present) odds are that no one has thought about it yet at all, or at least has written it down anywhere. Linking to the single point of truth would also be fine (or better) but also requires maintenance to keep the links live.

jacques-n:
WRT to the sorting discussion specifically, there are two options in the spec:

You choose the structured sql type sorts with asc/desc and nulls first/nulls last OR you choose a specific function to reference. 

https://github.com/substrait-io/substrait/blob/main/proto/substrait/algebra.proto#L857

The intention I had was to formally declare a default comparison operation within the Substrait spec for known types but also allow one to use any function one wants for sorting using a direct function reference that is specified as returning -1, 0 or 1. We should add more content to the formalization of this but I feel like it allows for arbitrary alternative collations, etc while also having a more meaningful representation. I'd also be open to enhancing this so that if you choose the asc/desc/nf/nl paradigm, we could have multiple default collations to avoid having to use opaque function references if you're non-default.

richtia:
> WRT to the sorting discussion specifically, there are two options in the spec:
> 
> You choose the structured sql type sorts with asc/desc and nulls first/nulls last OR you choose a specific function to reference.
> 
> https://github.com/substrait-io/substrait/blob/main/proto/substrait/algebra.proto#L857
> 
> The intention I had was to formally declare a default comparison operation within the Substrait spec for known types but also allow one to use any function one wants for sorting using a direct function reference that is specified as returning -1, 0 or 1. We should add more content to the formalization of this but I feel like it allows for arbitrary alternative collations, etc while also having a more meaningful representation. I'd also be open to enhancing this so that if you choose the asc/desc/nf/nl paradigm, we could have multiple default collations to avoid having to use opaque function references if you're non-default.



> WRT to the sorting discussion specifically, there are two options in the spec:
> 
> You choose the structured sql type sorts with asc/desc and nulls first/nulls last OR you choose a specific function to reference.
> 
> https://github.com/substrait-io/substrait/blob/main/proto/substrait/algebra.proto#L857
> 
> The intention I had was to formally declare a default comparison operation within the Substrait spec for known types but also allow one to use any function one wants for sorting using a direct function reference that is specified as returning -1, 0 or 1. We should add more content to the formalization of this but I feel like it allows for arbitrary alternative collations, etc while also having a more meaningful representation. I'd also be open to enhancing this so that if you choose the asc/desc/nf/nl paradigm, we could have multiple default collations to avoid having to use opaque function references if you're non-default.

Do you have a suggestion of how to handle the sorting in the yaml spec.  For example, if the default were lexicographical, how would I specify a natural sorting option?

Maybe for this PR we could also just try to get in what the function signatures look like and we can document/follow up on the sorting expectations for types via a github issue.

westonpace:
> Do you have a suggestion of how to handle the sorting in the yaml spec. For example, if the default were lexicographical, how would I specify a natural sorting option?

This is unclear to me as well.  How would a custom compare function be provided to a scalar function like the ones specified here (or lt or gt)?

> Maybe for this PR we could also just try to get in what the function signatures look like and we can document/follow up on the sorting expectations for types via a github issue.

A follow-up issue seems reasonable to me given we already have functions like `lt` or `gt` that rely on the comparison of their inputs.

jvanstraten (dismissed):
LGTM as is considering the description states how strings are to be ordered, despite the discussion surrounding ordering. I could see the string versions be superseded by something more generic at some point, though.

richtia:
Needed to fix the commit messages to pass linting check

extensions/functions_comparison.yaml
```diff
@@ -83,4 +83,89 @@ scalar_functions:
...
+      - args:
+        - value: fp64
+        variadic:
+          min: 1
+        return: fp64
+      - args:
+        - value: "string"
+        variadic:
+          min: 1
+        return: "string"
```
cpcloud, line 128: Should we add the rest of the types here?

I would add everything except:

* `interval_year`
* `interval_day`
* `struct`
* `list`
* `map`

richtia: Just added all the other ones, except for the ones you listed and boolean.

jvanstraten: @cpcloud Why don't you consider intervals to be comparable? Substrait effectively defines them as a number of months and a number of microseconds respectively, so I don't see why they're special. I would personally argue that only UUIDs and maps make little sense to order, though it's perfectly possible to define an ordering for them (for maps because they are basically just defined to behave like `list<struct<K, V>>`, and both of those can be compared using tie).

IMO this should just be `least(T) -> T`/`greatest(T) -> T`. Likewise for all normal comparison functions. If an engine doesn't consider a type orderable, they will already have had to solve this problem and return suitable errors for sort relations with unorderable sort keys and no custom comparator function.

jvanstraten: This kind of slipped through the cracks because I thought we had resolved this offline (more or less), but I must admit I never considered this example.

At some point in the past, when I clarified what the built-in types mean, I made the assertion that it should be allowed to store interval types using a single number. In other words, 1 day, 24 hours, 1440 minutes, or 86400 seconds, should all mean exactly the same thing. I figured we could do this because year-month and day-second were separate types already anyway (but, again, didn't consider this one). I did this to make the description of the types at all compatible with Arrow's types, because you could fairly easily construct an interval type that stores the components separately using structs anyway, and frankly, because (even after this example) I remain convinced that it's good enough.

For this particular example, I would argue that the result depends on whether you're adding the interval to a `timestamp` or `timestamp_tz`. `timestamp` has no timezone awareness, so in both cases you would get `2022-01-02 12:00:00`. `timestamp_tz` instead represents "real" time, where DST just doesn't exist. You simply get the timestamp that occurs 24 hours/1 day later. *When represented in this particular timezone* that might result in `2022-01-01 12:00:00` -> `2022-01-02 13:00:00`, but represented in UTC it might be `2022-01-01 11:00:00Z` -> `2022-01-02 11:00:00Z` if that timezone happened to be CET. Timezone shenanigans in general are captured by the conversion between `timestamp` and `timestamp_tz`, and need not exist anywhere else.

This leaves only the more fundamental problem that months are not always the same length in our calendar, which is already covered by having different types for year-month and day-second, and the lack of overlap between these ranges. Even leap seconds (if we would want to consider those) are defined to always happen at the end of a month.

What I *don't* know is whether existing query engines and SQL operations are defined with sufficient sanity to be encompassed by this logic. I'm going to hazard a guess based on recent experiences with null and say no, so maybe we need to revise these types again. But in any case, the current definition of Substrait interval types is encompassed by a number with some implicit time unit associated with it (i.e. seconds or months), which makes them trivially ordered.

extensions/functions_comparison.yaml
```diff
@@ -214,3 +214,26 @@ scalar_functions:
         variadic:
           min: 2
         return: any1
+  -
+    name: "least"
```
cpcloud, line 218: These functions (least and greatest) should follow the signature of coalesce, i.e., variadic.

richtia: added. thanks!

extensions/functions_comparison.yaml
```diff
@@ -214,3 +214,30 @@ scalar_functions:
...
+    description: >-
+      Returns the smallest value in `List<T>`. Only return null if 'all' arguments evaluate to null.
+
+      String comparison is done in lexicographical ordering, one character at a time, from left to right.
+      Uppercase letters are less than lowercase letters.
+    impls:
+      - args:
+          - value: List<T>
```
jacques-n, line 226: I don't think the value should be a List<T>. I think it should just be a T. Variadic means that we can execute something like: 

```least(4,6,8,10).```

I don't understand what something like the below would mean. This is what you are currently indicating with the arg type of  `List<T>`, min:2.

```
least([2,4],[5,8])
```

Comment (richtia) [reply extensions/functions_comparison.yaml]: Updated! Thanks!

**Comment by richtia** [comment]
@jacques-n @cpcloud

**Review by westonpace — COMMENTED** [review]
Hmm...the return type here suffers from the same potential clarification as coalesce.  The return type is nullable only if all of the inputs are nullable.  In other words, the return type should be `T&?` instead of `T` according to the solution we brainstormed in https://github.com/substrait-io/substrait/pull/340

Then again...no one has approved #340 beyond myself so that's still up in the air.

**Comment by westonpace** [comment]
Also, the proposal here actually invalidates my example in #340 slightly (we can fix #340 if we agree least/greatest should skip nulls).  I had assumed in #340 that `least` would return null if there were any nulls in the input.

**Comment by EpsilonPrime** [comment]
If null means "I don't know" then the function should return null if any
nulls are present in the input (because the value may be less or more than
the others present).

If null means "I decline to provide a value" then skipping nulls makes
sense and the function should only return null if all the values are null
(or if the list is of zero length).

Returning null as the least value implies some semantics I haven't been
considering.  There are some sorts that return null after the largest value
which would imply that greatest would return the nulls if there were any
nulls in the input.

Given that there are four different semantics do we require four methods or
does this elevate the null handling behavior to an option?

**Comment by julianhyde** [comment]
The SQL vendors are inconsistent. In Postgres, GREATEST and LEAST return null only if all arguments are null; in BigQuery, Oracle, MySQL, they return null if any argument is null. See https://issues.apache.org/jira/browse/CALCITE-5634.

**Comment by richtia** [comment]
Would it make sense to provide an option to decide when null should be returned? (`all` arguments are null, or `any` arguments are null)

File: extensions/functions_comparison.yaml [hunk 4]
```diff
@@ -214,3 +214,30 @@ scalar_functions:
         variadic:
           min: 2
         return: any1
+  -
+    name: "least"
+    description: >-
+      Returns the smallest value. Only return null if 'all' arguments evaluate to null.
+
+      String comparison is done in lexicographical ordering, one character at a time, from left to right.
```
EpsilonPrime, line 222: By lexicographic is the assumption the "C" locale?  The ordering could be different if an alternative locale is chosen.

extensions/functions_comparison.yaml
```diff
@@ -215,3 +215,49 @@ scalar_functions:
...
+    impls:
+      - args:
+          - value: T
+        variadic:
+          min: 2
+        return: T
+  -
+    name: "greatest"
+    description: >-
+      Returns the largest value. Only return null if 'all' arguments evaluate to null.
```
EpsilonPrime, line 254: Should probably document that there is no greatest_skip_null because it behaves the same as greatest.

EpsilonPrime (dismissed):

extensions/functions_comparison.yaml
```diff
@@ -215,3 +215,46 @@ scalar_functions:
...
           min: 2
         return: any1
+  -
+    name: "least"
+    description: >-
+      Returns the smallest value.  The function will return null if any argument evaluates
+      to null.
+
+      String comparison is done in lexicographical ordering, one character at a time, from
+      left to right. Uppercase letters are less than lowercase letters.
```
westonpace, line 225:
```suggestion
      Evaluates each argument and returns the smallest one.  The function will return null if any argument evaluates
      to null.
```

extensions/functions_comparison.yaml
```diff
@@ -215,3 +215,46 @@ scalar_functions:
...
+          min: 2
+        return: T|?
+  -
+    name: "least_skip_null"
+    description: >-
+      Returns the smallest value. The function will return null only if all arguments evaluate
+      to null.
+
+      String comparison is done in lexicographical ordering, one character at a time, from
+      left to right. Uppercase letters are less than lowercase letters.
```
westonpace, line 239:
```suggestion
      Evaluates each argument and returns the smallest one. The function will return null only if all arguments evaluate
      to null.
```

extensions/functions_comparison.yaml
```diff
@@ -215,3 +215,46 @@ scalar_functions:
...
+        return: T&?
+  -
+    name: "greatest"
+    description: >-
+      Returns the largest value. Only return null if 'all' arguments evaluate to null.
+
+      String comparison is done in lexicographical ordering, one character at a time, from left to right.
+      Uppercase letters are less than lowercase letters.
+
+      There is no greatest_skip_null function because it behaves the same as greatest.
```
westonpace, line 254: I disagree with this conclusion.  There is no existing engine I am aware of that behaves this way.  For example, in both Oracle and MySQL `GREATEST(1, 3, NULL)` yields `NULL`.  The theory is that "if any one of the inputs is unknown then I cannot know which is the greatest value because it may be the unknown one"

I think we do need a `greatest_skip_null` variant and this variant should not skip nulls.

westonpace (changes requested):
In yesterday's community meeting we discussed this PR.  I think we came to the following conclusions (but I am paraphrasing here so @EpsilonPrime can feel free to correct me):

 * Being able to specify a custom comparator for sorting is a problem that affects several functions.  We should not hold up this PR while we figure that out.  In the meantime, we should assume that all types have a default comparison method and we should not explicitly mention how values are compared.
 * We should support both "skip null" and "don't skip null" variants as two different functions instead of one function with two options.

With that said, I think these descriptions need updated.  I also think we need a `greatest_skip_null`.

richtia:
@westonpace Thanks for the suggestion!  I included them and added the greatest_skip_null function

EpsilonPrime (approved):

westonpace (approved):

PR #248, opened 2022-07-15
feat: add string trimming and padding functions

richtia:
PR to add definitions for string trimming and padding functions.


extensions/functions_string.yaml
```diff
@@ -163,3 +165,146 @@ scalar_functions:
...
+        return: "string"
+  -
+    name: lpad
+    description: Left-pad the string with the characters until the specified length of the string has been reached.
+    impls:
+      - args:
+          - value: "varchar<L1>"
+          - value: i32
+          - value: "varchar<L2>"
+        return: "string"
```
richtia, line 230: The return for these padding functions kind of confused me.  I guess the return would be a varchar, where the length is the i32 input?

jvanstraten: You can't make a return type depend on an argument value. Imagine for instance a project relation that applies this function to three separate columns; what would the type of the resulting column be? You could do this if Substrait would support something akin to non-type template arguments in C++, but currently it doesn't.

extensions/functions_string.yaml
```diff
@@ -163,3 +165,146 @@ scalar_functions:
...
         return: "BOOLEAN"
+  -
+    name: ltrim
+    description: >-
+      Remove any occurrence of the characters from the left side of the string.
+      If no characters are specified, spaces are removed.
+    impls:
+      - args:
+          - value: "varchar<L1>"
+          - value: "varchar<L2>"
```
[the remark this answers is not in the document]
richtia: Oh yeah, I wasn't aware that I could do this.  I'll make these changes.  I came across a few other functions that probably could be more descriptive like this as well.  I can submit a separate PR for those later.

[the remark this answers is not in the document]
richtia: Would it make more sense to do something like this directly in the function description instead of repeating it for all implementations of the function?

```
    name: ltrim
    description: >-
      Remove any occurrence of the characters from the left side of the string.
      If no characters are specified, spaces are removed.

      arg0: input - The string to remove characters from.
      arg1: characters - The set of characters to remove.```

File: extensions/functions_string.yaml [hunk 3]
```diff
@@ -163,3 +165,236 @@ scalar_functions:
...
+          - value: i32
+            name: "length"
+            description: "The length of the output string."
+          - value: "fixedchar<L2>"
+            name: "characters"
+            description: "The set of characters to use for padding."
+        return: "string"
+  -
+    name: center
+    description: Pad the string with characters from each side until the specified length of the string has been reached.
```
thisisnic, line 334: Would this need an option to determine which side to add more items to if the number of character to add was an odd number? Or should the consumer just decide that?

extensions/functions_string.yaml
```diff
@@ -163,3 +165,236 @@ scalar_functions:
           - value: "fixedchar<L1>"
           - value: "varchar<L2>"
         return: "BOOLEAN"
+  -
+    name: ltrim
```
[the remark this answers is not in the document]
richtia: My assumption was as-is, since that seems to be how Postgresql and DuckDB do it.  For regex, they use different functions.  I was planning on looking into those functions in another PR.

extensions/functions_string.yaml
richtia: I'm not actually sure either.  From what i've seen, the default is that for uneven padding, the lesser amount of padding goes on the left.  I can at least specify that in the description.

thisisnic: So the reason I brought it up is that when I was writing R bindings for Arrow, this came up and I think we ended up requesting it being added as an option in Acero as the R library handled  it differently to Acero. Not sure how common this is though - definitely worth checking as it's a pain to work around.

jvanstraten:
See https://github.com/substrait-io/substrait/issues/251#issuecomment-1191249269, it's mostly inspired by this PR.

extensions/functions_string.yaml
```diff
@@ -248,3 +251,177 @@ scalar_functions:
...
+      - args:
+          - value: "varchar<L1>"
+            name: "input"
+            description: "The string to pad."
+          - value: i32
+            name: "length"
+            description: "The length of the output string."
+          - value: "varchar<L2>"
+            name: "characters"
+            description: "The set of characters to use for padding."
```
jvanstraten, line 333: I understand the "set of characters" concept for trim functions (`trim("foobar", "of") -> "bar"`), but not for padding. "Set" implies unordered. I don't think I've ever seen these functions in forms where you can specify more than one character either, but I guess Substrait doesn't really have a character type except maybe `fixedchar<1>`. Do these functions just repeat the padding string as many times as needed? The question then becomes what they do if the number of characters needed is uneven, especially for the `center` functions, where I can think of at least two similarly efficient algorithms that would have different results: `center("x", 10, "abc")` -> `replace_slice(substring(repeat("abc", 4), 1, 10), 5, 1, "x")` -> `"abcaxcabca"`, and `center("x", 10, "abc")` -> `concat(substring(repeat("abc", 2), 1, 4), "x", substring(repeat("abc", 2), 1, 5))` -> `"abcaxabcab"`.

jvanstraten (changes requested):

extensions/functions_string.yaml
richtia: Good point.  I'll update the wording to 'string'.  DuckDB only allows one character, but postgresql let's you use a string.  I figured we would use whatever is more flexible. 

```
Do these functions just repeat the padding string as many times as needed? The question then becomes what they do if the number of characters needed is uneven
```
Yeah, if the uneven case occurs, the padding will get truncated (for postgresql): 
lpad('abc', 10, 'XYZ') -> 'XYZXYZXabc'

To be honest, I wasn't too convinced on the center function.  I couldn't really find many places that implement it.  I imagine the expectation is that someone could easily use a combination of lpad and rpad.  I wonder if we should omit here as well?  @iancook

richtia: Another thought is that maybe we have center only take single character padding.  Multiple character strings used for padding can be just for lpad/rpad

richtia: @jvanstraten in the interest of getting a bunch of these other more common functions in, i've removed center for now.  I'll create a separate task to look more into center and if/how other SQL variants deal with it.

jvanstraten (approved):
LGTM

PR #250, opened 2022-07-20
feat: add product function

vibhatha:
This PR includes aggregate function `product`.

### DIFF
--- a/extensions/functions_arithmetic.yaml
+++ b/extensions/functions_arithmetic.yaml
@@ -637,6 +637,63 @@ aggregate_functions:
         decomposable: MANY
         intermediate: fp64?
         return: fp64?
+  - name: "product"
+    description: Product of a set of values. Returns 1 for empty input.
+    impls:
+      - args:
+          - name: overflow
+            options: [ SILENT, SATURATE, ERROR ]
+            required: false
+          - value: i8
+        nullability: MIRROR
+        decomposable: MANY
+        intermediate: i64
+        return: i8
+      - args:
+          - name: overflow
+            options: [ SILENT, SATURATE, ERROR ]
+            required: false
+          - value: i16
+        nullability: MIRROR
+        decomposable: MANY
+        intermediate: i64
+        return: i16
+      - args:
+          - name: overflow
+            options: [ SILENT, SATURATE, ERROR ]
+            required: false
+          - value: i32
+        nullability: MIRROR
+        decomposable: MANY
+        intermediate: i64
+        return: i32
+      - args:
+          - name: overflow
+            options: [ SILENT, SATURATE, ERROR ]
+            required: false
+          - value: i64
+        nullability: MIRROR
+        decomposable: MANY
+        intermediate: i64
+        return: i64
+      - args:
+          - name: rounding
+            options: [ TIE_TO_EVEN, TIE_AWAY_FROM_ZERO, TRUNCATE, CEILING, FLOOR ]
+            required: false
+          - value: fp32
+        nullability: MIRROR
+        decomposable: MANY
+        intermediate: fp64
+        return: fp32
+      - args:
+          - name: rounding
+            options: [ TIE_TO_EVEN, TIE_AWAY_FROM_ZERO, TRUNCATE, CEILING, FLOOR ]
+            required: false
+          - value: fp64
+        nullability: MIRROR
+        decomposable: MANY
+        intermediate: fp64
+        return: fp64
 window_functions:
   - name: "row_number"
     description: "the number of the current row within its partition."


extensions/functions_arithmetic.yaml
```diff
@@ -362,3 +362,54 @@ aggregate_functions:
...
+        intermediate: i64?
+        return: i64?
+      - args:
+          - options: [ SILENT, SATURATE, ERROR ]
+            required: false
+          - value: fp32
+        nullability: DECLARED_OUTPUT
+        decomposable: MANY
+        intermediate: fp64?
+        return: fp64?
```
jvanstraten, line 407: As with other PRs, why the promotion? For floats especially since all floats represent all real numbers (to some degree of accuracy), but likewise for integers.

See https://github.com/substrait-io/substrait/issues/251 (you're not the first to do it this way and probably won't be the last)

extensions/functions_arithmetic.yaml
```diff
@@ -362,3 +362,54 @@ aggregate_functions:
...
         intermediate: fp64?
         return: fp64?
+  - name: "product"
+    description: Product of a set of values.
+    impls:
+      - args:
+          - options: [ SILENT, SATURATE, ERROR ]
+            required: false
+          - value: i8
+        nullability: DECLARED_OUTPUT
```
jvanstraten, line 372: Why not `MIRROR`? I would expect the product of zero elements to be the multiplicative identity, i.e. 1, rather than null. The only other reason I see for the result to be null is if the input has nulls, which implies `MIRROR`.

extensions/functions_arithmetic.yaml
```diff
@@ -362,3 +362,54 @@ aggregate_functions:
         decomposable: MANY
         intermediate: fp64?
         return: fp64?
+  - name: "product"
+    description: Product of a set of values.
+    impls:
+      - args:
+          - options: [ SILENT, SATURATE, ERROR ]
```
jvanstraten, line 369: I'm not sure this should be solved in this PR immediately because it probably also applies to existing stuff, but something to think about for `SATURATE` in the context of these types of aggregates... What if the internal accumulator overflows but the net result would be within range? For example, `sum:opt_i64(2^63-1, 2^63-1, -2^63-1, -2^63-1)`. Is that *required* to always yield 0? Or for `ERROR`, is it *required* to throw an error (after all, if the second and third element are swapped it wouldn't overflow)?

extensions/functions_arithmetic.yaml
vibhatha: One question, wouldn't that be applicable for sum operation?

jvanstraten: Same applies there, but with additive identity i.e. 0. If it's defined like you defined product here then IMO that's equally wrong and that should probably be a separate or followup issue.

extensions/functions_arithmetic.yaml
vibhatha: Thanks for pointing this out. I also had the same doubt and I just followed the promotion. Waiting to see how the discussion progress on the promotion conclusion.

extensions/functions_arithmetic.yaml
vibhatha: Yes this is a valid point. But I guess this is hard to determine. I would say it is better to leave room for overflows. But I could be wrong.

extensions/functions_arithmetic.yaml
vibhatha: Sure, I will fix this one. Shall make a minor PR to fix the SUM operation?

vibhatha:
@jvanstraten I updated the PR with some modifications.

extensions/functions_arithmetic.yaml
jvanstraten: Maybe we should wait with this until we get some more opinions in #251. I'm inclined to say the intermediate value should be the same type, if only because picking i64 feels like an arbitrary "eh, that's probably good enough" decision that doesn't actually solve the problem, but that just makes the overflow question all the more important.

vibhatha: Sure, let’s do that.

extensions/functions_arithmetic.yaml
```diff
@@ -441,6 +441,57 @@ aggregate_functions:
...
+            required: false
+          - value: i8
+        nullability: MIRROR
```
jvanstraten, line 451: To follow the result of the sum discussion, this should be `DECLARED_OUTPUT`.

extensions/functions_arithmetic.yaml
```diff
@@ -441,6 +441,57 @@ aggregate_functions:
...
+        intermediate: fp64?
+        return: fp32?
+      - args:
+          - options: [ SILENT, SATURATE, ERROR ]
+            required: false
+          - value: fp64
+        nullability: MIRROR
+        decomposable: MANY
+        intermediate: fp64?
+        return: fp64?
```
jvanstraten, line 494: I missed this before, but these should also have the floating-point rounding options rather than overflow.

extensions/functions_arithmetic.yaml
```diff
@@ -441,6 +441,57 @@ aggregate_functions:
         decomposable: MANY
         intermediate: fp64?
         return: fp64?
+  - name: "product"
+    description: Product of a set of values.
```
jvanstraten, line 445:
```suggestion
    description: Product of a set of values. Returns null for empty input.
```

jvanstraten (changes requested):

extensions/functions_arithmetic.yaml
vibhatha: Should be okay now.

extensions/functions_arithmetic.yaml
vibhatha: Included floating-point rounding options

julianhyde:
It would be clearer if you said 'arithmetic product'. Many people would assume cartesian product.

What value does this function return if the input is empty? The documentation should specify.

In the discussion for `sum`, the consensus seemed to be that sum should return 0 if it weren't for the SQL standard requiring that `SUM` returns null. In this case there is no such precedent. I would vote for 1.

jvanstraten:
> What value does this function return if the input is empty? The documentation should specify.

It does, we mimicked sum and had it return null.

IMO we should either embrace the SQL way and apply its null nonsense consistently across all similar arithmetic aggregations (and also define a `product1` alongside `sum0`) or not at all.

@vibhatha, you'll need to rewrite your commit messages (or just rebase them all into one) to make CI pass.

vibhatha:
@jvanstraten I will rebase with an accurate commit msg.

julianhyde:
SQL seems to be getting smarter. For example `ARRAG_AGG` on an empty input returns an empty array, not NULL. Similarly `LISTAGG`. See [Snowflake](https://docs.snowflake.com/en/sql-reference/functions/listagg.html).

For statistical functions such as `AVG` and `STDEV` there are valid reasons to return NULL on empty input. So I think that `SUM` is the only one it got wrong.

vibhatha:
@jvanstraten I squashed to a single commit, let's see the CIs.

extensions/functions_arithmetic.yaml
```diff
@@ -573,6 +573,63 @@ aggregate_functions:
...
+        return: i32?
+      - args:
+          - name: rounding
+            options: [ TIE_TO_EVEN, TIE_AWAY_FROM_ZERO, TRUNCATE, CEILING, FLOOR ]
+            required: false
+          - value: i64
+        nullability: DECLARED_OUTPUT
+        decomposable: MANY
+        intermediate: i64?
+        return: i64?
```
jvanstraten, line 614: You also changed the options for the integer versions... Things that operate on integers need integer overflow options, because integers can overflow but (besides overflow) can always represent the result exactly; things that operate on floats need rounding options, because they can't overflow (they'll use +/-infinity instead) but may not be able to represent the result of the computation exactly.

extensions/functions_arithmetic.yaml
```diff
@@ -573,6 +573,63 @@ aggregate_functions:
         decomposable: MANY
         intermediate: fp64?
         return: fp64?
+  - name: "product"
+    description: Product of a set of values. Returns null for empty input.
```
jvanstraten, line 577: Having thought about this some more, let's just go with the sane version of product that returns 1 on empty input like @julianhyde is suggesting. If sum really is the only function that the SQL spec messed up, let's not use it as precedent to mess up other functions as well. That is, update the description, remove the `?` from the intermediate and return types, and remove the `nullability: DECLARED_OUTPUT` lines (`MIRROR` is the default mode).

jvanstraten (changes requested):

vibhatha:
@jvanstraten I updated the PR with `nullability` and `return` type updates.

extensions/functions_arithmetic.yaml
jvanstraten:
```suggestion
    description: Product of a set of values. Returns 1 for empty input.
```

extensions/functions_arithmetic.yaml
```diff
@@ -637,6 +637,63 @@ aggregate_functions:
...
+        return: i32
+      - args:
+          - name: rounding
+            options: [ TIE_TO_EVEN, TIE_AWAY_FROM_ZERO, TRUNCATE, CEILING, FLOOR ]
+            required: false
+          - value: i64
+        nullability: MIRROR
+        decomposable: MANY
+        intermediate: i64
+        return: i64
```
jvanstraten, line 678: The comment thread about options went out of date but it still applies.

```suggestion
      - args:
          - name: overflow
            options: [ SILENT, SATURATE, ERROR ]
            required: false
          - value: i8
        nullability: MIRROR
        decomposable: MANY
        intermediate: i64
        return: i8
      - args:
          - name: overflow
            options: [ SILENT, SATURATE, ERROR ]
            required: false
          - value: i16
        nullability: MIRROR
        decomposable: MANY
        intermediate: i64
        return: i16
      - args:
          - name: overflow
            options: [ SILENT, SATURATE, ERROR ]
            required: false
          - value: i32
        nullability: MIRROR
        decomposable: MANY
        intermediate: i64
        return: i32
      - args:
          - name: overflow
            options: [ SILENT, SATURATE, ERROR ]
            required: false
          - value: i64
        nullability: MIRROR
        decomposable: MANY
        intermediate: i64
        return: i64
```

extensions/functions_arithmetic.yaml
jvanstraten: continued here https://github.com/substrait-io/substrait/pull/250#discussion_r932376903

jvanstraten (approved):

PR #253, opened 2022-07-20
fix: remove cast function definition

almann:
Given that the `Expression` node has an explicit `Cast` variant that was added in #88, the `cast` function definition (added in #152) seems redundant.

This change is technically breaking as there could be dependent plans that are modeling a `cast` via function call over the expression.


jvanstraten (approved):
I'm +1 on removing this, and also on calling it a breaking change.

jacques-n:
Looks good, thanks @almann !

PR #255, opened 2022-07-22
fix: Minor fix sum aggregate

vibhatha:
Adding `MIRROR` for `nullability` instead of the existing `DECLARED_OUTPUT`. 
Discussed [here](https://github.com/substrait-io/substrait/pull/250#discussion_r925461283)

### DIFF
--- a/extensions/functions_arithmetic.yaml
+++ b/extensions/functions_arithmetic.yaml
@@ -191,24 +191,24 @@ aggregate_functions:
           - value: i8
         nullability: MIRROR
         decomposable: MANY
-        intermediate: i8?
-        return: i8?
+        intermediate: i64?
+        return: i64?
       - args:
           - options: [ SILENT, SATURATE, ERROR ]
             required: false
           - value: i16
         nullability: MIRROR
         decomposable: MANY
-        intermediate: i16?
-        return: i16?
+        intermediate: i64?
+        return: i64?
       - args:
           - options: [ SILENT, SATURATE, ERROR ]
             required: false
           - value: i32
         nullability: MIRROR
         decomposable: MANY
-        intermediate: i32?
-        return: i32?
+        intermediate: i64?
+        return: i64?
       - args:
           - options: [ SILENT, SATURATE, ERROR ]
             required: false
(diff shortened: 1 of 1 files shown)


jacques-n:
I don't believe this change is correct. A straight aggregate can be aggregated over zero rows of non-nullable data. In this case, I believe sum returned is null (and therefore the output is null able even if the input is not).

vibhatha:
@jacques-n thanks for pointing this out. I was merely responding to a review comment mentioned in the PR descirption. 
May be close this PR if it is not accurate?

vibhatha:
@jacques-n I have the same question for the [`aggregate:product`](https://github.com/substrait-io/substrait/pull/250) function I added recently. Is it the same case. What do you think?

jvanstraten:
> In this case, I believe sum returned is null (and therefore the output is null able even if the input is not).

Double-checked to verify that this is correct: https://www.db-fiddle.com/f/uDycvdBUd7TeeupFv1J3Zi/0. Sorry for setting you on the wrong track, @vibhatha.

I find this extremely surprising to say the least, though, and would really like to see a note about this in the description. In mathematics the sum of zero values is zero, and the product of zero values is one. Even generalizing beyond numbers, I don't think it's possible to define addition or multiplication in a way that avoids introducing such additive and multiplicative identities; for example, the identity for concatenation is the empty string/list, for set union it's the empty set and for intersection it's the set with all possible values, for matrices of some dimension it's the zero matrix for addition and the identity matrix for multiplication, etc. This seems to me like a SQL/database idiosyncrasy at best.

jvanstraten:
[Everything about this stackoverflow is terrible.](https://stackoverflow.com/questions/2802713/changing-a-sum-returned-null-to-zero) Do we really want to copy this behavior from SQL?

jacques-n:
In Calcite (and in several systems I've worked on), we've also introduced a function called SUM0. SUM0 starts from zero and counts up. It is always non-nullable whether the input is nullable or not. 

Since sum has a really strong meaning in SQL, I don't think we should redefine it's behavior here. Closing this ticket out.

cpcloud:
I believe the empty-input-as-`NULL` of `sum` are explicit in the SQL standard unfortunately.

In SQLite the mathematical sum is called `total`.

+1 for having two functions to capture the different semantics.

PR #256, opened 2022-07-22
feat: add string containment functions

richtia:
PR to add definitions for string containment functions.


extensions/functions_string.yaml
```diff
@@ -163,3 +163,355 @@ scalar_functions:
...
+      - args:
+          - value: "fixedchar<L1>"
+            name: "input"
+            description: The input string.
+          - value: "varchar<L1>"
+            name: "substring"
+            description: The substring to search for.
+        return: i8
+  -
+    name: regexp_strpos
```
richtia, line 245: For these regular expression ones, I'm not sure whether the input and regex pattern can have mismatched types, so i just kept it consistent for now.

extensions/functions_string.yaml
```diff
@@ -163,3 +163,355 @@ scalar_functions:
...
+            name: "input"
+            description: Input string.
+          - value: "fixedchar<L1>"
+            name: "pattern"
+            description: The regular expression pattern.
+        return: i8
+  -
+    name: regexp_matches
+    description: >-
+      Returns true if part of the input string matches the regular expression pattern.
```
jvanstraten, line 393: The term "matches" is a bit overloaded when it comes to regexes. Does `bananas` match the regex `nan`? Or what about `ban`? Basically, is there an implicit `^` at the start and/or an implicit `$` at the end of the regex or not?

extensions/functions_string.yaml
```diff
@@ -163,3 +163,355 @@ scalar_functions:
...
+          - value: "fixedchar<L1>"
+            name: "input"
+            description: Input string.
+          - value: "fixedchar<L1>"
+            name: "pattern"
+            description: The regular expression pattern.
+        return: "BOOLEAN"
+  - name: replace
+    description: >-
+      Replace any occurrence of the substring with the replacement string
```
jvanstraten, line 427: "any" as in one of them or as in all of them? If the former, which one?

I think if you're being properly pedantic about what this says it unambiguously means all occurrences, but I'd rather see it written as "replaces all occurrences" to avoid confusion.

Also, nit: missing `.` at the end of the sentence.

extensions/functions_string.yaml
```diff
@@ -163,3 +163,355 @@ scalar_functions:
...
+          - value: "fixedchar<L1>"
+            name: "input"
+            description: Input string.
+          - value: "fixedchar<L1>"
+            name: "substring"
+            description: The substring to replace.
+          - value: "fixedchar<L1>"
+            name: "replacement"
+            description: The replacement string.
+        return: "fixedchar<L1>"
```
jvanstraten, line 461: This one in particular is fun for fixedchar. I think it works if `substring` and `replacement` have length `L2` and the input and return value have length `L1`, but for differing `substring` and `replacement` lengths you'd get a data-dependent return type.

extensions/functions_string.yaml
```diff
@@ -163,3 +163,355 @@ scalar_functions:
...
+            name: "input"
+            description: The input string.
+          - value: "varchar<L1>"
+            name: "substring"
+            description: The substring to search for.
+        return: i8
+  -
+    name: regexp_strpos
+    description: >-
+      Return the position of the first occurrence of a regular expression pattern in another
```
jvanstraten, line 247: > a regular expression pattern

Which flavor? There are many, and some vary greatly in terms of performance.

extensions/functions_string.yaml
```diff
@@ -163,3 +163,355 @@ scalar_functions:
...
+    impls:
+      - args:
+          - options: [ CASE_SENSITIVE, CASE_INSENSITIVE ]
+            required: false
+          - value: "string"
+            name: "input"
+            description: The input string.
+          - value: "string"
+            name: "pattern"
+            description: The regular expression pattern.
```
jvanstraten, line 259: We might want to consider requiring the regex to be constant. If not, we need to consider what happens if the pattern fails to compile on a row-by-row basis.

extensions/functions_string.yaml
```diff
@@ -163,3 +163,355 @@ scalar_functions:
...
+            required: false
+          - value: "string"
+            name: "input"
+            description: Input string.
+          - value: "string"
+            name: "pattern"
+            description:  The regular expression for matching.
+          - value: "string"
+            name: "replacement"
+            description: The string to replace the regular expression match with.
```
jvanstraten, line 482: Do we support backreferences in replacements? Named or only numbered?

jvanstraten (commented):
Note that many of my comments apply all over the place, this list is not exhaustive. Regexes in particular are a huge can of worms to open if we want any usable level of precision as to what needs to be supported and what doesn't. Maybe it's better to postpone the regex stuff to a later PR.

richtia:
> Note that many of my comments apply all over the place, this list is not exhaustive. Regexes in particular are a huge can of worms to open if we want any usable level of precision as to what needs to be supported and what doesn't. Maybe it's better to postpone the regex stuff to a later PR.

Thanks for all the input!  It definitely seems like there should be a longer discussion/writeup before the regex stuff is attempted.  I'll remove those from this PR.  And we can use your initial comments on here later on.

extensions/functions_string.yaml
```diff
@@ -163,3 +163,355 @@ scalar_functions:
           - value: "fixedchar<L1>"
           - value: "varchar<L2>"
         return: "BOOLEAN"
+  -
+    name: strpos
+    description: >-
+      Return the position of the first occurrence of a string in another string. The first
+      character of the string is at position 1. If no occurrence is found, 0 is returned.
```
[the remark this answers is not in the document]
richtia: Yeah...this seems to be what DuckDB and postgresql do :( 

https://www.postgresqltutorial.com/postgresql-string-functions/postgresql-position/
https://duckdb.org/docs/sql/functions/char -> strpos

extensions/functions_string.yaml
```diff
@@ -163,3 +163,355 @@ scalar_functions:
...
+    impls:
```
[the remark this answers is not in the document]
richtia: I thought it may have been a little overkill too.  Wasn't actually sure if the same stuff applied though.  Removed a bunch of them.

extensions/functions_string.yaml
```diff
@@ -163,3 +163,355 @@ scalar_functions:
...
+          - value: "fixedchar<L1>"
+            name: "input"
+            description: The input string.
+          - value: "fixedchar<L1>"
+            name: "pattern"
+            description: The regular expression pattern.
+        return: i8
+  -
+    name: count_substring
+    description: Returns the number of times a substring occurs in an input string.
```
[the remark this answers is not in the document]
richtia: Good question.  I couldn't find if there was a standard for databases around this, but my first guess is that default would be non-overlapping, since that seems like a more common thing to do.  Would it make sense to have non-overlapping as default and provide an options for this function?

This was in the Acero code, but it doesn't specify about overlap either:
https://github.com/apache/arrow/blob/master/cpp/src/arrow/compute/kernels/scalar_string_ascii.cc#L1875-L1880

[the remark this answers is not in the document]
jvanstraten: I also figured it'd be without overlap, but it wouldn't hurt to specify. I'm fine with either just putting that in the description or adding an option for it.

extensions/functions_string.yaml
```diff
@@ -163,3 +163,87 @@ scalar_functions:
...
+          - value: "fixedchar<L1>"
+            name: "input"
+            description: The input string.
+          - value: "fixedchar<L2>"
+            name: "substring"
+            description: The substring to count.
+        return: i64
+  - name: replace
+    description: >-
+      Replace all occurrence of the substring with the replacement string.
```
jvanstraten, line 226:
```suggestion
      Replace all occurrences of the substring with the replacement string.
```

jvanstraten:
I'll leave the regex discussions as unresolved, but aside from above grammar nit and overlapping/non-overlapping option/description for `count_substring` I'm good with this.

richtia:
> I'll leave the regex discussions as unresolved, but aside from above grammar nit and overlapping/non-overlapping option/description for `count_substring` I'm good with this.

Updated both!  I decided to stick with just the description update for the overlap/non-overlap for now, since I couldn't actually find anything like this that provides that option as part of a function.

jvanstraten (approved):

extensions/functions_string.yaml
```diff
@@ -163,3 +163,87 @@ scalar_functions:
...
+            description: The substring to replace.
+          - value: "string"
+            name: "replacement"
+            description: The replacement string.
+        return: "string"
+      - args:
+          - value: "varchar<L1>"
+            name: "input"
+            description: Input string.
+          - value: "varchar<L1>"
```
jacques-n, line 243: Same as above.

richtia: Updated both

jacques-n (dismissed):
Thanks @richtia !

jacques-n:
Looks like a check is failing. @richtia , can you address? Thx!

richtia:
> Looks like a check is failing. @richtia , can you address? Thx!

@jacques-n Yep. Just updated one of my commit messages due to linting issues.

extensions/functions_string.yaml
```diff
@@ -163,3 +163,87 @@ scalar_functions:
...
+            name: "input"
+            description: Input string.
+          - value: "varchar<L2>"
+            name: "substring"
+            description: The substring to replace.
+          - value: "varchar<L3>"
+            name: "replacement"
+            description: The replacement string.
+        return: "varchar<L1 - L2 + L3>"
```
jvanstraten, line 249: Not sure I agree with that one. The actual length depends on the number of replacements, this arbitrarily assumes there was one replacement. I'd just put L1 there.

richtia: Ahh...good point.  Updated.  Thanks!

jvanstraten (approved):

Issue #251, opened 2022-07-20
Implicit casts/promotions in functions within Substrait's core extensions


jvanstraten:
This has come up in a number of PRs now, and it's becoming a bit of a mess. I feel like we need to come to some collective decision about how to approach this, and then document clearly what Substrait's stance on this is:

**Given for example something like `multiply(i32, i32)`, should it A) return `i32` for consistency, or B) promote to `i64` because `i32` can overflow?**

Note that most examples are far more subtle, such as a `sum(i8) -> i64` aggregate or `power(fp32, fp32) -> fp64`, because they can still overflow or lose precision but are just less likely to do so with the larger return type.

I personally feel quite strongly that functions should never promote, i.e. option A, no exceptions. This is how virtually all programming languages behave: you define only `(i8, i8) -> i8`, `(i16, i16) -> i16`, ..., `(i64, i64) -> i64`, and then have a separate mechanism to promote or convert between representations. In most languages that promotion is implicit (in C in particular integers are usually i32 by default, so i8 + literal -> i32 and thus you might not notice), in Substrait all such promotions must be explicit via a cast operation (which is much easier on consumers). The net result is the same. Just as it is up to the producer to make sure that there are no implicit casts, it should also be up to the producer (or user) to promote if needed.

Here's a non-exhaustive list of problems and inconsistencies I see with option B:

 - What if we later add i128, fp80/fp128, bigint, etc? Do we then impose a breaking change for basically all of Substrait to maintain consistency, or do we make new functions? If the latter, why not also versions that promote only up to 32-bit values?
 - What if a particular consumer adds those types as extensions instead?
 - What about `add(i8, i8)`? The correct result type would be `i9`.
 - What about aggregations, which can overflow no matter how large you make the result type? (unless we would introduce bigints and arbitrary-precision floats)
 - What if a consumer doesn't support the larger types for whatever reason? (unlikely with the types currently defined by Substrait, just pointing out the inconsistencies here)

[...]

PR #257, opened 2022-07-25
feat: add standard deviation functions

vibhatha:
This PR includes standard_deviation definition.

### DIFF
--- a/extensions/functions_arithmetic.yaml
+++ b/extensions/functions_arithmetic.yaml
@@ -694,6 +694,23 @@ aggregate_functions:
         decomposable: MANY
         intermediate: fp64
         return: fp64
+  - name: "std_dev"
+    description: Calculates standard-deviation for a set of values.
+    impls:
+      - args:
+          - name: rounding
+            options: [ TIE_TO_EVEN, TIE_AWAY_FROM_ZERO, TRUNCATE, CEILING, FLOOR ]
+            required: false
+          - value: fp32
+        nullability: DECLARED_OUTPUT
+        return: fp32?
+      - args:
+          - name: rounding
+            options: [ TIE_TO_EVEN, TIE_AWAY_FROM_ZERO, TRUNCATE, CEILING, FLOOR ]
+            required: false
+          - value: fp64
+        nullability: DECLARED_OUTPUT
+        return: fp64?
 window_functions:
   - name: "row_number"
     description: "the number of the current row within its partition."
--- a/text/simple_extensions_schema.yaml
+++ b/text/simple_extensions_schema.yaml
@@ -200,7 +200,7 @@ $defs:
         items:
           type: object
           additionalProperties: false
-          required: [ intermediate, return ]
+          required: [ return ]
           properties:
             args:
               $ref: "#/$defs/arguments"


jvanstraten:
#251 applies. I would expect to see only fp32 -> fp32 and fp64 -> fp64 stddev, mean, etc.

vibhatha:
> #251 applies. I would expect to see only fp32 -> fp32 and fp64 -> fp64 stddev, mean, etc.

I thought about this, since the discussio is active here, I decided to keep the promotion. I am also aligned to non-promotion. I will update this.

extensions/functions_arithmetic.yaml
```diff
@@ -362,3 +362,54 @@ aggregate_functions:
...
+        nullability: DECLARED_OUTPUT
+        decomposable: MANY
+        intermediate: "fp64?"
+        return: i64?
+      - args:
+          - options: [ SILENT, SATURATE, ERROR ]
+            required: false
+          - value: fp32
+        nullability: DECLARED_OUTPUT
+        decomposable: MANY
```
jvanstraten, line 405: I'm pretty sure standard deviation isn't decomposable. AFAIK it can only be computed once the mean is already known.

extensions/functions_arithmetic.yaml
```diff
@@ -362,3 +362,54 @@ aggregate_functions:
...
+        intermediate: "fp64?"
+        return: i32?
+      - args:
+          - options: [ SILENT, SATURATE, ERROR ]
+            required: false
+          - value: i64
+        nullability: DECLARED_OUTPUT
+        decomposable: MANY
+        intermediate: "fp64?"
+        return: i64?
```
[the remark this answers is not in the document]
vibhatha: Initially, I looked into the usage in higher level, so even if we use integers the output should be in float. That's why I included the return as `fp64`. I am not sure about a practical use case for this, assuming we are excluding a data type, does that mean Substrait doesn't approve standard deviation calculated on integer values?

extensions/functions_arithmetic.yaml
vibhatha: Please correct me if I am wrong. 

Standard deviation is the `sort(variance)`. 

Variance can be calculated by using `mean` (`avg`) and deviation for each sample. 

So if we partition the data, we can calculate the mean for each partition in an embarassingly parallel manner. Then using an `allreduce` (aggregate in parallel) call we can find the global `mean`. Then we can calculate the `deviation` for each sample in parallel for each partition. Again we can use `allreduce` to sum these values up and also get the count of values in each partition. Now we have total sample count and deviation sum for all values. In the rank=0 process we can divide the sum by sample count. Then taking the `sqrt` we can calculate `std_dev`. 

Is this wrong? Am I not following the definition of decomposability?

vibhatha: Even if this is correct, my definition is wrong, there has to be a `struct<V1, V2>`, I need to fix this either way.

jvanstraten: That works for parallelizing stddev in general, but violates causality for the way Substrait deals with decomposition. Decomposition in Substrait means separating the aggregation between independent plans, so for instance one plan is pushed to several storage nodes and does the first part of the aggregation, and then another plan operates on a column containing the results from those plans and finishes the computation. That's why the intermediate type matters for as far as Substrait is concerned; how it's parallelized at the level of a single plan is not Substrait's concern. When applied to stddev, you can first indeed compute the mean in a decomposable manner, but then you need that mean *and* all individual rows in order to compute the variance (which again you could do in parallel, but you don't have the individual rows anymore at this point). At best the intermediate type could be something like `struct<fp64, list<fp64>>` to track the mean and all the values, but that seems a bit silly; at that point you can make all aggregates "decomposable" by just making the intermediate type the list of all input values thus far.

jacques-n: For reference, this is the way that Calcite does [decomposition of these functions](https://calcite.apache.org/javadocAggregate/org/apache/calcite/rel/rules/AggregateReduceFunctionsRule.html): 

```
AVG(x) → SUM(x) / COUNT(x)
STDDEV_POP(x) → SQRT( (SUM(x * x) - SUM(x) * SUM(x) / COUNT(x)) / COUNT(x))
STDDEV_SAMP(x) → SQRT( (SUM(x * x) - SUM(x) * SUM(x) / COUNT(x)) / CASE COUNT(x) WHEN 1 THEN NULL ELSE COUNT(x) - 1 END)
VAR_POP(x) → (SUM(x * x) - SUM(x) * SUM(x) / COUNT(x)) / COUNT(x)
VAR_SAMP(x) → (SUM(x * x) - SUM(x) * SUM(x) / COUNT(x)) / CASE COUNT(x) WHEN 1 THEN NULL ELSE COUNT(x) - 1 END
COVAR_POP(x, y) → (SUM(x * y) - SUM(x, y) * SUM(y, x) / REGR_COUNT(x, y)) / REGR_COUNT(x, y)
COVAR_SAMP(x, y) → (SUM(x * y) - SUM(x, y) * SUM(y, x) / REGR_COUNT(x, y)) / CASE REGR_COUNT(x, y) WHEN 1 THEN NULL ELSE REGR_COUNT(x, y) - 1 END
REGR_SXX(x, y) → REGR_COUNT(x, y) * VAR_POP(y)
REGR_SYY(x, y) → REGR_COUNT(x, y) * VAR_POP(x)
```

From there, SUM is decomposed into SUM0 in numerous places.

vibhatha: Thanks for the detailed explanation @jacques-n 
Should these be included in an Appendix section somewhere in the documentation. For instance, there are terminologies/standards that the Substrait community is trying to bring forward. So it would be very helpful if pointers to such definitions are there with more facts.

westonpace: Who is the audience for the decomposition field?  Is this advice to consumers / planners on how they might be able to decompose something?  Is it a restriction that limits what decompositions are allowed?

vibhatha:
@jvanstraten updated the PR with the recent suggestions.

extensions/functions_arithmetic.yaml
jvanstraten: I think the audience is a hypothetical optimizer & task distribution engine. If it gets a plan with a decomposable aggregation function, it can use the decomposability information and intermediate type to split the plan up and distribute it over nodes. Using `[X]` to signify a column of type `X`, saying an aggregate function from `[X] -> Z` is `ONE`- or `MANY`-decomposable with intermediate type `Y` implies that `[X] -> Y` (`INITIAL_TO_INTERMEDIATE`) and `[Y] -> Z` (`INTERMEDIATE_TO_RESULT`) can also be used in plans, and that piping the results of a number of `[X] -> Y` invocations to a single `[Y] -> Z` invocation yields the same result as doing `[X] -> Z` on all `X`s in one go. If it's `MANY`-decomposable `[Y] -> Y` (`INTERMEDIATE_TO_INTERMEDIATE`) also exists, allowing a distributor to use more exotic distribution patterns as well. Selection between these modes of operation happen using the "aggregation phase" option when binding the aggregate function (the default being `INITIAL_TO_RESULT` for the `[X] -> Z` variant), so effectively you're declaring several aggregate functions in one go by defining one to be decomposable.

How an engine implements a function internally (parallelization strategy, distribution over nodes, etc.) is out of scope for Substrait. It's ultimately just an interface, after all; such things are implementation details that are intentionally hidden by that interface.

jvanstraten (commented):
I still don't know how I feel about the integer variants of these functions. I don't really see a use case for them; in fact I'd sooner implement them for decimals than for integers.

extensions/functions_arithmetic.yaml
```diff
@@ -497,3 +536,6 @@ window_functions:
         decomposable: NONE
         return: i64?
         window_type: PARTITION
+        decomposable: MANY
+        intermediate: "fp64?"
+        return: fp64?
```
jvanstraten, line 541: This doesn't look right.

extensions/functions_arithmetic.yaml
```diff
@@ -441,6 +441,45 @@ aggregate_functions:
...
+            required: false
+          - value: fp32
+        nullability: DECLARED_OUTPUT
+        return: fp32?
+      - args:
+          - options: [ SILENT, SATURATE, ERROR ]
+            required: false
+          - value: fp64
+        nullability: DECLARED_OUTPUT
+        return: fp64?
```
jvanstraten, line 482: Please use the [floating point rounding options](https://github.com/substrait-io/substrait/blob/d6b9b344f0f0865573a79feb6ec818c146b47f62/extensions/functions_arithmetic.yaml#L255-L257) rather than integer overflow options for floats.

jvanstraten (changes requested):

extensions/functions_arithmetic.yaml
vibhatha: Typo, fixed.

extensions/functions_arithmetic.yaml
vibhatha: Just to make sure, if I understood this right, we are not including this for integer inputs?

extensions/functions_arithmetic.yaml
jvanstraten: No, what doesn't look right is that these three lines came out of nowhere at the bottom of the file. They should be removed.

extensions/functions_arithmetic.yaml
jvanstraten: It's still an open question to me if the integer input versions should even exist (I feel pretty strongly that they shouldn't).

If they do, well, you're going to have to specify something about rounding for sure, just not the floating point options because those are specific to floating points. I mean, you're doing divisions and a sqrt at the end of the computation there. Division *is* defined for integers but won't do what you want (it rounds toward zero or negative infinity depending on who you ask, which in itself is a problem, but also means it's statistically biased), but sqrt certainly isn't.

Anyone intending to use the integer versions is still going to expect floating point return types from these. But then the question is which, i.e. fp32 or fp64? Or even some decimal? The solution to that, from a user-perspective, would be to just cast to the real-number type of your choice first, and then use that version of the function. So the integer versions would never be used.

vibhatha: That make sense, let me remove the integer types from this.

vibhatha: Removed the integer types.

extensions/functions_arithmetic.yaml
vibhatha: Yes, I didn't catch that sorry. Now it should be okay.

jvanstraten (approved):
LGTM now.

jvanstraten:
Hm, CI doesn't agree with me though.
 - You'll need to rebase and make all your commits pass commitlint for the semantic release check. IMO it's a bit silly because your commits will be squashed anyway, but still...
 - The YAML validation error is because the intermediate type is erroneously marked as required. This has come up in an issue before, but I forget which. Apparently that one didn't get merged yet. Can you loosen the restriction by changing https://github.com/substrait-io/substrait/blob/d4cfbe014e9c126ac008094323a2baca9f47c42d/text/simple_extensions_schema.yaml#L203 to just:
   ```yaml
   required: [ return ]
   ```
   to make CI pass?

jacques-n:
@jvanstraten, I think the intermediate change was only for window functions. Looks like the same should be applied to aggregates.

jvanstraten:
Ah, I forgot those were separate things in the YAML schema. That makes sense.

vibhatha:
@jacques-n I squashed it to a single commit message. Let's check the CI's. 
Do I need to make any changes to CI rules? Did I misundertood [this](https://github.com/substrait-io/substrait/pull/257#issuecomment-1197044193)

jvanstraten (approved):
I just made the schema change myself. LGTM now.

jacques-n:
@vibhatha , can you rebase so there aren't merge conflicts?

vibhatha:
@jacques-n rebased and resolved conflicts. Please check it once more.

jacques-n (approved):
Thanks @vibhatha !

Issue #258, opened 2022-07-25
Clarify intention that a single grouping set aggregate does not output a group id


jacques-n:
We have this currently in the docs: 

> The list of distinct columns from each grouping set (ordered by their first appearance) followed by the list of measures in declaration order, followed by an integer describing the associated particular grouping set the value is derived from.

I think we should update it to clarify two things:
- In the case of a single grouping set, the integer is not output.
- The integer is zero-indexed.

jacques-n:
CC @westonpace and @jvanstraten

jacques-n:
The changes above were part of the original intention but somehow failed to make it into the documentation.

jvanstraten:
Usually I would scream "breaking change!" but for this one I don't even know, since I'm pretty sure the validator is literally the only implementation doing this "correctly." I can submit a patch if you want.

jvanstraten:
Just to clarify,

> In the case of a single grouping set, the integer is not output.

I'm assuming you mean the case for zero grouping sets (i.e. only measures) as well.

jacques-n:
> "breaking change!"

Yeah, I have no idea how this information got lost. I'm wondering if the reason Weston thinks the same thing I do is that we had a slack discussion post his reading of the docs and the two got merged together. I hunted around for a better description in our docs and as far as I can tell, just failed to write this down... I'm okay calling this a breaking change. On the flipside, it will make more things compliant :D

> I'm assuming you mean the case for zero grouping sets (i.e. only measures) as well.

Yup, good point.

A patch would be great.

PR #260, opened 2022-07-25
docs: better explain aggregate relations

jvanstraten:
BREAKING CHANGE: the grouping set index column now only exists if there is more
than one grouping set.


jvanstraten:
Closes #258

jvanstraten:
Apparently buf is unable to build the protos within two minutes now...? FWIW I've also seen it be ridiculously slow on my machine.

cpcloud:
The buf failure looked like an intermittent network problem, possibly related to https://github.com/bufbuild/buf/issues/824.

jvanstraten:
It was hanging on the `buf generate` call though, so unless it was still downloading dependencies by then I don't think that was the issue just now. Weird. Add it to the pile of protobuf-related issues I suppose :man_shrugging:

jacques-n (approved):
Thanks!

Issue #259, opened 2022-07-25
Introduce a SUM0 Aggregate


jacques-n:
SUM0 is an aggregate that behaves the same as Sum in all cases except when no values are added to the sum. In SUM, the output is null. In SUM0, the output is zero.

jacques-n:
This came up while discussing #255 .

julianhyde:
+1 I made the same point in #262 before I saw this issue.

jacques-n:
Note that I believe @cpcloud said this was called `total` in Postgres. That seems better than using the internal Calcite name. Are there other common names for this function in databases or dataframes? No strong opinion from my side but would be good to be as consistent with prior art as possible.

jvanstraten:
I prefer `sum0`; `total` feels like something much more generic than a sum to me. But maybe that's just me.

PR #262, opened 2022-07-25
docs: clarify result of sum() aggregate for zero elements

jvanstraten:


jacques-n (approved):

julianhyde:
It's good to clarify what the behavior is, so I have no argument with this PR.

And I will note that this behavior is consistent with SQL's `SUM` aggregate function.

Still, there is a good argument that `sum()` should return zero, not null, for zero elements. Or that there should be another function `sum0` with that behavior. It easier to compute (because less internal state is required). It is easier to consume (because the caller doesn't have to deal with nulls). And anyone who wants the SQL-like behavior can write `case count(x) when 0 then null else sum0(x) end`. In Calcite we find `SUM0` to be a more useful primitive, and `SUM` is derived, like `AVG`.

jvanstraten:
See #259; adding `sum0` as well is planned.

julianhyde:
Good to hear. Sorry I didn't notice #255 and #259 earlier.

FWIW, you made some remarks in another case about the need for additive and multiplicative identities. I agree that identities are useful. But it wasn't totally clear whether you were thinking of NULL the additive identity. I have given a lot of thought to the algebraic aspects of SQL and I think it is wrong to think of NULL as an identity. Most SQL aggregate functions, `SUM` included, do not even see input values that are NULL.

jvanstraten:
I was talking about the mathematical identities, i.e. the x in x + y = y + x = y for addition. So, zero.

PR #264, opened 2022-07-26
chore: add commitlint to pre-commit hooks

jvanstraten:


cpcloud (approved):
Neat, we were just talking about something like this for ibis.

PR #265, opened 2022-07-26
feat: support for simple extensions dependencies

jvanstraten:


jvanstraten:
This one has been on my mental TODO list for a while now: it's currently impossible for an extension to define functions that make use of types from another extension. Some problems with that:

 - The Substrait core extensions are split over a whole bunch of files because, well, it'd be an absolute mess otherwise, but consumers that define their own extension types currently can't follow the same paradigm for any function that uses any extension type.
 - Consumer X may want to make use of the types defined by consumer Y for extra compatibility. For example, Arrow is/will be defining extension types for all the Arrow types that don't exist in Substrait core. Execution engines besides Arrow/Acero may want to define functions for those types as well, if they use Arrow's type system internally.

AFAICT, the only missing link for supporting this as far as the specification is concerned is the lack of a way to define and refer to extension URI dependencies in YAML, so that's what this PR adds.

jacques-n:
Seems reasonable. Do you want to also write up some markdown for this?

jvanstraten:
Yep, good point.

site/docs/extensions/index.md
```diff
@@ -23,6 +23,27 @@ A Substrait plan can reference one or more YAML files via URI for extension. In
 | Type Variation     | The name as defined on the type variation object.            |
 | Function Signature | In a specific YAML, if there is only one function implementation with a specific name, a extension type declaration can reference the function using either simple or compound references. Simple references are simply the name of the function (e.g. `add`). Compound references (e.g. `add:i8_i8`)are described below. |
 
+A YAML file can also reference types and type variations defined in another YAML file. To do this, it must declare the YAML file it depends on using a key-value pair in the `dependencies` key, where the value is the URI to the YAML file, and the key is a valid identifier that can then be used as an identifier-safe alias for the URI. This alias can then be used as a `.`-separated namespace prefix wherever a type class or type variation name is expected.
+
+For example, if the YAML file at `/extension_types.yaml` defines a type called `point`, a different YAML file can use the type in a function declaration as follows:
+
+```yaml
+dependencies:
+  ext: /extension_types.yaml
```
jacques-n, line 32: Is this a valid URI?

EpsilonPrime: It is a valid URI reference which we tend to accept in implementations.  To make this a better example we could just append `file:`.

EpsilonPrime: I've made that change.

EpsilonPrime (dismissed):

PR #269, opened 2022-07-28
refactor(arithmetic): specify FP overflow and domain options for rema…

gforsyth:
…ining ops

Addition, substraction, multiplication and division were added before we'd settled on / had
conversations around what made sense for floating-point overflow and domain error options.  Adding
in appropriate overflow options for those four ops and also adding the appropriate domain error
option for division.

BREAKING CHANGE: Options SILENT, SATURATE, ERROR are no longer valid for use with floating point
arguments to add, subtract, multiply or divide

### DIFF
--- a/extensions/functions_arithmetic.yaml
+++ b/extensions/functions_arithmetic.yaml
@@ -30,13 +30,15 @@ scalar_functions:
           - value: i64
         return: i64
       - args:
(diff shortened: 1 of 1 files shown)


jvanstraten (approved):
LGTM, thanks!

jacques-n:
Thanks for putting this together @gforsyth !

Issue #137, opened 2022-02-05
How should we express pushdown filtering?


westonpace:
The `ReadRel` has a filter property defined (in the markdown) as:

> A boolean Substrait expression that describes the filter of a iceberg dataset. TBD: define how field referencing works.

I assume this is intended for pushing filtering down into the read.  However, it isn't clear to me if it is expected that the consumer is able to fully satisfy the filter or not.  There are many cases where filters cannot be pushed down, and some of those cases won't be known at plan creation time.

For example, if the source files are CSV then no filter pushdown is possible.  If the source files are parquet but there are no column statistics or bloom filters enabled for a particular source file.

As a result, in Arrow, we typically define plans with both a pushdown filter and a filter node.  The pushdown filter is basically a best-effort hint.

Is this the same intent in Substrait?  Or, if a consumer can't satisfy the filter in the ReadRel, is it expected that the consumer will do the additional filtering in-memory?

jacques-n:
The TBD part of the definition is whether the filter is defined based on the original schema of the data or the mask selection. We need to choose one.


For the optionality: hmmm...

The current intended model is that a filter expressed must be applied. In the past, for this kind of optionality, I've defined it inside the scan operator rather than a separate concept in the plan. This is because one doesn't want to reapply the filter in the case that you can be done as part of the read (or part of it can) since that would waste time/cycles. 

I suppose we could create something which changes filter behavior from `required` to `best effort`. That being said, I'm initially inclined to model as an internal concern of the read though, instead of something that is exposed. What do you think?

westonpace:
Sorry for the delay in coming back to this.  I think leaving it a concern of the read operator initially is fine.  If there is ever a reason we need to delay the optional filtering we can add the complexity later (or just use hints)

westonpace:
I'll create a PR to clarify the wording.

> The TBD part of the definition is whether the filter is defined based on the original schema of the data or the mask selection. We need to choose one.

My two cents would be to use the original schema.  The masked schema is defined as `the portions of the content that should be read`.  The purpose of having a pushdown filter is that sometimes we can filter contents without reading them.  So it seems natural we'd be able to filter on contents that are not read.

jacques-n:
Sounds good. Makes sense to me.

curino:
[...]

PR #271, opened 2022-07-29
docs: add details to the read operator stating that the filter applies to the pre-masked schema and must be fully satisfied

westonpace:


westonpace:
Closes #137

site/docs/relations/logical_relations.md
```diff
@@ -19,11 +19,15 @@ The read operator is an operator that produces one output. A simple example woul
...
 | Projection        | A masked complex expression describing the portions of the content that should be read | Optional, defaults to all of schema  |
 | Output properties | Declaration of orderedness and/or distribution properties this read produces. | Optional, defaults to no properties. |
 | Properties        | A list of name/value pairs associated with the read.         | Optional, defaults to empty          |
 
+### Read Filtering
+
+Consumers can often take advantage of a ReadRel's filter property, combined with file metadata, statistics, and indices to reduce the amount of data the needs to be read.  This technique is often referred to as "pushdown filtering".  In many cases this is inexact and a filter can only be partially satisfied.  In these cases the consumer must apply some kind of in-memory filtering operation to fully satisfy the filter.
```
jacques-n, line 29: I think we need this to have a boolean called filter_best_effort which defaults to false (and is therefore backwards compatible with the current definition/intention). If set to true, the scanner may only do best effort filtering. As you put it right now, it sounds like the correct interpretation of a filter is best effort (e.g. may be completed ignored). That wasn't the intention and doesn't feel consistent with least surprise.

westonpace: I thought the plan was to add such a field later.  But I don't mind adding one now.

> As you put it right now, it sounds like the correct interpretation of a filter is best effort

I need to work on my phrasing then.  I was trying to imply the exact opposite.

westonpace: Should this be a boolean options or an enum with a 2 values?  I can't think of anything between "best_effort" and "exact" at the moment but maybe someone would want something where a certain class of expressions must be satisfied? (e.g. "equalities" to mean that "x == 0 and is_cool(y)" should at least evaluate "x == 0"?)

jacques-n: Lol. If you meant the opposite, I'm fine without the field. I also think it is fine to add the field as this definitely is fairly common.

jacques-n: I suppose an alternate approach per your comments is to add a second field that is best_effort_filter, which if populated, may be applied. This allows people to separate the two if they want. (your in-between case)

westonpace: I've updated my wording to be more clear.  I think it's best to leave the `best_effort_filter` out at the moment.  I'm not sure if this is just laziness on my part but my rationalization is that this will present producers with a difficult dilemma picking which one to use and we should wait until there is a concrete case so we can better understand.

site/docs/relations/logical_relations.md
```diff
@@ -19,11 +19,17 @@ The read operator is an operator that produces one output. A simple example woul
...
 | Projection        | A masked complex expression describing the portions of the content that should be read | Optional, defaults to all of schema  |
-| Output properties | Declaration of orderedness and/or distribution properties this read produces. | Optional, defaults to no properties. |
+| Output Properties | Declaration of orderedness and/or distribution properties this read produces. | Optional, defaults to no properties. |
 | Properties        | A list of name/value pairs associated with the read.         | Optional, defaults to empty          |
 
+### Read Filtering
+
+Consumers can often take advantage of a ReadRel's filter property, combined with file metadata, statistics, and indices to reduce the amount of data the needs to be read. This technique is often referred to as "pushdown filtering". In many cases this is inexact and a filter can only be partially satisfied. The specification requires that this filter be exactly satisfied. This means that the consumer will often need to apply some kind of in-memory filtering operation to fully satisfy the filter.
```
jacques-n, line 29:
```suggestion
If a filter is defined, a consumer must guarantee that all records returned from the scan do not match the filter condition. 
```

jacques-n: I think we either need to support the case of inexact or remove all the wording referring to it. I'm fine with either. Having a bunch of wording about the dynamics of how someone applies a filter feels out of place (in memory, footer stats, etc).

In general, I've come to the conclusion that the second filter field makes more sense than any kind of boolean (whenever we want to move forward on this). I agree with your producer choice complexity concern but also think that it has a lot to do with how physical/logical someone is working. It's trivial rewrite for a system to replace a scan with a required filter with a scan best-effort filter + filter rel as needed. The only reason you typically do that is right before physical execution (which thus means you need specific understanding of the underlying execution engines capabilities).

westonpace: We can leave the entire `Read Filtering` section out entirely and just change the property definition to:

> A boolean Substrait expression that describes a filter that must be applied to the data. The filter should be interpreted against the direct schema.

Do you still think the sparse `Read Filtering` section you suggested would add value or would it be redundant?  I mainly added the explanatory paragraph since I had heard a few comments that it would be helpful if the documentation assumed a bit less SQL knowledge but perhaps there is a better place for such "informational" content.

> I agree with your producer choice complexity concern but also think that it has a lot to do with how physical/logical someone is working

Indeed, I am coming at this from the consumer / pure-physical end of things and so the fact that the filter must be fully satisfied was actually the more surprising default to me.

> It's trivial rewrite for a system to replace a scan with a required filter with a scan best-effort filter + filter rel as needed. The only reason you typically do that is right before physical execution

Yes, this is somewhat of an aside but I'm wondering how we want to tackle this in Acero.  Currently we are sticking to a "Acero does no optimization" philosophy so the most faithful approach would be to reject plans that have `filter` specified and only allow plans that have `best_effort_filter`.  I fear, however, that this will make Acero somewhat unusable by initial producers until optimization / rewrites come along (which I don't think will happen for a bit of time yet).  The rewrite is easy enough for us to do internally and so we'll probably just do that.

curino: Hey Folks, what is the conclusion on this one? I like the idea that we can express a "best_effort_filter", I can see scenarios where we want to push down to computational storage type devices where only certain operators might be implemented (e.g., the "equalities" you list a few comments above).

westonpace: I added best_effort_filter and reworked this paragraph

westonpace:
I have taken the approach @jacques-n suggested and added a best_effort_filter to the ReadRel

jacques-n (approved):
LGTM. Thanks for writing all of this down @westonpace !

PR #277, opened 2022-08-03
docs: add mkdocs plugin to generate markdown from yaml doc for website

richtia:
To see what the html pages look like on your local, you can run `mkdocs serve`

edit: latest screenshot 8/16
Screen Shot 2022-08-16 at 3 44 30 PM


jacques-n:
I'm going to ask the same thing of you that I asked of another ticket. Can we please use mkdocs extension points for rendering as opposed to having a manual task people have to run and checking in generated content/code? I suspect there is a relatively way to expose this same thing with mkdocs. It may just be we put the python script in some directory that mkdocs triggers?

richtia:
> I'm going to ask the same thing of you that I asked of another ticket. Can we please use mkdocs extension points for rendering as opposed to having a manual task people have to run and checking in generated content/code? I suspect there is a relatively way to expose this same thing with mkdocs. It may just be we put the python script in some directory that mkdocs triggers?

Sure.  I'll look into that.

richtia:
@jacques-n, I was able to set up a new workflow that runs the python script if any of the function yamls are updated or the python script it.  I tested this out in my repo/branch, which you can see in this commit: https://github.com/substrait-io/substrait/pull/277/commits/84f5e6b169fba1afcc2d4d897aea155178f7dd2d

jacques-n:
> @jacques-n, I was able to set up a new workflow that runs the python script if any of the function yamls are updated or the python script it. I tested this out in my repo/branch, which you can see in this commit: [84f5e6b](https://github.com/substrait-io/substrait/commit/84f5e6b169fba1afcc2d4d897aea155178f7dd2d)

Thanks. Your change integrates these docs within a github workflow. I'm hoping you can integrate with mkdocs. That way mkdocs serve shows the content correctly (and we can inspect this page like any other).

richtia:
> > @jacques-n, I was able to set up a new workflow that runs the python script if any of the function yamls are updated or the python script it. I tested this out in my repo/branch, which you can see in this commit: [84f5e6b](https://github.com/substrait-io/substrait/commit/84f5e6b169fba1afcc2d4d897aea155178f7dd2d)
> 
> Thanks. Your change integrates these docs within a github workflow. I'm hoping you can integrate with mkdocs. That way mkdocs serve shows the content correctly (and we can inspect this page like any other).

Got it.  I was under the impression that the push action in this workflow would then trigger the site.yml workflow (I'm not certain of this though) since modifications will be made in the 'site/**' path.  

I can also add these 2 actions from site.yml into this new documentation workflow as well.
```
      - name: Generate Static Site
        run: mkdocs build
        working-directory: ./site
      - name: Deploy Static Site to GitHub
        uses: peaceiris/actions-gh-pages@v3
        with:
          external_repository: substrait-io/substrait.io
          publish_branch: main
          deploy_key: ${{ secrets.SUBSTRAIT_SITE_DEPLOY_KEY }}
          publish_dir: ./site/site
          cname: substrait.io
```

Let me know what you think.

jacques-n:
Hey @richtia , I think you're confusing the generation of the prod site for substrait.io consumption purposes (prod: `mkdocs build`) versus the ability to preview the site while doing development (dev: `mkdocs serve`). I agree your changes should make it so that prod is handled. You did this by adding an additional Github workflow. 

What you haven't addressed is the development use case. This is because you didn't integrate your generation to work within mkdocs. Development/previewing of the site is done using `mkdocs serve`. I'd like to have you figure out how to use mkdocs to both cases (`build` and `serve`) as opposed to your current solution which uses Github workflows for prod and manual steps for dev.

richtia:
> Hey @richtia , I think you're confusing the generation of the prod site for substrait.io consumption purposes (prod: `mkdocs build`) versus the ability to preview the site while doing development (dev: `mkdocs serve`). I agree your changes should make it so that prod is handled. You did this by adding an additional Github workflow.
> 
> What you haven't addressed is the development use case. This is because you didn't integrate your generation to work within mkdocs. Development/previewing of the site is done using `mkdocs serve`. I'd like to have you figure out how to use mkdocs to handle both cases (`build` and `serve`) as opposed to your current solution which uses Github workflows for prod and manual steps for dev. In that solution, no github workflow alterations should be necessary.

Ahh...yeah, you're right. I was definitely confusing the 2.  Upon initial investigation, I couldn't find a solution that was using the `mkdocs serve` outside of local development.  Is the intention to allow the reviewers to easily see a preview of the site before approving changes?

One way of handling of prod versus dev that I came across was using a utility that allows multiple versions of mkdocs:
https://squidfunk.github.io/mkdocs-material/setup/setting-up-versioning/#publishing-a-new-version
https://github.com/jimporter/mike 

By default the site can be directed to the prod version, and there can be a drop down that points to any other version.  However, this does still require the changes to be pushed in order to preview.

Please let me know your thoughts on this.

jacques-n:
@rymurr, you had previous success on integrating something into mkdocs. Any pointers you can provide @richtia? I'd really like to have mdkocs serve work across all of our content.

rymurr:
hey @jacques-n and @richtia  check out https://github.com/rymurr/mkdocs-protobuf and https://www.mkdocs.org/dev-guide/plugins/

its surprisingly simple to write a plugin for mkdocs.

richtia:
@jacques-n I realized I may have misunderstood your ask originally.  My latest change uses a plugin (https://oprypin.github.io/mkdocs-gen-files/) that will run the markdown generation python script via mkdocs.yml.

Whenever the mkdocs command (`serve` or `build`) is run, the script will be run now.

Does this sound closer to what you're looking for?

jvanstraten:
Some notes (not exhaustive, haven't reviewed extensively yet):

 - When running `mkdocs serve` the first time, the generated pages don't exist. Looking at the docs, half of it is about needing to use `mkdocs_gen_files.open()` to open the files you generate, which I don't think you're doing (that `MdUtils` thing is doing the file writing), so that's probably why.
 - Changes to the YAML files don't automatically trigger a rebuild, like changes to the markdown files do.
 - The argument order is documented incorrectly. Options behave like regular positional arguments; `add(opt_enum, i8, i8)` (as defined) != `add(i8, i8, opt_enum)` (as documented).
 - The generated docs could still use some TLC to make them look a bit nicer and integrate better with the rest of the website. For example...
    - IMO there's no reason to generate a ToC in each page; none of the other pages do that. mkdocs should generate the ToC instead for the sidebar. I don't know for sure why it's not doing that right now, but it's probably because the first header in each generated markdown file disagrees with the filename. You're also only supposed to have one level-1 header in each file, I think (you have two right underneath each other now, even).
    - The new "Functions" section of the website is pretty out-of-place. I would expect to find it organized under the "Extensions" section, or perhaps be part of a new "Reference" section. Either way it should be ordered before the much more meta "Community" section.
    - The "scalar_functions", "aggregate_functions", and "window_functions" section headers could just be written title case like everything else on the website. The page titles are also not using title case.
    - Whenever an option doesn't have a name associated with it, it's probably better to inline the options in the signature rather than using `opt_enum:name_placeholder` as a reference.
    - The "DESCRIPTION:" header is redundant since there's already a header above it, and nothing else on the website uses all-caps paragraph headers.
    - The list of implementations is probably the first thing you want to see when looking for functions, so it should probably not be in a details/summary block, and should maybe even come before the description (like cppreference).

richtia:
Thanks for the suggestion.  Made a few changes:

* Table of contents should now be appearing on the sidebar
* Implementations no longer in details/summary block
* Fixed options position
* Update top level site.yml to regenerate docs when the extension yamls are changed

Still need to address a few of the other comments:
* Not sure if i can combine mkdocs_gen_files with mdutils.  Once the markdown files are there for the first time, the issue with the generated pages not being there shouldn't be a problem.  Hopefully that's not a big deal?
* One of my other changes got rid of the function signatures, need to update that again.
* Will update the nameplaceholder default option name as well.

richtia:
> 

Thanks for the feedback!

Made a few updates:
* The pages should be viewable even when running `mkdocs serve` for the first time now.  mkdocs_gen_files doesn't actually create the, it's more for just the ability to preview the site by setting up a temp file and using that to display.  I'm using it by writing the generated files into those temp files now.
* Updated the site.yml workflow to trigger the site build when the function yamls are modified as well.
* Moved the options to the proper position
* ToC should be on the left sidebar now
* Options are inlined with the function signature if they don't have a name associated with them now
* Implementations moved to the be first
* Syntax fixes

jvanstraten:
> The pages should be viewable even when running `mkdocs serve` for the first time now.

Cannot reproduce, unfortunately. Running `git clean -fdx; mkdocs serve` still breaks it for me.

> Updated the site.yml workflow to trigger the site build when the function yamls are modified as well.

:white_check_mark:

However, it doesn't trigger live updates from `mkdocs serve`. I'm not sure if that's even possible though.

> Moved the options to the proper position

It will still break if I were to define a function like `f(opt_enum, value, opt_enum)`. You have to build the `options_and_arg_names` and `arg_values` lists while iterating over `args_list` or you'll lose the ordering information.

I also noticed that you're not handling type arguments at all. I don't think any functions in core are using them, but they are specified and thus can be used. Maybe throw an error if you encounter an argument object that your script doesn't recognize, so it doesn't silently document the function incorrectly?

> ToC should be on the left sidebar now

:white_check_mark:

Page titles are still a bit odd. Is it possible to just name them e.g. "functions_aggregate_approx.yaml" instead of "Functions aggregate approx"?

> Options are inlined with the function signature if they don't have a name associated with them now

You might still want to not also document the option in this case though. This looks a bit odd:

> - ['SILENT', 'SATURATE', 'ERROR'] ['SILENT', 'SATURATE', 'ERROR']

> Implementations moved to the be first

:white_check_mark: Looks better now IMO.

You missed this one I think:

> The new "Functions" section of the website is pretty out-of-place. I would expect to find it organized under the "Extensions" section, or perhaps be part of a new "Reference" section. Either way it should be ordered before the much more meta "Community" section.

A landing page for the section with some explanation would also be nice, so it doesn't just immediately send you to whatever the first generated page happens to be.

richtia:
> It will still break if I were to define a function like `f(opt_enum, value, opt_enum)`. You have to build the `options_and_arg_names` and `arg_values` lists while iterating over `args_list` or you'll lose the ordering information.
> 

Made some changes to fix this.

> I also noticed that you're not handling type arguments at all. I don't think any functions in core are using them, but they are specified and thus can be used. Maybe throw an error if you encounter an argument object that your script doesn't recognize, so it doesn't silently document the function incorrectly?
> 

Yeah, I wasn't sure how they are used so I kind of just ignored it yet.  How would they be specified in a function?


> Page titles are still a bit odd. Is it possible to just name them e.g. "functions_aggregate_approx.yaml" instead of "Functions aggregate approx"?
> 

Updated.
> > Options are inlined with the function signature if they don't have a name associated with them now
> 
> You might still want to not also document the option in this case though. This looks a bit odd:
> 
> > * ['SILENT', 'SATURATE', 'ERROR'] ['SILENT', 'SATURATE', 'ERROR']

Removed.

> You missed this one I think:
> 
> > The new "Functions" section of the website is pretty out-of-place. I would expect to find it organized under the "Extensions" section, or perhaps be part of a new "Reference" section. Either way it should be ordered before the much more meta "Community" section.
> 
> A landing page for the section with some explanation would also be nice, so it doesn't just immediately send you to whatever the first generated page happens to be.

Decided to just move it under extensions.

jvanstraten:
> I'll try from a fresh environment and see what's going on.

:heavy_check_mark: Not sure if you changed anything besides moving it to a different category (maybe that was just the broken bit in mkdocs), but it seems to work for me now.

> Made some changes to fix this.

:heavy_check_mark: I haven't thoroughly checked the new logic but LGTM now.

> Yeah, I wasn't sure how they are used so I kind of just ignored it yet. How would they be specified in a function?

When the function is bound, you just pass a type instead of an expression in protobuf. Otherwise, they behave pretty much the same AFAICT. You'd use them sort of like typename template arguments in C++. In the docs I suppose you would just have to annotate somehow when something is a type argument. Come to think of it, you might also want to annotate constant value arguments; they're sort of analogous to value template arguments, though you can't use them in derivation expressions. I think that's just specified with a `constant: true` field on the argument, but check the schema to be sure.

> Updated.

:heavy_check_mark:

> Removed.

:heavy_check_mark:

> Decided to just move it under extensions.

Could still use some handwritten text to link to these pages and specify what these extensions are, but I should probably submit something for #251 one of these days, as well as something to explain the difference between core and user extensions (and why this difference is fundamentally needed), so I don't have to keep repeating myself. I'll probably end up rewriting a lot of the existing text then anyway. So as far as I'm concerned this is :heavy_check_mark: for now.

richtia:
> When the function is bound, you just pass a type instead of an expression in protobuf. Otherwise, they behave pretty much the same AFAICT. You'd use them sort of like typename template arguments in C++. In the docs I suppose you would just have to annotate somehow when something is a type argument. Come to think of it, you might also want to annotate constant value arguments; they're sort of analogous to value template arguments, though you can't use them in derivation expressions. I think that's just specified with a `constant: true` field on the argument, but check the schema to be sure.

Do you have an example of how the type argument would show up in the yaml so I could have a point of reference for what to look for?  Do they just go after value in the args? Or I guess an option for what to return as well?

jvanstraten:
For example, you could describe a cast like this (if there wouldn't have been a dedicated expression for it):

```yaml
args:
- type: to
- value: from
return: to
```

> Do they just go after value in the args? Or I guess an option for what to return as well?

I think I lost you here.

richtia:
> ```yaml
> - type: to
> ```

Ahh...ignore that second part, I was misunderstanding something...but your example cleared it up.  Added some logic to display a message about unrecognized arguments for the implementation if we come across one.

jvanstraten:
That's sort of what I meant, but IMO it's better if it fails hard by just throwing an exception so you'll get a CI failure if this happens, rather than silently inserting a warning in the generated docs. I don't think anyone's intending to regularly read through all the generated pages to check that these messages aren't there.

richtia:
> That's sort of what I meant, but IMO it's better if it fails hard by just throwing an exception so you'll get a CI failure if this happens, rather than silently inserting a warning in the generated docs. I don't think anyone's intending to regularly read through all the generated pages to check that these messages aren't there.

Ahh got it.  Made an update so the script will throw an exception instead now.

jacques-n:
Thanks for working through this @richtia , this will be a great improvement to the site.

jvanstraten (approved):
Functionally this LGTM now, so I say merge :) I'm sure I could nitpick about implementation details and aesthetics until the cows come home, but IMO it's better to just merge now and fix/add/improve things later, since it's obviously already a big documentation improvement.

richtia:
@jacques-n let me know if there are any other changes you'd like to see before this initial function documentation update.  Feel free to add anyone else to review as well.

PR #278, opened 2022-08-04
feat: add median and count_distinct aggregation functions

vibhatha:
This PR includes yaml configs for `median`, `approx_median` and `count_distinct` for aggregate functions.

### DIFF
--- a/extensions/functions_arithmetic.yaml
+++ b/extensions/functions_arithmetic.yaml
@@ -1134,9 +1134,24 @@ aggregate_functions:
         nullability: DECLARED_OUTPUT
         return: fp64?
   - name: "median"
-    description: Calculate the median for a set of values.
+    description: > 
+      Calculate the median for a set of values.
+
+      Returns null if applied to zero records. For the integer implementations,
+      the rounding option determines how the median should be rounded if it ends
+      up midway between two values. For the floating point implementations,
+      they specify the usual floating point rounding mode.
     impls:
       - args:
+          - name: precision
+            description: >
+              Based on required operator performance and configured optimizations
+              on saving memory bandwidth, the precision of the end result can be
+              the highest possible accuracy of an approximation.
+
+                - EXACT: provides the highest accurate output
+                - APPROXIMATE: provides a sub-optimal output
+            options: [ EXACT, APPROXIMATE ]
           - name: rounding
             options: [ TIE_TO_EVEN, TIE_AWAY_FROM_ZERO, TRUNCATE, CEILING, FLOOR ]
             required: false
@@ -1144,6 +1159,16 @@ aggregate_functions:
         nullability: DECLARED_OUTPUT
         return: i8?
       - args:
+          - name: precision
+            description: >
+              Based on required operator performance and configured optimizations
+              on saving memory bandwidth, the precision of the end result can be
+              the highest possible accuracy of an approximation.
+
+                - EXACT: provides the highest accurate output
+                - APPROXIMATE: provides a sub-optimal output
+            options: [ EXACT, APPROXIMATE ]
+            required: true
           - name: rounding
             options: [ TIE_TO_EVEN, TIE_AWAY_FROM_ZERO, TRUNCATE, CEILING, FLOOR ]
             required: false
@@ -1151,6 +1176,16 @@ aggregate_functions:
         nullability: DECLARED_OUTPUT
         return: i16?
       - args:
+          - name: precision
+            description: >
+              Based on required operator performance and configured optimizations
+              on saving memory bandwidth, the precision of the end result can be
+              the highest possible accuracy of an approximation.
+
+                - EXACT: provides the highest accurate output
+                - APPROXIMATE: provides a sub-optimal output
+            options: [ EXACT, APPROXIMATE ]
+            required: true
           - name: rounding
             options: [ TIE_TO_EVEN, TIE_AWAY_FROM_ZERO, TRUNCATE, CEILING, FLOOR ]
             required: false
@@ -1158,6 +1193,16 @@ aggregate_functions:
         nullability: DECLARED_OUTPUT
         return: i32?
       - args:
+          - name: precision
+            description: >
+              Based on required operator performance and configured optimizations
+              on saving memory bandwidth, the precision of the end result can be
+              the highest possible accuracy of an approximation.
+
+                - EXACT: provides the highest accurate output
+                - APPROXIMATE: provides a sub-optimal output
+            options: [ EXACT, APPROXIMATE ]
+            required: true
           - name: rounding
             options: [ TIE_TO_EVEN, TIE_AWAY_FROM_ZERO, TRUNCATE, CEILING, FLOOR ]
             required: false
@@ -1165,6 +1210,16 @@ aggregate_functions:
         nullability: DECLARED_OUTPUT
         return: i64?
       - args:
+          - name: precision
+            description: >
+              Based on required operator performance and configured optimizations
+              on saving memory bandwidth, the precision of the end result can be
+              the highest possible accuracy of an approximation.
+
+                - EXACT: provides the highest accurate output
+                - APPROXIMATE: provides a sub-optimal output
+            options: [ EXACT, APPROXIMATE ]
+            required: true
           - name: rounding
             options: [ TIE_TO_EVEN, TIE_AWAY_FROM_ZERO, TRUNCATE, CEILING, FLOOR ]
             required: false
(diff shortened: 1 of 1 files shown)


extensions/functions_arithmetic.yaml
```diff
@@ -742,6 +742,107 @@ aggregate_functions:
...
+  - name: "median"
+    description: Calculate the median for a set of values.
+    impls:
+      - args:
+          - name: rounding
+            options: [ TIE_TO_EVEN, TIE_AWAY_FROM_ZERO, TRUNCATE, CEILING, FLOOR ]
+            required: false
+          - value: i8
+        nullability: DECLARED_OUTPUT
+        return: i8?
```
cpcloud, line 754: I think this should be `fp64?`. There's no way to guarantee that the result is going to be an `i8`. Similarly for the remaining overloads as well as `approx_median`.

extensions/functions_arithmetic.yaml
```diff
@@ -742,6 +742,107 @@ aggregate_functions:
...
+        return: fp64?
+  - name: "count_distinct"
+    description: Count of unique values in a set of values
+    impls:
+      - args:
+          - options: [SILENT, SATURATE, ERROR]
+            required: false
+          - value: any
+        nullability: DECLARED_OUTPUT
+        decomposable: MANY
```
cpcloud, line 843: Can you give a short description of how count distinct can be decomposed using the intermediate type you wrote below (`i64`)?

extensions/functions_arithmetic.yaml
```diff
@@ -742,6 +742,107 @@ aggregate_functions:
...
+        return: fp32?
+      - args:
+          - name: rounding
+            options: [ TIE_TO_EVEN, TIE_AWAY_FROM_ZERO, TRUNCATE, CEILING, FLOOR ]
+            required: false
+          - value: fp64
+        nullability: DECLARED_OUTPUT
+        return: fp64?
+  - name: "count_distinct"
+    description: Count of unique values in a set of values
```
cpcloud, line 836:
```suggestion
    description: Count of unique values in a set of values.
```

extensions/functions_arithmetic.yaml
jvanstraten: There's no reason for the i8/i16/i32 variants to exist at all IMO. f(integer) -> integer variants for these statistical functions is not really wrong (you'd just have to round a lot more aggressively than you would for floats...) but make little sense in practice. i8/i16/i32 can be losslessly converted to fp32 and/or fp64 prior to applying this function whenever you actually meant to use floats. f(i8/i16/i32) -> fp64 is problematic because you have to pick one of the number types more or less at random (what if we ever add an fp128 type, do we now change the return type? see also #251). An argument could be made for f(i64) -> fp64 because it cannot be losslessly converted to an fp64 (and we did this for a similar function at some point, though I can't remember which off the top of my head), though it has that same problem in the end, and I doubt anyone's going to implement that in a way that doesn't just convert the i64 inputs to fp64 directly anyway.

jvanstraten: Scratch that. As @vibhatha pointed out offline, median is a bit special among the statistical functions because it returns one of the inputs exactly.

extensions/functions_arithmetic.yaml
vibhatha: You're correct @cpcloud. What I have written is wrong. I mixed up a stateful parallel computation with this. I will fix this one.

extensions/functions_arithmetic.yaml
vibhatha: A little doubt on the median, what if the data distribution is integer and why should we return floats in that case?

vibhatha:
@jvanstraten would it be better introduce an option like `precision` or and introduce `EXACT` and `APPROXIMATE` instead of writing two separate functions. I did this today for another PR. Not sure if it is a good way to handle it. WDYT?

extensions/functions_arithmetic.yaml
```diff
@@ -742,6 +742,105 @@ aggregate_functions:
           - value: fp64
         nullability: DECLARED_OUTPUT
         return: fp64?
+  - name: "median"
+    description: Calculate the median for a set of values.
```
jvanstraten, line 746: IMO this should include something like "Returns null if applied to zero records. For the integer implementations, the rounding option determines how the median should be rounded if it ends up midway between two values. For the floating point implementations, they specify the usual floating point rounding mode."

jvanstraten (changes requested):
Sorry, I forgot to update this.

> would it be better introduce an option like precision or and introduce EXACT and APPROXIMATE instead of writing two separate functions

I don't really have an opinion, but unless there's a statistical definition I'm not aware of, I do believe we need to specify what "approximate" means. Is 100.0 an acceptable approximate median of [101.2, 101.3, 101.4]? Probably not, so where's the limit?

extensions/functions_arithmetic.yaml
vibhatha: You mean the `description`?

jvanstraten: Yes.

jvanstraten:
@vibhatha Please extend the descriptions. Here's a good example of the level of descriptiveness I'm looking for: https://github.com/substrait-io/substrait/blob/add698f3ddca2475286355f532212f82fd2871b2/extensions/functions_string.yaml#L571-L608 Better to be verbose and explicit than to assume the reader already knows what the function does, especially in odd corner cases, and in this case for what "approximate" actually means in terms of accuracy.

Also, IIRC we went for an option for approximate vs precise for other statistical functions? Not sure if they're merged yet but I remember that being the conclusion.

vibhatha:
@jvanstraten we had a function for `population` and `sample` and I did write a option recently for `precision`. I am going to add it here as well. Let me push what I wrote, then we can probably re-word it and standardize to re-use. 

I am also wondering can we reference such options under a general options attribute. Rather than re-writing the description and definitions over and over again. It could save spacing and readability. I am not sure if this is possible, but just a suggestion. Not asking about array of items, just 

```YAML
operator_options:
     - population
     - precision
     - ....

```

And the `operator_options` can be read from somewhere else. How about that?

extensions/functions_arithmetic.yaml
```diff
@@ -1071,6 +1071,125 @@ aggregate_functions:
...
+    impls:
+      - args:
+          - name: precision
+            description: >
+              Based on required operator performance and configured optimizations
+              on saving memory bandwidth, the precision of the end result can be
+              the highest possible accuracy of an approximation.
+
+                - EXACT: provides the highest accurate output
+                - APPROXIMATE: provides a sub-optimal output
```
jvanstraten, line 1091: This still doesn't specify *how* approximate the result may be. Is this what you mean, or is this too broad?

```suggestion
                - APPROXIMATE: provides only an estimate; the result must lie
                  between the minimum and maximum values in the input
                  (inclusive), but otherwise the accuracy is left up to the
                  consumer
```

jvanstraten (changes requested):
> I am also wondering can we reference such options under a general options attribute. Rather than re-writing the description and definitions over and over again. It could save spacing and readability.

We've run into similar things with the YAMLs before, and the conclusion has always been that it's easier on consumers of the YAML files to just repeat everything. If at some point it becomes really annoying we could always just generate the YAML files. However, the descriptions of the standard options we use all over the core extensions (like overflow, rounding, and I guess precision and population) could IMO just be defined on the website instead of in the YAMLs. A computer that's parsing these YAML files won't care about the description anyway. There's no page for this yet, but I'm planning to rewrite/add to the extension pages of the website soon, and there will absolutely be a section for this. So I'm fine with just omitting the descriptions of `precision` for now.

extensions/functions_arithmetic.yaml
```diff
@@ -1071,6 +1071,125 @@ aggregate_functions:
...
+      the rounding option determines how the median should be rounded if it ends
+      up midway between two values. For the floating point implementations,
+      they specify the usual floating point rounding mode.
+    impls:
+      - args:
+          - name: precision
+            description: >
+              Based on required operator performance and configured optimizations
+              on saving memory bandwidth, the precision of the end result can be
+              the highest possible accuracy of an approximation.
```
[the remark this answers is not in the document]
vibhatha: Yes, thanks for cathing this one.

extensions/functions_arithmetic.yaml
vibhatha: I wanted to mean this one, but may be we could elaborate this a bit? 

Although, this is broad since the optimization strategy is very hard to specify. Again I think the choice of approximation is up to the engine to decide and that depends on various optimization techniques. But that's not the job of Substrait to define that optimization strategy. Substrait can only specify that it is going to be an approximation. So I wanted to put forward that idea. 

Should we enhance more?

jvanstraten: I don't know that we can do better than what I suggested to constrain the approximation method further, but you're welcome to try. I at least can't think of any approximate median algorithm that doesn't at least satisfy that constraint, and only trivial constraints is better than no constraints, IMO.

vibhatha: Let's keep the suggested once for now. I will update this.

jvanstraten (approved):

jacques-n:
Sorry, just saw this come through. I think we should revert the addition of count distinct here. Distinct isn't generally a property of the function, it is a property of the aggregate. That's why we have [distinct as a property of function invocation](https://github.com/substrait-io/substrait/blob/main/proto/substrait/algebra.proto#L1166).

You achieve count distinct by combining count + that property.

jvanstraten:
Oops, you're right, I didn't think of that. #311

PR #279, opened 2022-08-04
feat: add quantile aggregate function

vibhatha:
This PR includes the `quantile` and `approx_quantile` functions for aggregate functions.

### DIFF
--- a/extensions/functions_arithmetic.yaml
+++ b/extensions/functions_arithmetic.yaml
@@ -1269,104 +1269,74 @@ aggregate_functions:
         nullability: DECLARED_OUTPUT
         return: fp64?
   - name: "quantile"
-    description: Calculates quantile for a set of values.
+    description: >
+      Calculates quantiles for a set of values.
+      
+      This function will divide the aggregated values (passed via the
+      distribution argument) over N equally-sized bins, where N is passed
+      via a constant argument. It will then return the values at the
+      boundaries of these bins in list form. If the input is appropriately
+      sorted, this computes the quantiles of the distribution.
+      
+      The function can optionally return the first and/or last element of
+      the input, as specified by the `boundaries` argument. If the input is
+      appropriately sorted, this will thus be the minimum and/or maximum
+      values of the distribution.
+      
+      When the boundaries do not lie exactly on elements of the incoming
+      distribution, the function will interpolate between the two nearby
+      elements. If the interpolated value cannot be represented exactly,
+      the `rounding` option controls how the value should be selected or
+      computed.
+      
+      The function fails and returns null in the following cases:
+        - `n` is null or less than one;
+        - any value in `distribution` is null.
+      
+      The function returns an empty list if `n` equals 1 and `boundaries` is
+      set to `NEITHER`.
+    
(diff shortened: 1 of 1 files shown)


jvanstraten:
From just the descriptions, implementations, and a few minutes of googling I'm not at all sure what these functions and their arguments actually do. Can you please clarify in the descriptions and include argument names?

vibhatha:
@jvanstraten I updated the PR.

jvanstraten (changes requested):
We discussed this function at length offline. Could you update accordingly? Right now, the description and definition just isn't good enough IMO.

vibhatha:
@jvanstraten I updated the PR and also for the approximate type, I created a new field called `precision` and added a description. Please check that description see if it requires modifications?

jvanstraten (approved):
~~LGTM now!~~

jvanstraten (changes requested):
Actually, correction (sorry), the same comments I gave on the other PR for precision apply here. I guess for approximate the list items should at least be monotonously increasing/decreasing if the input is sorted ascending/descending? Again, I'm also fine with just omitting its description for now since it will be a standard option.

vibhatha:
> Actually, correction (sorry), the same comments I gave on the other PR for precision apply here. 

I updated the description of the `precision`. 

> I guess for approximate the list items should at least be monotonously increasing/decreasing if the input is sorted ascending/descending? Again, I'm also fine with just omitting its description for now since it will be a standard option.

Shall we include options about the ordering too? If so, it shouldn't be an option of `precision` but the operator itself, right?

extensions/functions_arithmetic.yaml
```diff
@@ -1071,6 +1071,86 @@ aggregate_functions:
...
+            options: [ NEITHER, MINIMUM, MAXIMUM, BOTH ]
+            required: true
+          - name: precision
+            description: >
+              Based on required operator performance and configured optimizations
+              on saving memory bandwidth, the precision of the end result can be
+              the highest possible accuracy of an approximation. 
+
+                - EXACT: provides the highest accurate output
+                - APPROXIMATE: provides a sub-optimal output
```
jvanstraten, line 1118: > I updated the description of the `precision`.

I mean here. Or maybe you forgot to push?

jvanstraten (changes requested):
> Shall we include options about the ordering too? If so, it shouldn't be an option of precision but the operator itself, right?

No, because aggregate function bindings already include specialized options for sorting prior to aggregation in protobuf land.

Also needs a rebase.

jvanstraten:
We also need to add `ordered: true` to all these implementations. I didn't realize that wasn't the default behavior. See also #312

extensions/functions_arithmetic.yaml
vibhatha: Is it okay now?

vibhatha:
> We also need to add `ordered: true` to all these implementations. I didn't realize that wasn't the default behavior. See also #312

@jvanstraten updated.

jvanstraten (approved):
I think this looks good now.

PR #280, opened 2022-08-08
feat(set): add basic set membership operations

gforsyth:
Adds `is_in` and `index_in` functions to determine whether a value is a
member of a given set of values, and the index of that value if it
exists, respectively.

I put these in a new yaml file to start with, but I think we could
_probably_ stick them in `comparisons`?  Open to thoughts on where these
belong.  If we are planning to add other `set` operations (union,
difference, intersection, etc) then we may be better served by having a
separate file.

### DIFF
--- a/extensions/functions_set.yaml
+++ b/extensions/functions_set.yaml
@@ -1,0 +1,24 @@
+%YAML 1.2
+---
+scalar_functions:
+  -
+    name: "index_in"
+    description: >
+      Checks the membership of a value in a list of values
+
+      Returns the first 0-based index value of some input `T` if `T` is equal to
+      any element in `List<T>`.  Returns `NULL` if not found.
+
+      If `T` is `NULL`, returns `NULL`.
+
+      If `T` is `NaN`:
+        - Returns 0-based index of `NaN` in `List<T>` (default)
+        - Returns `NULL` (if `NAN_IS_NOT_NAN` is specified)
+    impls:
+      - args:
+          - options: [ NAN_IS_NAN, NAN_IS_NOT_NAN ]
+            required: false
+          - value: T
+          - value: List<T>
+        nullability: DECLARED_OUTPUT
+        return: int64?


extensions/functions_set.yaml
```diff
@@ -0,0 +1,89 @@
...
+  -
+    name: "is_in"
+    description: >
+      Checks the membership of x in values
+
+      Returns True if found, False if not
+    impls:
+      - args:
+          - value: i8
+            name: "x"
```
gforsyth, line 13: this is not a great name, but `value` and `values` seem too easily confused

extensions/functions_set.yaml
```diff
@@ -0,0 +1,89 @@
...
+      Checks the membership of x in values
+
+      Returns the 0-based index value of `x` if it is present, returns NULL if not found.
+    impls:
+      - args:
+          - value: i8
+            name: "x"
+          - value: i8
+            name: "values"
+        return: int64?
```
gforsyth, line 59: I have all of these returning a nullable int64 which I think makes sense

jacques-n:
I haven't looked at the details here but saw the commit message. `is_in` doesn't need a function, the orlist expression type already provides this functionality, I belive.

extensions/functions_set.yaml
```diff
@@ -0,0 +1,89 @@
...
+            name: "values"
+        return: bool
+      - args:
+          - value: fp64
+            name: "x"
+          - value: fp64
+            name: "values"
+        return: bool
+  -
+    name: "index_in"
```
jacques-n, line 48: I'd expect these to either be variadic or have an argument that is a list of items. I'd also tend to think these should be defined generically e.g. 

```
- args: 
  - value: T
  - value: List<T>
```

gforsyth:
> haven't looked at the details here but saw the commit message. `is_in` doesn't need a function, the orlist expression type already provides this functionality, I belive.

Thanks, I thought that might be the case but erred on the side of inclusion.  Removed and updated.

extensions/functions_set.yaml
```diff
@@ -0,0 +1,15 @@
+%YAML 1.2
+---
+scalar_functions:
+  -
+    name: "index_in"
+    description: >
+      Checks the membership of a value in a set of values
+
+      Returns the 0-based index value of some input `T` if it is present,
```
cpcloud, line 9: Thanks for clarifying the index base :)

I think there are a couple things to resolve in the description:

1. What does `is present` mean? I'll suggest that it means the needle element is equal to (according substrait's `equals` function) an element in the haystack list.
2. What is the behavior when the needle is `NULL` (regardless of whether there's a `NULL` in the `list`). I believe as a consequence of point 1 it'll be `NULL` since `NULL = NULL` returns `NULL`.

cpcloud (changes requested):
Looks mostly good, just needs a bit of clarification in the description around what it means for an element to be present.

extensions/functions_set.yaml
cpcloud: This looks to be resolved.

extensions/functions_set.yaml
cpcloud: There's also `NaN` too ... yay?

gforsyth: `NaN` -> ... `NULL` ?

cpcloud: Gosh, this is really a fraught issue. See https://github.com/cockroachdb/cockroach/issues/18860 for example.

It seems like the decision to make `nan = nan` be `true` is from postgres's decision:

> in order to allow numeric values to be sorted and used in tree-based indexes, PostgreSQL treats NaN values as equal, and greater than all non-NaN values.

Many other incumbent databases also behave this way.

Pros of IEEE semantics:

1. We don't violate expectations of people who aren't steeped in SQL practicalities.

Pros of not-IEEE semantics:

1. NaNs (can) have a well-defined ordering
1. NaNs can be used in indexes

Thoughts?

cpcloud (dismissed):

jvanstraten (changes requested):
Nullability should be set to `DECLARED_OUTPUT`, or the `?` flag is ignored (at least by my interpretation) since the default is `MIRROR`.

Presumably this will return the index of the *first* element that matches. After all, `list` allows for duplicates. Or is it allowed to return any matching index (better for parallelization)? The description should be explicit about this.

LGTM otherwise.

gforsyth:
For now I've opted to declare that it returns the first matching value's index.  We can extend the functionality later (returning all of the matching indices?).
And I've added the nullability declaration

jvanstraten (approved):

cpcloud (approved):

PR #281, opened 2022-08-08
feat: add regexp_replace function

richtia:
PR to regexp_replace function.

Addressed some of the comments made by @jvanstraten here: https://github.com/substrait-io/substrait/pull/256

Regular expression `modifiers` are made options here.  If there's a better way, i'm open to suggestions.

Once we can come to an agreement on how to handle this regex function, i'll submit another PR to add a few of the others.

### DIFF
--- a/extensions/functions_string.yaml
+++ b/extensions/functions_string.yaml
@@ -821,7 +821,7 @@ scalar_functions:
     impls:
       - args:
           - name: case_sensitivity
-            options: [ CASE_SENSITIVE, CASE_INSENSITIVE]
+            options: [ CASE_SENSITIVE, CASE_INSENSITIVE, CASE_INSENSITIVE_ASCII]
             required: false
           - name: multiline
             options: [ MULTILINE_DISABLED, MULTILINE_ENABLED]
@@ -847,7 +847,7 @@ scalar_functions:
         return: "string"
       - args:
           - name: case_sensitivity
-            options: [ CASE_SENSITIVE, CASE_INSENSITIVE]
+            options: [ CASE_SENSITIVE, CASE_INSENSITIVE, CASE_INSENSITIVE_ASCII]
             required: false
           - name: multiline
             options: [ MULTILINE_DISABLED, MULTILINE_ENABLED]
--- a/extensions/functions_string.yaml
+++ b/extensions/functions_string.yaml
@@ -795,6 +795,83 @@ scalar_functions:
             name: "input"
         return: i64
   -
+    name: regexp_replace
+    description: >-
+      Search a string for a substring that matches a given regular expression pattern and replace
+      it with a replacement string. The regular expression pattern should follow the
+      International Components for Unicode implementation (https://unicode-org.github
+      .io/icu/userguide/strings/regexp.html). The occurrence of the pattern to be replaced is
+      specified using the `occurrence` argument. Specifying `1` means only the first occurrence
+      will be replaced, `2` means the second occurrence, and so on. Specifying `0` means all
+      occurrences will be replaced. The number of characters from the beginning of the string to
+      begin starting to search for pattern matches can be specified using the `position` argument.
+      Specifying `1` means to search for matches starting at the first character of the input
+      string, `2` means the second character, and so on. The `position` argument should be a
+      positive non-zero integer. The replacement string can capture groups using numbered
+      backreferences.
+
+      The `case_sensitivity` option specifies case-sensitive or case-insensitive matching.
+      Enabling the `multiline` option will treat the input string as multiple lines.  This makes
+      the `^` and `$` characters match at the beginning and end of any line, instead of just the
+      beginning and end of the input string. Enabling the `dotall` option makes the `.` character
+      match line terminator characters in a string.
+
+      Behavior is undefined if the regex fails to compile, the replacement contains an illegal
+      back-reference, the occurrence value is out of range, or the position value is out of range.
+    impls:
+      - args:
+          - name: case_sensitivity
+            options: [ CASE_SENSITIVE, CASE_INSENSITIVE]
+            required: false
+          - name: multiline
+            options: [ MULTILINE_DISABLED, MULTILINE_ENABLED]
+            required: false
+          - name: dotall
+            options: [ DOTALL_DISABLED, DOTALL_ENABLED]
+            required: false
+          - value: "string"
+            name: "input"
+            description: The input string.
+          - value: "string"
+            name: "pattern"
+            description: The regular expression to search for within the input string.
+          - value: "string"
+            name: "replacement"
+            description: The replacement string.
+          - value: i64
+            name: "position"
+            description: The position to start the search.
+          - value: i64
+            name: "occurrence"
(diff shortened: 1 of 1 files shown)


extensions/functions_string.yaml
```diff
@@ -251,6 +251,104 @@ scalar_functions:
...
+          - name: multiline
+            options: [ MULTILINE_DISABLED, MULTILINE_ENABLED]
+            required: false
+          - name: dotall
+            options: [ DOTALL_DISABLED, DOTALL_ENABLED]
+            required: false
+          - value: "varchar<L1>"
+            name: "input"
+            description: The input string.
+          - value: "varchar<L1>"
```
jvanstraten, line 313: No need for the pattern to have the same max length as the input.
```suggestion
          - value: "varchar<L2>"
```

extensions/functions_string.yaml
```diff
@@ -251,6 +251,104 @@ scalar_functions:
...
+            name: "pattern"
+            description: The regular expression to search for within the input string.
+          - value: "varchar<L1>"
```
jvanstraten, line 316:
```suggestion
          - value: "varchar<L3>"
```

extensions/functions_string.yaml
```diff
@@ -251,6 +251,104 @@ scalar_functions:
...
+      Search a string for a substring that matches a given regular expression pattern and replace
+      it with a replacement string. The regular expression pattern should follow the
+      International Components for Unicode implementation (https://unicode-org.github
+      .io/icu/userguide/strings/regexp.html). The occurrence of the pattern to be replaced is
+      specified using the `occurrence` argument. Specifying `1` means only the first occurrence
+      will be replaced and specifying `0` means all occurrences will be replaced. The number of
+      characters from the beginning of the string to begin starting to search for pattern matches
+      can be specified using the `position` argument.  Specifying `1` means to search for matches
+      starting at the first character of the input string. The replacement string can capture
+      groups using numbered backreferences.
```
jvanstraten, line 266: It'd be nice to be more explicit about how > 1 works, e.g. "Specifying `1` means only the first occurrence will be replaced, `2` means the second occurrence, and so on. Specify `0` to replace all occurrences." and likewise for the string indices.

What happens if I were to pass negative numbers or 0 for the string index?

extensions/functions_string.yaml
```diff
@@ -251,6 +251,104 @@ scalar_functions:
             name: "replacement"
             description: The replacement string.
         return: "varchar<L1>"
+  -
+    name: regexp_replace
+    description: >-
+      Search a string for a substring that matches a given regular expression pattern and replace
+      it with a replacement string. The regular expression pattern should follow the
+      International Components for Unicode implementation (https://unicode-org.github
+      .io/icu/userguide/strings/regexp.html). The occurrence of the pattern to be replaced is
```
jvanstraten, line 260: Awesome, I didn't know that one. Nice and concise, and an actual standard rather than the usual "it behaves like the matcher in this random tool or language".

extensions/functions_string.yaml
```diff
@@ -251,6 +251,104 @@ scalar_functions:
...
+          - value: "fixedchar<l1>"
+            name: "replacement"
+            description: The replacement string.
+          - value: i64
+            name: "position"
+            description: The position to start the search.
+          - value: i64
+            name: "occurrence"
+            description: Which occurrence of the match to replace.
+        return: "fixedchar<l1>"
```
jvanstraten, line 351: I don't think a regex replace makes a whole lot of sense for `fixedchar`s. In any case, `pattern` and `replacement` would need different sizes. The return type would remain problematic.

jvanstraten (changes requested):
Nice, this *almost* looks good to me.

extensions/functions_string.yaml
richtia: Removed `fixedchar`

extensions/functions_string.yaml
richtia: Updated.  For string index, it states that it should be a positive non-zero integer now.  How does that sound?

jvanstraten: I already understood that I'm not supposed to do it, but it still doesn't really specify what would happen if I would :) The same actually applies for what happens if the regex fails to compile, or the replacement has a back-reference that doesn't exist. Maybe this is overly pedantic and it should be up to the engine to decide whether to invoke `MAYBE_HALT_AND_CATCH_FIRE`, throw a segfault, or do something more gentle, but otherwise I see three options:

 - throw an error and fail the entire plan if any invocation fails;
 - change nullability behavior to `SPECIFIED_OUTPUT` and always return nullable, so you can fail invocations independently by returning null;
 - allow either of the above by means of an option enum.

richtia: Yeah, I was originally under the assumption that the engine would decide, which I guess would open up a lot of variability between engine implementations. Do you have a preference for any of those three options?

richtia: Seems like not that many actually have this as an argument that can be passed.  MySQL does though and if you pass a negative input, it throws and error saying incorrect parameters.

jvanstraten: > How do you suggest we define the option that allows 1 or 2? [ DECLARED_OUTPUT, ERROR]?

I was thinking something like an option named "on_error" with values `ERROR` and `NULL`.

> And would we need to explicitly set the nullability for the implementations?

Yes, and that's actually the primary reason I'm kind of on the fence about it. There's no way to specify that the function should have MIRROR behavior for `ERROR` and `DECLARED_OUTPUT` behavior for `NULL`, so you'd have to do `DECLARED_OUTPUT` for both. That's kind of nasty and will probably not work well with the functions in existing query engines.

I may well be overthinking it. Maybe we should just add something like "Behavior is undefined if the regex fails to compile, the replacement contains an illegal back-reference, the occurrence value is out of range, or the position value is out of range." to leave it up to the engine. Or just keep things as is and not specify anything, since that statement is arguably redundant (though I'd prefer to be overly explicit to show that we did actually think about these edge cases).

I guess I'll leave this up to you and @jacques-n (or anyone else stumbling upon this who might have better ideas).

richtia: I went with adding your error handling message you wrote there to the description.  I'm also on the fence about potentially having this nullability situation here.

jvanstraten (approved):
I'll mark this as approved because I'm good with this, but whoever's merging this might want to have a look at the unresolved thread and weigh in before bashing "merge".

jvanstraten (changes requested):
@richtia Needs rebase due to other merges. Also, can you add a `CASE_INSENSITIVE_ASCII` option to the end of `case_sensitivity` as we discussed offline at some point?

richtia:
> @richtia Needs rebase due to other merges. Also, can you add a `CASE_INSENSITIVE_ASCII` option to the end of `case_sensitivity` as we discussed offline at some point?

Updated!  This one just updates the options for regexp_replace.  https://github.com/substrait-io/substrait/pull/293 handles the rest of them.

jvanstraten (approved):

PR #282, opened 2022-08-08
feat: add center function

richtia:
PR to add center function.

Originally worked on in https://github.com/substrait-io/substrait/pull/248.

Instead of allowing a string with multiple characters to be used as padding, like lpad and rpad, center will only allow a single character for padding.

### DIFF
--- a/extensions/functions_string.yaml
+++ b/extensions/functions_string.yaml
@@ -382,10 +382,12 @@ scalar_functions:
   -
     name: center
     description: >-
-      Center the input string by padding the sides with the given `character` until the specified
+      Center the input string by padding the sides with a single `character` until the specified
       `length` of the string has been reached. By default, if the `length` will be reached with
       an uneven number of padding, the extra padding will be applied to the right side.
       The side with extra padding can be controlled with the `padding` option.
+
+      Behavior is undefined if the number of characters passed to the `character` argument is not 1.
     impls:
       - args:
           - name: padding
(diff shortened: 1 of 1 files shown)


jvanstraten (changes requested):
The description is not very explicit about it needing to be a single character; `character` vs `characters` could easily be misinterpreted as a typo. This is especially problematic because the prototypes allow anything from 0 to 2^31-1 characters to be specified for `character`. The `varchar<L2>` could at least be made a `varchar<1>` to make it 0 or 1 character in that case, though that doesn't really solve the problem completely and doesn't solve anything for the `string` variant... The closest thing to a character type that Substrait has is `fixedchar<1>`, but using that bends the rules of #251.

I'm on the fence about what the right approach is, but if we stick to the implementations as currently defined, the description should state what happens if != 1 characters are passed to the `character` argument (throw error that fails the whole query? return null? use first character for >1 and use space for 0 characters? something else?)

richtia:
> The description is not very explicit about it needing to be a single character; `character` vs `characters` could easily be misinterpreted as a typo. This is especially problematic because the prototypes allow anything from 0 to 2^31-1 characters to be specified for `character`. The `varchar<L2>` could at least be made a `varchar<1>` to make it 0 or 1 character in that case, though that doesn't really solve the problem completely and doesn't solve anything for the `string` variant... The closest thing to a character type that Substrait has is `fixedchar<1>`, but using that bends the rules of #251.
> 
> I'm on the fence about what the right approach is, but if we stick to the implementations as currently defined, the description should state what happens if != 1 characters are passed to the `character` argument (throw error that fails the whole query? return null? use first character for >1 and use space for 0 characters? something else? explicitly undefined/up to the consumer to decide?)

Updated the description and arg implementation to specify 1 character.  For now, i've also just explicitly stated the undefined behavior if the number of characters passed to the `character` argument is not 1.

jvanstraten (approved):
LGTM. CI failure is due to a network failure.

PR #284, opened 2022-08-09
Implementation for ReferenceRel

curino:
This is a simple implementation for the ReferenceRel, based on conversation with @jacques-n. The key idea is that a simple positional pointer to the `repeated Rel relations` in `Plan`. This is minimal and sufficient to support DAGs and Spool, and combined with #128 this supports multi-outout without redundant execution.

### DIFF
--- a/proto/substrait/algebra.proto
+++ b/proto/substrait/algebra.proto
@@ -368,7 +368,6 @@ message Rel {
 message NamedObjectWrite {
   // The list of string is used to represent namespacing (e.g., mydb.mytable).
   // This assumes shared catalog between systems exchanging a message.
-  // 
   repeated string names = 1;
   substrait.extensions.AdvancedExtension advanced_extension = 10;
 }
--- a/proto/substrait/algebra.proto
+++ b/proto/substrait/algebra.proto
@@ -1164,4 +1164,10 @@ message AggregateFunction {
     // Use only distinct values in the aggregation calculation.
     AGGREGATION_INVOCATION_DISTINCT = 2;
   }
+
+  // This rel is used  to create references,
+  // in case we refer to a RelRoot field names will be ignored
+  message ReferenceRel {
+    int32 subtree_ordinal = 1;
+  }
 }
--- a/site/docs/relations/logical_relations.md
+++ b/site/docs/relations/logical_relations.md
@@ -321,19 +321,19 @@ We could use the `ReferenceRel` to highlight the shared `A JOIN B` between the t
 One expressing `A JOIN B` (in position 0 in the plan), one using reference as follows: `ReferenceRel(0) JOIN C` and a third one
 doing `ReferenceRel(0) JOIN D`. This allows to avoid the redundancy of `A JOIN B`.
 
-| Signature            | Value           |
-| -------------------- | --------------- |
-| Inputs               | 1               |
-| Outputs              | 0               |
-| Property Maintenance | N/A (no output) |
-| Direct Output Order  | N/A (no output) |
+| Signature            | Value                                 |
+| -------------------- |---------------------------------------|
+| Inputs               | 1                                     |
+| Outputs              | 1                                     |
+| Property Maintenance | Maintains all properties of the input |
+| Direct Output Order  | Maintains order                       |
 
 
 ### Reference Properties
 
-| Property                    | Description                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                 | Required                    |
-|-----------------------------|---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------| --------------------------- |
-| Referred Rel                | A positional reference to a `Rel` defined within the same `Plan`.  | Required                    |
+| Property                    | Description                                                                    | Required                    |
+|-----------------------------|--------------------------------------------------------------------------------| --------------------------- |
+| Referred Rel                | A zero-indexed positional reference to a `Rel` defined within the same `Plan`. | Required                    |
 
 === "ReferenceRel Message"
 
--- a/site/docs/relations/logical_relations.md
+++ b/site/docs/relations/logical_relations.md
@@ -420,6 +420,8 @@ The operator that defines modifications of a database schema (CREATE/DROP/ALTER 
     ```proto
 %%% proto.algebra.DdlRel %%%
     ```
+=======
+>>>>>>> e041822 (feat: removing spurious change. (#283))
 
 ## Discussion Points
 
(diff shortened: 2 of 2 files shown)


site/docs/relations/logical_relations.md
```diff
@@ -310,6 +310,36 @@ If at least one grouping expression is present, the aggregation is allowed to no
...
+As a concrete example think about two queries `SELECT * FROM A JOIN B JOIN C` and `SELECT * FROM A JOIN B JOIN D`,
+We could use the `ReferenceRel` to highlight the shared `A JOIN B` between the two queries, by creating a plan with 3 `Rel`. 
+One expressing `A JOIN B` (in position 0 in the plan), one using reference as follows: `ReferenceRel(0) JOIN C` and a third one
+doing `ReferenceRel(0) JOIN D`. This allows to avoid the redundancy of `A JOIN B`.
+
+| Signature            | Value           |
+| -------------------- | --------------- |
+| Inputs               | 1               |
+| Outputs              | 0               |
+| Property Maintenance | N/A (no output) |
```
jacques-n, line 328: I would expect the reference operator to maintain the property and output order of it's reference. Can we update this and below to state that?

site/docs/relations/logical_relations.md
```diff
@@ -310,6 +310,36 @@ If at least one grouping expression is present, the aggregation is allowed to no
...
+|-----------------------------|---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------| --------------------------- |
+| Referred Rel                | A positional reference to a `Rel` defined within the same `Plan`.  | Required                    |
```
jacques-n, line 336: I suggest we state that this is a zero-indexed reference.

jacques-n (commented):
@curino , this looks really solid to me. Just a few minor things and then I think we can merge it. Thanks!

curino:
I addressed @jacques-n  comment. Anything else? @westonpace  / @jcamachor do you think this covers the key use cases for Spooling / DAGs / view-like things? 

I agree with @jacques-n argument during our community call, that this is a good simple and flexible primitive to have in core spec/proto, while in each of the language wrappers we could provider other facilities that are more idiomatic (e.g., spools) in that language/use-case.

jacques-n (approved):
I feel good about this being a great next step. We can always evaluate additional ways to enhance this pattern as we see them.

curino:
Thanks @jacques-n for the help devising the solution and quick turnaround on the PR.

proto/substrait/algebra.proto
```diff
@@ -1164,4 +1163,10 @@ message AggregateFunction {
     // Use only distinct values in the aggregation calculation.
     AGGREGATION_INVOCATION_DISTINCT = 2;
   }
+
+  // This rel is used  to create references,
+  // in case we refer to a RelRoot field names will be ignored
```
westonpace, line 1168: Would referring to RelRoot make sense?  Wouldn't that lead to a circular tree?

site/docs/relations/logical_relations.md
```diff
@@ -310,6 +310,37 @@ If at least one grouping expression is present, the aggregation is allowed to no
...
+## Reference Operator
+
+The reference operator is used to construct DAGs of operations. In a `Plan` we can have multiple Rel representing various 
+computations with potentially multiple outputs. The `ReferenceRel` is used to express the fact that multiple `Rel` might be
+sharing subtrees of computation. This can be used to express arbitrary DAGs as well as represent multi-query optimizations.
+
+As a concrete example think about two queries `SELECT * FROM A JOIN B JOIN C` and `SELECT * FROM A JOIN B JOIN D`,
+We could use the `ReferenceRel` to highlight the shared `A JOIN B` between the two queries, by creating a plan with 3 `Rel`. 
+One expressing `A JOIN B` (in position 0 in the plan), one using reference as follows: `ReferenceRel(0) JOIN C` and a third one
+doing `ReferenceRel(0) JOIN D`. This allows to avoid the redundancy of `A JOIN B`.
```
westonpace, line 322: I'm confused by this example because it seems to imply we can have multiple queries expressed in a single plan.  What is the root of the DAG here?  I think this is two DAGs.  I thought a reference rel would be more related to subqueries.

site/docs/relations/logical_relations.md
```diff
@@ -310,6 +310,37 @@ If at least one grouping expression is present, the aggregation is allowed to no
...
+
+| Signature            | Value                                 |
+| -------------------- |---------------------------------------|
+| Inputs               | 1                                     |
```
westonpace, line 326: Wouldn't this be 0?

proto/substrait/algebra.proto
```diff
@@ -1164,4 +1163,10 @@ message AggregateFunction {
...
+  message ReferenceRel {
```
westonpace, line 1169: Doesn't this need to be added to the `oneof` in the `Rel` message?

site/docs/relations/logical_relations.md
```diff
@@ -310,6 +310,37 @@ If at least one grouping expression is present, the aggregation is allowed to no
...
+| Outputs              | 1                                     |
+| Property Maintenance | Maintains all properties of the input |
+| Direct Output Order  | Maintains order                       |
+
+
+### Reference Properties
+
+| Property                    | Description                                                                    | Required                    |
+|-----------------------------|--------------------------------------------------------------------------------| --------------------------- |
+| Referred Rel                | A zero-indexed positional reference to a `Rel` defined within the same `Plan`. | Required                    |
```
westonpace, line 336: You call this "Referred Rel" here but "subtree_ordinal" in the proto.  Should we be more consistent?

westonpace (commented):
Sorry for the belated review.

proto/substrait/algebra.proto
jvanstraten: IIRC Jacques told me at some point that in the original design there is only supposed to be one `RelRoot`, and that it should indeed be the root of a relation, while the normal `Rel`s in `PlanRel` were exclusively intended to be used for this referencing logic. @jacques-n please correct me if I'm misremembering or if I misunderstood.

Personally I don't like this at all; IMO this structure should have been reflected in the protobuf message structure rather than making a repeated oneof, i.e. something like `PlanRel { RootRel root, repeated Rel subtrees }`. But that's just my opinion, and it's a bit late for that.

Since none of the above was ever documented AFAICT, and since rel references didn't exist at all until this proposal, I think most readers (including @curino I think) would interpret plans the way Arrow's consumer does it now and the way I initially interpreted it, being:

 - a `Plan` actually consists of one or more plans;
 - the only functional difference between `RootRel` and a normal `Rel` is that the former attaches column names and the other does not.

in which case it'd be perfectly fine to refer to a `RelRoot` as well.

IMO the better way to avoid circular references and to make parsing the plans easier is to require that you can only refer to relation trees with an index < the one you're referring from. Thus far everything in Substrait could be parsed with a single depth-first tree traversal (with carefully chosen field order); without this requirement this would no longer be the case.

proto/substrait/algebra.proto
jvanstraten: +1

PR #286, opened 2022-08-10
feat: add is_finite and is_infinite

thisisnic:
This PR adds definitions for is_infinite and is_finite.

### DIFF
--- a/extensions/functions_comparison.yaml
+++ b/extensions/functions_comparison.yaml
@@ -83,4 +83,28 @@ scalar_functions:
       - args:
           - value: fp64
         return: BOOLEAN
+  -
+    name: "is_finite"
+    description: Whether a value is finite (neither infinite nor NaN).
+    impls:
+      - args:
+          - value: fp32
+        nullability: DECLARED_OUTPUT
+        return: BOOLEAN
+      - args:
+          - value: fp64
+        nullability: DECLARED_OUTPUT
+        return: BOOLEAN
+  -
+    name: "is_infinite"
+    description: Whether a value is infinite.
+    impls:
+      - args:
+          - value: fp32
+        nullability: DECLARED_OUTPUT
+        return: BOOLEAN
+      - args:
+          - value: fp64
+        nullability: DECLARED_OUTPUT
+        return: BOOLEAN
 


extensions/functions_comparison.yaml
```diff
@@ -83,4 +83,36 @@ scalar_functions:
       - args:
           - value: fp64
         return: BOOLEAN
+  -
+    name: "is_finite"
+    description: Whether a value is finite.
+    impls:
+      - args:
+          - name: value_for_null
```
cpcloud, line 91: What does `value_for_null` mean? Does it mean the value that should be returned from this function when a `NULL` is encountered?

extensions/functions_comparison.yaml
```diff
@@ -83,4 +83,36 @@ scalar_functions:
       - args:
           - value: fp64
         return: BOOLEAN
+  -
+    name: "is_finite"
+    description: Whether a value is finite.
```
cpcloud, line 88: This should probably describe the meaning of finite here. For example, is NaN finite?

extensions/functions_comparison.yaml
thisisnic: Yep, that's exactly it.  This PR is still a draft as I've not finished working out exactly what options might be needed with these functions.

cpcloud: Interesting, I'm curious in what scenario the ability to return anything other than `NULL` for `is_(in)finite(NULL)` would be useful.

thisisnic: There might not be any - currently looking at various implementations of this function to see what options are provided there, so I may end up concluding that it's not needed.

thisisnic: Doesn't look like it's needed, removed now.

extensions/functions_comparison.yaml
thisisnic: Updated now.

extensions/functions_comparison.yaml
```diff
@@ -83,4 +83,24 @@ scalar_functions:
...
+  -
+    name: "is_infinite"
+    description: Whether a value is infinite.
+    impls:
+      - args:
+          - value: fp32
+        return: BOOLEAN
+      - args:
+          - value: fp64
+        return: BOOLEAN
```
cpcloud, line 105: I think the `BOOLEAN` and `fp32`/`fp64` need to be nullable since `NULL` would be the return value of a `NULL` input.

After that I think this PR is good to go.

cpcloud: Perhaps I am reading the spec wrong but under the [nullability handling section](https://github.com/substrait-io/substrait/blob/3d64e4aa41c9cfb5b5575973d23c07272eb902f9/site/docs/expressions/scalar_functions.md#nullability-handling)

I interpret

> The nullability of the output function is whatever the return type expression states

with `return: BOOLEAN` to mean that no matter whether the input is nullable the output will always be `true` or `false` and never `NULL`.

Am I interpreting that correctly?

cpcloud: It looks like nothing else declares `return: T?`, so this is fine according to the current conventions.

cpcloud (approved):

PR #288, opened 2022-08-15
fix: various fixes for #252 (write and DDL relations) and #284 (relation references)

jvanstraten:
- add new relation types to rel_type to make them usable
 - add constraints to prevent cyclic relation references
 - document how relation references work in the relation basics section
 - s/tuples/records/g for naming consistency
 - move ReferenceRel out of the AggregateFunction message scope

BREAKING CHANGE: various messages and semantics that were not yet reachable from the
Plan message were changed

### DIFF
--- a/proto/substrait/algebra.proto
+++ b/proto/substrait/algebra.proto
@@ -340,9 +340,10 @@ message ExchangeRel {
 //
 // This is for use at the root of a `Rel` tree.
 message RelRoot {
-  // A relation
+  // A relation.
   Rel input = 1;
-  // Field names in depth-first order
+  // Field names in depth-first order. If the relation doesn't return anything
+  // (zero outputs, for example DdlRel), this should be empty.
   repeated string names = 2;
 }
 
@@ -361,6 +362,9 @@ message Rel {
     ExtensionMultiRel extension_multi = 10;
     ExtensionLeafRel extension_leaf = 11;
     CrossRel cross = 12;
+    WriteRel write = 13;
+    DdlRel ddl = 14;
+    ReferenceRel reference = 15;
   }
 }
 
@@ -436,12 +440,13 @@ message WriteRel {
   }
 
   // The schema of the table (must align with Rel input (e.g., number of leaf fields must match))
+  // FIXME: I don't really have a problem with it being here because Substrait defines all kinds of things redundantly already, but why explicitly specify the whole schema here but not for RelRoot? It just specifies the names, which is enough, because the types can be derived from the input relation.
   NamedStruct table_schema = 3;
 
   // The type of operation to perform
   WriteOp op = 4;
 
-  // The relation that determines the tuples to add/remove/modify
+  // The relation that determines the records to add/remove/modify
   // the schema must match with table_schema. Default values must be explicitly stated
   // in a ProjectRel at the top of the input. The match must also
   // occur in case of DELETE to ensure multi-engine plans are unequivocal.
@@ -452,27 +457,39 @@ message WriteRel {
 
   enum WriteOp {
     WRITE_OP_UNSPECIFIED = 0;
-    // The insert of new tuples in a table
+    // The insert of new records in a table
     WRITE_OP_INSERT = 1;
-    // The removal of tuples from a table
+    // The removal of records from a table
+    // FIXME: how are the records to be deleted identified? It's stated that schema(input) must match the table schema, so is it equality? Does that mean that it's impossible to differentiate between different records with the same contents, i.e. it's not possible to match positionally?
     WRITE_OP_DELETE = 2;
-    // The modification of existing tuples within a table
+    // The modification of existing records within a table
+    // FIXME: likewise for WRITE_OP_DELETE, how are the records that are to be updated identified?
     WRITE_OP_UPDATE = 3;
-    // The Creation of a new table, and the insert of new tuples in the table
+    // The Creation of a new table, and the insert of new records in the table
     WRITE_OP_CTAS = 4;
   }
 
   enum OutputMode {
     OUTPUT_MODE_UNSPECIFIED = 0;
-    // return no tuples at all
+    // return no records at all
     OUTPUT_MODE_NO_OUTPUT = 1;
     // this mode makes the operator return all the tuple INSERTED/DELETED/UPDATED by the operator.
     // The operator returns the AFTER-image of any change. This can be further manipulated by operators upstreams
-    // (e.g., retunring the typical "count of modified tuples").
+    // (e.g., retunring the typical "count of modified records").
     // For scenarios in which the BEFORE image is required, the user must implement a spool (via references to
     // subplans in the body of the Rel input) and return those with anounter PlanRel.relations.
-    OUTPUT_MODE_MODIFIED_TUPLES = 2;
+    // FIXME: if I'm interpreting "before-image", "after-image", and that "spool" mechanism correctly, I'm not sure how those things help. For example, in an update operation, I'd expect the operator to return the previous values of the replaced records, because that's the information that would otherwise be thrown away. Likewise for a delete, it should return the rows that are deleted, not the emptiness that I would call the "after-image".
+    OUTPUT_MODE_MODIFIED_RECORDS = 2;
   }
+}
+
+// This rel is used to create references,
+// in case we refer to a RelRoot field names will be ignored
+message ReferenceRel {
+  // Zero-based index into the Plan.relations list. Must be less than the index
+  // of the relation tree that this reference appears in, in order to avoid
+  // cycles and forward references.
+  int32 rel_reference = 1;
 }
 
 // The argument of a function
@@ -1162,11 +1179,5 @@ message AggregateFunction {
 
     // Use only distinct values in the aggregation calculation.
     AGGREGATION_INVOCATION_DISTINCT = 2;
-  }
-
-  // This rel is used  to create references,
-  // in case we refer to a RelRoot field names will be ignored
-  message ReferenceRel {
-    int32 subtree_ordinal = 1;
   }
 }
--- a/site/docs/relations/logical_relations.md
+++ b/site/docs/relations/logical_relations.md
@@ -323,7 +323,7 @@ doing `ReferenceRel(0) JOIN D`. This allows to avoid the redundancy of `A JOIN B
 
 | Signature            | Value                                 |
 | -------------------- |---------------------------------------|
-| Inputs               | 1                                     |
+| Inputs               | 0                                     |
 | Outputs              | 1                                     |
 | Property Maintenance | Maintains all properties of the input |
 | Direct Output Order  | Maintains order                       |
@@ -333,7 +333,7 @@ doing `ReferenceRel(0) JOIN D`. This allows to avoid the redundancy of `A JOIN B
 
 | Property                    | Description                                                                    | Required                    |
 |-----------------------------|--------------------------------------------------------------------------------| --------------------------- |
-| Referred Rel                | A zero-indexed positional reference to a `Rel` defined within the same `Plan`. | Required                    |
+| Referred Rel                | A zero-indexed positional reference to a `Rel` defined within the same `Plan`. The index must be less than the index of the relation tree that the reference appears in; put differently, you can only refer to trees that have already been declared. This avoids cyclic dependencies and forward references. | Required                    |
 
 === "ReferenceRel Message"
 
@@ -343,14 +343,14 @@ doing `ReferenceRel(0) JOIN D`. This allows to avoid the redundancy of `A JOIN B
 
 ## Write Operator
 
-The write operator is an operator that consumes one output and writes it to storage. This can range from writing to a Parquet file, to INSERT/DELETE/UPDATE in a database. 
+The write operator is an operator that consumes one output and writes it to storage. Currently, only named tables and extensions are supported, but the intention is to also support writing files in the future.
 
-| Signature            | Value                                                   |
-| -------------------- |---------------------------------------------------------|
-| Inputs               | 1                                                       |
-| Outputs              | 1                                                       |
-| Property Maintenance | Output depends on OutputMode (none, or modified tuples) |
-| Direct Output Order  | Unchanged from input                                    |
+| Signature            | Value                                                    |
+| -------------------- |----------------------------------------------------------|
+| Inputs               | 1                                                        |
+| Outputs              | 1                                                        |
+| Property Maintenance | Output depends on OutputMode (none, or modified records) |
+| Direct Output Order  | Unchanged from input                                     |
 
 ### Write Properties
 
@@ -360,8 +360,8 @@ The write operator is an operator that consumes one output and writes it to stor
 | Write Type                 | Definition of which object we are operating on (e.g., a fully-qualified table name).                                                                                                                                                                                                                                                                                                                                                                                               | Required                                            |
 | CTAS Schema                | The names of all the columns and their type for a CREATE TABLE AS.                                                                                                                                                                                                                                                                                                                                                                                                                 | Required only for CTAS  |
 | Write Operator             | Which type of operation we are performing (INSERT/DELETE/UPDATE/CTAS).                                                                                                                                                                                                                                                                                                                                                                                                             | Required                                            |
-| Rel Input                  | The Rel representing which tuples we will be operating on (e.g., VALUES for an INSERT, or which tuples to DELETE, or tuples and after-image of their values for UPDATE).                                                                                                                                                                                                                                                                                                           | Required                                            |
-| Output Mode | For views that modify a DB it is important to control, which tuples to "return". Common default is NO_OUTPUT where we return nothing. Alternatively, we can return MODIFIED_TUPLES, that can be further manipulated by layering more rels ontop of this WriteRel (e.g., to "count how many tuples were updated"). This also allows to return the after-image of the change. To return before-image (or both) one can use the reference mechanisms and have multiple return values. | Required for VIEW CREATE/CREATE_OR_REPLACE/ALTER    |
+| Rel Input                  | The Rel representing which records we will be operating on (e.g., VALUES for an INSERT, or which records to DELETE, or records and after-image of their values for UPDATE).                                                                                                                                                                                                                                                                                                           | Required                                            |
+| Output Mode | For views that modify a DB it is important to control, which records to "return". Common default is NO_OUTPUT where we return nothing. Alternatively, we can return MODIFIED_RECORDS, that can be further manipulated by layering more rels ontop of this WriteRel (e.g., to "count how many records were updated"). This also allows to return the after-image of the change. To return before-image (or both) one can use the relation reference mechanisms to use the relation subtree passed to the write input more than once. | Required for VIEW CREATE/CREATE_OR_REPLACE/ALTER    |
 
 
 ### Write Definition Types
--- a/proto/substrait/plan.proto
+++ b/proto/substrait/plan.proto
@@ -14,9 +14,13 @@ option java_package = "io.substrait.proto";
 // Either a relation or root relation
 message PlanRel {
   oneof rel_type {
-    // Any relation (used for references and CTEs)
+    // A relation with an output that should be referred to by one or more
+    // relation references in subsequent relation trees, but for which the
+    // result is not logically returned by the plan.
     Rel rel = 1;
-    // The root of a relation tree
+    // The root of a relation tree. The result is logically returned by the
+    // plan using the enclosed field names. The result may also be used by
+    // relation references in subsequent relation trees.
     RelRoot root = 2;
   }
 }
--- a/site/docs/relations/basics.md
+++ b/site/docs/relations/basics.md
@@ -1,6 +1,12 @@
 # Basics
 
-Substrait is designed to allow a user to construct an arbitrarily complex data transformation plan. The plan is composed of one or more relational operations. Relational operations are well-defined transformation operations that work by taking zero or more input datasets and transforming them into zero or more output transformations. Substrait defines a core set of transformations, but users are also able to extend the operations with their own specialized operations.
+Substrait is designed to allow a user to construct an arbitrarily complex data transformation plan. The plan is composed of one or more relational operations. Relational operations are well-defined transformation operations that work by taking zero or more input datasets and transforming them, typically returning a single result set. Substrait defines a core set of transformations, but users are also able to extend the operations with their own specialized operations.
+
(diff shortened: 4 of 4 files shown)


jvanstraten:
To address problems that @westonpace and I found in #252 and #284 after they were already merged.

@curino, I was pretty confused by some of the proto comments relating to `WriteRel`. I wrote `FIXME` lines where applicable. Could you please elaborate in review threads or something? I'll mark this as ready for review when I'm clear on what's going on and have removed those lines.

site/docs/relations/logical_relations.md
```diff
@@ -333,7 +333,7 @@ doing `ReferenceRel(0) JOIN D`. This allows to avoid the redundancy of `A JOIN B
 
 | Property                    | Description                                                                    | Required                    |
 |-----------------------------|--------------------------------------------------------------------------------| --------------------------- |
-| Referred Rel                | A zero-indexed positional reference to a `Rel` defined within the same `Plan`. | Required                    |
+| Referred Rel                | A zero-indexed positional reference to a `Rel` defined within the same `Plan`. The index must be less than the index of the relation tree that the reference appears in; put differently, you can only refer to trees that have already been declared. This avoids cyclic dependencies and forward references. | Required                    |
```
jacques-n, line 336: I'm not convinced we need the constraint of no forward references. I'm fine with it, just not convinced it is necessary. If you feel strongly, sounds good.

proto/substrait/algebra.proto
```diff
@@ -452,29 +457,41 @@ message WriteRel {
 
   enum WriteOp {
     WRITE_OP_UNSPECIFIED = 0;
-    // The insert of new tuples in a table
+    // The insert of new records in a table
     WRITE_OP_INSERT = 1;
-    // The removal of tuples from a table
+    // The removal of records from a table
+    // FIXME: how are the records to be deleted identified? It's stated that schema(input) must match the table schema, so is it equality? Does that mean that it's impossible to differentiate between different records with the same contents, i.e. it's not possible to match positionally?
```
jacques-n, line 463: @curino and I had a long conversation about this offline. We should have done a better job of capturing that discussion here. Yeah, I agree this is a problem. In some circumstances you can imagine from input that there is a implicit rowId column that comes from the input to the delete. However, there are also scenarios where you might be using a different pattern and those should be illegal. @curino wasn't as sold on the problem so we punted on it until someone was actually trying to implement delete.

jacques-n (commented):
Thanks for picking up these cleanups @jvanstraten.

proto/substrait/algebra.proto
jvanstraten: SQL delete is based on a condition, which makes sense to me, so given this structure, I would expect a delete operator in Substrait to take a schema with a single boolean as its relation input... though on second thought, that doesn't really make sense either, because that boolean could be derived from a different table entirely. It's also not obvious at all how you'd make use of indices in that case.

Okay, I guess I see the problem. I imagine @curino had a structure like `read(table X) -> filter(condition) -> write(delete from table X)` in mind for deletions? The problem with that is if I would write `read(table Y) -> filter(condition) -> write(delete from table X)` it would be legal by these rules if table X and Y happen to have the same schema, but equal schema doesn't mean equal table. Likewise I could stick a projection in there to make the schema match and it'd still be nonsense.

I feel like the problem is that we're trying to push a square peg in a round hole here. Why are we trying to make these fundamentally different operators satisfy normal relation semantics, and the exact same semantics across a number of them to boot? I didn't want to butt into this discussion at the time because I didn't feel qualified, but maybe I should have...

IMO, update and delete should be a different operator entirely, specifically one without an input. But if we would want to stick to a single relation type for modifications, I would define it like this, based on my limited knowledge of SQL:

```proto
message DdlRel {
  RelCommon common = 1;

  // Table to operate on.
  oneof read_type {
    LocalFiles local_files = 2;
    NamedTable named_table = 3;
    ExtensionTable extension_table = 4;
  }

  // Current table schema. If not specified, the table does not currently
  // exist.
  NamedStruct current_schema = 5;

  // Optional row deletion expression. Operates on the table schema directly.
  // Only legal for preexisting tables, i.e. base_schema is specified. Must
  // yield a non-nullable boolean if specified; if true, a record is deleted,
  // if false, it is retained. If not specified, all records are retained
  // (same as literal false). Set to literal true to model a SQL truncate
  // operation.
  Expression delete = 6;

  // Optional filter expression for update operations. Operates on the table
  // schema directly. Only legal for preexisting tables, i.e. base_schema is
  // specified. If delete is also specified, this is evaluated only on the rows
  // that were retained. Must yield a non-nullable boolean if specified; if
  // true, a record is selected for updating, if false, it is not selected. If
  // not specified, all records are updated (same as literal true).
  Expression filter = 7;
  
  // Specifies the new value for each column. If specified, there must be
  // exactly as many Column messages as there are columns in the (altered)
  // schema. Must be specified when the schema is altered. Only legal for preexisting tables, i.e. base_schema is specified.
  repeated Column update = 8;

  message Column {
    // Optional expression to use to update the column. The return type must
    // match the column type as per the (altered) schema exactly. If not
    // specified, the column is not updated. All columns must be specified if
    // the schema is being altered and delete is not set to literal true. The
    // expression operates on the original table schema, and is only evaluated
    // for records where the filter expression (if specified) returned true
    // and the delete expression (if specified) returned false.
    Expression expression = 1;
  }
  
  // Optional relation to produce records that are to be inserted into the
  // table. The schema returned by the relation must match the (altered) schema
  // of the table exactly. For ordered storage, the rows are expected to be
  // appended to the end of preexisting records.
  Rel insert = 9;

  // If specified for a preexisting table, the table schema is altered to the
  // given schema. Cannot be used in conjunction with filter; i.e. it's illegal
  // to specify both. Must be specified if there is no preexisting table, i.e.
  // current_schema is not specified.
  NamedStruct new_schema = 10;

  // Extension for additional schema information, such as index specifications.
  substrait.extensions.AdvancedExtension new_schema_extension = 11;
  
  // SQL operations vs. fields populated:
  //
  // | Operation   | current_schema | delete | filter | update | insert | new_schema |
  // |-------------|----------------|--------|--------|--------|--------|------------|
  // | CREATE      | -              | -      | -      | -      | ?      | specified  |
  // | INSERT INTO | specified      | ?      | ?      | ?      | VALUES | -          |
  // | UPDATE      | specified      | ?      | WHERE  | SET    | ?      | -          |
  // | DELETE FROM | specified      | WHERE  | ?      | ?      | ?      | -          |
  // | ALTER       | specified      | ?      | -      | *      | ?      | specified  |
  // | TRUNCATE    | specified      | true   | -      | -      | ?      | ?          |
  //
  // - : should not be specified/illegal
  // ? : optional, but not possible to specify using SQL
  // * : new columns would be set to literal null for a normal ALTER, but any
  //     initial value or projection can be used
  
  // What to return. Required.
  oneof return_type {
    // Return an empty schema with zero rows (as close to nothing as
    // we can get in Substrait's relation tree system).
    google.protobuf.Empty nothing = 12;

    // Return the affected rows (deleted or updated) as they were before the
    // update. Inserted rows are not returned.
    google.protobuf.Empty pre_image = 13;

    // Return the updated rows as they were after the update. Deleted rows are
    // not returned.
    google.protobuf.Empty post_image = 14;

    // Return the field-wise concatenation of the pre-image and post-image,
    // with all top-level fields aka column types made nullable. The pre-image
    // part of a record will be nullified for inserted columns, the post-image
    // part of a record will be nullified for deleted columns. Only affected
    // rows (deleted, updated, or inserted) are returned.
    google.protobuf.Empty affected_pre_and_post = 15;
    
    // Like affected_pre_and_post, but also returns rows that were not
    // affected.
    google.protobuf.Empty all_pre_and_post = 16;
  }
}
```

But I understand if it's a bit late for that.

site/docs/relations/logical_relations.md
curino: Seems like a fine sufficient condition, not necessary, but loop-checks are maybe trickier?

proto/substrait/algebra.proto
curino: @jvanstraten  thanks for looking into this. I think the challenge @jacques-n  and I were faced with is that SQL supports more complex statements where you can write a nearly arbitrary query that affects a `DELETE` or `UPDATE` statement.  A couple of examples:

```
UPDATE wine w
SET stock = stock - (
                     SELECT SUM (quantity)
                     FROM order
                     WHERE date = CURRENT_DATE AND order.wine_name = w.name
                    )

```

```
DELETE table-name1
  FROM table-name1 
  JOIN table-name2 ON column-name3 = column-name4
 WHERE condition
```

This made us prefer reusing the `Rel` structure, vs building out a separate dedicated scheme. You are right that you could construct a non-sensical plan, but this is true in general (hence your awesome validator work).  The intuition is what you describe, where the `input` selects the tuples we want to affect and either describe its after-image for updates, or the tuples to be deleted.  SQL being declarative there is no way to say "delete the first occurrence of this identical tuples", as a consequence folks are used to workarounds such as primary keys or rowIds (per @jacques-n's comment) in the less common scenario where separate handling of cloned tuples are required.  Also I believe the `Expression` approach above might suffer of the same limitation right? 

I am not attached to this implementation, and happy to brainstorm/support variations. 

I think we should also soon build-out a "golden dataset" of queries we aim to support so that we can iterate variations and compare them against that (or at least convince ourselves we can cover all cases that matter). Folks at MS are starting to using this in practical settings and we will stumble on any detail issue and come back and address them, but if you have ideas/time to work on this, super happy to collaborate.

curino: Also for the pre/post image I liked the structure we came up with with @jacques-n as it was rather minimal and allowed us to stitch together plans that do any sort of pre/after image with very few additions/fields.

I think beside better comments we should soon get to have a bunch of example SQL->Plan listed in the docs to explain what we mean (I have been hesitating not to overly SQL-ify our website).

jvanstraten: Okay, after a few hours of typing (most of which I've now edited out), I think I've narrowed down where my misunderstanding comes from. I now agree that, with some further specification, WriteRel could work, but with the following limitations:

 - It will only ever work on key-value-based storage. Anything that identifies rows *implicitly* by some index or pointer or whatever (columnar formats, row-oriented formats, stream-based systems, etc.) can never work with this, because the identifying information is by definition not part of the data.
 - An update operation can never modify key columns. I don't know if this is also a limitation of SQL update statements, but if so, IMO this would be a pretty big limitation to just copy-paste from SQL without further thought.
 - Records with equivalent data cannot be told apart, though I suppose that's a limitation of key-value-based storage in the first place.

These feel like pretty hefty limitations, considering that the rest of Substrait works just fine for non-key-value-based systems, and considering that Substrait currently doesn't even acknowledge the *concept* of key columns in the first place. However, I'm not the one who has experience implementing query engines, so I'll yield if you guys think these limitations are perfectly acceptable.

That being said, my proposal does *not* suffer from these limitations. The key difference is that an expression is defined to be evaluated for each record individually, so any implicit information identifying the record can stay implicit in the execution context of that expression. A consumer could even define a function extension that allows a user to query this implicit information for use in the expression; for example, you could define a function that returns the index of the record in the current iteration, if that is meaningful information for that particular system. It would be, for instance, for a read from a columnar format, which has implicit ordering (even if Substrait does not in general require a meaningful ordering from a read).

Also, I'm confident that I can convert everything that the current WriteRel supports to my proposal. You mention deletions and updates as pain points, so here goes. Delete:

```
delete = in_predicate_subquery(
  needles = (field references to key columns)
  haystack = (
    project(
      emit only key columns,
      <relation input goes here>
    )
  )
)
```

That's pretty much just making explicit what a key-value-based delete operator actually does. I imagine a key-value-based system would special-case that syntax.

Update is uglier:

```
update[i] = coalesce(
  scalar_subquery( // I'm assuming this yields null if there is no input
    fetch( // just to avoid nonsense if the relation yields equivalent rows
      count = 1,
      project(
        emit only column i,
        filter(
          condition = (key columns in inner query match key columns in outer query),
          <relation input goes here>
        )
      )
    )
  ),
  field(i)
)
```

but only because I'm lacking a proper struct-building expression (and this is not the first time I'm running into this by far; other things it would make redundant are the repeated expressions in aggregate grouping sets, the 2D version of or_list, and struct literals; you could do the same for list and map too and avoid lots of literal shenanigans in fact). It would be *much* more natural to have only one update expression and require it to yield a struct that matches the schema, rather than requiring individual expressions. That would also solve the problem of differentiating between "don't update the record" (would be a null struct) and "replace a field in the record with null" (would be a struct containing null). The whole thing in Substrait where the schema of a row is defined using the same abstraction as a struct is awesome; I don't see why we're not using it more. But I digress. Assuming that expression would exist, it would become:

```
update = coalesce(
  scalar_subquery(
    fetch(
      count = 1,
      project(
        expression[0] = struct(field(0) ... field(n-1)),
        emit only above expression,
        filter(
          condition = (key columns in inner query match key columns in outer query),
          <relation input goes here>
        )
      )
    )
  ),
  struct(field(0) ... field(n-1))
)
```

Even better would be if the field reference would be generalized to be able to return the entire record as a struct, rather than requiring at least one index operation. Then the `struct(field(0) ... field(n-1))` subexpressions would just become that generalized reference type. I suppose technically you could already do it with a mask expression, not sure if it would be considered legal to apply that to RootReference and OuterReference though.

The particular update query you listed can be represented much more naturally, though,

```
update[row index of stock] = subtract(field(row index of stock), scalar_subquery(...))
```

but if SQL supports IMO arcane syntax like the delete statement for updates as well, I imagine it would be harder to generate them like this from SQL in general. That being said, I don't think we should be considering a SQL-to-Substrait producer as the golden standard for how things should be represented in Substrait anyway. There are more producers being worked on than just Isthmus; if Substrait intends to be a generalized interchange format for queries, I don't see why it should receive special treatment. Substrait is not supposed to just be a SQL syntax tree, right?

curino: I think I understand your concerns, I don't agree with some but some are fundamental (e.g., the update of PK issue). 

There is maybe one misalignment of how we think about this. I am not envisioning an engine to first perform the `input` query, then match the output tuples and perform the delete operation by key lookup. The `input` is a declarative specification of which tuples to operate on. As such I think has similar properties to what you write for expression in your variant of `update`. Imagine I want to delete 10% of the tuples at random, in the body of the `input` I can have `rand()>0.9` and any tuple that matches gets deleted, similarly I think you can have the special function returning the index/position/rowId and a match/no-match rule to pick which tuples to delete. So I think of the plan as a declarative specification of what needs to get done, not a mandatory execution path. This might be due to different mindsets of how sophisticated/opinionated is the engine behind it. We might need to have a "logical" (what I have) and "physical" (some variant of what you propose) version of these statements. 

On the other hand, the limitation of an implicit reference to the PK for the update remains valid as it is supported by SQL and because we have potentially mingled it in the body of `input` and that is confusing. One option is to simplify things further and implement  `UPDATE`  out of `DELETE+INSERT` as follows:

1) We "delete" all tuples we want to update
2) We output all before-image of deleted tuples
3) We decorate those with whatever changes we fill are necessary (including PK changes)
4) We "insert" them back in. 

Obviously most engines would not execute it this this way, but interpret the plan and perform whatever efficient update operation they can support. I think this should have the same expressivity of what you were proposing but without the convoluted `UPDATE` implementation (again maybe a logical/physical mindset here too). 

We could also use some variant of your "delete" statement if it helps other engines and remains similarly natural to convert in/out of all the SQL-speaking systems (which I would really love to capture as there are many and with a massive market and I think can be a big win to Substrait if well supported). I am happy to work with Ashvin/Jesus to look at how some variant of your delete can be converted to/from DB plans. We can also look at how SQL Server or other DBMSs represent this in their internal plan data structures, there might be some ideas there.

jvanstraten: Sorry, another long one...

> There is maybe one misalignment of how we think about this. I am not envisioning an engine to first perform the input query, then match the output tuples and perform the delete operation by key lookup. The input is a declarative specification of which tuples to operate on.

I'm not really concerned with how the engine will actually perform the query. My problem is with how to specify what the behavior should be. If I were to hand `int a = 1; int b = a + 3; int c = b * 4; return b` to a C compiler, I would expect it to emit a program that just returns 4 straight away, but the specification says nothing about that. It just says what those individual statements mean out of context and how they combine, by induction specifying the behavior of every possible C program, and then says "a C compiler can emit whatever program it wants, as long as all visible behavior matches the specified behavior of the input program." 

Likewise, we have to specify the behavior of every possible Substrait plan. Unless we want to special-case the input subtrees of update/delete, the behavior of the relations needs to satisfy the contract of yielding a list of records with some defined schema when executed (and *only* yielding that). Or we could change the contract to include more information, but that means updating the specification of all existing relations.

Again to illustrate, let's say I write the following:

```
WriteRel {
  named table A,
  schema (a: i32, b: i32),
  op: update,
  input: ReadRel {
    named table B,
    schema (c: i32, d: i32)
  }
  output: modified tuples
}
```

and my tables are:

| a | b |
|---|---|
| 1 | 1 |
| 2 | 1 |
| 2 | 2 |
| 2 | 2 |

| c | d |
|---|---|
| 2 | 2 |
| 2 | 3 |
| 3 | 3 |

The question that the spec has to answer for this query, and every other possible query, is A) what should be returned by the WriteRel (if anything, or maybe an error should be thrown), B) how is table A modified (if at all), and C) why? It does this for the ReadRel just fine, but for the update WriteRel this is all kinds of broken.

The first problem is that Substrait does not conceptually define what keys are. The whole point of adding schemas to the plan in the first place is to make plans not sensitive to context, so that means we also can't make use of key information in the behavior description of this plan. Which means we're completely lost, because we can't both specify the desired before- and after-image of the update in the input *and* satisfy the constraint that the schema of input must match the schema of the mutated table. We could add a field to WriteRel that specifies which columns should be matched and which shouldn't, but then you'd still have the problem that you can't update key columns.

Likewise, to a lesser extent, for delete. If we take the above plan and change only the operation to delete, again, what should happen? Should only the first `(2, 2)` row be deleted, or both? Or maybe `b` is not a key column, and two of the three `a=2` rows should be deleted? Which? And so on. I would be fine with a specification like

> A delete operation removes all records present in both the input and the mutated table from the mutated table. Records in the input that do not exist in the mutated table are ignored. If a record appears multiple times in the mutated table, only one record is to be deleted for every matching record in the input. In `MODIFIED_TUPLES` output mode, the write relation returns the deleted records.

but I imagine this conflicts pretty badly with what key-value-based engines currently do. I would expect them to only match the (primary) key column and call it a day, so specifying deletion of `(2, 2)` might end up deleting `(2, 1)` if the first column is the primary key. We could fix that by specifying which columns are keys and saying that only those are to be matched, but that really begs the question why those columns would need to be included in the input schema in the first place. And then still, if the engine uses hidden row IDs to match rows to allow for non-unique primary keys, then the above query wouldn't work at all because the to-be-deleted records don't originate from table A.

The output mode is another problem. For updates the relevant information is the before-image, because the after-image is just a copy of the input query. That could be obtained via some yet to be defined relation reference system, but the before-image can't be. Likewise for deletions; I would imagine anyway that the after-image of a delete operator is always empty.

I'm pretty sure the correct answer to the above query is "you can't do that," since I don't see any way (based on the SQL syntax you've showed me) to update or delete based on a different table. I think this is precisely where the problem lies. The SQL syntax works because it's not just a tree where information flows from leaf to root (or a DAG/forest/whatever), it's a flat thing with a number of possible modifiers. You can convert SELECT, INSERT, and CREATE statements to a tree because information only flows in one direction, but you just can't do that for UPDATE and DELETE. In tree form, the UPDATE/DELETE node would have to somehow live at both the root *and* a leaf of the tree.

We could define some kind of dummy relation to act as a placeholder for the leaf part, and require the input of WriteRel to have exactly one of those placeholders in it. We'd have to be careful though, because WriteRels could be nested, rendering the placeholder ambiguous (we could just disallow nesting of WriteRels, or we could use some anchor/reference trick or something). Besides matching SQL semantics more closely, an advantage of this is that the table specification in the WriteRel doesn't have to be duplicated in the plan for the ReadRel. An IMO big disadvantage is that these are still relatively context-sensitive rules.

This on its own doesn't solve the problem, though; the origin of the records now necessarily has to at least in part come from the mutated table, but we still haven't specified how to match them. One solution could be to embrace an implicit row ID concept in the specification, and define for each relation how those implicit row IDs propagate through the relations (for some this is obvious; for others, like aggregates or joins, not so much). That effectively just changes their contract to return not only a list of records, but also some record ID along with each record. Then the delete and update operators can make use of that information. This would also work on its own, without the placeholder relation; for the above query, the row IDs returned by the records from table B would by definition be distinct from those returned by table A, so nothing would be deleted.

The expression-based approach I outlined in my previous comment doesn't suffer from these issues, because expressions are already evaluated in record context rather than in table context, which makes it much easier to specify how they should behave. That's the only reason why I suggested it. So, as for

> We could also use some variant of your "delete" statement if it helps other engines and remains similarly natural to convert in/out of all the SQL-speaking systems (which I would really love to capture as there are many and with a massive market and I think can be a big win to Substrait if well supported).

I honestly don't know if using expressions is better or worse for query engines in general than beating relations into submission. My background, if anything, is limited to streaming systems where all nodes are pure stream operators, where updates and deletions just don't exist as primitive operations. In the end, I only care that we specify these things in a way that generalizes to all possible plans, and not just some subset that at first glance resembles SQL syntax.

Actually, to clarify "generalizing to all possible plans"... I'm *not* saying that all possible plans must be valid, I'm saying that the specification must for every possible plan answer whether the plan *is* valid, and if so, what constraints a consumer must satisfy in terms of behavior if it accepts it. Those constraints don't even need to be as strict as "it HAS to do EXACTLY this;" see also unspecified options. But if something *is* left to the imagination of the consumer, that should explicitly be unspecified behavior. Just literally not saying anything about it in the specification is technically valid, but then there's no way for a human reading the spec to judge whether A) they missed something, B) the authors of the spec missed something, or C) the intention is actually that the consumer can decide what to do with the plan.

> One option is to simplify things further and implement UPDATE out of DELETE+INSERT [...]

I'd personally be fine with that, though I imagine it would make writing a consumer much more difficult due to the degree of context-sensitivity involved in extracting the update relation from that.

We'd also have to be careful to specify order of operations. Actually, I guess that applies in general: if I were to join the output of two WriteRels acting on the same table, and a consumer naively accepts the plan, what happens? Is the plan invalid and is it therefore wrong for it to accept it, is behavior unspecified (i.e. it's up to the consumer to pick order of operations or something in between), or does it maybe have to be evaluated left-to-right? WriteRel and DdlRel are the first relations to be defined that have side effects, after all.

Semi-related to this discussion; during the sync meeting @jacques-n asked me to split the obvious changes off from the disputed ones. I agreed at the time, but I'm not so sure anymore now, because I think we at least agree that update is currently broken to some extent. I'd rather get the spec sorted before adding the relations to the Rel oneof and updating the comments.

jvanstraten: Okay, sorting this out in a call (at least to converge on the basics) works for me.

jvanstraten: Hijacking this for another thing I only just put one and one together for while writing [this](https://github.com/substrait-io/substrait/pull/239#issuecomment-1246535134): I don't see how WriteRel can ever consistently work for write-only files, because it relies heavily on associative table and read-modify-write semantics. That means it's not the complementary operation of ReadRel. That's, honestly, terrible and confusing. Can we just rename it to UpdateRel or something and reserve WriteRel for a pure sink operation? AFAICT a pure WriteRel would replace CTAS, though...?

curino: @jvanstraten  and @jacques-n we should find time to talk!

jacques-n:
Closing due to lack of progress in a year and another ticket solving the catalyst issue.

PR #289, opened 2022-08-16
feat: add case_sensitivity option to string functions

richtia:
Add an option to control case sensitivity to string functions that may need it.

BREAKING CHANGE: added option arguments to existing functions.

### DIFF
--- a/extensions/functions_string.yaml
+++ b/extensions/functions_string.yaml
@@ -22,7 +22,7 @@ scalar_functions:
     impls:
       - args:
           - name: case_sensitivity
-            options: [ CASE_SENSITIVE, CASE_INSENSITIVE]
+            options: [ CASE_SENSITIVE, CASE_INSENSITIVE ]
             required: false
           - value: "varchar<L1>"
             name: "input"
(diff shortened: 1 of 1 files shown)


jvanstraten (dismissed):
Nice, and the added documentation makes things look a lot better already. Note that this is a breaking change though (the `required: false` is not what you might think it means), so I've edited the description accordingly so it ends up in the squash commit (at least I *think* it does).

Looking at this I also realized that for another PR I got pedantic about making an option for ASCII vs full unicode case conversions, and the same logic applies to these options, so maybe there should be a `CASE_INSENSITIVE_ASCII` option added to the end. But that should probably go in another PR.

extensions/functions_string.yaml
```diff
@@ -15,15 +15,32 @@ scalar_functions:
...
     name: like
-    description: Are two strings like each other.
+    description: >-
+      Are two strings like each other.
+
+      The `case_sensitivity` option applies to the `match` argument.
     impls:
       - args:
+          - name: case_sensitivity
+            options: [ CASE_SENSITIVE, CASE_INSENSITIVE]
```
cpcloud, line 25:
```suggestion
            options: [ CASE_SENSITIVE, CASE_INSENSITIVE ]
```

micro nit!

richtia: Good catch!  updated!

cpcloud (approved):

PR #290, opened 2022-08-17
feat: add hyperbolic functions

thisisnic:
This PR adds hyperbolic functions

### DIFF
--- a/extensions/functions_arithmetic.yaml
+++ b/extensions/functions_arithmetic.yaml
@@ -330,13 +330,15 @@ scalar_functions:
           - name: rounding
             options: [ TIE_TO_EVEN, TIE_AWAY_FROM_ZERO, TRUNCATE, CEILING, FLOOR ]
             required: false
-          - value: fp32
+          - name: x
+            value: fp32
         return: fp32
       - args:
           - name: rounding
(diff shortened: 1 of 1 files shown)


jvanstraten (changes requested):
Uh, hm. I guess I didn't catch this in #241, but it should be fp32 -> fp32 and fp64 -> fp64. Unless you're building huge lookup tables (which may or may not be feasible for fp32, not so much for fp64) I see zero reason to have an fp32 -> fp64 variant from a computational perspective, and it goes against #251. Can you change the new functions to fp32 -> fp32? I'll submit a patch to weed out inconsistencies between all the new functions in general (and at some point to codify #251, since there have been no objections) later.

thisisnic:
> Uh, hm. I guess I didn't catch this in #241, but it should be fp32 -> fp32 and fp64 -> fp64. Unless you're building huge lookup tables (which may or may not be feasible for fp32, not so much for fp64) I see zero reason to have an fp32 -> fp64 variant from a computational perspective, and it goes against #251. Can you change the new functions to fp32 -> fp32? I'll submit a patch to weed out inconsistencies between all the new functions in general (and at some point to codify #251, since there have been no objections) later.

Good catch, I've updated them in this PR now!

jvanstraten (approved):
Thanks, LGTM now.

jvanstraten (approved):

PR #291, opened 2022-08-17
feat: add nullif function

thisisnic:
This PR adds 2 conditional functions:  `if_else` and `nullif`.

### DIFF
--- a/extensions/functions_comparison.yaml
+++ b/extensions/functions_comparison.yaml
@@ -107,4 +107,12 @@ scalar_functions:
           - value: fp64
         nullability: DECLARED_OUTPUT
         return: BOOLEAN
+  -
+    name: "nullif"
+    description: If two values are equal, return null. Otherwise, return the first value.
+    impls:
+      - args:
+          - value: any1
+          - value: any1
+        return: any1
 


extensions/functions_comparison.yaml
```diff
@@ -107,4 +107,23 @@ scalar_functions:
           - value: fp64
         nullability: DECLARED_OUTPUT
         return: BOOLEAN
+  -
+    name: "if_else"
+    description: If true, return the first value, otherwise return the second.
```
thisisnic, line 112: I'm a bit unsure about phrasing here; does it make sense to refer to the second and third parameters as "first" and "second" values? Perhaps not. Not sure how best to go about this.

jvanstraten:
If-else is actually special-cased as an expression type, so there's no need for a function for it. See https://substrait.io/expressions/specialized_record_expressions/#if-expression

thisisnic:
> If-else is actually special-cased as an expression type, so there's no need for a function for it. See https://substrait.io/expressions/specialized_record_expressions/#if-expression

Thanks, removed now.

jvanstraten (approved):
LGTM

PR #292, opened 2022-08-17
fix: update descriptions and specify null behavior for bool funcs

gforsyth:
The existing descriptions state that these boolean functions are using Kleene logic, so I've added in the null handling assuming that `null` is a stand-in for "unknown" in these contexts.

`and` and `or` are currently variadic -- I've added descriptions of null handling for 2 arguments, but not more than that.


jvanstraten (commented):
I don't think the nullability change is correct. The default behavior is MIRROR, which means they use Kleene logic when any input is nullable but reduce to normal Boolean otherwise.

Technically none of the functions need a minimum variadic count; even with 0 args they would work. xor can technically also be variadic (it generalizes to an odd parity check) but I do think it's somewhat controversial to still call it xor at that point so I'm also fine with keeping it as is. For reference: and() -> true, and(x) -> x, or() -> false, or(x) -> x, xor() -> false, xor(x) -> x, xor(a, b...) -> xor(a, xor(b...)).

I agree though that and_not has no right to be variadic, though it could be interpreted as a regular and with all but the first argument inverted or something... IMO it shouldn't even be in core at all because it's hardly primitive, but lets not remove it unless it gets in the way, I guess.

gforsyth:
Thanks @jvanstraten -- I've updated the nullability (and marked the inputs as nullable), and removed the variadic minimum from `and` and `or`.  I'm going to leave `xor` alone, I think.

extensions/functions_boolean.yaml
```diff
@@ -3,43 +3,110 @@
...
+      In other words, in this context a null value really means "unknown", and
+      an unknown value `or` true is always true.
+
+      Behavior for 0 or 1 inputs is as follows:
+        or() -> false
+        or(x) -> x
     impls:
       - args:
-          - value: boolean
+          - value: boolean?
```
jvanstraten, line 29: Note that the added question marks here and elsewhere also aren't necessary, since MIRROR behavior overrides nullability. But I'm fine with it either way.

jvanstraten (approved):
Great; much better than it was, especially if you'd never heard of Kleene logic before.

PR #293, opened 2022-08-17
feat: add regexp_match_substring, regexp_strpos, and regexp_count_sub…

richtia:
PR to add remaining regexp string functions

### DIFF
--- a/extensions/functions_string.yaml
+++ b/extensions/functions_string.yaml
@@ -26,7 +26,7 @@ scalar_functions:
     impls:
       - args:
           - name: case_sensitivity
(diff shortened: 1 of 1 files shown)


jvanstraten (approved):
LGTM

jvanstraten (changes requested):
Sorry for the review change, but as discussed elsewhere, we should probably add `CASE_INSENSITIVE_ASCII` for consistency with the case conversion functions.

richtia:
> Sorry for the review change, but as discussed elsewhere, we should probably add `CASE_INSENSITIVE_ASCII` for consistency with the case conversion functions.

Updated!

jvanstraten (approved):

PR #295, opened 2022-08-22
feat: add population option to variance and standard deviation functions

vibhatha:
In this PR population variance and standard-deviation is added.

BREAKING CHANGE: signature change of `variance` and `std_dev`

### DIFF
--- a/extensions/functions_arithmetic.yaml
+++ b/extensions/functions_arithmetic.yaml
@@ -1040,16 +1040,20 @@ aggregate_functions:
           - name: rounding
             options: [ TIE_TO_EVEN, TIE_AWAY_FROM_ZERO, TRUNCATE, CEILING, FLOOR ]
             required: false
-          - name: x
-            value: fp32
+          - name: distribution
+            options: [ SAMPLE, POPULATION]
+            required: false
+          - value: fp32
         nullability: DECLARED_OUTPUT
         return: fp32?
       - args:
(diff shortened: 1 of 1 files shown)


gforsyth:
hmm, I wonder if `population` vs `sample` should be options to pass to the `variance` and `std-dev` functions (with a default of `sample`).
I don't think they need to be separate functions.

gforsyth:
And I know `postgres` defaults to `population` for things like `variance` but that doesn't make very much sense to me

vibhatha:
@gforsyth I added an option to choose the `distribution`.

jvanstraten:
To be perfectly honest, I have no idea what the difference is, so I have no opinion on the matter. But if you believe that there's no sane default for this, you can also just make the enum required, to force the producer to pick which option they want.

gforsyth:
`sample` is the sane default

jvanstraten (approved):
LGTM then.

vibhatha:
@jvanstraten any changes required?

jvanstraten (approved):
I guess your merge/rebase dismissed my previous approval.

jvanstraten:
Needs rebase due to previous merges.

vibhatha:
@jvanstraten I updated, I think the approval vanishes after pushing.

jvanstraten (approved):

PR #296, opened 2022-08-22
feat: add corr and mode aggregation functions

vibhatha:
In this PR `corr` and `mode` aggregation functions are added.

### DIFF
--- a/extensions/functions_arithmetic.yaml
+++ b/extensions/functions_arithmetic.yaml
@@ -1084,7 +1084,9 @@ aggregate_functions:
         nullability: DECLARED_OUTPUT
         return: fp64?
   - name: "corr"
-    description: Calculates correlation of two set of values.
+    description: >
+      Calculates correlation of two set of values. 
+      If there is no input, null is returned.
     impls:
       - args:
           - name: rounding
--- a/extensions/functions_arithmetic.yaml
+++ b/extensions/functions_arithmetic.yaml
@@ -1083,6 +1083,52 @@ aggregate_functions:
             value: fp64
         nullability: DECLARED_OUTPUT
         return: fp64?
+  - name: "corr"
+    description: Calculates correlation of two set of values.
+    impls:
+      - args:
+          - name: rounding
+            options: [ TIE_TO_EVEN, TIE_AWAY_FROM_ZERO, TRUNCATE, CEILING, FLOOR ]
+            required: false
+          - value: fp32
+          - value: fp32
+        nullability: DECLARED_OUTPUT
+        return: fp32?
+      - args:
+          - name: rounding
+            options: [ TIE_TO_EVEN, TIE_AWAY_FROM_ZERO, TRUNCATE, CEILING, FLOOR ]
+            required: false
+          - value: fp64
+          - value: fp64
+        nullability: DECLARED_OUTPUT
+        return: fp64?
+  - name: "mode"
+    description: Calculates mode for a set of values.
+    impls:
+      - args:
+          - value: i8
+        nullability: DECLARED_OUTPUT
+        return: i8?
+      - args:
+          - value: i16
+        nullability: DECLARED_OUTPUT
+        return: i16?
+      - args:
+          - value: i32
+        nullability: DECLARED_OUTPUT
+        return: i32?
+      - args:
+          - value: i64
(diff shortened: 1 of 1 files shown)


extensions/functions_arithmetic.yaml
```diff
@@ -742,6 +742,52 @@ aggregate_functions:
...
+      - args:
+          - name: rounding
+            options: [ TIE_TO_EVEN, TIE_AWAY_FROM_ZERO, TRUNCATE, CEILING, FLOOR ]
+            required: false
+          - value: fp64
+          - value: fp64
+        nullability: DECLARED_OUTPUT
+        return: fp64?
+  - name: "mode"
+    description: Calculates mode for a set of values.
```
jvanstraten, line 765: I feel like there should be some description of what should happen if the mode is not uniquely defined, for example for [1, 1, 2, 2]. Does that yield 1, 2, either 1 or 2 at the discretion of the consumer, or null?

The mode of a set of floating-point values is also weird, since mode requires a strict equality operator that languages usually shy away from defining for floats, though Substrait doesn't generally consider floats to be special and implicitly assumes an equality operator exists for all types (since there are no specified type constraints on set operations for example). Flipping that around actually, since mode *only* requires an equality operator, shouldn't the mode be defined for all types rather than only on simple numeric types?

extensions/functions_arithmetic.yaml
```diff
@@ -742,6 +742,52 @@ aggregate_functions:
           - value: fp64
         nullability: DECLARED_OUTPUT
         return: fp64?
+  - name: "corr"
+    description: Calculates correlation of two set of values.
```
jvanstraten, line 746: This should also be specifying that the function returns null if there is no input (I'm assuming from context that that's how it works, anyway)

jvanstraten (changes requested):

extensions/functions_arithmetic.yaml
vibhatha: You mean to add it to the description?

extensions/functions_arithmetic.yaml
vibhatha: What do you mean by simple numeric types?

jvanstraten: [Simple](https://substrait.io/types/type_classes/#simple-types) types representing a number, i.e. exactly the types you have now exclusively defined `mode` for. I'm saying the logic of `mode` does not depend on whether a type represents a number or not, so why not just define it for *all* types? For example, the mode of the strings `["a", "b", "c", "b"]` is `"b"`, because that value appears more often than the other values.

extensions/functions_arithmetic.yaml
jvanstraten: The descriptions should document how the functions behave, including corner cases, yes. And likewise for mode, which has even less obvious corner cases.

extensions/functions_arithmetic.yaml
vibhatha: so basically we could use `any` here right?

extensions/functions_arithmetic.yaml
vibhatha: Updated the description.

extensions/functions_arithmetic.yaml
jvanstraten: I think so. It might be a bit weird though nonetheless. Maybe we should just leave it as is for now.

extensions/functions_arithmetic.yaml
```diff
@@ -1071,6 +1071,54 @@ aggregate_functions:
...
+      - args:
+          - name: rounding
+            options: [ TIE_TO_EVEN, TIE_AWAY_FROM_ZERO, TRUNCATE, CEILING, FLOOR ]
+            required: false
+          - value: fp64
+          - value: fp64
+        nullability: DECLARED_OUTPUT
+        return: fp64?
+  - name: "mode"
+    description: Calculates mode for a set of values.
```
jvanstraten, line 1096: Please update for mode as well;

```suggestion
    description: >
      Calculates mode for a set of values.
      If there is no input, null is returned.
```

jvanstraten (changes requested):

jvanstraten (approved):

PR #297, opened 2022-08-23
feat: add string_agg aggregate function

vibhatha:
This PR includes the addition of `string_agg` operator to concatenate string column data with a separator.

### DIFF
--- a/extensions/functions_string.yaml
+++ b/extensions/functions_string.yaml
@@ -1263,3 +1263,18 @@ scalar_functions:
           - value: i32
             name: "count"
         return: "string"
+
+aggregate_functions:
+
+  -
+    name: string_agg
+    description: Concatenates a column of string values with a separator.
+    impls:
+      - args:
+          - value: "varchar<L1>"
+            name: "input"
+            description: "Column of string values."
+          - value: "varchar<L2>"
+            name: "separator"
+            description: "Separator for concatenated strings"
+        return: "varchar<L1 + L2>"
--- a/extensions/functions_string.yaml
+++ b/extensions/functions_string.yaml
@@ -1276,5 +1276,6 @@ aggregate_functions:
             description: "Column of string values."
           - value: "string"
             name: "separator"
+            constant: true
             description: "Separator for concatenated strings"
         return: "string"
--- a/extensions/functions_string.yaml
+++ b/extensions/functions_string.yaml
@@ -1271,10 +1271,10 @@ aggregate_functions:
     description: Concatenates a column of string values with a separator.
     impls:
       - args:
-          - value: "varchar<L1>"
+          - value: "string"
             name: "input"
             description: "Column of string values."
-          - value: "varchar<L2>"
+          - value: "string"
             name: "separator"
             description: "Separator for concatenated strings"
-        return: "varchar<L1 + L2>"
+        return: "string"


extensions/functions_string.yaml
```diff
@@ -403,3 +403,18 @@ scalar_functions:
...
+    name: string_agg
+    description: Concatenates a column of string values with a separator.
+    impls:
+      - args:
+          - value: "varchar<L1>"
+            name: "input"
+            description: "Column of string values."
+          - value: "varchar<L2>"
+            name: "separator"
+            description: "Separator for concatenated strings"
```
jvanstraten, line 419: I guess this should be marked as needing to be constant (`constant: true` IIRC), otherwise someone could pass something that varies from record to record, which makes little sense.

extensions/functions_string.yaml
```diff
@@ -403,3 +403,18 @@ scalar_functions:
...
+        return: "varchar<L1 + L2>"
```
jvanstraten, line 420: I'm not sure where L1 + L2 came from, but it seems rather arbitrary; "the resulting max length is that of the max length of one value and one separator"? `varchar<L1>` would make more sense to me. Also, why define this for `varchar`s and not for `string`s (or for both)?

jvanstraten (changes requested):

extensions/functions_string.yaml
vibhatha: Is it in the spec? I cannot find the definition for this.

extensions/functions_string.yaml
vibhatha: I added string

extensions/functions_string.yaml
jvanstraten: Yes,

> Constant: Whether this argument is required to be a constant for invocation. For example, in some system a regular expression pattern would only be accepted as a literal and not a column value reference.

https://substrait.io/expressions/scalar_functions/#value-argument-properties

That being said... it does appear to be missing from the schema. Huh. #305

extensions/functions_string.yaml
```diff
@@ -649,3 +649,18 @@ scalar_functions:
...
+    description: Concatenates a column of string values with a separator.
+    impls:
+      - args:
+          - value: "string"
+            name: "input"
+            description: "Column of string values."
+          - value: "string"
+            name: "separator"
+            description: "Separator for concatenated strings"
+        return: "string"
```
jvanstraten, line 666: Nit, and not sure why none of the linters are complaining about this, but there should probably be a newline at the end of the file.

jvanstraten (changes requested):
Depends on #305 for constant separator argument.

extensions/functions_string.yaml
```diff
@@ -649,3 +649,18 @@ scalar_functions:
           - value: "string"
           - value: i32
         return: "string"
+
+aggregate_functions:
+
+  -
+    name: string_agg
```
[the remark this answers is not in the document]
vibhatha: Not really, that was the name suggested when we discussed this before.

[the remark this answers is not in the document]
vibhatha: Also one point to mention, this name has been kind of taken by certain industry standards already. I guess that is compelling to keep the name as it is?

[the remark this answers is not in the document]
jvanstraten: Okay, if everyone's calling it `string_agg` already that's good enough for me.

jvanstraten (approved):

jacques-n:
I think this function needs to be updated to declare that it is ordered per property here: 

https://substrait.io/expressions/aggregate_functions/

vibhatha:
> I think this function needs to be updated to declare that it is ordered per property here:
> 
> https://substrait.io/expressions/aggregate_functions/

@jacques-n I am looking into this ...

jvanstraten:
Sorry, forgot to link here; #312

PR #300, opened 2022-08-24
feat: add factorial scalar function

gforsyth:
Adds a scalar function definition for `factorial(n)`

I've opted to only support `i32` and `i64`.

### DIFF
--- a/extensions/functions_arithmetic.yaml
+++ b/extensions/functions_arithmetic.yaml
@@ -469,6 +469,29 @@ scalar_functions:
       - args:
           - value: fp64
         return: fp64
+  -
+    name: "factorial"
+    description: >
+      Return the factorial of a given integer input.
+
+      The factorial of 0! is 1 by convention.
+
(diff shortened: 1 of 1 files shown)


jvanstraten (approved):
LGTM

extensions/functions_arithmetic.yaml
```diff
@@ -469,6 +469,27 @@ scalar_functions:
...
+    name: "factorial"
+    description: >
+      Return the factorial of a given integer input.
+
+      The factorial of 0! is 1 by convention.
+
+      Negative inputs will raise an error.
+    impls:
+      - args:
+          - options: [ SILENT, SATURATE, ERROR ]
```
richtia, line 482: Should we specify the name for the options here?
`- name: overflow`

gforsyth: yes!  nice catch.

jvanstraten (approved):

PR #302, opened 2022-08-25
feat: extend on_domain_error

thisisnic:
This PR extends the possible enum values for the `on_domain_error` option to include `LIMIT_OR_NAN`.

I've submitted this PR due to edge cases for division.  Different systems produce different results for calculating 1/0.  I did a brief investigation and found:

- [The IEEE 754 standard for floating point arithmetic "requires signed 0" and "real arithmetic with signed zeros can be considered a variant of the extended real number line such that 1/−0 = −∞ and 1/+0 = +∞; division is only undefined for ±0/±0 and ±∞/±∞"](https://en.wikipedia.org/wiki/Signed_zero)
- C (via https://cplayground.com/) works as per IEEE 754 (n.b. produces Inf/-Inf for floats only and not ints)
- R returns -Inf and Inf (for both floats and ints)
- DuckDB in a web shell (https://shell.duckdb.org/) returns blank values (NaN?) for both floats and integers
- PostGres - not sure but when I googled it, the results imply that any and all division by zero isn't going to work

Extending `on_domain_error` to include `LIMIT_OR_NAN` was proposed in a separate conversation by @jvanstraten as a possible solution, and in more detail would mean:

* `NAN`: if a domain error occurs, yield NaN.
* `LIMIT_OR_NAN`: if a domain error occurs due to a zero input, evaluate the function as if taking the one-sided limit as that input tends to zero, from the direction of the sign of the zero, and returning divergence to +/-infinity as inf. If the limit is not defined or this rule is not applicable, yield NaN. For example:
1/-0 -> lim x->0- 1/x -> inf
log(+0) -> lim x->0+ log(x) -> -inf
log(-0) -> lim x->0- log(x) -> undefined -> NaN
log(-3) -> limit rule does not apply -> NaN
* `ERROR`: if a domain error occurs, throw an exception.

I imagine that if we go with this approach, we'll need to update the value for `on_domain_error` elsewhere as well, but I wanted to start the discussion here before going through more of the code.

### DIFF
--- a/extensions/functions_arithmetic.yaml
+++ b/extensions/functions_arithmetic.yaml
@@ -160,7 +160,7 @@ scalar_functions:
             options: [ TIE_TO_EVEN, TIE_AWAY_FROM_ZERO, TRUNCATE, CEILING, FLOOR ]
             required: false
           - name: on_domain_error
-            options: [ NAN, ERROR ]
+            options: [ NAN, ERROR, LIMIT_OR_NAN ]
             required: false
           - value: fp32
           - value: fp32
@@ -170,7 +170,7 @@ scalar_functions:
             options: [ TIE_TO_EVEN, TIE_AWAY_FROM_ZERO, TRUNCATE, CEILING, FLOOR ]
             required: false
           - name: on_domain_error
-            options: [ NAN, ERROR ]
+            options: [ NAN, ERROR, LIMIT_OR_NAN ]
             required: false
           - value: fp64
           - value: fp64


jvanstraten:
> ieee754_horrors

Accurate.

Since I basically came up with this particular solution, someone else should review this. @cpcloud @jacques-n ?

I also think we should add a dedicated page to the website with details specific to the core extensions that includes the reasoning for these generalized options and what they mean, rather than repeating this ad nauseum. In my mind that depends on #277 though, so we don't have to rewrite it when that's merged and to avoid unnecessary merge conflicts.

pitrou:
I was asked to chime in here :-)

My own expectation would be that +1/0 always give +Inf (same for the negative counterpart), and that NaN would only be returned for undecidable cases such as 0/0 or Inf/Inf.

I would also expect to be able to ask for 0/0 or Inf/Inf to return an error, while having +1/0 return successfuly and give +Inf.

On the contrary, returning NaN for +1/0 sounds completely unexpected.

jvanstraten:
Thanks for having a look. I think `LIMIT_OR_NAN` agrees with your intuition, except for 0 / 0, where I guess it's ill-defined because in that case two of the arguments are zero. I hadn't considered that one, and naively assumed it would yield +inf like all other positive / +0.0.

Another thing I hadn't considered until now is that you might want the limit behavior but still throw an error if it fails. So, maybe the limit behavior should be a separate option.

In that case, I guess it's better to just make special cases for each function where this might apply, rather than trying to generalize as above.

pitrou:
Yes, I think these should be two separate options:
* the behaviour for `x / 0` where `x != 0`: either output `+-Inf`, or error out
* the behaviour for `0 / 0`, `Inf / Inf`, or any `NaN` input: either output  `NaN`, or error out

pitrou:
Sidenote: I don't understand what "Partial values are truncated" means in the function description.

jvanstraten:
It's from the initial commit, so I guess that was Jacques' way of describing that the integer variants truncate according to the usual rules... though everyone seems to disagree about what should happen for negative integers, so it's not very well-defined. Should probably be an option for that as well. Same for modulus.

By the way... I just checked DuckDB using Python, and there it returns `None`, so presumably that means null. That's very annoying... We can't accommodate that behavior without changing the nullability behavior, which would probably conflict with other engines.

pitrou:
>  though everyone seems to disagree about what should happen for negative integers, so it's not very well-defined

You probably need two different functions then: modulus and remainder :-)

> I just checked DuckDB using Python, and there it returns None, so presumably that means null. 

What exactly returns None?

jvanstraten:
> You probably need two different functions then: modulus and remainder :-)

No, same goes for division. It tends to be either round to zero (C in practice), round to negative infinity (eg Python), or "I don't care" (C spec). Everyone seems to have a different definition for modulo and remainder, too. IMO remainder is the better name, because however you define your division, remainder(x, y) = x - divide(x, y) * y, and that's normally how the `%` operator is defined (whatever you want to call it).

> What exactly returns None?

```python
import duckdb
con = duckdb.connect(database=':memory:')
con.execute("select 1 / 0")
print(con.fetchall())
```

pitrou:
Looks like a bug in DuckDB then...

jvanstraten:
Interestingly, `acos(2)` does give NaN (so it's not a printout problem), while `log(0)` and `log(-1)` throw an exception... Not much consistency there, so I'm gonna +1 on your assessment that this is either a bug or just something that nobody has cared about until now (or both).

pitrou:
Note that `log(0)` could also give `-Inf` ;-)

jvanstraten:
Yeah, that was kinda why I tried defining `LIMIT_OR_NAN` the way I did. It extends naturally to the corner cases, at least for unary functions. But it assumes that engines implement a fundamentally broken representation of real numbers "correctly".

pitrou:
Hmm, if we want to keep the generic wording, perhaps the distinction should be between "domain limit" and "domain error"?
* domain limit: emit infinite or error out
* domain error: emit NaN or error out

(of course, domain errors on integers would always error out?)

jvanstraten:
I think you'd end up with problems for division anyway, so I don't know that it's worth it anymore. It also probably makes more sense to people who're not necessarily used to mathematical definitions if we just special-case all the operations with domain errors with their own wording that makes more sense in context. Something like `ON_DIVISION_BY_ZERO` for division (with a note in the description that it does not apply to 0/0) and `ON_LOG_ZERO` for logarithms, etc. It's not like there's an inordinate amount of common math functions that have domain errors to worry about, anyway. I think it's just divide, sqrt, acos/asin, acosh, atanh, atan2, and the various logarithms. Of those, I think only divide and logarithms can return infinities for finite inputs. Strictly speaking tan could also fail, but (multiples of) pi/2 can never be exactly represented with floats so you can never run into that case, and either way it has no defined limits at its discontinuities.

@thisisnic, any thoughts?

thisisnic:
>  It also probably makes more sense to people who're not necessarily used to mathematical definitions if we just special-case all the operations with domain errors with their own wording that makes more sense in context.

Entirely agree on this; I think the least ambiguity possible in interpreting these things is going to be the best way to go here, and these special cases will achieve that.

thisisnic:
> > It also probably makes more sense to people who're not necessarily used to mathematical definitions if we just special-case all the operations with domain errors with their own wording that makes more sense in context.
> 
> Entirely agree on this; I think the least ambiguity possible in interpreting these things is going to be the best way to go here, and these special cases will achieve that.

If I go ahead and update those, shall I just close this PR and make the changes in a new one, linking back to this one, so it's clear how/why we decided to go that way?

jvanstraten:
Sure, that's fine.

PR #303, opened 2022-08-26
docs: improve descriptions of comparison and rounding functions

gforsyth:
Clarifying some null handling behavior and adding names to all arguments.

### DIFF
--- a/extensions/functions_comparison.yaml
+++ b/extensions/functions_comparison.yaml
@@ -3,59 +3,110 @@
 scalar_functions:
   -
     name: "not_equal"
-    description: Whether two values are not_equal (nulls are considered not not-equal).
+    description: >
+      Whether two values are not_equal.
+
+      `not_equal(x, y) := (x != y)`
+
+      If either/both of `x` and `y` are `null`, `null` is returned.
     impls:
       - args:
           - value: any1
+            name: x
           - value: any1
+            name: y
         return: BOOLEAN
   -
     name: "equal"
-    description: Whether two values are equal (nulls are considered unequal).
+    description: >
+      Whether two values are equal.
+
+      `equal(x, y) := (x == y)`
+
+      If either/both of `x` and `y` are `null`, `null` is returned.
     impls:
       - args:
           - value: any1
+            name: x
           - value: any1
+            name: y
         return: BOOLEAN
   -
     name: "is_not_distinct_from"
-    description: Whether two values are equal (nulls are considered equal).
+    description: >
+      Whether two values are equal.
+
+      This function treats `null` values as comparable, so
+
+      `is_not_distinct_from(null, null) == True`
+
+      This is in contrast to `equal`, in which `null` values do not compare.
     impls:
       - args:
           - value: any1
+            name: x
           - value: any1
+            name: y
         return: BOOLEAN
   -
     name: "lt"
-    description: Less than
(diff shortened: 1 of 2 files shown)


extensions/functions_rounding.yaml
```diff
@@ -3,21 +3,45 @@
...
+
+      Note that the named option `rounding` refers to floating-point rounding
+      behavior and does not effect the behavior of `ceil` itself
     impls:
       - args:
           - value: fp32
+            name: "x"
+          - name: rounding
+            options: [ TIE_TO_EVEN, TIE_AWAY_FROM_ZERO, TRUNCATE, CEILING, FLOOR ]
+            required: false
```
jvanstraten, line 17: Ceil and floor can't ever round; the result can always be exactly represented because it necessarily is one of the function's inputs.

extensions/functions_comparison.yaml
```diff
@@ -3,66 +3,119 @@
 scalar_functions:
   -
     name: "not_equal"
-    description: Whether two values are not_equal (nulls are considered not not-equal).
+    description: >
+      Whether two values are not_equal
```
jvanstraten, line 7: While we're at it, we should probably normalize to complete sentences with a period at the end for the descriptions (see also the other old comparison functions). All the new functions are already doing that, I think.

jvanstraten (changes requested):

jvanstraten (dismissed):
LGTM, thanks!

jvanstraten:
Needs rebase due to previous merges.

gforsyth:
@jvanstraten rebased against `main`

extensions/functions_comparison.yaml
```diff
@@ -89,51 +143,68 @@ scalar_functions:
...
+    description: >
+      Whether a value is finite (neither infinite nor NaN).
+
+      If `x` is `null`, `null` is returned.
     impls:
       - args:
           - value: fp32
-        nullability: MIRROR
+            name: x
+        nullability: DECLARED_OUTPUT
```
jvanstraten, line 174: https://github.com/substrait-io/substrait/pull/303/files#diff-13de9e7b81c09d871b8cc3c28440048686cf37aa76860e44e7155dd3df04e64dL100 (removing `nullability: MIRROR`) is fine because it's the default, but this isn't, because this means that the output will be non-nullable.

extensions/functions_comparison.yaml
```diff
@@ -89,51 +143,68 @@ scalar_functions:
...
   -
     name: "nullif"
     description: If two values are equal, return null. Otherwise, return the first value.
     impls:
       - args:
           - value: any1
+            name: x
           - value: any1
+            name: y
+        nullability: DECLARED_OUTPUT
```
jvanstraten, line 207: This doesn't seem right either, can't return null as per the description if both inputs are non-nullable and equal.

jvanstraten (changes requested):
Looks like some nullability behavior things got messed up in the rebase/merge with #285.

gforsyth:
Nothing like a bad rebase to make Friday morning interesting.  Thanks for catching those @jvanstraten -- I _think_ I've got them fixed up correctly now

jvanstraten (approved):
Keeps me sharp as well :) LGTM now, thanks!

PR #305, opened 2022-08-29
feat: support constant function arguments

jvanstraten:
Constant value arguments were already part of the specification, but were missing from the simple extension schema.


cpcloud (approved):

jacques-n:
Looking at this change it looks like we have another change we should do. It looks like we're allowing arbitrary options on arguments? (I didn't review the schema, just saw that it didn't choke on yes.)

PR #312, opened 2022-09-05
fix: mark string_agg aggregate as being sensitive to input order

jvanstraten:


jacques-n (approved):
Thanks @jvanstraten !

PR #316, opened 2022-09-08
fix(naming): add missing arg names in functions_aggregate_*.yaml

vibhatha:
This PR adds the missing argument names in `functions_aggregate_approx.yaml` and `functions_aggregate_generic.yaml`.

### DIFF
--- a/extensions/functions_aggregate_generic.yaml
+++ b/extensions/functions_aggregate_generic.yaml
@@ -17,7 +17,8 @@ aggregate_functions:
     description: "Count a set of records (not field referenced)"
     impls:
       - args:
-          - options: [SILENT, SATURATE, ERROR]
+          - name: x
+            options: [SILENT, SATURATE, ERROR]
             required: false
         nullability: DECLARED_OUTPUT
         decomposable: MANY
--- a/extensions/functions_aggregate_generic.yaml
+++ b/extensions/functions_aggregate_generic.yaml
@@ -5,7 +5,8 @@ aggregate_functions:
     description: Count a set of values
     impls:
       - args:
-          - options: [SILENT, SATURATE, ERROR]
+          - name: overflow
+            options: [SILENT, SATURATE, ERROR]
             required: false
           - name: x
             value: any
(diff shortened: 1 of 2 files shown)


jvanstraten:
The sole argument there would just be named `overflow`. The intended behavior of the function is to count all records when aggregating (so there's no need to reference a field) if that's what you were wondering about, but I don't really like that special case either if that's what you mean... nor do I like those functions having the same name, which is something I hadn't really considered as an option yet. But I guess that's legal because the compound names are different, so... more work for the validator :/

vibhatha:
> The sole argument there would just be named `overflow`. The intended behavior of the function is to count all records when aggregating (so there's no need to reference a field) if that's what you were wondering about, but I don't really like that special case either if that's what you mean... nor do I like those functions having the same name, which is something I hadn't really considered as an option yet. But I guess that's legal because the compound names are different, so... more work for the validator :/

So we could have the following for the unnamed count function?

```YAML
- name: "count"
    description: "Count a set of records (not field referenced)"
    impls:
      - args:
          - name: overflow
          - options: [SILENT, SATURATE, ERROR]
            required: false
        nullability: DECLARED_OUTPUT
        decomposable: MANY
        intermediate: i64
        return: i64
```

jvanstraten:
Uh, more or less, but without the `-` before `options`.

vibhatha:
@jvanstraten updated the PR.

extensions/functions_aggregate_generic.yaml
```diff
@@ -16,7 +17,8 @@ aggregate_functions:
     description: "Count a set of records (not field referenced)"
     impls:
       - args:
-          - options: [SILENT, SATURATE, ERROR]
+          - name: x
```
jvanstraten, line 20: In this case, it should actually be named `overflow`.

```suggestion
          - name: overflow
```

extensions/functions_aggregate_generic.yaml
```diff
@@ -7,7 +7,8 @@ aggregate_functions:
       - args:
           - options: [SILENT, SATURATE, ERROR]
```
jvanstraten, line 8: The `overflow` name is also missing here.

jvanstraten (changes requested):

vibhatha:
@jvanstraten updated the PR.

jvanstraten (approved):

PR #317, opened 2022-09-08
fix(naming): add/replace arg names in functions_boolean.yaml

vibhatha:
This PR adds missing arg names in the `functions_boolean.yaml`. I also replaced the name `x` to `a` to keep the naming consistency in the file. I have missed this when making the previous PR on `bool_or` and `bool_and`.


jvanstraten (approved):
I don't think the letter really matters for these simple arguments, to be honest. As long as it can be referred to in a description or when you're talking about it it's fine.

PR #319, opened 2022-09-08
fix: Add missing arg names in logarithmic functions

vibhatha:
This PR adds the missing argument names in `functions_logarithmic.yaml`.


vibhatha:
cc @jvanstraten is `name: x` is the right way or `name: "x"` ? I see these two patterns in the file edited in this PR.

jvanstraten (approved):
There's no difference, both parse to a string in YAML. I don't really have a strong preference; I prefer the cleanliness of not using quotes, but then you run the risk of clashing with names that are special-cased in YAML.

PR #321, opened 2022-09-08
feat: add any_value aggregate function

vibhatha:
This PR  adds `any_value` aggregate function.

### DIFF
--- a/extensions/functions_arithmetic.yaml
+++ b/extensions/functions_arithmetic.yaml
@@ -1378,6 +1378,52 @@ aggregate_functions:
         nullability: DECLARED_OUTPUT
         ordered: true
         return: LIST?<any>
+  
+  - name: "any_value"
+    description: >
+      Selects an arbitrary value from a group of values.
+
+      This function returns a single arbitrary value from an expression.
+      Main purpose of this function is to optimize a query containing a GROUP BY
+      caluse.
+
+      A comparison is not performed when this function is applied, hence it
+      is not necessary to be included in the GROUP BY clause.
+
+      This function returns NULL when expression produces NULL.
+      When the expression is NULL, it returns NULL for all values in the group.
+
+    impls:
+      - args:
+          - value: i8
+            name: x
+      nullability: DECLARED_OUTPUT
+      return: i8?
+      - args:
+          - value: i16
+            name: x
+      nullability: DECLARED_OUTPUT
+      return: i16?
+      - args:
+          - value: i32
+            name: x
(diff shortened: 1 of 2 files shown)


extensions/functions_arithmetic.yaml
```diff
@@ -1352,6 +1352,51 @@ aggregate_functions:
...
+  - name: "any_value"
+    description: >
+      Selects an arbitrary value from a group of values.
+
+      This function returns a single arbitrary value from an expression.
+      Main purpose of this function is to optimize a query containing a GROUP BY
+      caluse.
+
+      A comparison is not performed when this function is applied, hence it
+      is not necessary to be included in the GROUP BY clause.
```
jvanstraten, line 1365: I don't think this applies to Substrait, so I think it can be removed.

extensions/functions_arithmetic.yaml
```diff
@@ -1352,6 +1352,51 @@ aggregate_functions:
...
+
+      This function returns NULL when expression produces NULL.
+      When the expression is NULL, it returns NULL for all values in the group.
```
jvanstraten, line 1368: I'm not sure what's going on here. I would expect something like "If the input is empty, this returns null." and that's it, I don't see any special cases beyond that.

vibhatha:
@jvanstraten I updated the PR.

jvanstraten:
Sorry, I don't know why I didn't notice this before, but there's no reason for this to be limited to numeric types. It can just be `any_value(T) -> T`. It should also be in the file for generic aggregates.

vibhatha:
@jvanstraten updated the PR.

extensions/functions_arithmetic.yaml
```diff
@@ -1368,6 +1368,43 @@ aggregate_functions:
...
+      - args:
+          - name: x
+            value: fp32
+        nullability: DECLARED_OUTPUT
+        return: fp32?
+      - args:
+          - name: x
+            value: fp64
+        nullability: DECLARED_OUTPUT
+        return: fp64?
```
jvanstraten, line 1407: This is completely covered by the `f(any) -> any?` variant, and it isn't really an arithmetic function to begin with. So why is this still here?

jvanstraten (changes requested):

vibhatha:
@jvanstraten updated the PR. Is it okay now?

jvanstraten (approved):
Yeah, LGTM now, I think.

Issue #307, opened 2022-09-02
Establish intended scope of functions in `extensions/*.yaml`


ianmcook:
Some folks at Voltron Data have been working to add more functions in the [extensions YAML](https://github.com/substrait-io/substrait/tree/main/extensions). Along the way, we have gained a clearer understanding of when a function should be considered in scope versus out of scope for Substrait. Here are some guidelines that have emerged:

- Avoid adding a new function if is is merely syntactic sugar and if producers could express its behavior easily enough by composing other existing functions. However, it might be acceptable to add such a function if adding it can enable consumers to more efficiently execute plans, or if adding it can avoid producers deconstructing a syntactic construct that consumers then need to reconstruct.
  - Example: https://github.com/substrait-io/substrait/pull/287#discussion_r942705485
- Avoid adding functions that express behaviors that are already expressible with [specialized record expressions].(https://substrait.io/expressions/specialized_record_expressions/)
  - Example: the `if_else` function originally proposed in #291
- Prefer adding new options to existing functions instead of adding new functions.
  - Example: #289
  - Example: #295
- Aim for syntactic and semantic consistency with widely used SQL dialects, especially PostgreSQL.
  - Example: https://github.com/substrait-io/substrait/pull/285#discussion_r944542030

Are there any other important guidelines I'm missing here?

Should we document this somewhere?

ianmcook:
To share some further context about how the folks at Voltron Data chose which functions to add:

We went through the functions documentation for various SQL engines and other data processing libraries (PostgreSQL, DuckDB, MySQL, Hive, Impala, Acero, NumPy/pandas, R/dplyr, and more) and made a big list of the functions that exist in many of these systems. We then applied some subjective judgements based on our experience to make a few additions and removals, and set the names to (mostly) consistent with what things are named in PostgreSQL.

Here is the list that come out of that. Note that the categories in this list are more fine-grained than the categories used to organize these functions into YAML files. A checked box indicates that the function was deemed to be in scope. An unchecked box indicates that the function was deemed to be out of scope for the stated reason.

## Arithmetic functions
- [x] `abs`
- [x] `add`
- [x] `divide`
- [x] `factorial`
- [x] `modulus`
- [x] `multiply`
- [x] `negate`
- [x] `power`
- [x] `sign`
- [x] `sqrt`
- [x] `subtract`

## Rounding functions
- [x] `ceil`
- [x] `floor`
- [x] `round`
[...]

PR #322, opened 2022-09-08
feat: add round function

vibhatha:
This PR adds `round` function.

### DIFF
--- a/extensions/functions_rounding.yaml
+++ b/extensions/functions_rounding.yaml
@@ -27,3 +27,66 @@ scalar_functions:
           - value: fp64
             name: "x"
         return: fp64
+  -
+    name: "round"
+    description: >
+      Rounding the value `x` to s decimal places.
+    impls:
+      - args:
+          - name: rounding
+            options: [ TIE_TO_EVEN, TIE_AWAY_FROM_ZERO, TRUNCATE, CEILING, FLOOR, 
+              AWAY_FROM_ZERO, TIE_DOWN, TIE_UP, TIE_TOWARDS_ZERO, TIE_TO_ODD ]
+          - value: fp32
+            name: "x"
+            description: >
+              Numerical expression to be rounded.
+          - value: i32
+            name: "s"
+            description: >
+              Number of decimal places to be rounded to.
+
+              When `s` is a positive number, the rounding
+              is performed to a `s` number of decimal places.
+
+              When `s` is a negative number, the rounding is
+              performed to the left side of the decimal point
+              as specified by `s`.
+          - value: i32
+            name: "function"
+            description: >
+              This is an optional argument `function`, a numeric
+              value determines whether to perform round or truncate.
+              When the `function` value is set to 0, the function
+              performs rounding to `s` decimal places, otherwise it
+              performs truncate. The default value is 0.
+        nullability: DECLARED_OUTPUT
+        return: fp32?
+      - args:
+          - name: rounding
+            options: [ TIE_TO_EVEN, TIE_AWAY_FROM_ZERO, TRUNCATE, CEILING, FLOOR, 
+              AWAY_FROM_ZERO, TIE_DOWN, TIE_UP, TIE_TOWARDS_ZERO, TIE_TO_ODD ]
+          - value: fp64
+            name: "x"
+            description: >
+              Numerical expression to be rounded.
+          - value: i32
+            name: "s"
+            description: >
+              Number of decimal places to be rounded to.
+
+              When `s` is a positive number, the rounding
+              is performed to a `s` number of decimal places.
+
+              When `s` is a negative number, the rounding is
+              performed to the left side of the decimal point
+              as specified by `s`.
+          - value: i32
+            name: "function"
+            description: >
+              This is an optional argument `function`, a numeric
+              value determines whether to perform round or truncate.
+              When the `function` value is set to 0, the function
+              performs rounding to `s` decimal places, otherwise it
+              performs truncate. The default value is 0.
+        nullability: DECLARED_OUTPUT
+        return: fp64?
--- a/extensions/functions_rounding.yaml
+++ b/extensions/functions_rounding.yaml
@@ -33,8 +33,9 @@ scalar_functions:
       Rounding the value `x` to s decimal places.
     impls:
       - args:
-          - name: rounding
-            options: [ TIE_TO_EVEN, TIE_AWAY_FROM_ZERO, TRUNCATE, CEILING, FLOOR, 
+        options:
+          rounding:
+            values: [ TIE_TO_EVEN, TIE_AWAY_FROM_ZERO, TRUNCATE, CEILING, FLOOR,
               AWAY_FROM_ZERO, TIE_DOWN, TIE_UP, TIE_TOWARDS_ZERO, TIE_TO_ODD ]
           - value: fp32
             name: "x"
@@ -62,8 +63,9 @@ scalar_functions:
         nullability: DECLARED_OUTPUT
         return: fp32?
       - args:
-          - name: rounding
-            options: [ TIE_TO_EVEN, TIE_AWAY_FROM_ZERO, TRUNCATE, CEILING, FLOOR, 
+        options:
+          rounding:
+            values: [ TIE_TO_EVEN, TIE_AWAY_FROM_ZERO, TRUNCATE, CEILING, FLOOR,
               AWAY_FROM_ZERO, TIE_DOWN, TIE_UP, TIE_TOWARDS_ZERO, TIE_TO_ODD ]
           - value: fp64
             name: "x"
--- a/extensions/functions_rounding.yaml
+++ b/extensions/functions_rounding.yaml
@@ -33,6 +33,166 @@ scalar_functions:
       Rounding the value `x` to `s` decimal places.
     impls:
       - args:
+          - value: i8
+            name: "x"
+            description: >
+              Numerical expression to be rounded.
+          - value: i32
+            name: "s"
+            description: >
+              Number of decimal places to be rounded to.
+
+              When `s` is a positive number, the rounding
+              is performed to a `s` number of decimal places.
+
+              When `s` is a negative number, the rounding is
+              performed to the left side of the decimal point
+              as specified by `s`.
+        options:
+          rounding:
+            description: >
+              When a boundary is computed to lie somewhere between two values,
+              and this value cannot be exactly represented, this specifies how
+              to round it.
+
+                - TIE_TO_EVEN: round to nearest value; if exactly halfway, tie
+                  to the even option.
+                - TIE_AWAY_FROM_ZERO: round to nearest value; if exactly
+                  halfway, tie away from zero.
+                - TRUNCATE: always round toward zero.
+                - CEILING: always round toward positive infinity.
+                - FLOOR: always round toward negative infinity.
+                - AWAY_FROM_ZERO: round negative values with FLOOR rule, round positive values with CEILING rule
+                - TIE_DOWN: round ties with FLOOR rule
+                - TIE_UP: round ties with CEILING rule
+                - TIE_TOWARDS_ZERO: round ties with TRUNCATE rule
+                - TIE_TO_ODD: round to nearest value; if exactly halfway, tie
+                  to the odd option.
+            values: [ TIE_TO_EVEN, TIE_AWAY_FROM_ZERO, TRUNCATE, CEILING, FLOOR,
+              AWAY_FROM_ZERO, TIE_DOWN, TIE_UP, TIE_TOWARDS_ZERO, TIE_TO_ODD ]
+        nullability: DECLARED_OUTPUT
+        return: i8?
+      - args:
+          - value: i16
+            name: "x"
+            description: >
+              Numerical expression to be rounded.
+          - value: i32
+            name: "s"
+            description: >
+              Number of decimal places to be rounded to.
+
+              When `s` is a positive number, the rounding
+              is performed to a `s` number of decimal places.
+
+              When `s` is a negative number, the rounding is
+              performed to the left side of the decimal point
+              as specified by `s`.
+        options:
+          rounding:
+            description: >
+              When a boundary is computed to lie somewhere between two values,
+              and this value cannot be exactly represented, this specifies how
+              to round it.
+
+                - TIE_TO_EVEN: round to nearest value; if exactly halfway, tie
+                  to the even option.
+                - TIE_AWAY_FROM_ZERO: round to nearest value; if exactly
+                  halfway, tie away from zero.
+                - TRUNCATE: always round toward zero.
+                - CEILING: always round toward positive infinity.
+                - FLOOR: always round toward negative infinity.
+                - AWAY_FROM_ZERO: round negative values with FLOOR rule, round positive values with CEILING rule
+                - TIE_DOWN: round ties with FLOOR rule
+                - TIE_UP: round ties with CEILING rule
+                - TIE_TOWARDS_ZERO: round ties with TRUNCATE rule
+                - TIE_TO_ODD: round to nearest value; if exactly halfway, tie
+                  to the odd option.
+            values: [ TIE_TO_EVEN, TIE_AWAY_FROM_ZERO, TRUNCATE, CEILING, FLOOR,
+              AWAY_FROM_ZERO, TIE_DOWN, TIE_UP, TIE_TOWARDS_ZERO, TIE_TO_ODD ]
+        nullability: DECLARED_OUTPUT
+        return: i16?
+      - args:
+          - value: i32
+            name: "x"
+            description: >
+              Numerical expression to be rounded.
+          - value: i32
+            name: "s"
+            description: >
+              Number of decimal places to be rounded to.
+
+              When `s` is a positive number, the rounding
+              is performed to a `s` number of decimal places.
+
+              When `s` is a negative number, the rounding is
+              performed to the left side of the decimal point
+              as specified by `s`.
+        options:
+          rounding:
+            description: >
+              When a boundary is computed to lie somewhere between two values,
+              and this value cannot be exactly represented, this specifies how
+              to round it.
+
+                - TIE_TO_EVEN: round to nearest value; if exactly halfway, tie
+                  to the even option.
+                - TIE_AWAY_FROM_ZERO: round to nearest value; if exactly
+                  halfway, tie away from zero.
+                - TRUNCATE: always round toward zero.
+                - CEILING: always round toward positive infinity.
+                - FLOOR: always round toward negative infinity.
+                - AWAY_FROM_ZERO: round negative values with FLOOR rule, round positive values with CEILING rule
+                - TIE_DOWN: round ties with FLOOR rule
+                - TIE_UP: round ties with CEILING rule
+                - TIE_TOWARDS_ZERO: round ties with TRUNCATE rule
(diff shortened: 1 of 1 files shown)


jvanstraten:
Where did this come from? Based on #307 I thought we weren't going to do rounding to multiple. Rounding to some number of digits is exactly the same thing, just restricted to powers of 10.

That rounding argument is also really weird, there's no such thing as an optional value argument and using an integer for this is really weird. If we *are* going to do a round-to-multiple I feel like we should just reuse the existing rounding mode option to determine how to round.

Based on #307 I would expect simply `round(f32) -> f32` and `trunc(f32) -> f32` (and likewise for `f64`).

vibhatha:
> Where did this come from? Based on #307 I thought we weren't going to do rounding to multiple. Rounding to some number of digits is exactly the same thing, just restricted to powers of 10.
> 
> That rounding argument is also really weird, there's no such thing as an optional value argument and using an integer for this is really weird. If we _are_ going to do a round-to-multiple I feel like we should just reuse the existing rounding mode option to determine how to round.
> 
> Based on #307 I would expect simply `round(f32) -> f32` and `trunc(f32) -> f32` (and likewise for `f64`).

Referring to: https://github.com/substrait-io/substrait/pull/230/files#diff-b6317d19776a0381379a5980e1be0ffcde22082a034b891cb7a072fdbab32327

vibhatha:
@jvanstraten I was referring to the SQL definition [here](https://docs.microsoft.com/en-us/sql/t-sql/functions/round-transact-sql?view=sql-server-ver16).

I see your point on `function` argument. In Duckdb there is no such parameter. May be I will remove that option and also add the rounding options here too. 

WDYT? cc @ianmcook

ianmcook:
IIUC, I think we can just implement an rounding mode option, and make `TRUNCATE` one of the available modes, alongside the others like `DOWN`, `UP`, `TOWARDS_ZERO`, etc. Does that sound right to you @jvanstraten ?

jvanstraten:
I figured we could just reuse the floating point rounding mode. I also generalized it to integers for quantile earlier. Wouldn't `TRUNCATE` and `TOWARDS_ZERO` be the same thing, anyway?

Note that dedicated ceil and floor functions already exist. I guess we should get rid of those if we go the option route.

ianmcook:
Ah, yes, I think that all makes sense.

Since a lot of thought went into ensuring that [Acero's `RoundMode`](https://arrow.apache.org/docs/cpp/api/compute.html#_CPPv49RoundMode) can represent the behavior of any other mainstream system's round function, I'll list out the options it provides and map them onto the options used in the Substrait functions that take a `rounding` argument. To see an example showing the behavior of these different Acero `RoundMode`, look at [this test](https://github.com/apache/arrow/blob/dd0988b49cb6726cf915bb9f53d7320e3a97b00b/python/pyarrow/tests/test_compute.py#L1509-L1521).

- `DOWN` - corresponds to Substrait's `FLOOR`
- `UP` - corresponds to Substrait's `CEILING`
- `TOWARDS_ZERO` - corresponds to Substrait's `TRUNCATE`
- `TOWARDS_INFINITY` - no corresponding option in Substrait
- `HALF_DOWN` -  no corresponding option in Substrait
- `HALF_UP` - no corresponding option in Substrait
- `HALF_TOWARDS_ZERO` - no corresponding option in Substrait
- `HALF_TOWARDS_INFINITY` - corresponds to Substrait's `TIE_AWAY_FROM_ZERO`
- `HALF_TO_EVEN` - corresponds to Substrait's `TIE_TO_EVEN` 
- `HALF_TO_ODD` - no corresponding option in Substrait

So this raises some questions:
- Should we add a Substrait option `AWAY_FROM_ZERO`? This corresponds to Acero's `TOWARDS_INFINITY `.
- Should we add a Substrait option `TIE_TOWARDS_NEGATIVE_INFINITY `? This corresponds to Acero's `HALF_DOWN`.
- Should we add a Substrait option `TIE_TOWARDS_POSITIVE_INFINITY `? This corresponds to Acero's `HALF_UP`.
- Should we add a Substrait option `TIE_TOWARDS_ZERO`? This corresponds to Acero's `HALF_TOWARDS_ZERO`.
- Should we add a Substrait option `TIE_TO_ODD`? This corresponds to Acero's `HALF_TO_ODD`.

jvanstraten:
Hm. I guess that makes sense for round, and adding more variants is cheap. I don't think it makes sense to do it for all floating point functions though, since IEEE is pretty explicit about what the possible rounding modes are.

vibhatha:
cc @jvanstraten I added the round options. Does it reflect what you expected?

jvanstraten:
No, because you didn't add the rounding modes @ianmcook was talking about, and the supposedly-optional `function` argument still exists.

vibhatha:
@jvanstraten I added other options. Should we create documentation for this too? In this PR itself or a separate PR?

jvanstraten:
It should just be in the description of the function in this case; this option argument is basically specific to `round`. The rounding mode should probably be a required argument, too, since it makes little sense for a producer to allow a consumer to round however it wants for a function that is specifically about rounding (unlike the usual floating point rounding mode).

vibhatha:
@jvanstraten I had the doubt though. I will update it.

westonpace (changes requested):
@vibhatha when you get a moment can you update this to the new format for optional arguments?

vibhatha:
> @vibhatha when you get a moment can you update this to the new format for optional arguments?

sure @westonpace, thanks for the ping. I will update it.

extensions/functions_rounding.yaml
```diff
@@ -27,3 +27,68 @@ scalar_functions:
...
+  -
+    name: "round"
+    description: >
+      Rounding the value `x` to s decimal places.
+
+    impls:
+      - args:
+          - name: rounding
+            options: [ TIE_TO_EVEN, TIE_AWAY_FROM_ZERO, TRUNCATE, CEILING, FLOOR, AWAY_FROM_ZERO,
+              TIE_DOWN, TIE_UP, TIE_TOWARDS_ZERO, TIE_TOWARDS_ZERO, TIE_TO_ODD ]
```
EpsilonPrime, line 39: TIE_TOWARDS_ZERO is listed twice on this line.

extensions/functions_rounding.yaml
```diff
@@ -27,3 +27,68 @@ scalar_functions:
...
+              This is an optional argument `function`, a numeric
+              value determines whether to perform round or truncate.
+              When the `function` value is set to 0, the function
+              performs rounding to `s` decimal places, otherwise it
+              performs truncate. The default value is 0.
+        nullability: DECLARED_OUTPUT
+        return: fp32?
+      - args:
+          - name: rounding
+            options: [ TIE_TO_EVEN, TIE_AWAY_FROM_ZERO, TRUNCATE, CEILING, FLOOR ]
```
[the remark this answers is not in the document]
vibhatha: no that is a typo.

extensions/functions_rounding.yaml
vibhatha: Thanks for noticing this, it is a typo.

extensions/functions_rounding.yaml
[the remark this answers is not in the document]
vibhatha: @EpsilonPrime  I updated the PR. Please take a look.

extensions/functions_rounding.yaml
```diff
@@ -27,3 +27,68 @@ scalar_functions:
...
+              performed to the left side of the decimal point
+              as specified by `s`.
+          - value: i32
+            name: "function"
+            description: >
+              This is an optional argument `function`, a numeric
+              value determines whether to perform round or truncate.
+              When the `function` value is set to 0, the function
+              performs rounding to `s` decimal places, otherwise it
+              performs truncate. The default value is 0.
```
westonpace, line 58: Why isn't this an enum?

vibhatha: @westonpace I think roughly I get your point. But could you please give me an example here, I don't think my understanding is very precise.

westonpace (changes requested):
Sorry, my previous comments were terse and I will explain in more detail:

There are currently three value arguments, `x`, `s`, and `function`.  No engine I am aware of implements a 3-arg round function.  I have a few concerns:

1. `function` is redundant and should be a bool or an enum instead of an `i32`.  It currently is used to determine whether we are rounding or truncating.  However, "truncate" is (as far as I know) equivalent to "round" with `rounding = FLOOR`.

So I think we can get rid of `function`.  If I am misunderstanding it, and we choose to keep it, then it should not be `i32` but something like:

```
value: bool
name: "should_truncate"
description: >
  If true perform truncation.  If false then round to `s` decimal places.  The default value is false.
```

2. Should `s` be an option or a value argument?

In other words, is `round` a unary function with an option telling it how much to round to?  Or is `round` a binary function?

My initial thought was "unary function" since that is how it is implemented in Acero.  However, looking further, it appears that other engines have implemented this as a binary function.  So I think binary function is the correct answer and this comment is no longer valid.

Let me know how you'd like to proceed on question 1.

westonpace:
Oops, one minor clarification. I said `function != 0` is equivalent to `rounding = FLOOR` but I meant `function != 0` is equivalent to `rounding = TRUNCATE`.

vibhatha:
> Sorry, my previous comments were terse and I will explain in more detail:
> 
> There are currently three value arguments, `x`, `s`, and `function`. No engine I am aware of implements a 3-arg round function. I have a few concerns:
> 
> 1. `function` is redundant and should be a bool or an enum instead of an `i32`.  It currently is used to determine whether we are rounding or truncating.  However, "truncate" is (as far as I know) equivalent to "round" with `rounding = FLOOR`.
> 
> **So I think we can get rid of `function`.** If I am misunderstanding it, and we choose to keep it, then it should not be `i32` but something like:
> 
> ```
> value: bool
> name: "should_truncate"
> description: >
>   If true perform truncation.  If false then round to `s` decimal places.  The default value is false.
> ```
> 
>  2. Should `s` be an option or a value argument?
> 
> In other words, is `round` a unary function with an option telling it how much to round to? Or is `round` a binary function?
> 
> My initial thought was "unary function" since that is how it is implemented in Acero. However, looking further, it appears that other engines have implemented this as a binary function. So I think binary function is the correct answer and this comment is no longer valid.
> 
> Let me know how you'd like to proceed on question 1.

@westonpace looking into the new options added (after addressing reviews), I think your comment on leaving functions from this definition would be better. Because the `truncate`  and `floor` are already defined as an option. 

But keeping `s` as optional, I am not so sure, we may need to include it, referring to [snowflake round function](https://docs.snowflake.com/en/sql-reference/functions/round.html).

extensions/functions_rounding.yaml
```diff
@@ -27,3 +27,52 @@ scalar_functions:
           - value: fp64
             name: "x"
         return: fp64
+  -
+    name: "round"
+    description: >
+      Rounding the value `x` to s decimal places.
```
westonpace, line 33:
```suggestion
      Rounding the value `x` to `s` decimal places.
```
This isn't markdown so I don't know if there is value in using backticks like this.  Do we do this elsewhere?  If we are using backticks consistently then `s` should also be quoted.

westonpace (changes requested):
I agree that `s` needs to remain a second value argument given other engines work this way.  There are a few unique rounding options here that are not part of the IEEE rounding options.  These are `AWAY_FROM_ZERO`, `TIE_DOWN`, `TIE_UP`, `TIE_TOWARDS_ZERO`, `TIE_TO_ODD`.

Can we describe what these do somewhere?

vibhatha:
> I agree that `s` needs to remain a second value argument given other engines work this way. There are a few unique rounding options here that are not part of the IEEE rounding options. These are `AWAY_FROM_ZERO`, `TIE_DOWN`, `TIE_UP`, `TIE_TOWARDS_ZERO`, `TIE_TO_ODD`.
> 
> Can we describe what these do somewhere?

Do you propose a documentation update or describe here itself in the function spec?

westonpace:
> Do you propose a documentation update or describe here itself in the function spec?

In the function spec

EpsilonPrime:
I like the description of the rounding options found in functions_arithmetic.yaml.  Perhaps we can utilize those here?

extensions/functions_rounding.yaml
```diff
@@ -27,3 +27,92 @@ scalar_functions:
...
+                  halfway, tie away from zero.
+                - TRUNCATE: always round toward zero.
+                - CEILING: always round toward positive infinity.
+                - FLOOR: always round toward negative infinity.
+                - AWAY_FROM_ZERO: round negative values with FLOOR rule, round positive values with CEILING rule
+                - TIE_DOWN: round ties with FLOOR rule
+                - TIE_UP: round ties with CEILING rule
+                - TIE_TOWARDS_ZERO: round ties with TRUNCATE rule
+                - TIE_TO_ODD: round to nearest value; if exactly halfway, tie
+                  to the odd option.
```
vibhatha, line 72: @westonpace @EpsilonPrime I added the spec here. And filled the missing fields. 

I copied the textual description 1:1 from the arithmetic PR. But I am not sure if we should state it 1:1 or describe differently. 

Maybe we need to update the arithmeitc rounding options too?

extensions/functions_rounding.yaml
```diff
@@ -27,3 +27,92 @@ scalar_functions:
...
+              performed to the left side of the decimal point
+              as specified by `s`.
+        options:
+          rounding:
+            description: >
+              When a boundary is computed to lie somewhere between two values,
+              and this value cannot be exactly represented, this specifies how
+              to round it. For floating point numbers, it specifies the IEEE
+              754 rounding mode (as it does for all other floating point
+              operations). For integer types:
```
westonpace, line 58:
```suggestion
              to round it:
```

westonpace: These options are a superset of the IEEE 754 options.  Also, I'm not sure what you mean by "For integer types" since this function only applies to floating point types.  I think it's just best if we remove this part altogether.

westonpace (changes requested):
Do we want implementations for integers?  It could be useful when `s` is negative.

extensions/functions_rounding.yaml
```diff
@@ -27,3 +27,92 @@ scalar_functions:
           - value: fp64
             name: "x"
         return: fp64
+  -
+    name: "round"
+    description: >
+      Rounding the value `x` to `s` decimal places.
+    impls:
+      - args:
+          - value: fp32
```
EpsilonPrime, line 36: Are there other pairs of argument types that we want to support?  I could see fp64 and i8 as potential options here.  With the ability to round anywhere in the number we may also want to include integer types for x.

westonpace:
@vibhatha can you update this when you get a chance?

vibhatha:
@westonpace will update it tomorrow. Sorry about the delay.

extensions/functions_rounding.yaml
vibhatha: Not entirely sure about `i8`, although with s being `i32` we have already added `fp32` and `fp64`. 
cc @westonpace

vibhatha:
@westonpace sorry about the long wait. I updated the PR, although I see `branch is out-of-date with base branch`. I have rebased and did `--force` push.

extensions/functions_rounding.yaml
westonpace: You removed this test below (in the fp64/i32 kernel).  Can you also remove it here (in the fp32/i32 kernel)

extensions/functions_rounding.yaml
westonpace: For the moment let's assume it is safe to mandate that `s` be `i32`.  If we need to add additional kernels for `s` later then we can do so without much strife I hope.

However, we should allow `x` to be `i8, i16, i32, i64`.  So please add the additional 4 kernels.

westonpace (changes requested):
Thanks for the update.  It still says out of date but a rebase should have fixed that.  Did you make sure you had updated your local `main` branch before rebasing?

vibhatha:
> Thanks for the update. It still says out of date but a rebase should have fixed that. Did you make sure you had updated your local `main` branch before rebasing?

`main` and `round_func` branches are rebased with the `upstream/main`. Not sure what is going on here?

vibhatha:
@westonpace I think now the branch is in sync with the base branch.

extensions/functions_rounding.yaml
```diff
@@ -27,3 +27,248 @@ scalar_functions:
...
+            name: "x"
+            description: >
+              Numerical expression to be rounded.
+          - value: i32
+            name: "s"
+            description: >
+              Number of decimal places to be rounded to.
+
+              When `s` is a positive number, the rounding
+              is performed to a `s` number of decimal places.
```
westonpace, line 46:
```suggestion
              When `s` is a positive number, nothing will happen
              since `x` is an integer value.
```

extensions/functions_rounding.yaml
```diff
@@ -27,3 +27,248 @@ scalar_functions:
...
+
+              When `s` is a negative number, the rounding is
+              performed to the left side of the decimal point
+              as specified by `s`.
```
westonpace, line 50:
```suggestion
              When `s` is a negative number, the rounding is
              performed to the nearest multiple of `10^(-s)`.
```

extensions/functions_rounding.yaml
```diff
@@ -27,3 +27,248 @@ scalar_functions:
...
+                  halfway, tie away from zero.
+                - TRUNCATE: always round toward zero.
+                - CEILING: always round toward positive infinity.
+                - FLOOR: always round toward negative infinity.
+                - AWAY_FROM_ZERO: round negative values with FLOOR rule, round positive values with CEILING rule
+                - TIE_DOWN: round ties with FLOOR rule
+                - TIE_UP: round ties with CEILING rule
+                - TIE_TOWARDS_ZERO: round ties with TRUNCATE rule
+                - TIE_TO_ODD: round to nearest value; if exactly halfway, tie
+                  to the odd option.
```
westonpace, line 70: What does `TIE_TO_EVEN` and `TIE_TO_ODD` mean for negative values of `s`?

CC @ianmcook

westonpace (changes requested):
Thanks for adding integer kernels.  I think the description will need to be slightly different for these kernels.

extensions/functions_rounding.yaml
vibhatha: Ref: https://docs.oracle.com/cd/E57185_01/AGOHP/apbs03s18.html

vibhatha: It is not quite clear from this definition either: https://dev.mysql.com/doc/refman/5.7/en/mathematical-functions.html#function_round
:thinking:

westonpace: I don't think we need to clarify.  I don't think we can easily do so without writing a novel.

westonpace (approved):
Alright, I think this looks good.  @EpsilonPrime did you want to take another look?

extensions/functions_rounding.yaml
vibhatha: I played with Mysql a bit, but with the round function, can we specify such options?

vibhatha:
@westonpace a CI is failing and this happened a few times. Seems like unrelated, though  not very sure about it.

EpsilonPrime (commented):
LGTM

westonpace:
> @westonpace a CI is failing and this happened a few times. Seems like unrelated, though not very sure about it.

Unrelated to this PR.  I created #439 which should address it.

PR #326, opened 2022-09-10
feat: add arg_min and arg_max

vibhatha:
This PR includes adding aggregate functions `arg_min` and `arg_max` to `functions_arithmetic.yaml`

### DIFF
--- a/extensions/functions_arithmetic.yaml
+++ b/extensions/functions_arithmetic.yaml
@@ -1368,6 +1368,38 @@ aggregate_functions:
         nullability: DECLARED_OUTPUT
         ordered: true
         return: LIST?<any>
+  - name: "arg_min"
+    description: >
+      Finds the value `y` which minimizes the `x`.
+
+    impls:
+      - args:
+          - name: x
+            description: >
+              Expression to be minimized.
+            value: any
+          - name: y
+            description: >
+              Expression to be returned.
+            value: any
+        nullability: DECLARED_OUTPUT
+        return: any?
+  - name: "arg_max"
+    description: >
+      Finds the value `y` which maximizes the `x`.
+
+    impls:
+      - args:
+          - name: x
+            description: >
+              Expression to be maximized.
+            value: any
+          - name: y
+            description: >
+              Expression to be returned.
+            value: any
+        nullability: DECLARED_OUTPUT
+        return: any?
 
 window_functions:
   - name: "row_number"


jacques-n:
What systems use these names?

Snowflake uses least and greatest I believe.

vibhatha:
> What systems use these names?
> 
> Snowflake uses least and greatest I believe.

Some of the references I looked into are as follows;

[duckdb](https://duckdb.org/docs/sql/aggregates) uses `arg_min` and `arg_max` but also they refer to them as `max_by` and `min_by`. 

Also in `clickhouse` [arg_min](https://clickhouse.com/docs/en/sql-reference/aggregate-functions/reference/argmin/) and [arg_max](https://clickhouse.com/docs/en/sql-reference/aggregate-functions/reference/argmax/)

Microsoft [argmin](https://docs.microsoft.com/en-us/azure/data-explorer/kusto/query/arg-min-aggfunction) and [argmax](https://docs.microsoft.com/en-us/azure/data-explorer/kusto/query/arg-max-aggfunction) 

Velox [argmin](https://facebookincubator.github.io/velox/functions/aggregate.html#min_by) and [argmax](https://facebookincubator.github.io/velox/functions/aggregate.html#max_by)

I also played a bit with `clickhouse` `argMin` and `argMax` too.

vibhatha:
@jacques-n most of the frameworks did call it `arg_min` or `arg_max` as I observed.

jvanstraten:
I couldn't figure out what these functions do from the description you gave, so I ended up looking in the DuckDB documentation instead. I think I get it now, but:

 - The description is worded like this is some hill-climbing optimization function, which it decidedly isn't. It should be something like "find the record that minimizes `x`, then return `y` for that record."
 - Not all types in Substrait have obvious comparison operators to construct min/max with, and there's no way to pass custom sort operator here. How does null compare?
 - What happens when there are multiple minimum/maximum values? `y` need not necessarily be the same in this case. What happens when there is no input?
 - I *think* you have the arguments swapped around (at least compared to DuckDB, I don't know if others do this differently). This is one of those cases where it does help to name arguments more intelligently than just `x` and `y`... For the avoidance of doubt, I'm using your convention in this comment.

I get the feeling that, instead of `arg_min` and `arg_max`, we should just make an order-sensitive aggregate function named `first_value` (and maybe a complementary `last_value`), or something along those lines. Combined with the sort logic that's part of the aggregate function binding, you can then construct `arg_min` and `arg_max` from that. For example, `arg_min(x, y)` reduces to `first_value(x)` with an ascending sort on `y`. Using sort semantics immediately covers all of the special cases I mentioned (except no input, but that should just be null like any other aggregate), and lets you choose for the not-so-obvious ones (how null compares, ability to specify a custom sort function, sorting by multiple keys to disambiguate when there are multiple minima/maxima).

ianmcook:
> I get the feeling that, instead of `arg_min` and `arg_max`, we should just make an order-sensitive aggregate function named `first_value` (and maybe a complementary `last_value`), or something along those lines. Combined with the sort logic that's part of the aggregate function binding, you can then construct `arg_min` and `arg_max` from that. For example, `arg_min(x, y)` reduces to `first_value(x)` with an ascending sort on `y`. Using sort semantics immediately covers all of the special cases I mentioned (except no input, but that should just be null like any other aggregate), and lets you choose for the not-so-obvious ones (how null compares, ability to specify a custom sort function, sorting by multiple keys to disambiguate when there are multiple minima/maxima).

But would this allow a single operation in a single query to compute `arg_min` and `arg_max` on multiple different `x`, `y` pairs?

ianmcook:
The docs should explicitly say what the functions return if there are 2+ values tied for minimum/maximum.

jvanstraten:
> But would this allow a single operation in a single query to compute arg_min and arg_max on multiple different x, y pairs?

With the way Substrait defines aggregate measures at least, I don't see why not. The sort order is part of the aggregate function binding, not part of AggregateRel, so different measures can have different sorts. I could imagine this definition being annoying for query engines that don't do it this way (probably the majority, if not all of them...) but they'd need to deal with this anyway, at least in order to throw an error when different measures use different sorts and this isn't supported. The way I see it, they may as well convert special cases like things that map to `arg_min` and `arg_max` to their internal special cases during this step.

ianmcook:
I think it would be instructive to look at some query engine explain plans from a query containing `arg_min` or `arg_max` to see whether they treat them just like any other aggregate functions, or whether they do something different. As discussed in #307, it can be worthwhile adding new functions in cases like this if adding them can avoid producers deconstructing syntactic constructs that consumers then just need to reconstruct.

vibhatha:
@jvanstraten about description, I am inclined towards the current simple form of description (may be my grammar is wrong, we should fix if it is). At first, I also had doubts, but as I read it once more (while reading other's definitions) it is just saying mathematically `find the value y which maximizes x` that is as simple as that we can define things mathematically (correct me if I am wrong). In description, I am not sure we should explain every little detail in length. I observed the descriptions in a few standard platforms for many functions and what I saw is they keep it very simple and on-point.

vibhatha:
> * Not all types in Substrait have obvious comparison operators to construct min/max with, and there's no way to pass custom sort operator here. How does null compare?

I propose adding null-handling options. I am not sure if we can generalize this for all functions. This is missing from my definition as well. But my suggestion is to standardize null-handling, but not to generalize it.

> * What happens when there are multiple minimum/maximum values? `y` need not necessarily be the same in this case. What happens when there is no input?

This is a case I missed (ignored probably). We could return `LIST<T>`. 

> * I _think_ you have the arguments swapped around (at least compared to DuckDB, I don't know if others do this differently). This is one of those cases where it does help to name arguments more intelligently than just `x` and `y`... For the avoidance of doubt, I'm using your convention in this comment.

Yes you're right about swapping, I have swapped them. I will correct it.

Regrading naming arguments, I thought about this too, but decided to add a description than changing the names. Because at the moment we use `x` and `y` or `a` and `b`. That is also not consistent either. So I propose names (`arg`, `val`) or (`argument` , `value`). Any thoughts?
> 
> I get the feeling that, instead of `arg_min` and `arg_max`, we should just make an order-sensitive aggregate function named `first_value` (and maybe a complementary `last_value`), or something along those lines. Combined with the sort logic that's part of the aggregate function binding, you can then construct `arg_min` and `arg_max` from that. For example, `arg_min(x, y)` reduces to `first_value(x)` with an ascending sort on `y`. Using sort semantics immediately covers all of the special cases I mentioned (except no input, but that should just be null like any other aggregate), and lets you choose for the not-so-obvious ones (how null compares, ability to specify a custom sort function, sorting by multiple keys to disambiguate when there are multiple minima/maxima).

Why not use ordering direction options here? If that's just for the output, why not define a set of options. Is there something wrong with that approach? I could be wrong here, just thinking out loud. 

Another thing is why we should avoid using the already used names in functions but introduce our own names. If we are to take a different route because of the restrictions and models we have to follow when defining things in Susbtrait, we must document such exceptions properly. What I meant is to have a separate section pointing to what the such exception cases are and how Substrait defines them.

jvanstraten:
> At first, I also had doubts, but as I read it once more (while reading other's definitions) it is just saying mathematically `find the value y which maximizes x` that is as simple as that we can define things mathematically (correct me if I am wrong).

No, that's not what it does.

"Find the value y which maximizes x" is mathematical language for "given a function x := f(y), find the value y that maximizes x" (you'd customarily swap x and y, but I'll roll with it). There are a myriad of problems with that. There are entire *fields* of mathematics dedicated to answering that question in a whole number of ways (computing f'(y) and intersecting that with zero, decision problem solving, hill-climbing, neural networks, monte-carlo, etc), none of which apply generally, because computing f(y) for all possible values of x ranges from intractable to impossible. Not to mention that you cannot represent f(x) in Substrait in the first place. Substrait functions take *values* as arguments, not functions or expressions, and certainly not lambda functions with arguments (not yet anyway). They happen to be *computed* using an expression, but that's not the function's concern. I don't really like the way you've started describing arguments either for this reason; it implies that you're passing something like a lambda to the arguments rather than a value, which isn't really the case.

What `arg_max` *actually* does is "given val := f(record) and arg := f(record), compute val for all records, find the/a/the first record for which val has the max value, and then compute and return arg for that same record." Functions of the current record are just scalar expressions, and computing them for all records (in a certain group) is exactly what an aggregate function does for its arguments. Mapping that onto Substrait, the function gets two sequences of values, one for val and one for arg, each with one element for each record in the group. It finds the record index for which val[index] has the highest value, and then returns arg[index]. In practice, the execution engine will probably choose to only compute arg[index] rather than all of arg, but (possibility for missed errors aside) that doesn't change the behavior. I still feel like we need an argument flag to indicate whether the function may choose not to compute it, so an optimizer knows it can't just move the expression somewhere else and replace it with a FieldRef, but either way, other functions in Substrait already do this as well (like `and`), so it's nothing new.

> I propose adding null-handling options. [...]

Assuming we don't go for the `first_value` thing I proposed, I'm fine with that. Assuming `MIRROR` or `DECLARED_OUTPUT`-nullable behavior, I guess you could:

 - ignore records for which val is null (so if val is null for all records, return null to signal empty input);
 - treat null as min/max (so if there is a null val, arg will be returned for its record);
 - treat null as not min/max (so if there is at least one non-null val, arg will be returned for a non-null val, but if there are only nulls, arg will still be returned for a record where val was null, which may end up being non-null since arg and val are completely unrelated expressions)

For `first_value`, the latter two options are chosen using NULLS_FIRST/NULLS_LAST. I guess the former would require a slightly dubious option on `first_value` to turn it into `first_non_null_value` instead, or maybe that should just be a different function.

> This is a case I missed (ignored probably). We could return `LIST<T>`.

I guess, but that's probably not what the execution engines actually do. It's probably more sensible to just say that in that case *any* value of arg for which val is maximum may be returned. I guess you could have an option to explicitly choose between the first record (probably what most do in practice anyway) or any record (what you're likely to get when you parallelize/distribute), or maybe null or an error for ambiguous cases instead. A snag here, though, is that the ordered flag of the function now depends on an option argument (first -> ordered, any -> not ordered). I guess another way of doing it is dropping the option, setting the ordered flag, and describing that it should return the first value if a sort is specified, but may return any value if not; but then you still haven't covered the case where you just want the first value in the original order, for example if the input from the AggregateRel comes from a dedicated SortRel. Or we could just use the option and pessimistically set the ordered flag.

> Regrading naming arguments, I thought about this too, but decided to add a description than changing the names. Because at the moment we use x and y or a and b. That is also not consistent either. So I propose names (arg, val) or (argument , value). Any thoughts?

The a,b,c vs x,y,z thing is... well, I'm finding it hard to care about it, to be honest. Like, I'd personally use things like `lhs` and `rhs` for binary operators, but it's just a preference. I'd also have been fine not naming them at all, but we decided to just always name arguments, no matter how trivial, just so we don't end up discussing whether an argument is trivial *enough* to not be named. If you feel more strongly about this and want to do a consistency pass to rename a,b,c to x,y,z or vice versa though, I won't stop you.

Either way, you should only do the x,y,z thing for arguments where the order doesn't matter and where they are the only logical inputs to the function. For a unary like `negate(x: i32) -> i32`, well, who cares what the argument is called, right? For `add(overflow: opt, x: i32, y: i32) -> i32` that prototype on its own is sufficient to be able to get an impression of what the function does; you don't need special names for x and y. But compare that to `add(x: opt, y: i32, z: i32) -> i32`; well, now what the heck is the option for? Does it somehow get added as well? Maybe it's an extra carry flag or it optionally negates z (both very common in hardware)? Or `replace(x: string, y: string, z: string) -> string`, well, who *knows* what the argument order for that one is? Now you actually have to read the description just to get a general idea, and hope that the description is thorough enough to actually specify.

`arg_max(x: any, y: any) -> any` decidedly isn't among the self-evident functions, or we wouldn't be having this conversation. The types are actually also wrong (if they were correct they would also disambiguate in this case) but I'll get to that. Anyway, `arg_max(val, arg) -> any` is much more descriptive, because the general logic for `arg_max`-like functions is to find the maximum value of some sequence, but rather than returning this value, return something that identifies the value. Usually that's just an index, but for database stuff indices are not very meaningful, so having an `arg` argument instead makes sense.

I don't really have an opinion about abbreviated vs. long names.

> Why not use ordering direction options here? If that's just for the output, why not define a set of options. Is there something wrong with that approach? I could be wrong here, just thinking out loud.

I'm not 100% sure what you're saying here. Sort options for `arg_min`/`arg_max` don't make sense (except maybe to disambiguate when there are multiple minima/maxima). If you *would* use the sort options in the aggregate binding to describe how to order the `val` of `arg_min`/`arg_max`, the function you naturally end up with is `first_value` (or its complement `last_value`, of course) because `val` just becomes the sort key (which isn't a normal argument), and the minimum/maximum of a sorted list is just the first/last element. As a boinus, that function is much more widely applicable than just `arg_min`/`arg_max`.

If you mean, "why don't we just mimic the sort options with optional enumeration function arguments," well, the sort options for aggregate function bindings are much more comprehensive than what you can accomplish with those. The sort options let you specify a *list* of sort keys, each of which has an expression for the key, *and* a direction option (ascending/descending with nulls first/last) or a custom sort function reference. The variadic nature of that list is a problem, the options are a problem (because variadic), and the function reference for a custom sort is a problem. None of those things are possible with function arguments.

We can decide that just having a single sort key and not supporting a custom sort function reference is *good enough*; then the distinction between `arg_min` and `arg_max` chooses the order, and an option could choose the null behavior. That's what the typical execution engine does, I imagine (though I'm not sure because I don't actually know what the usage of these things looks like syntactically in SQL). But you'll never get the generality of `first_value` with a sort this way, unless we add both a struct expression (to wrap multiple sort keys into `val`) and the means to pass lambda functions/expressions to function arguments. I would argue that we need those features anyway to be fair, but we definitely don't have them yet. And even then, you can't really choose to reverse the order of one or some of the keys without a custom sort.

> Another thing is why we should avoid using the already used names in functions but introduce our own names. If we are to take a different route because of the restrictions and models we have to follow when defining things in Susbtrait, we must document such exceptions properly. What I meant is to have a separate section pointing to what the such exception cases are and how Substrait defines them.

I'm not sure I understand what you're saying, but I think you mean that if we go the `first_value` route, we should have a thing in the docs that says you should use that for `arg_min` and `arg_max`. In that case I wholeheartedly agree; we could and *should* add a page to the core extensions section introduced by #310 with recommended patterns for common things that we don't explicitly have functions for.

> The types are actually also wrong (I realized that when I thought more about this function later) but I'll get to that.

`arg_max(arg: any, val: any) -> any?` means that `arg` and `val` need to have the same type. It should be `arg_max(arg: any1, val: any2) -> any1?` instead, or something like `arg_max(arg: T, val: V) -> T?` if you prefer (the names don't matter as long as they're unique; I personally don't like `any` because it implies to me that it's not actually a binding, but there's lots of precedent for it).

westonpace:
I think these functions are valuable.  In hopes of restarting momentum on this let me take a pass at some of these discussion points:

> The description is worded like this is some hill-climbing optimization function, which it decidedly isn't. It should be something like "find the record that minimizes x, then return y for that record."

To be fair, this appears to be pretty closely to the way Microsoft describes the function.  However, I agree, this is not the most intuitive description.  I think I prefer the description from clickhouse "Calculates the arg value for a minimum val value. If there are several different values of arg for minimum values of val, returns the first of these values encountered."  I agree the wording should be updated.

> Not all types in Substrait have obvious comparison operators to construct min/max with, and there's no way to pass custom sort operator here. How does null compare?

I agree that the ability to specify a custom sort operator would be valuable.  This can be added as an additional constant argument.  However, we do not yet have the capability for a "function literal".  So let's defer for now. Worst case, we need to introduce `arg_min_custom` and `arg_max_custom` when we figure that out.

> What happens when there are multiple minimum/maximum values? y need not necessarily be the same in this case. What happens when there is no input?

An option should control this.  Either we take the first instance or we error.  We should not return a list in this case.

>  we should just make an order-sensitive aggregate function named first_value

Aggregation functions are, by definition (admittedly, my definition), order insensitive.  Order sensitive aggregates fall into a category similar to window functions and I agree that an order-sensitive `first` and `last` should exist.  However, that does not remove the need for this function.

Order-sensitive aggregates are expensive to compute, often prohibitively so.  The ability to calculate the first and last in a distributed order-insensitive fashion is a powerful optimization.

Next steps:

 * If we can fix up the wording then I think we can move forward with merging this PR.

jacques-n:
Closing given this is out of date. I'm supportive of the functions but someone needs to own taking this over the finish line.

PR #327, opened 2022-09-11
feat: adding bitwise and, or, xor 

vibhatha:
This PR includes the bitwise operations [`and`, `or`, `xor`]. 
The functions are included in 
`functions_arithmetic.yaml` and `functions_boolean` yaml.

### DIFF
--- a/extensions/functions_arithmetic.yaml
+++ b/extensions/functions_arithmetic.yaml
@@ -1527,6 +1527,84 @@ aggregate_functions:
         nullability: DECLARED_OUTPUT
         ordered: true
         return: LIST?<any>
+  - name: "bit_or"
+    description: >
+      Calculates the bitwise OR of a set of values.
+      Null values are ignored.
+      If there is no input, null is returned.
+    impls:
+      - args:
+          - name: x
+            value: i8
+        nullability: DECLARED_OUTPUT
+        return: i8?
+      - args:
+          - name: x
+            value: i16
+        nullability: DECLARED_OUTPUT
+        return: i16?
+      - args:
+          - name: x
+            value: i32
+        nullability: DECLARED_OUTPUT
+        return: i32?
+      - args:
+          - name: x
+            value: i64
+        nullability: DECLARED_OUTPUT
+        return: i64?
+  - name: "bit_and"
+    description: >
+      Calculates the bitwise AND of a set of values.
+      Null values are ignored.
+      If there is no input, null is returned.
+    impls:
+      - args:
+          - name: x
+            value: i8
+        nullability: DECLARED_OUTPUT
+        return: i8?
+      - args:
+          - name: x
+            value: i16
+        nullability: DECLARED_OUTPUT
+        return: i16?
+      - args:
+          - name: x
+            value: i32
+        nullability: DECLARED_OUTPUT
+        return: i32?
+      - args:
+          - name: x
+            value: i64
+        nullability: DECLARED_OUTPUT
+        return: i64?
+  - name: "bit_xor"
+    description: >
+      Calculates the bitwise XOR of a set of values.
+      Null values are ignored.
+      If there is no input, null is returned.
+    impls:
+      - args:
+          - name: x
+            value: i8
+        nullability: DECLARED_OUTPUT
+        return: i8?
+      - args:
+          - name: x
+            value: i16
+        nullability: DECLARED_OUTPUT
+        return: i16?
+      - args:
+          - name: x
+            value: i32
+        nullability: DECLARED_OUTPUT
+        return: i32?
+      - args:
+          - name: x
+            value: i64
+        nullability: DECLARED_OUTPUT
+        return: i64?
 
 window_functions:
   - name: "row_number"
--- a/extensions/functions_boolean.yaml
+++ b/extensions/functions_boolean.yaml
@@ -138,3 +138,42 @@ aggregate_functions:
             name: a
         nullability: DECLARED_OUTPUT
         return: boolean?
+
+  -
+    name: "bit_or"
+    description: >
+      Calculates the bitwise OR of a set of values.
+      Null values are ignored.
+      If there is no input, null is returned.
+    impls:
+      - args:
+          - name: x
+            value: boolean
+        nullability: DECLARED_OUTPUT
+        return: boolean?
+
+  -
+    name: "bit_and"
+    description: >
+      Calculates the bitwise AND of a set of values.
+      Null values are ignored.
+      If there is no input, null is returned.
+    impls:
+      - args:
+          - name: x
+            value: boolean
+        nullability: DECLARED_OUTPUT
+        return: boolean?
+
+  -
+    name: "bit_xor"
+    description: >
+      Calculates the bitwise XOR of a set of values.
+      Null values are ignored.
+      If there is no input, null is returned.
+    impls:
+      - args:
+          - name: x
+            value: boolean
+        nullability: DECLARED_OUTPUT
+        return: boolean?


extensions/functions_arithmetic.yaml
```diff
@@ -1369,6 +1369,109 @@ aggregate_functions:
...
+      - args:
+          - name: x
+            value: fp32
+        nullability: DECLARED_OUTPUT
+        return: fp32
+      - args:
+          - name: x
+            value: fp64
+        nullability: DECLARED_OUTPUT
+        return: fp64
```
jvanstraten, line 1405: Bitwise operations on floats make no sense and should be removed.

extensions/functions_arithmetic.yaml
```diff
@@ -1369,6 +1369,109 @@ aggregate_functions:
...
         return: LIST?<any>
 
+  - name: "bit_or"
+    description: >
+      Calculates the bitwise OR of all values in set of values.
+    impls:
+      - args:
+          - name: x
+            value: i8
+        nullability: DECLARED_OUTPUT
```
jvanstraten, line 1379: `DECLARED_OUTPUT` means that the input *can* be nullable. So how does this function deal with nulls? I imagine it ignores them, but this should be documented.

extensions/functions_arithmetic.yaml
```diff
@@ -1369,6 +1369,109 @@ aggregate_functions:
...
+        return: i8
```
jvanstraten, line 1380: The output is non-nullable, so what do these aggregates do when there is no input?

jvanstraten (changes requested):

ianmcook:
We will also want to add _non-aggregate_ bitwise functions [`and`, `or`, `xor`]. What names should we pick for those? Unfortunately it's not much help looking at what SQL databases call them, because they use binary operators like `&`, `|`, `#` or else they use names that closely resemble the names of the aggregate bitwise functions that this PR adds.

Before we merge this, we should decide on a good naming convention for the aggregate and non-aggregate bitwise functions.

jvanstraten:
I've been thinking lately that maybe what we need is just new expression types that evaluate aggregate functions on a variadic function argument list and on `LIST<T>` input in scalar context. That should *hopefully* avoid a lot of this duplication (and name disambiguation), although it does seem like SQL insists on different null behavior for the different invocation styles, so maybe we can't have nice things. Otherwise, though, if the functions are defined in different YAML files the URI could maybe be used to disambiguate them.

In general, we should figure out a sane subdivision of the functions over the YAML files and refactor accordingly. I would personally not consider bitwise operators to belong to the class of arithmetic operators, for example. Likewise for trigonometric functions, at least if we consider logarithms to be special (while simultaneously grouping exponentiation with arithmetic again).

julianhyde:
Despite the name, aggregate functions are more like relational operators than functions. You can't rely on the input being small enough, and in the same place at the same time, to be able to form a list. So aggregate functions are invariably implemented by decomposing them into smaller functions.

One common decomposition is what the functional folks call 'fold'. For example, SUM decomposes to fold of '+', MIN decomposes to fold of 'least'. The bitwise aggregate functions could be defined by folding 'and', 'or' and 'xor'.

Monoid decomposition defines an aggregate function by means of subtotals, and a means to merge subtotals. Such aggregate functions scale well because they can merge results that are computed in parallel. The sketch algorithms (such as HyperLogLog which implements approximate-distinct-count) are monoids.

Perhaps Substrait could give people a means to define their own aggregate function by providing a (initialize, fold, finalize) operations, or (initialize, add, merge, finalize) operations. Then people will stop asking for bitwise aggregate functions.

jvanstraten:
I was thinking about it the other way around, i.e. if an aggregate function exists that can operate on huge inputs, it should also work just fine on small inputs. Okay, perhaps not efficiently if it ends up spawning a worker process for just two integers or whatever. Either way, I really don't think Substrait should concern itself with whether something can actually be computed before the heat death of the universe, at least for logical plans; it's up to the consumer to execute a plan in a sane way, or alternatively up to the producer or an optimizer that sits somewhere in between to emit a plan that uses only consumer-specific functions and relations that actually perform. I'd much rather generalize everything and end up with a few corner cases that you just shouldn't use because computation isn't feasible, than defining a new special case for everything that *is* feasible, like we're kind of forced to for scalar variadic vs scalar list vs aggregate functions now.

But, again, differences in null handling remain a problem.

I personally like the idea of going the more functional route and spam lambdas and functional operators everywhere instead of aggregates, but that's just not feasible. It would require essentially a complete rewrite of expressions and most relations. You might as well start from scratch.

vibhatha:
Thinking about this a bit as @jvanstraten said, null-handling is kind of a hard to generalize thing per all functions. How about defining options to handle null. Depending on the function we could provide options to handle null. That's whether to exclude in the input, so the null records are not used in computation, or use them in the computation and the engine decides the logic. Also we already define whether it could return null or not. Not sure if this is a fine approach, but just a suggestion.

Regarding bit operations just for aggregates, as far as I understand, why shouldn't we allow it as an arithmetic function? What is stopping us providing that for numerics? If we are not defining it for numerics, how are suppose to provide that definition for users who want to use it with numerics? If we don't do that we should define limitations or exceptions in our documentation.

extensions/functions_arithmetic.yaml
vibhatha: Ah... thanks for catching this one.

jvanstraten:
We're digressing here. We used `bool_and` etc for the aggregate version of boolean AND, and just `and` for the scalar version. That seems to come from postgres. We can just do the same here.

As for nullability behavior, let's just define them in a way that is consistent with the rest of the functions now. The logic seems to be pretty consistent among aggregate functions (just not consistent with the scalar interpretation of null), i.e. ignore nulls at the input and return null if there are no non-null inputs.

That's not to say that we shouldn't be thinking about more generic ways of solving this, and we absolutely need to refactor the URIs and the files the functions are defined in because that's an inconsistent mess right now, but it'd be a shame to let everything stagnate now just because we might refactor later. Before we can really do that properly though, we first need to decide if we're going to A) keep doing what we're doing and accept the lack of scalability as a cost of compatibility with current systems, B) do something along the lines of what I first proposed, i.e. add more function binding expressions for different broadcast methods to merge add/sum/list-sum into one (having thought about this some more, the nullability behavior of all these SQL aggregates is just "ignore nulls, return null if input is empty," so that's easily generalized), C) go the lambda function route, or D) make some cursed combination thereof, so we know whether we have to think about name conflicts between aggregate and scalar functions in the first place. The more I think about those alternatives though, the less sure I am about any of them; not because the current system is better, but because I'm still not sure we can come up with a *clean* system without throwing out a lot of what we already have. We can tack something like lambdas on pretty easily by just adding a lambda data type and a few expression types here and there, but if we're not going to or can't replace any existing functionality that would be naturally covered by lambdas and functional programming with that without starting over, it's going to make things *less* consistent, not more.

vibhatha:
> We're digressing here. We used `bool_and` etc for the aggregate version of boolean AND, and just `and` for the scalar version. That seems to come from postgres. We can just do the same here.
> 
> As for nullability behavior, let's just define them in a way that is consistent with the rest of the functions now. The logic seems to be pretty consistent among aggregate functions (just not consistent with the scalar interpretation of null), i.e. ignore nulls at the input and return null if there are no non-null inputs.
> 
> That's not to say that we shouldn't be thinking about more generic ways of solving this, and we absolutely need to refactor the URIs and the files the functions are defined in because that's an inconsistent mess right now, but it'd be a shame to let everything stagnate now just because we might refactor later. Before we can really do that properly though, we first need to decide if we're going to A) keep doing what we're doing and accept the lack of scalability as a cost of compatibility with current systems, B) do something along the lines of what I first proposed, i.e. add more function binding expressions for different broadcast methods to merge add/sum/list-sum into one (having thought about this some more, the nullability behavior of all these SQL aggregates is just "ignore nulls, return null if input is empty," so that's easily generalized), C) go the lambda function route, or D) make some cursed combination thereof, so we know whether we have to think about name conflicts between aggregate and scalar functions in the first place. The more I think about those alternatives though, the less sure I am about any of them; not because the current system is better, but because I'm still not sure we can come up with a _clean_ system without throwing out a lot of what we already have. We can tack something like lambdas on pretty easily by just adding a lambda data type and a few expression types here and there, but if we're not going to or can't replace any existing functionality that would be naturally covered by lambdas and functional programming with that without starting over, it's going to make things _less_ consistent, not more.

This is definitely something to think about and broadly discuss in the community. I am glad we started this conversation.

vibhatha:
@jvanstraten could you please look into the updated description on nullability? I just kept it simple only in the description and didn't add any fancy options for now. I thought that's a topic for another day. Maybe I am wrong. But WDYT?

extensions/functions_boolean.yaml
```diff
@@ -138,3 +138,42 @@ aggregate_functions:
...
+            value: boolean
+        nullability: DECLARED_OUTPUT
+        return: boolean
+
+  -
+    name: "bit_xor"
+    description: >
+      Calculates the bitwise XOR of all values in set of values.
+      Null values doesn't affect the output and can be considered ignored.
+      If the output is null it means that there is no output for the passed input.
```
westonpace, line 173:
```suggestion
      Calculates the bitwise XOR of a set of values.
      Null values are ignored.
      If there is no input, null is returned.
```

extensions/functions_boolean.yaml
```diff
@@ -138,3 +138,42 @@ aggregate_functions:
...
+    description: >
+      Calculates the bitwise OR of all values in set of values.
+      Null values doesn't affect the output and can be considered ignored.
+      If the output is null it means that there is no output for the passed input.
+    impls:
+      - args:
+          - name: x
+            value: boolean
+        nullability: DECLARED_OUTPUT
+        return: boolean
```
westonpace, line 153: I believe `DECLARED_OUTPUT` with a return type of `boolean` means that there cannot be nulls in the output.  Yet, the description claims "If the output is null it means there is no output for the passed input."  Do you perhaps want `return: boolean?`

westonpace (changes requested):
Some minor wording suggestions to keep the wording similar to the other functions in this file.

extensions/functions_boolean.yaml
vibhatha: I agree. We should do that.

EpsilonPrime:
These operations were added in #370.

PR #329, opened 2022-09-12
feat: add options for behaviour when dividing by zero or calculating log zero

thisisnic:
This PR adds options which allow specification of behaviour when dividing by 0 or when trying to calculate log of 0.  See discussion on https://github.com/substrait-io/substrait/pull/302 for more detail.

### DIFF
--- a/extensions/functions_arithmetic.yaml
+++ b/extensions/functions_arithmetic.yaml
@@ -175,12 +175,20 @@ scalar_functions:
         return: fp64
   -
     name: "divide"
-    description: "Divide one value by another. Partial values are truncated."
+    description: >
+      Divide x by y. In the case of integer division, partial values are truncated.
+      The `on_division_by_zero` option governs behaviour in cases where y is 0 and x is not 0.
+      The `on_domain_error` option governs behaviour for 0/0 and Inf/Inf etc.
     impls:
       - args:
           - name: overflow
             options: [ SILENT, SATURATE, ERROR ]
             required: false
+          - name: on_division_by_zero
+            options: [ NAN, ERROR, LIMIT ]
+            required: false
+          - name: on_domain_error
+            options: [ NAN, ERROR ]
           - name: x
             value: i8
           - name: y
@@ -190,6 +198,11 @@ scalar_functions:
           - name: overflow
             options: [ SILENT, SATURATE, ERROR ]
             required: false
+          - name: on_division_by_zero
+            options: [ NAN, ERROR, LIMIT ]
+            required: false
+          - name: on_domain_error
+            options: [ NAN, ERROR ]
           - name: x
             value: i16
           - name: y
@@ -199,6 +212,11 @@ scalar_functions:
           - name: overflow
             options: [ SILENT, SATURATE, ERROR ]
             required: false
+          - name: on_division_by_zero
+            options: [ NAN, ERROR, LIMIT ]
+            required: false
+          - name: on_domain_error
+            options: [ NAN, ERROR ]
           - name: x
             value: i32
           - name: y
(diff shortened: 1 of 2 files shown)


extensions/functions_arithmetic.yaml
```diff
@@ -175,12 +175,20 @@ scalar_functions:
...
     impls:
       - args:
           - name: overflow
             options: [ SILENT, SATURATE, ERROR ]
             required: false
+          - name: on_division_by_zero
+            options: [ NAN, ERROR, LIMIT ]
+            required: false
+          - name: on_domain_error
+            options: [ NAN, ERROR ]
```
jvanstraten, line 191: NaNs and infinities are, thankfully, not a thing for integers (and Substrait is strongly typed, so the function can't decide to return those things anyway). So the integer variants have no choice but to fail for divisions by zero. I think there *are* actually programming languages that define the result of an integer division by zero to zero (which is perfectly fine mathematically, you can define a division by zero to whatever you want), but let's not go that far unless there are also *query engines* that do that.

extensions/functions_arithmetic.yaml
```diff
@@ -175,12 +175,20 @@ scalar_functions:
         return: fp64
   -
     name: "divide"
-    description: "Divide one value by another. Partial values are truncated."
+    description: >
+      Divide x by y. In the case of integer division, partial values are truncated.
+      The `on_division_by_zero` option governs behaviour in cases where y is 0 and x is not 0.
```
jvanstraten, line 180: Let's stick to American English for the documentation for consistency (behaviour -> behavior).

thisisnic: Oops, updated now!

extensions/functions_arithmetic.yaml
thisisnic: That's useful to know, updated now!

extensions/functions_arithmetic.yaml
```diff
@@ -175,12 +175,20 @@ scalar_functions:
...
+      The `on_domain_error` option governs behaviour for 0/0 and Inf/Inf etc.
```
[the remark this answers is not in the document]
jvanstraten: It looks fine to me, but cc @pitrou 

PS. I honestly can't tell if github actually added this comment when I first reviewed this because the UI was glitching all to hell, but it seems to be missing for me now, so... Please excuse me if this ends up being duplicated.

extensions/functions_arithmetic.yaml
```diff
@@ -175,7 +175,10 @@ scalar_functions:
         return: fp64
   -
     name: "divide"
-    description: "Divide one value by another. Partial values are truncated."
+    description: >
+      Divide x by y. In the case of integer division, partial values are truncated.
```
pitrou, line 179: Can we describe more precisely what "partial values are truncated" means?
For example, what would be the result of `1 / -3`?

jvanstraten: +1, we use "truncation" for rounding toward zero (in contrast with ceil for +infinity and floor for -infinity) but it bears repeating.

extensions/functions_arithmetic.yaml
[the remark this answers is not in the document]
thisisnic: That makes sense, updated now!

extensions/functions_arithmetic.yaml
thisisnic: I have not seen truncation defined before as anything other than "round towards zero", but adding a few extra words does no harm, so updated here.

jvanstraten (approved):
Looks good now, thanks!

PR #333, opened 2022-09-16
feat(spec): change projection output order to intuitive expectation

westonpace:
BREAKING CHANGE: output order of the project relation has changed.

### DIFF
--- a/site/docs/relations/logical_relations.md
+++ b/site/docs/relations/logical_relations.md
@@ -137,14 +137,14 @@ The project operation will produce one or more additional expressions based on t
 | Inputs               | 1                                                            |
 | Outputs              | 1                                                            |
 | Property Maintenance | Distribution maintained, mapped by emit. Orderedness: Maintained if no window operations. Extended to include projection fields if fields are direct references. If window operations are present, no orderedness is maintained. |
-| Direct Output Order  | The field order of the input + the list of new expressions in the order they are declared in the expressions list. |
+| Direct Output Order  | One field for each expression in the order they are declared in the expressions list. |
 
 ### Project Properties
 
 | Property    | Description                                          | Required                         |
 | ----------- | ---------------------------------------------------- | -------------------------------- |
 | Input       | The relational input.                                | Required                         |
-| Expressions | List of one or more expressions to add to the input. | At least one expression required |
+| Expressions | List of one or more expressions to output.           | At least one expression required |
 
 === "ProjectRel Message"
 


westonpace:
I expect this one will be somewhat controversial and perhaps the time for making such a breaking change has passed.  However, I have heard consistently that the output order of the projection relation was surprising to people.

For example, Acero's engine only emitted one column per expression initially and had to add support for this scheme later on.

As another example, DuckDb's producer currently does not appear to respect the correct output order for projection.

westonpace:
> This seems like a convenience interface on libraries, not a reason to change the spec.

The spec should be intuitive where possible.  I certainly would not propose something like this after 1.0.  As I said, I'm not really sure if the time has passed on this one or not but I wanted to see how people felt about it.

cpcloud:
I'm not sure I follow the current spec here.

Are all projections equivalent to `SELECT *, <new expressions>`?

jacques-n:
The current spec is: 

The output of this is mapped to the projection expressions. There are no other output. E.g.

```SELECT <new expressions> FROM t```

This change proposes switching to 

```SELECT t.*, <new expressions> FROM t```

**Comment by cpcloud** [comment]
It's funny, I read it as the exact opposite.

**Comment by jacques-n** [comment]
My main feeling is that we're falling down on building convenience apis. The result is people are trying to change the spec to solve problems as opposed to just adding a convenience api. This isn't the only place this has happened. I'm not religiously against this but it does suck. In some places it can be without disruption...

I believe this change is a breaking change for entirely cosmetic reasons. Nothing new can be represented. Nothing previously representable is lost.

The tone of that last paragraph sounds harsher than it is. I'm trying to state fact. We can choose to do this. It just feels like the start of a  never-ending journey.

**Comment by cpcloud** [comment]
The current spec says:

> The field order of the input + the list of new expressions in the order they are declared in the expressions list.

To me `field order of the input` means `*`, and `list of new expressions` means `<new expressions>`. Is that wrong?

**Comment by westonpace** [comment]
I interpret the current spec the same as @cpcloud

**Comment by jacques-n** [comment]
> One field for each expression in the order they are declared in the expressions list.

> It's funny, I read it as the exact opposite.

Hmm. It doesn't mention anything but the expression list in the spec. Open to idea for how we can make it clearer.

**Comment by westonpace** [comment]
Isthmus (and I believe the validator) also both agree with @cpcloud 's interpretation:

```
(base) pace@pace-desktop:~/dev/substrait-java$ ./isthmus/build/graal/isthmus -c "CREATE TABLE T1(foo int, bar bigint)" "SELECT * FROM T1"
{
  "extensionUris": [],
  "extensions": [],
  "relations": [{
    "root": {
      "input": {
        "project": {
          "common": {
            "emit": {
              "outputMapping": [2, 3]
            }
          },
          "input": {
            "read": {
              "common": {
                "direct": {
                }
              },
              "baseSchema": {
                "names": ["FOO", "BAR"],
                "struct": {
                  "types": [{
                    "i32": {
                      "typeVariationReference": 0,
                      "nullability": "NULLABILITY_NULLABLE"
                    }
                  }, {
                    "i64": {
                      "typeVariationReference": 0,
                      "nullability": "NULLABILITY_NULLABLE"
                    }
                  }],
                  "typeVariationReference": 0,
                  "nullability": "NULLABILITY_REQUIRED"
                }
              },
              "namedTable": {
                "names": ["T1"]
              }
            }
          },
          "expressions": [{
            "selection": {
              "directReference": {
                "structField": {
                  "field": 0
                }
              },
              "rootReference": {
              }
            }
          }, {
            "selection": {
              "directReference": {
                "structField": {
                  "field": 1
                }
              },
              "rootReference": {
              }
            }
          }]
        }
      },
      "names": ["FOO", "BAR"]
    }
  }],
  "expectedTypeUrls": []
}
```

Note that the emit on `project` is referring to `[2, 3]` which only makes sense if `[0, 1]` are the input columns.

cpcloud:
It sounds like this may be just a mis-mis-interpretation of the spec and is not actually a breaking change.

cpcloud:
If we all agree that the behavior should be `SELECT <new expression>` the only thing to do is make that clear in the spec language. No need for any more back and forth!

jacques-n:
Okay, I misread this patch proposal in the opposite way (going from `new => * + new` as opposed to what it is actually saying: going from `* + new => new`. 

Either case, I think we should leave it as is. Sorry for added noise of my confusion. That's what one gets for looking at things on the phone...

jvanstraten:
I don't think that there eventually hopefully being an API to wrap a confusing format specification is a reasonable excuse to willfully ignore that the specification is confusing. That's just adding a whole new layer of flimsiness and questionable conversions and interpretations on top of a known-to-be flimsy foundation. If we all agree that, in retrospect, the way Substrait was defined is more confusing than it has any right to be, we need to rip the band-aid off and fix things properly, not apply more band-aids. It doesn't *need* to be confusing, we're not defining "brainfuck: SQL edition" here.

jacques-n:
I wrote this part of the spec. I wrote it based out of experience working on a number of relational algebra systems (and actually going through this exact process of building systems to system-agnostically serialize algebraic trees several times). Each time I've looked to improve upon weaknesses of the previous implementation. The spec isn't solely trying to copy "how things have been done" because in many situations, things have been done suboptimally for years. There are real design choices in this project to try to avoid loading down Substrait with historical missteps. There are a number of things that were designed a certain way to try to build a new, better way. Many of those are different than what people are familiar with. Some examples of design choices that are atypical compared with systems that aren't focused on a serialized common and a clean standard include:

- Create a new emit concept to avoid using project to trim or reorder columns in a plan. This was done to minimize project explosion (something that happens frequently in systems that can't trim fields on every relation) as well as to avoid requirement of trivial projects for potential column-reordering transformations (e.g. swapping the sides of a join). It was kept single level (without complex references) to keep the representation small (and allow very compact representations for narrower plans).
- Avoid the use of field names within standard relational structures. This keeps things compact and avoids introduction of confusing equality semantics and potential inconsistencies.
- Use one dimensional field references in the join operation: this avoids the need to create specialized scalar expressions that can only be used in the join.
- Exclusion of named struct from the type system and aggregate invocations from the expression system: done to reduce the need for semantic analysis of proper construction of expressions trees.
- Avoid storing relation-level output schema information in the protobuf: this reduces the size of messages and potential inconsistencies. It was done because this causes massive bloat in plans with very large schemas (a majority of the plan serialization time, cost, wire and memory size are dealing with these schemas in these cases).
- Define a default projection pattern that includes inputs where feasible. This is true for most non-collapsing operations such as filter, join, union, sort and yes, project. This avoids having a lot of boilerplate for the most common pattern: passing through input data. In the case of project this changes the common representation from having to potentially create many a long list of expression unions + field reference expressions to using a small list of integers (smaller & faster).

In every one of these cases, these choices make Substrait harder (or less intuitive) to use at the protobuf format level. Does that mean we should immediately remove emit, add field names, create a special join field expression that gives relation input offset in addition to field offset? I don't think it does. 

We should make the protobuf as easy to use as possible within the constraints of our design choices (and--of course--feel open to reevaluating our design choices). However, that isn't the only tool we have in our tool belt. Independent of any protobuf polish, we *must* implement language-specific APIs to work with Substrait. If we don't, the project will fail. Protobuf will always be unintuitive and painful in many ways. To put it bluntly, nobody should have to hand construct function pointers and relation unions.

In many ways this tension (format vs api) is similar to the alignment requirements and string representation in Arrow. Do they make it harder to write to Arrow vectors directly? Yes. Does that mean we should alter the design? No, there were good reasons to do what we did. Instead, we try to hide that complexity with user-friendly apis. We should be comfortable with the fact that here as well. 

I'm happy to have conversations here about the design choices above and others. We can all disagree about the relative advantages and disadvantages of those design choices. There is lots we (and especially me) got wrong in the current (and previous) implementation(s). I'm not even interested in dying on this particular project hill . It was a deliberate choice. The proposed *breaking* change seems to be more about style than substance. Committing it sets a pretty low bar for breaking changes. We can do that if people think it is important.

Let's just do our best to assume the positive intentions and rationale of others.


------
> excuse to willfully ignore that the specification is confusing...  we're not defining "brainfuck: SQL edition" here

@jvanstraten, I feel like the messages quoted above and some others in this thread are inappropriate. I'd like to keep this focused on a productive and respectful conversation. You've turned my comments into a strawman to teardown. I'm not willfully ignoring anything.

> I happen to think none of those things are true

@paleolimbot, it would better to just ask the rationale and wait for an answer as opposed to assuming weak intention.

jvanstraten:
> I feel like the messages quoted above and some others in this thread are inappropriate.

Yeah, that's fair, I could have worded things better and I apologize. Not that it's an excuse, but I'd had a bit to drink before I saw this thread and it seemed to be going really fast all of a sudden, so I didn't want to just ignore it in case it'd suddenly be merged.

I do stand by the hypothetical I was talking about (again, sorry if it wasn't clear enough that it was a hypothetical and not an assertion). *If* at some point we all come to an agreement that things are weird at a level this fundamental (which, earlier in the thread, we seemed to have), I believe we should iterate on Substrait as a whole in one go, rather than potentially making things more confusing by changing one thing at a time, or pretending there is no problem in the first place or there is no solution. We *could* also choose not to and instead hide the weirdness with an API, but I don't think that's a good idea, because now you basically have two different specifications (API and binary) that disagree on certain things. Regardless, it doesn't seem like we've reached that point, so the hypothetical doesn't apply.

I also want to point out that I stand by the majority of the design choices you listed there. I think Substrait does a lot of things right already, but there are nevertheless fundamental pain points, too. Project+emit (mostly the latter) is IMO one of them.

To comment on your reason for doing it this way, you're talking about these relations as if they are things to be implemented as atomic operations. Now, I don't question that setting up a query engine that way would make sense; you certainly have more experience on that front than I do. But it'd be an implementation detail of that particular (hypothetical) engine, and as an interchange format, I don't think Substrait should favor one implementation over another so fundamentally. Likewise for the post-join filter and the read filter and projection. The latter especially feels odd: why don't all relations have a MaskExpression rather than a column-only emit reduction? There is of course the pragmatic argument, i.e. "well most consumers do it this way so it's just easier all around if we make a field for it;" for filter and mask in ReadRel I could see that, but I disagree for project+emit because *literally everyone else* seems to do it in the way suggested by this PR instead. As for the API argument, you can abstract to as many different conventions as you like, so you might as well use the convention most people are already familiar with as the underlying format (replacing project) and convert to the unconventional ones (appending project + emit) as needed.

IMO the format should have the means to easily facilitate consumers making implementation choices like these, but it shouldn't force a choice on them like it does now. As an example of why, in Acero, we can now end up with IIRC up to 5 projections back to back from a single project relation unless at some point we write an optimizer in our own IR (or leverage a generic one), because we chose to cluster everything related to expression evaluation and column removal/swizzling in a single relation. I don't think that's any less sane than what Substrait is implicitly suggesting a consumer does, but now we have to deal with both the conversion from a logical context where everything is an expression (for which the onus should indeed be on us, because in logical context that's perfectly sensible) *and* dealing with compatibility logic for the implementation detail fields that snuck into Substrait. Simultaneously, someone else might have a consumer where every relation has full filter and project functionality after each relation. They now have the opposite problem, because this is *more* general than Substrait.

In general, I worry that having a particular set of implemetation choices embedded like this stiffles innovation, because with Substrait as a starting point, why would you shoot yourself in the foot by doing things differently and thus needing complex conversions? For Acero we even discussed changing internal semantics after the fact, just to relieve some Substrait conversion pain points. Even if we make a public API to represent things in different ways, that'd always be more brittle than using the native representation, and it would still dissuade out-of-the-box thinking for representations we never even considered.

IMO a good way for Substrait to be agnostic to these design choices *and* be able to represent where the boundaries of the physical relations should be *and* not have to resort to advanced extensions for this, is to make all relations primitive operations, and introduce a relation group semantic to point out the physical boundaries. So, when some implementation supports pushdown filtering for reads, this can be represented by a ReadRel grouped with a FilterRel. If each relation can choose to emit only certain fields, you'd group a ProjectRel (or we could keep project semantics as append and represent the usual project implementation as a ProjectRel grouped with an EmitRel or MaskExpressionRel or whatever). You could also convey physical attributes like join strategy in the group that way and treat this as an assertion that the contained relations have a specific, compliant structure, so we wouldn't need to define a bunch of implementation-specific physical join relation types as different proto messages. But what I like most about this abstraction is that it's still consumer-agnostic; if a consumer supports a different set of physical relations it can just ignore the groups and rerun optimization and mapping.

Making the relations more primitive also *severely* reduces the amount of special cases you'd need for an optimizer, but I've digressed more than enough already.

The point I'm trying to make is that I've been holding back on proposing/submitting PRs for things like this despite being convinced that they would be an improvement, because they don't mesh properly with existing core concepts, and the time for changing such things incrementally has long passed. So they come out as bottled-up frustrations from time to time instead. I'm aware that I get very rant-y, blunt, opinionated, verbose etc when that happens; I try not to, but I can't help myself sometimes. Please know that I don't *intend* to personally offend, though, even if it might sometimes come across that way.

Now, I'd just try to suck it up and deal with it if it were just me, but judging by the rest of the thread, I'm hardly alone in these frustrations. I don't think anyone doubts your intentions or experience (I don't, anyway), but people aren't perfect, so it's frustrating that the community as a whole seems stuck with decisions you seem to have made on your own over a year ago. That's just not a healthy dynamic for what's supposed to be an unbiased interchange specification. Users shouldn't *have* to go out of their way to ask what the reasoning or intention of a single person was when they don't understand or agree with a specification; they should be able to look up the originally-open discussion thread or RFC or whatever that they resulted from. This is just further exacerbated by the continued lack of a formalized governance description.

Again, I'm not saying you did anything wrong here (well, maybe the procrastination on the governance document :). A project needs a starting point, and clearly this starting point was good enough that it garnered lots of support. But I think passive-agressive PRs and issues like this one, the "move away from protobuf" rant, or IMO more naive things like "why have non-nullable types at all" are inevitable until the bubble pops and we either iterate on Substrait as a community in a more structured way, or people get silently or vocally fed up and make their own "Superstrait" or "Substrait++" instead (and that's just SQL dialects with extra steps).

westonpace:
This thread seems to have gotten out of hand.  Saturday morning is a time for cartoons and not dramatic discussion 😄.

### Re: Stuff that isn't project

TL:DR; Let's make changes incrementally where it makes sense and try to avoid over-generalizing problems and spreading FUD.

> If at some point we all come to an agreement that things are weird at a level this fundamental (which, earlier in the thread, we seemed to have)

As a general principle for open source projects, it is a matter of fairness and openness, that one not arrive at conclusions within 24 hours.  I'd further recommend at least 24 business hours 😆 

> I believe we should iterate on Substrait as a whole in one go, rather than potentially making things more confusing by changing one thing at a time

> and the time for changing such things incrementally has long passed

I strongly disagree.  This is why I advocated early on for pre-release versioning.  Breaking changes are going to be the norm for quite some time.  I don't see any advantage in bunching them up.  I think we've all agreed that some things will have to change (e.g. in the community sync meeting we just discussed moving to named function arguments which seems like a good idea to me).

> Users shouldn't have to go out of their way to ask what the reasoning or intention of a single person was when they don't understand or agree with a specification; they should be able to look up the originally-open discussion thread or RFC or whatever that they resulted from.

This seems inevitable.  If this project had started with "What is a projection?" then I don't think we would be where we are today.  The project started with a sizable donation.  Most projects I'm aware of do.  It makes sense that it will take some time to transfer ownership and knowledge of this project away from the original author and disseminate into a group.

> But I think passive-agressive PRs and issues like this one, the "move away from protobuf" rant, or IMO more naive things like "why have non-nullable types at all" are inevitable until the bubble pops and we either iterate on Substrait as a community in a more structured way

What is more structured than bringing up issues one at a time, seeing who has interest in them, and hashing out opinions?  I don't see these as passive-aggressive PRs but instead as a way to explore and discover intentions and reasoning and get it out into the open and documented.

### Re: Project

I've been thinking about this more and I believe I have changed my stance.  I have intuitively mapped `PROJECT` in my mind to `SELECT` in SQL.  However, those aren't the same thing.  Yes, the SELECT is a point where new columns can be added but it is also the final declaration of what a plan needs to create, which goes beyond just projection.

When a user writes a query they need some set of columns.  In order to satisfy that query, various other columns may be introduced along the way.  For example, in order to satisfy a query for `column X filtered by column Y` we must load column Y even though we can throw it away after the filter node.

Generalizing this, it makes sense that each node in the graph might be the terminal node for some column of data.  So emit makes sense.  I don't think I've ever had a problem with emit.  Using a masked expression for emit might also make sense in some scenarios.  For example, `SELECT X.Y FROM table WHERE X.Z > 50`.

Given that we have emit, and it serves a justified purpose, I think it then makes a lot of sense that the project node's only real need for existing is to add new columns.

> IMO a good way for Substrait to be agnostic to these design choices and be able to represent where the boundaries of the physical relations should be and not have to resort to advanced extensions for this, is to make all relations primitive operations, and introduce a relation group semantic to point out the physical boundaries.

What we have today results in fairly compact graphs.  When we look at implementation we may have to introduce or combine nodes in the graph (e.g. because Acero doesn't support emit) but the transformations so far have been pretty straightforward.  The relation grouping idea would, I fear, become too general and difficult to implement.  If we generalize too much then we start to head down the route of re-implementing LLVM I fear.

drin:
As someone who is new to substrait, I find it very hard to read the spec. having (mostly) followed this thread, I'm now curious what the relationship is between emit and project.

The website puts project under "logical relations" and emit under "relation basics". Can any relation have an emit "clause"? Can I emit without using project? Can I emit new columns without using project? Finding the answer to these questions is difficult and makes it hard to understand why both emit and project exist.

jacques-n:
> As someone who is new to substrait, I find it very hard to read the spec. having (mostly) followed this thread, I'm now curious what the relationship is between emit and project.
> 
> The website puts project under "logical relations" and emit under "relation basics". Can any relation have an emit "clause"? Can I emit without using project? Can I emit new columns without using project? Finding the answer to these questions is difficult and makes it hard to understand why both emit and project exist.

All relations have an emit property that defaults to conventional behavior but can be overridden. "Relation basics" is trying to cover general topics as opposed to the details of a specific relation. (Open to ways to make this clearer) This allows any relation to trim excess columns or reorder without having to introduce a bunch of project operations. A couple of common examples of where this is useful:

- A filter where you only need the condition for filter evaluation, not to be propagated further down the plan.
- A join where you only want the join keys for the join, not for user consumption.
- Swapping left and right sides of the join, where you can use emit to reorder the columns.

Patches much appreciated to improve our deficits for the next person who experiences what you experienced.

PR #334, opened 2022-09-19
feat(proto): add physical hashJoin & mergeJoin physical relations

chaojun-zhang:
This PR includes adding hashJoin and mergeJoin physical plan relations according the spec for [hashJoin](https://substrait.io/relations/physical_relations/#hash-equijoin-operator) and [mergeJoin](https://substrait.io/relations/physical_relations/#merge-equijoin-operator).

### DIFF
--- a/proto/substrait/algebra.proto
+++ b/proto/substrait/algebra.proto
@@ -189,7 +189,7 @@ message CrossRel {
 
 // The relational operator representing LIMIT/OFFSET or TOP type semantics.
 message FetchRel {
-  RelCommon common = 1;
+  RelCommon commhon = 1;
   Rel input = 2;
   // the offset expressed in number of records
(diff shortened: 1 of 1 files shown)


jvanstraten (changes requested):
LGTM as implementation of the current spec (I'm not qualified to check whether the spec is actually correct so I'll assume that it is) except for CI violations:

 - the field names in `Rel` must be `lower_case`;
 - I *think* buf will also complain about duplication of the `JoinType` enum once above is fixed, because protobuf enums are not namespaced as they appear to be;
 - the titles of the individual commits must satisfy commitlint according to [conventional commit](https://www.conventionalcommits.org/en/v1.0.0/) (I know this doesn't really make sense, but I don't have authority to overrule CI failures).

PR #335, opened 2022-09-19
fix: type promotion issue in aggregate function sum

vibhatha:
This PR includes a fix for the type promotion on aggregation function `sum`.

### DIFF
--- a/extensions/functions_arithmetic.yaml
+++ b/extensions/functions_arithmetic.yaml
@@ -775,8 +775,8 @@ aggregate_functions:
             value: i8
         nullability: DECLARED_OUTPUT
         decomposable: MANY
-        intermediate: i64?
-        return: i64?
+        intermediate: i8?
+        return: i8?
       - args:
           - name: overflow
             options: [ SILENT, SATURATE, ERROR ]
@@ -785,8 +785,8 @@ aggregate_functions:
             value: i16
         nullability: DECLARED_OUTPUT
         decomposable: MANY
-        intermediate: i64?
-        return: i64?
+        intermediate: i16?
+        return: i16?
       - args:
           - name: overflow
             options: [ SILENT, SATURATE, ERROR ]
@@ -795,8 +795,8 @@ aggregate_functions:
             value: i32
         nullability: DECLARED_OUTPUT
         decomposable: MANY
-        intermediate: i64?
-        return: i64?
+        intermediate: i32?
+        return: i32?
       - args:
           - name: overflow
             options: [ SILENT, SATURATE, ERROR ]
@@ -815,8 +815,8 @@ aggregate_functions:
             value: fp32
         nullability: DECLARED_OUTPUT
         decomposable: MANY
-        intermediate: fp64?
-        return: fp64?
+        intermediate: fp32?
+        return: fp32?
       - args:
           - name: overflow
             options: [ SILENT, SATURATE, ERROR ]


vibhatha:
cc @jvanstraten @jacques-n please check this one.

jacques-n:
This would be a breaking change and I think it is undesirable. It looks like you're removing sum type promotion (unless my phone viewing of diffs is wrong).

vibhatha:
No what you saw is correct. @jacques-n I think there was a discussion regarding type promotion in Substrait and I recall aligned with the idea that we don't support promotion. The rest of the aggregate functions were written in that way.

Or is this only should be an exception for `sum`? If so why?

jvanstraten:
I'm with @vibhatha on this one. See also #251. I also listed `sum(i8) -> i64` there specifically; I didn't use it as primary example but the same logic applies. tl;dr of my reasoning: currently, there's no way for a producer to specify what it wants the return type to be. Maybe the particular system defines an `i128`, or a bigint, or maybe it wants some decimal; after all, `i64` can still overflow. Or, in the opposite direction, maybe `i32` is good enough. The only way to deal with all these special cases once and for all is to just not do implicit promotion, and require the consumer to cast to the desired type first.

I've always found the fp32 -> fp64 promotion to be particularly weird. If you're using fp32 for something you're probably doing that for performance reasons and don't just want a random promotion to occur, and while precision diminishes, floats can deal with enormous values, so you're much less likely to overflow (or rather, receive +/-infinity) than you are with integers regardless of whether you're using fp32 or fp64.

It *is* a breaking change, though. I can't deny that. As a compromise we could also just deprecate the promoting variants or at least mark that they're considered to be a special case, but I'd much rather things just be consistent.

westonpace:
Ah, I answered on the Acero issue so I'll copy my reasoning here but I'm -1 on this change at the moment.  Aggregate functions are different than scalar functions because there isn't very much output and so it almost never matters how large that output is (outside of some group by situations where you have millions of groups but those aren't usually combined with `sum`).

Here's my comment from Acero:

Aggregate functions typically have very small outputs compared to the input (e.g. the sum of 1 million rows is a single value) and so it very often makes sense for the output type to be larger than the input type.

One could argue that you can simply cast beforehand. However, you would have to cast the entire array of inputs (e.g. the 1 million rows) and this could be rather costly.

Finally, we are mirroring SQL here (which is not, by itself, necessarily a good thing, but it is worth noting). From the [postgres docs](https://www.postgresql.org/docs/8.2/functions-aggregate.html) for sum the return type is:

    bigint for smallint or int arguments, numeric for bigint arguments, double precision for floating-point arguments, otherwise the same as the argument data type

vibhatha:
@westonpace I understand your point and didn't thought about it in this angle. For the record, we have all other aggregate functions written this way. And if we standardize aggregate function usage it means either way the change required is going to be a breaking change. Please correct me if I am wrong. As @jvanstraten said, we still have to work on the function definitions and finalize some of these cases.

westonpace:
> For the record, we have all other aggregate functions written this way.

I disagree.  Some aggregate functions (e.g. `std_dev`, `variance`, `corr` only accept floating point values, `quantile` returns a list)  I do not think there is an expectation of consistency among functions.

I think `sum` and `product` are fundamentally different than `min`, `max`, `median`, `avg` because the output for `sum` and `product` is usually highly affected by the number of rows and the output for the others does not depend on the number of rows at all.

When in doubt we should use existing engines as precedent:

| Engine | Behavior |
| --- | --- |
| [Postgres](https://www.postgresql.org/docs/9.5/functions-aggregate.html) | Promotes int8/int16/int32 to int64.  Promotes int64 to Decimal (avoids overflow) |
| [SQL Server](https://learn.microsoft.com/en-us/sql/t-sql/functions/sum-transact-sql?view=sql-server-ver16) | Promotes int8/int16 to int32.  Leaves int32 and int64 alone. |
| [Redshift](https://docs.aws.amazon.com/redshift/latest/dg/r_SUM.html) | Promotes all integer types to int64 |
| [Presto](https://prestodb.io/docs/current/functions/aggregate.html#sum) | Leaves types as-is |
| Sqlite | N/A (Sqlite uses 8-byte integers for all arithmetic) |

So it appears there is some precedent for treating `sum` as a special function.  For the record, all of these engines return the same type as the input for `min` and `max`.  Sadly, there does not appear to be any consistency across the engines on what exactly the special `sum` rules should be :)

westonpace:
Also, whatever we decide to do for `sum` we should also do for `product`

jvanstraten:
> Aggregate functions typically have very small outputs compared to the input (e.g. the sum of 1 million rows is a single value) and so it very often makes sense for the output type to be larger than the input type.

I agree with your idea here. However, if an engine has first-class support for an `i128` type or an unlimited-size integer or whatever, it would make sense for them to use that one, not `i64`. Such an engine could add a `sum(i128) -> i128` function implementation (or we could technically add it to the spec later) without conflict, but can't add `sum(i64) -> i128` without a breaking change. Well, okay, it can, because functions are scoped by URI and they can't adjust core extensions in the first place, but if *we* ever embrace a type like `i128` or arbitrarily-sized numbers we'd have to break sum or be hugely inconsistent.

I discussed this (among many other things) verbally with @cpcloud last Thursday, and he suggested to just remove the promoting variants and leave only the `sum(i64) -> i64` variant for integers, which I think is a sensible enough middle ground. At that point you could argue to just leave the promoting integer implementations because they're not in the way, but I'd at least want to see a deprecation notice then.

For floating point though, I strongly believe that `sum(fp32) -> fp32` has a right to exist. Even half-precision floats can be massive -- you just lose precision. All floating point numbers suffer from this, `fp64` is no exception. In fact, any IEEE floating point type can be *infinitely* large, and therefore could be argued to not be capable of overflow at all. Making matters worse, we don't specify the minimum size of the accumulator register for the sum, nor the order of computation. For example, I could sum 2^-50, 2^-50, 2^-49, 2^-48, ..., 2^49, and 2^50 and get the correct 2^51 result with just fp32, but swap the order and it gets much more tricky due to precision loss (it might still give the same value, but proving it does would be much more difficult). Usually, accumulators and floating point registers behaving like `fp64` are actually ~80 bits in processors so precision is much less likely to degrade unexpectedly like this; they only reduce when writing to memory.

Anyway, usually the reason for choosing `fp32` over `fp64` is because you've already made the performance vs precision tradeoff and found `fp32` to be precise enough for what you're doing. An engine shouldn't then randomly override that decision and give you a less performant solution without asking. A user-facing SQL-esque interface might still want to do that promotion by default as a service to users who don't know or care about this stuff, but Substrait isn't user-facing. Such logic belongs at a higher abstraction layer.

> One could argue that you can simply cast beforehand. However, you would have to cast the entire array of inputs (e.g. the 1 million rows) and this could be rather costly.

Any and every expression could be rather costly if implemented naively. Heck, I could argue that the cast should cost exactly nothing on a 64-bit processor, because the registers are already 64 bits in size. You could then argue that there's probably a SIMD instruction in x86 that can load a number of 8-bit values from memory and sum them in a single instruction or something. Then someone else could come in and argue something for GPUs. But this is all beyond the point, because these are all implementation details. Substrait should not be opinionated about how someone implements a particular behavior description, just like how SQL engines are expected to optimize the query provided by the user to the best of their abilities.

If an engine can do the `sum(cast<i64>(x))` pattern with a single kernel faster than by doing the operations sequentially or independently, it should recognize that pattern and do it. It should also publish its own extension function that captures this behavior exactly. But Substrait should not have this behavior built into it because it *may* make sense to one or more engines, while ignoring the fact that other engines could forward the exact same set of arguments for a different return type.

> Finally, we are mirroring SQL here (which is not, by itself, necessarily a good thing, but it is worth noting). From the postgres docs for sum the return type is: [...]

I don't see how this helps your argument. We can't mirror Postgres exactly here even if we would decide that it is the golden standard, because it uses a different type system. In general, Substrait has a much stronger type system than SQL, so most of these things are and will always be naturally different in Substrait unless we start over with a more loose and dynamic type system.

> [table] So it appears there is some precedent for treating sum as a special function.

Again, I don't see how this helps your argument. What I see here is that everyone does this differently, so this only strengthens *my* argument that Substrait should not be opinionated about any particular implementation. The only way for Substrait to represent all of them is by means of a `sum(cast(x))` pattern, or (hypothetically) by adding a type argument to `sum` for the return type and exhaustively specifying all the combinations that make sense. A type trait system could also be massively helpful in conjunction with that (it would prevent the exhaustive list and would let engines implicitly extend the core functions based on their type extensions), but we don't have anything like that even in sketch form yet.

EpsilonPrime:
@vibhatha It seems like we aren't going to pursue this option.  Should we close?

vibhatha:
> @vibhatha It seems like we aren't going to pursue this option. Should we close?

This is accurate. Let's close this one.

jacques-n:
Closing per conversation

PR #336, opened 2022-09-19
feat(proto): add physical hashJoin & mergeJoin physical relations

chaojun-zhang:
This PR includes hashJoin and mergeJoin physical plan relations according the spec

### DIFF
--- a/proto/substrait/algebra.proto
+++ b/proto/substrait/algebra.proto
@@ -365,6 +365,10 @@ message Rel {
     ExtensionMultiRel extension_multi = 10;
     ExtensionLeafRel extension_leaf = 11;
     CrossRel cross = 12;
+
+    //Physical relations
+    HashJoinRel hash_join = 13;
+    MergeJoinRel merge_join = 14;
   }
 }
 
@@ -477,6 +481,60 @@ message WriteRel {
     // subplans in the body of the Rel input) and return those with anounter PlanRel.relations.
     OUTPUT_MODE_MODIFIED_TUPLES = 2;
   }
+}
+
+// The hash equijoin join operator will build a hash table out of the right input based on a set of join keys.
+// It will then probe that hash table for incoming inputs, finding matches.
+message HashJoinRel {
+  RelCommon common = 1;
+  Rel left = 2;
+  Rel right = 3;
+  repeated Expression.FieldReference left_keys = 4;
+  repeated Expression.FieldReference right_keys = 5;
+  Expression post_join_filter = 6;
+
+  JoinType type = 7;
+
+  enum JoinType {
+    JOIN_TYPE_UNSPECIFIED = 0;
+    JOIN_TYPE_INNER = 1;
+    JOIN_TYPE_OUTER = 2;
+    JOIN_TYPE_LEFT = 3;
+    JOIN_TYPE_RIGHT = 4;
+    JOIN_TYPE_LEFT_SEMI = 5;
+    JOIN_TYPE_RIGHT_SEMI = 6;
+    JOIN_TYPE_LEFT_ANTI = 7;
+    JOIN_TYPE_RIGHT_ANTI = 8;
+  }
+
+  substrait.extensions.AdvancedExtension advanced_extension = 10;
+}
+
+// The merge equijoin does a join by taking advantage of two sets that are sorted on the join keys.
+// This allows the join operation to be done in a streaming fashion.
+message MergeJoinRel {
+  RelCommon common = 1;
+  Rel left = 2;
+  Rel right = 3;
+  repeated Expression.FieldReference left_keys = 4;
+  repeated Expression.FieldReference right_keys = 5;
+  Expression post_join_filter = 6;
+
+  JoinType type = 7;
+
+  enum JoinType {
+    JOIN_TYPE_UNSPECIFIED = 0;
+    JOIN_TYPE_INNER = 1;
+    JOIN_TYPE_OUTER = 2;
+    JOIN_TYPE_LEFT = 3;
+    JOIN_TYPE_RIGHT = 4;
+    JOIN_TYPE_LEFT_SEMI = 5;
+    JOIN_TYPE_RIGHT_SEMI = 6;
+    JOIN_TYPE_LEFT_ANTI = 7;
+    JOIN_TYPE_RIGHT_ANTI = 8;
+  }
+
+  substrait.extensions.AdvancedExtension advanced_extension = 10;
 }
 
 // The argument of a function
--- a/site/docs/relations/physical_relations.md
+++ b/site/docs/relations/physical_relations.md
@@ -17,14 +17,14 @@ The hash equijoin join operator will build a hash table out of the right input b
 
 ### Hash Equijoin Properties
 
-| Property            | Description                                                  | Required                 |
-| ------------------- | ------------------------------------------------------------ | ------------------------ |
-| Left Input          | A relational input.                                          | Required                 |
-| Right Input         | A relational input.                                          | Required                 |
-| Join Expression     | A boolean condition that describes whether each record from the left set "match" the record from the right set. The condition must only include the following operations: AND, ==, field references, is not distinct from. Field references correspond to the direct output order of the data. | Required.                |
+| Property            | Description                                                                                                                                                                                                            | Required                 |
+|---------------------|------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|--------------------------|
+| Left Input          | A relational input.(Probe-side)                                                                                                                                                                                        | Required                 |
+| Right Input         | A relational input.(Build-side)                                                                                                                                                                                        | Required                 |
+| Left Keys           | References to the fields to join on in the left input.                                                                                                                                                                 | Required                 |
+| Right Keys          | References to the fields to join on in the right input.                                                                                                                                                                | Required                 |
 | Post Join Predicate | An additional expression that can be used to reduce the output of the join operation post the equality condition. Minimizes the overhead of secondary join conditions that cannot be evaluated using the equijoin keys. | Optional, defaults true. |
-| Join Type           | One of the join types defined in the Join operator.          | Required                 |
-
+| Join Type           | One of the join types defined in the Join operator.                                                                                                                                                                    | Required                 |
 
 
 ## NLJ Operator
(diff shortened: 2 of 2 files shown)


chaojun-zhang:
@jvanstraten , sorry for my mistake  that  close the original PR , I can't reopen it so raise a new PR here

westonpace:
Thank you for contributing these.  I'm curious about the duplication here, though I don't know if it's a bad thing entirely.

In #331 we decided that trying too hard to avoid duplication (e.g. by basing as-of join on a logical join) was probably a bad idea.  However, in this PR we are providing direct physical specializations of the logical join.  Do you think it would make sense to use inheritance here?  Is there a reason you chose not to?  Can you imagine the physical specializations diverging from the logical specifications (other than just adding on new fields) at some point?

jvanstraten:
I don't know if this is true, but if all physical joins are likely to use the same fields, shouldn't the physical implementation just be an optional field on `JoinRel`?

If not, for some of the cases like this for the validator-specific protos (especially scalar vs aggregate vs window functions) I made separate messages for the common fields, in this case for example `JoinCommon` or something. That way, if we add logical-specific or physical-specific fields later, we can do that outside that common structure, but also avoid duplication. It would be a breaking change for `JoinRel` to add this indirection, though.

chaojun-zhang:
To avoid duplicate, I think we may have two options: 

- option 1: add new field inside JoinRel like below to distinguish logical join , hash join , merge join  and nested loop join
 ```
JoinKind kind = 7;

  enum JoinKind {
    JOIN_KIND_LOGICAL = 0;
    JOIN_KIND_HASH = 1;
    JOIN_KIND_MERGE = 2;
    JOIN_KIND_NESTED_LOOP = 3;
  }
```
- option 2: extract  join fields to JoinCommon message, and use inheritance for logical join ,hash join ,merge join and nested loop join 

I'm not sure which option is prefer to,  my thinking the option 1  may be better  since these join operator share the same attributes  according to the spec in substrait.io?

File: proto/substrait/algebra.proto [hunk 1]
```diff
@@ -477,6 +481,60 @@ message WriteRel {
   }
 }
 
+// The hash equijoin join operator will build a hash table out of the right input based on a set of join keys.
+// It will then probe that hash table for incoming inputs, finding matches.
+message HashJoinRel {
+  RelCommon common = 1;
+  Rel left = 2;
+  Rel right = 3;
+  Expression expression = 4;
```
jacques-n, line 490: I think we should specialize these fields for what a dumb consumer would want. I think that is probably:

FieldReference[] left_keys (hash table)
FieldReference[] left_keys
Expression during_join_filter (join conditions)
Expression post_join_filter (afterwards)

We should also validate that the join types are valid in this situation.

proto/substrait/algebra.proto
```diff
@@ -477,6 +481,60 @@ message WriteRel {
...
+  substrait.extensions.AdvancedExtension advanced_extension = 10;
+}
+
+// The merge equijoin does a join by taking advantage of two sets that are sorted on the join keys.
+// This allows the join operation to be done in a streaming fashion.
+message MergeJoinRel {
+  RelCommon common = 1;
+  Rel left = 2;
+  Rel right = 3;
+  Expression expression = 4;
```
jacques-n, line 517: Same as above, I think we need to specialize for dumb consumers when producing physical rels like this.

jacques-n (changes requested):
In general, I think we should avoid collapsing these. I think they actually need to be different. They have different physical property maintenance characterististics and will probably develop in different directions.

Once we get a good definition, let's make sure we also update the associated site docs.

proto/substrait/algebra.proto
westonpace: I like this idea.  @jvanstraten @vibhatha and I have struggled at times with the mismatch that Acero expects simple field references (except in the project relation) in many places the official spec contains full expressions.  This comment helped me crystallize that this is because those spots with full expressions are simply more logical than physical.

jvanstraten: +1; I don't personally know what it should look like, but for physical relations I wasn't expecting expressions at all. I'd go as far as to say that the and-tree-of-equalities expression should be entirely reduced to just a bunch of FieldReferences, so a consumer doesn't have to match and unpack something involving something as complex as extension functions just to understand what the producer meant.

westonpace: > the and-tree-of-equalities expression should be entirely reduced to just a bunch of FieldReferences

For reference, in Acero, we currently only support a vector of field references (one from the left and one from the right) and, for each field, an enum of `equal` or `is_not_distinct_from`.  Though that would limit this to equi-joins.

proto/substrait/algebra.proto
chaojun-zhang: updated

proto/substrait/algebra.proto
chaojun-zhang: Reading  [velox ](https://github.com/facebookincubator/velox)code,  hash join and merge join also support a vector of field references for 'lefyKeys' and 'rightKeys', but how can I pass the [outputType ](https://github.com/facebookincubator/velox/blob/main/velox/core/PlanNode.h#L1086)from proto , how can I pass it from proto ?

proto/substrait/algebra.proto
chaojun-zhang: updated

proto/substrait/algebra.proto
jacques-n: I think it is valuable to support the equal/is not distinct concepts for each of the key pairings here. Probably makes most sense to use a message similar to what it sounds like acero does (e.g. left_key, right_key, equality_condition)

WRT to join type, I'm inclined to only introduce the join types people are using for each of these. We can introduce more as people have them. As such, I'd suggest we have separate enums for each (HashJoinType, MergeJoinType, etc). What do others think?

jvanstraten: > As such, I'd suggest we have separate enums for each (HashJoinType, MergeJoinType, etc). What do others think?

I agree with the thought because they seem different enough (and also agree with your other paragraph). However, I'd personally opt against adding new enum types in the protos, because enums in general are pretty fragile compared to everything else in protobuf (lack of proper scoping, annoyingly long names in JSON, representing unknown values with invalid enum values in C, easy for consumers to forget to handle unrecognized values, can't add options that contain values later like you can with e.g. Rust enums, etc). A message containing a `oneof`s with (initially) empty messages seems superior to me in every way. I don't think it even costs more in the binary serialization.

jacques-n: I'm actually pretty supportive of this pattern generally. It feels much more like rust enums.

site/docs/relations/physical_relations.md
```diff
@@ -17,13 +17,15 @@ The hash equijoin join operator will build a hash table out of the right input b
...
+|---------------------|-------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|----------------------------------------------------------------------------|
+| Left Input          | A relational input.                                                                                                                                                                                                                                                                             | Required                                                                   |
+| Right Input         | A relational input.                                                                                                                                                                                                                                                                             | Required                                                                   |
```
westonpace, line 23: Since this is a physical relation do we want to explicitly declare that the left side is the build side and the right side is the probe side (or vice versa)?  It looks like you are already making this assumption in your description of left keys and right keys.

site/docs/relations/physical_relations.md
```diff
@@ -17,13 +17,15 @@ The hash equijoin join operator will build a hash table out of the right input b
...
+ | Left Keys           | A collection of fields reference in probe table                                                                                                                                                                                                                                                 | Optional, defaults to blank if during join predicate is specified          |
```
westonpace, line 24:
```suggestion
 | Left Keys           | References to the fields to join on in the left input.                                                                                                                                                                                                                                                 | Optional, defaults to blank if during join predicate is specified          |
```

Same change for right keys

site/docs/relations/physical_relations.md
```diff
@@ -17,13 +17,15 @@ The hash equijoin join operator will build a hash table out of the right input b
...
+ | Right Keys          | A collection of fields reference in build table                                                                                                                                                                                                                                                 | Optional, defaults to blank if during join predicate is specified          |             
+| During Join Predicate | A boolean condition that describes whether each record from the left set "match" the record from the right set. The condition must only include the following operations: AND, ==, field references, is not distinct from. Field references correspond to the direct output order of the data.  | Optional, defaults true if left keys and right keys are both specified.    |
```
westonpace, line 26: I don't see this in the proto definition

proto/substrait/algebra.proto
westonpace: > updated, do we need to extract JoinType as shared message?


I think that would be a good idea.

westonpace (commented):
This review is long overdue.  I have a few more questions.

site/docs/relations/physical_relations.md
chaojun-zhang: I think we can leave it to the user to decide which side is probe or build , your thoughts?

site/docs/relations/physical_relations.md
chaojun-zhang: Removed

site/docs/relations/physical_relations.md
westonpace: Picking the correct side is generally an optimizer task (requires statistics / heuristics).  My assumption for more physical nodes is that they've already gone through an optimizer.

For a single optimizer/consumer pair this wouldn't be a problem.  They can both agree offline which side should be build and probe.  However, imagine an engine-independent substrait optimizer created this plan and hoped for it to be consumed by many different engines.  In that case the different consumers might pick left/right build/probe differently.

That being said, I haven't worked much with optimizers.  CC @jacques-n who has more experience here.

chaojun-zhang:
Please ignore this PR since lots of commit message can't pass the Lint check

chaojun-zhang:
@jacques-n @westonpace Please take a further review and help trigger the check again.

site/docs/relations/physical_relations.md
```diff
@@ -17,14 +17,14 @@ The hash equijoin join operator will build a hash table out of the right input b
...
+| Property            | Description                                                                                                                                                                                                             | Required                 |
+|---------------------|-------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|--------------------------|
+| Left Input          | A relational input.                                                                                                                                                                                                     | Required                 |
+| Right Input         | A relational input.                                                                                                                                                                                                     | Required                 |
```
westonpace, line 23: Either we should assign these as probe and build (my preference would be to be explicit) or we should specifically explain that consumers should assign probe & build based on whatever criteria they want.  As it is, I think this is ambiguous.

chaojun-zhang: I found that the documentation has mentioned that the right input is build-side, see [this](https://substrait.io/relations/physical_relations/#hash-equijoin-operator)

I also updated in this section [https://substrait.io/relations/physical_relations/#hash-equijoin-properties]

chaojun-zhang:
@westonpace I have update the document and help trigger the check again.

westonpace:
:( It seems our release script is failing (unrelated to this change) I will see if I can fix that real quick.

westonpace:
Fixed.  Please rebase and I will run again.

westonpace (dismissed):

westonpace:
This looks good to me.  Will also need approval from @jacques-n or Jacques

chaojun-zhang:
@westonpace @jacques-n I have rebased and please help trigger the check again

chaojun-zhang:
@westonpace PR rebased, Please take a look again. Thanks!

jacques-n (approved):
LGTM. With @westonpace approval, I'm going to merge.

PR #337, opened 2022-09-19
fix: change temporal add and subtract

rok:
BREAKING CHANGE: change return type of temporal `add` and adds `add/subtract` with `timezone`.

### DIFF
--- a/extensions/functions_datetime.yaml
+++ b/extensions/functions_datetime.yaml
@@ -35,7 +35,12 @@ scalar_functions:
         return: i64
   -
     name: "add"
-    description: Add an interval to a date/time type.
+    description: >-
+      Add an interval to a date/time type.
+
+      Timezone strings must be as defined by IANA timezone database (https://www.iana.org/time-zones).
+      Examples: "Pacific/Marquesas", "Etc/GMT+1".
+      If timezone is invalid an error is thrown.
     impls:
       - args:
           - name: x
@@ -48,7 +53,10 @@ scalar_functions:
             value: timestamp_tz
           - name: y
             value: interval_year
-        return: timestamp
+          - name: timezone
+            description: Timezone string from IANA tzdb.
+            value: string
+        return: timestamp_tz
       - args:
           - name: x
             value: date
@@ -66,7 +74,7 @@ scalar_functions:
             value: timestamp_tz
           - name: y
             value: interval_day
-        return: timestamp
+        return: timestamp_tz
       - args:
           - name: x
             value: date
@@ -91,7 +99,12 @@ scalar_functions:
         return: interval_year
   -
     name: "subtract"
-    description: Subtract an interval from a date/time type.
+    description: >-
+      Subtract an interval from a date/time type.
+
+      Timezone strings must be as defined by IANA timezone database (https://www.iana.org/time-zones).
+      Examples: "Pacific/Marquesas", "Etc/GMT+1".
+      If timezone is invalid an error is thrown.
     impls:
       - args:
           - name: x
@@ -104,6 +117,15 @@ scalar_functions:
             value: timestamp_tz
           - name: y
             value: interval_year
+        return: timestamp_tz
+      - args:
+          - name: x
+            value: timestamp_tz
+          - name: y
+            value: interval_year
+          - name: timezone
+            description: Timezone string from IANA tzdb.
+            value: string
         return: timestamp_tz
       - args:
           - name: x


jvanstraten:
Just to be clear, the return type changes are merely fixes for rather obvious copy-paste mistakes, but are nevertheless breaking changes.

I failed to consider that the addition of the timezone argument is also a breaking change. Maybe that should be in yet another PR, just so we can be clearer about what the PRs actually change, i.e. something like "feat: add timezone argument to timezone_tz +/- interval_year function" and "fix: correct return types in datetime add function". I don't know if that's helpful or just adds clutter, though.

jvanstraten (dismissed):
If we don't split this up this looks good to me, but requires an SMC vote according to the proposed governance document. @jacques-n @cpcloud @westonpace

extensions/functions_datetime.yaml
```diff
@@ -104,6 +117,9 @@ scalar_functions:
             value: timestamp_tz
           - name: y
             value: interval_year
+          - name: timezone
```
jacques-n, line 120: I'm inclined to suggest that we just make a new variation that has timezone rather than change the existing here.

extensions/functions_datetime.yaml
```diff
@@ -66,7 +74,7 @@ scalar_functions:
             value: timestamp_tz
           - name: y
             value: interval_day
-        return: timestamp
+        return: timestamp_tz
```
jacques-n, line 77: good catch, thanks.

extensions/functions_datetime.yaml
```diff
@@ -48,7 +53,10 @@ scalar_functions:
             value: timestamp_tz
           - name: y
             value: interval_year
-        return: timestamp
+          - name: timezone
```
jacques-n, line 56: As below, I'm inclined to introduce a new variant as opposed to change the existing.

jacques-n: Also, I think we should probably make these timezone arguments constant values.

extensions/functions_datetime.yaml
jvanstraten: The reason we added it is because, as @rok noted, adding a month/year interval to an instant is ill-defined without timezone information. For example, an instant may be February 28th in one timezone and March 1st in another, so to add a month you'd have to add either 28 or 31 days.

That being said, we could still keep the original one with a note stating that it assumes UTC.

extensions/functions_datetime.yaml
jvanstraten: +1 on marking them as constant. @rok, that's just adding `constant: true` at the same level as the return field.

extensions/functions_datetime.yaml
jacques-n: > That being said, we could still keep the original one with a note stating that it assumes UTC.

Yes, that is what I was requesting.

julianhyde: If someone is adding 3 months to an instant, what makes you think that they want to do it in UTC? I don't think defaulting to UTC helps anyone.

julianhyde: @jvanstraten I pretty much agree. As you may know, I am one of these extremists who believes that instant, local time, and time-with-time-zone should be kept strictly separate, and therefore I would be inclined to remove the original method.

extensions/functions_datetime.yaml
rok: Added. Are you sure that is the right level?

rok:
Pushed the feedback suggestions and rebased. @jvanstraten

extensions/functions_datetime.yaml
jvanstraten: Uh, no. Not sure what I was thinking there, sorry... :sweat_smile: Like this:

```suggestion
          - name: timezone
            constant: true
```

instead of the one I told you to add before.

rok: Fixed :).

extensions/functions_datetime.yaml
westonpace: I'm inclined to agree with @julianhyde .  I'd rather only add new functions and implementations when needed and not for the sake of convenience overloads (I also agree that the convenience here is rather dubious).  Otherwise our set of functions gets unwieldy.

There will be a time when we are forced to leave functions around for deprecation reasons but I don't think we are there yet.

@jacques-n are you -1 on removing the old function?  Or was that simply a recommendation?

westonpace:
Why is the timezone not required when adding a day?  If a time zone has DST that might mean adding 23 or 25 hours correct?

rok:
> Why is the timezone not required when adding a day? If a time zone has DST that might mean adding 23 or 25 hours correct?

Forgive me, it's been a while since I looked at this. I vaguely recall (and perhaps incorrectly) that the day interval is a calendrical unit as opposed to a physical one so want to subtract calendar days not physical units.

westonpace:
> Forgive me, it's been a while since I looked at this. I vaguely recall (and perhaps incorrectly) that the day interval is a calendrical unit as opposed to a physical one so want to subtract calendar days not physical units.

Hmm, refreshing my memory I do see this in the spec for interval_day:

> Usually stored as separate integers for various components, but only the total number of microseconds is significant, i.e. 1d 0s is considered equal to 0d 86400s.

I suppose that means that we, indeed, do not need a time zone.

extensions/functions_datetime.yaml
```diff
@@ -48,7 +53,11 @@ scalar_functions:
...
+            constant: true
```
westonpace, line 57: Remove.  Pending ML discussion on constant arguments in scalar functions.

extensions/functions_datetime.yaml
```diff
@@ -35,7 +35,12 @@ scalar_functions:
         return: i64
   -
     name: "add"
-    description: Add an interval to a date/time type.
+    description: >-
+      Add an interval to a date/time type.
+
+      Timezone strings must be as defined by IANA timezone database (https://www.iana.org/time-zones).
+      Examples: "Pacific/Marquesas", "Etc/GMT+1".
+      If timezone is not present then UTC is assumed. If timezone is invalid an error is thrown.
```
westonpace, line 43: How would it be possible now for the timezone to not be present?

rok: By "present" I mean provided as a parameter. I'll rephrase.

extensions/functions_datetime.yaml
rok: Removed.

rok:
Thanks for the review @westonpace. I think I addressed all the points.

extensions/functions_datetime.yaml
westonpace: In Substrait, currently, arguments are always required and only options are optional.  So there is no way to call this function without including the timezone as a parameter.  I think this is fine, and we should just remove the "default to UTC" wording.  If a producer wants to default to UTC then they can do so very easily.  However, the plan should be as simple and explicit as reasonably possible.

rok: That makes sense. Changed `If timezone is not provided as parameter then UTC is assumed. If timezone is invalid an error is thrown.` to `If timezone is invalid an error is thrown.`

westonpace (approved):

westonpace:
Thanks for figuring this out.

PR #338, opened 2022-09-20
docs: clarify correlation measure

thisisnic:
This PR clarifies the correlation measure.  I half wondered if this should be an option (i.e. Pearson's/Kendall/Spearman) but Postgres seems to use the Pearson Correlation Coefficient, but I concluded that adding in the possible choice might be a bit niche for what we're trying to do here.


jvanstraten (approved):
LGTM, if we want to make this an option we can always do so later (preferably after having figured out a way to avoid every option addition being a breaking change).

PR #340, opened 2022-09-21
chore: improve coalesce nullability handling

rui-mo:
Improve coalesce nullability handling by describing the semantics of coalesce more accurately.


jvanstraten:
Coalesce is already covered by a function: https://github.com/substrait-io/substrait/blob/9a921a0bfde74b41efea73a360c6ef7fb8651f06/extensions/functions_comparison.yaml#L204-L216

Do you have a particular reason why it should be a special-cased expression instead?

I could forward a small reason actually, being that the nullability of the return type of coalesce should match the last argument. I don't think that can be described with a variadic function. Right now, the function simply returns nullable if and only if any argument is nullable.

If we go for the expression version, the coalesce function should probably be removed. Also, documentation for the addition should be added [here](https://github.com/substrait-io/substrait/blob/main/site/docs/expressions/specialized_record_expressions.md).

rui-mo:
@jvanstraten Sorry I did not notice the coalesce function in yaml. Maybe this PR can be closed if the function already exists.
> the nullability of the return type of coalesce should match the last argument

About the nullability of the return type, from my view, below sentence may be more reasonable because if any argument is not nullable, the returned column would not contain null value.

> Coalesce is nullable if all of its children are nullable, or if it has no children.

How do you think so? If using Scalar function, maybe we can set the correct nullabilty to the output type.

jvanstraten:
> Coalesce is nullable if all of its children are nullable, or if it has no children.

For functions I'd use the term "arguments" rather than "children", but otherwise I think that logic is sound and boils down to the same thing I said for sane usages of coalesce. After all, it makes no sense for any argument except the last to be non-nullable, because any arguments that follow would never be used. I'm also not sure what a coalesce call with zero arguments would be useful for, since it would only ever return null (and it would also problematic with Substrait's type system, in which null is typed, so type information is always needed from at least one argument in this case).

In any case, the type derivation system isn't expressive enough to represent either of our interpretations of nullability behavior. With the derivation overhaul I'm working on (but that I've not yet completed enough to propose) we could get a bit closer, but still not all the way:

```yaml
- name: "coalesce" 
  description: >- 
    Evaluate arguments from right to left (!) and return the first argument encountered
    that is not null. Once a non-null argument is found, the remaining arguments are
    not evaluated. 
  
    If all arguments are null, return null. 
  impls: 
  - args: 
    - value: T?nullable
    - value: T?
    variadic:
      parameterConsistency: CONSISTENT
    return: T?nullable
    nullability: DISCRETE
```

Here, the syntax `T?nullable` is new and means nullable or non-nullable (with nullability captured as a boolean by the `nullable` name). So this means that the return type will be exactly the same as the first argument (including nullability), and the rest of the arguments must be nullable. Problem is, of course, that the arguments would need to be evaluated right to left rather than left to right for this to make sense. Not a problem for the semantics of Substrait, but certainly unconventional and bound to cause confusion and consumer/producer conversion ugliness.

We could of course extend the semantics of variadic functions to allow the variadic argument slot to be any of the arguments rather than only the last to solve that, but that seems horribly overkill for this. Or we could just define coalesce as a two-argument function and require them to be nested for more than 2 inputs (coalesce is left-associative after all), but that might be a bit too much of the purist approach.

So, the more I think about it, the more I think it might make sense to just special-case coalesce as a dedicated expression type and deprecate the function. In which case I think what you wrote is fine, though the type field is redundant and could probably just be removed. Other than that, I think it just needs a bit of documentation, and a deprecation notice in the coalesce function description.

rui-mo:
@jvanstraten Thanks a lot for your great thinking on this!

> I think it just needs a bit of documentation, and a deprecation notice in the coalesce function description.

I noticed there was a description on this [page](https://substrait.io/extensions/functions_comparison/#coalesce), and am I capable of changing it?

gforsyth:
@rui-mo -- sure, those descriptions are pulled directly out of the extension yaml files.  For the `coalesce` description you linked above, see https://github.com/substrait-io/substrait/blob/main/extensions/functions_comparison.yaml#L205-L210

rui-mo:
> @rui-mo -- sure, those descriptions are pulled directly out of the extension yaml files. For the `coalesce` description you linked above, see https://github.com/substrait-io/substrait/blob/main/extensions/functions_comparison.yaml#L205-L210

@gforsyth Thank you. Yes, I removed the description for `coalesce` in this PR. Is that OK?

jvanstraten:
What I meant was just adding something along the lines of "This function is deprecated; use the specialized coalesce expression type instead." Actually removing it would make this a breaking change, which is a bit more difficult procedurally because it requires an SMC vote. I'm not opposed to it, though. @jacques-n @westonpace @cpcloud 

Reminder that the new expression type should also be added to the docs, preferably with some explanation about what the expected argument nullability should be and what nullability the return type would have based on that. That would go in [this file](https://github.com/substrait-io/substrait/blob/main/site/docs/expressions/specialized_record_expressions.md).

rui-mo:
> What I meant was just adding something along the lines of "This function is deprecated; use the specialized coalesce expression type instead." Actually removing it would make this a breaking change, which is a bit more difficult procedurally because it requires an SMC vote. I'm not opposed to it, though. @jacques-n @westonpace @cpcloud
> 
> Reminder that the new expression type should also be added to the docs, preferably with some explanation about what the expected argument nullability should be and what nullability the return type would have based on that. That would go in [this file](https://github.com/substrait-io/substrait/blob/main/site/docs/expressions/specialized_record_expressions.md).

Understood. I think I can mark it as deprecated firstly to make sure do not bring sudden change for now. And will update the doc. Thanks.

rui-mo:
> Reminder that the new expression type should also be added to the docs, preferably with some explanation about what the expected argument nullability should be and what nullability the return type would have based on that.

@jvanstraten I updated the doc and fixed the comments. Would you like to review again?

site/docs/expressions/specialized_record_expressions.md
```diff
@@ -68,10 +68,15 @@ Multi Value:
 For single value expressions, these are a compact equivalent of `expression = value1 OR expression = value2 OR .. OR expression = valueN`. When using an expression of this type, two things are required; the types of the test expression and all value expressions that are related must be of the same type. Additionally, a function signature for equality must be available for the expression type used.
 
 
+## Coalesce Expression
 
+Coalesce expression evaluates arguments from left to right and return the first argument encountered that is not null. Once a non-null argument is found, the remaining arguments are not evaluated. If all arguments are null, return null.
 
+The arguments could be both nullable and non-nullable. If there is any argument that is non-nullable, the return type would be non-nullable. If all the arguments are nullable, or there is no argument, the return type would be nullable.
```
jvanstraten, line 75: I'd add something stating that, nullability aside, the types must match, and that at least one argument is required (our null value is typed, so there is no generic null for an empty coalesce() expression to return). I'd also use less passive language in general, but that's more of a nit.

```suggestion
The arguments may be any combination of nullable and non-nullable, but must otherwise be of the same type. If any argument is non-nullable, the return type is non-nullable. If all the arguments are nullable, or there is no argument, the return type is be nullable. Nullability aside, the return type matches the type of the arguments. At least one argument is required, in order for the return type to be defined.
```

jvanstraten (commented):
Already looks good enough to me, but I think the docs could be worded a bit more precisely.

rui-mo:
@jvanstraten Thanks for your comment. Updated the doc based on your suggestion.

jvanstraten (dismissed):
LGTM now. @jacques-n @westonpace any strong opinions? This is a newly special-cased construct after all. If not I'll merge in a few days.

rui-mo:
Hi @jvanstraten, it has been a long time, and sorry for the delay on CLA. Could you help review this PR again? The CLA is ready. And I'm not sure how should I change the PR title/description to make the checks work. Thanks for your help!

westonpace:
Apologies.  The CLA is fine.

This seems to be a very small reason (nullability handling) to make coalesce a specialized expression.  Couldn't there be other expressions that share this behavior?  For example (I am making these up but I think they could be valid):

```
coalesce_min // return the smallest non-null value
coalesce_max // return the largest non-null value
```

I would rather have a special (but rarely used) syntax for this return style.  For example:

```
return: any1|? # Return type is nullable if any of the variadic inputs are nullable
```
```
return: any1&? # Return type is nullable if all variadic inputs are nullable
```

In the worst case I think we can err on the side of returning nullable even if something might be non-nullable.

rui-mo:
> This seems to be a very small reason (nullability handling) to make coalesce a specialized expression. Couldn't there be other expressions that share this behavior? For example (I am making these up but I think they could be valid):
> 
> ```
> coalesce_min // return the smallest non-null value
> coalesce_max // return the largest non-null value
> ```
> 
> I would rather have a special (but rarely used) syntax for this return style. For example:
> 
> ```
> return: any1|? # Return type is nullable if any of the variadic inputs are nullable
> ```
> 
> ```
> return: any1&? # Return type is nullable if all variadic inputs are nullable
> ```
> 
> In the worst case I think we can err on the side of returning nullable even if something might be non-nullable.

@westonpace Thank you for the great advice. I agree it is more proper to improve the return type instead of making coalesce a specialized expression. Changed to use `any1&?` in the return type.

westonpace:
We need to document the new syntax in the markdown files.  I will open a PR into your branch with a proposal.

rui-mo:
> We need to document the new syntax in the markdown files. I will open a PR into your branch with a proposal.

@westonpace Thank you! Merged into this PR.

westonpace:
I think this is good now.  +1 from me.

@jacques-n / @cpcloud any opinions here?  I believe you both have had to parse these type strings in the past.

westonpace:
Also CC @mbrobbel as I imagine there will be implications for the validator.

EpsilonPrime (approved):

EpsilonPrime:
@westonpace @mbrobbel Are we good to merge this now?

EpsilonPrime:
This was discussed in the Oct 11, 2013 community meeting (as well as the April 23, 2013 community meeting).  The plan there is to instead use the same language used to define the decimal type used instead of complicating the grammar.  There will be an issue filed to track that effort.  Once complete this PR will need to be updated to use that logic.

jacques-n:
Closing as abandon. We can reopen if becomes active.

PR #341, opened 2022-09-21
docs: remove outdated TBD and reference to iceberg in filter description of read relation

mbrobbel:
I assume that this should not be specific but generic.


jvanstraten (approved):
I'm not sure this was intended to be a reference to Apache Iceberg, but either way it's better to just clear the ambiguity.

PR #342, opened 2022-09-22
feat: add names to args at invocation and allow omitting optional args

westonpace:
BREAKING CHANGE: optional arguments may now be omitted.  However, if they are specified then they MUST be named.

### DIFF
--- a/extensions/functions_aggregate_generic.yaml
+++ b/extensions/functions_aggregate_generic.yaml
@@ -5,11 +5,11 @@ aggregate_functions:
     description: Count a set of values
     impls:
       - args:
-          - name: overflow
-            options: [SILENT, SATURATE, ERROR]
-            required: false
           - name: x
             value: any
+        options:
+          overflow:
+            values: [SILENT, SATURATE, ERROR]
         nullability: DECLARED_OUTPUT
         decomposable: MANY
         intermediate: i64
@@ -17,10 +17,9 @@ aggregate_functions:
   - name: "count"
     description: "Count a set of records (not field referenced)"
     impls:
-      - args:
-          - name: overflow
-            options: [SILENT, SATURATE, ERROR]
-            required: false
+      - options:
+          overflow:
+            values: [SILENT, SATURATE, ERROR]
         nullability: DECLARED_OUTPUT
         decomposable: MANY
         intermediate: i64
--- a/extensions/functions_datetime.yaml
+++ b/extensions/functions_datetime.yaml
@@ -6,30 +6,30 @@ scalar_functions:
     description: Extract portion of a date/time value.
     impls:
       - args:
-          - options: [ YEAR, MONTH, DAY, SECOND ]
-            name: The part of the value to extract.
-            required: true
+          - name: component
+            options: [ YEAR, MONTH, DAY, SECOND ]
+            description: The part of the value to extract.
           - name: x
             value: timestamp
         return: i64
       - args:
-          - options: [ YEAR, MONTH, DAY, SECOND ]
-            name: The part of the value to extract.
-            required: true
+          - name: component
+            options: [ YEAR, MONTH, DAY, SECOND ]
+            description: The part of the value to extract.
           - name: x
             value: timestamp_tz
         return: i64
       - args:
-          - options: [ YEAR, MONTH, DAY ]
-            name: The part of the value to extract.
-            required: true
+          - name: component
+            options: [ YEAR, MONTH, DAY ]
+            description: The part of the value to extract.
           - name: x
             value: date
         return: i64
       - args:
-          - options: [ SECOND ]
-            name: The part of the value to extract.
-            required: true
+          - name: component
+            options: [ SECOND ]
+            description: The part of the value to extract.
           - name: x
             value: time
         return: i64
--- a/proto/substrait/algebra.proto
+++ b/proto/substrait/algebra.proto
@@ -4,7 +4,6 @@ syntax = "proto3";
 package substrait;
 
 import "google/protobuf/any.proto";
-import "google/protobuf/empty.proto";
 import "substrait/extensions/extensions.proto";
 import "substrait/type.proto";
 
@@ -482,17 +481,27 @@ message WriteRel {
 // The argument of a function
 message FunctionArgument {
   oneof arg_type {
-    Enum enum = 1;
+    string enum = 1;
     Type type = 2;
     Expression value = 3;
   }
+}
 
-  message Enum {
-    oneof enum_kind {
-      string specified = 1;
-      google.protobuf.Empty unspecified = 2;
-    }
-  }
+// An optional function argument.  Typically used for specifying behavior in
+// invalid or corner cases.
+message FunctionOption {
+  // Name of the option to set. If the consumer does not recognize the
+  // option, it must reject the plan. The name is matched case-insensitively
+  // with option names defined for the function.
+  string name = 1;
+
+  // List of behavior options allowed by the producer. At least one must be
+  // specified; to leave an option unspecified, simply don't add an entry to
+  // `options`. The consumer must use the first option from the list that it
+  // supports. If the consumer supports none of the specified options, it
+  // must reject the plan. The name is matched case-insensitively and must
+  // match one of the option values defined for the option.
+  repeated string preference = 2;
 }
 
 message Expression {
@@ -647,13 +656,15 @@ message Expression {
     //    the expression in that must yield a value of a type that a function
     //    overload is defined for.
     //  - Type arguments must be bound using FunctionArgument.type.
-    //  - Required enum arguments must be bound using FunctionArgument.enum
+    //  - Enum arguments must be bound using FunctionArgument.enum
     //    followed by Enum.specified, with a string that case-insensitively
     //    matches one of the allowed options.
-    //  - Optional enum arguments must be bound using FunctionArgument.enum
-    //    followed by either Enum.specified or Enum.unspecified. If specified,
-    //    the string must case-insensitively match one of the allowed options.
     repeated FunctionArgument arguments = 4;
+
+    // Options to specify behavior for corner cases, or leave behavior
+    // unspecified if the consumer does not need specific behavior in these
+    // cases.
+    repeated FunctionOption options = 5;
 
     // Must be set to the return type of the function, exactly as derived
     // using the declaration in the extension.
@@ -679,13 +690,15 @@ message Expression {
     //    overload is defined for.
     //  - Type arguments must be bound using FunctionArgument.type, and a
     //    function overload must be defined for that type.
-    //  - Required enum arguments must be bound using FunctionArgument.enum
+    //  - Enum arguments must be bound using FunctionArgument.enum
     //    followed by Enum.specified, with a string that case-insensitively
     //    matches one of the allowed options.
-    //  - Optional enum arguments must be bound using FunctionArgument.enum
-    //    followed by either Enum.specified or Enum.unspecified. If specified,
-    //    the string must case-insensitively match one of the allowed options.
     repeated FunctionArgument arguments = 9;
+
+    // Options to specify behavior for corner cases, or leave behavior
+    // unspecified if the consumer does not need specific behavior in these
+    // cases.
+    repeated FunctionOption options = 11;
 
     // Must be set to the return type of the function, exactly as derived
     // using the declaration in the extension.
@@ -1126,13 +1139,18 @@ message AggregateFunction {
   //    overload is defined for.
   //  - Type arguments must be bound using FunctionArgument.type, and a
   //    function overload must be defined for that type.
-  //  - Required enum arguments must be bound using FunctionArgument.enum
+  //  - Enum arguments must be bound using FunctionArgument.enum
   //    followed by Enum.specified, with a string that case-insensitively
   //    matches one of the allowed options.
   //  - Optional enum arguments must be bound using FunctionArgument.enum
   //    followed by either Enum.specified or Enum.unspecified. If specified,
   //    the string must case-insensitively match one of the allowed options.
   repeated FunctionArgument arguments = 7;
+
+  // Options to specify behavior for corner cases, or leave behavior
+  // unspecified if the consumer does not need specific behavior in these
+  // cases.
+  repeated FunctionOption options = 8;
 
   // Must be set to the return type of the function, exactly as derived
   // using the declaration in the extension.
--- a/site/docs/expressions/scalar_functions.md
+++ b/site/docs/expressions/scalar_functions.md
@@ -18,12 +18,11 @@ A function is a scalar function if that function takes in values from a single r
 
 ## Argument Types
 
-There are four main types of arguments: value arguments, type arguments, required enumerations, and optional enumerations.
+There are three main types of arguments: value arguments, type arguments, and enumerations.  Every defined arguments must be specified in every invocation of the function.  When specified, the position of these arguments in the function invocation must match the position of the arguments as defined in the YAML function definition.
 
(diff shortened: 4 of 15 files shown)


westonpace:
Hopefully this breaking change won't be quite so contentious.  I imagine there are other ways we could implement this as well (e.g. we could require all arguments to be named).  Or, if we must, we could open the door to optional argument inference strategies (though I'd rather not go down that road).  I'm open to alternative suggestions.

I'm filing this because it is becoming a problem in Acero.  Acero currently expects all arguments (including optional arguments) to always be specified (I believe this is the validator's interpretation as well).  Unfortunately, neither the Isthmus or DuckDb producers behave in this way.

As an example:

```
isthmus -c "CREATE TABLE T1(foo int)" "SELECT foo + 1 FROM T1"
{
  ...
          "expressions": [{
            "scalarFunction": {
              ...
              "arguments": [{"value": { ... }}, {"value": { ... }}], // Missing overflow
  ...

}
```

westonpace:
As an added bonus, and something that was brought up in the last community sync, this change should enable new optional arguments to be added to a function without breaking backwards compatibility.

jacques-n:
I'm strongly inclined to pull the options out of the arguments array. It's a bigger breaking change but seems cleaner. When @jvanstraten mentioned names on the sync, that was more what I was expecting.

westonpace:
> I'm strongly inclined to pull the options out of the arguments array.

That's fine with me.  I don't think it's any more work either way for us.  Proto3 actually supports [maps](https://developers.google.com/protocol-buffers/docs/proto3#maps).  Would we want to express the named options as a map then?

jvanstraten:
> I'm strongly inclined to pull the options out of the arguments array. It's a bigger breaking change but seems cleaner. When @jvanstraten mentioned names on the sync, that was more what I was expecting.

+1, that's what I was intending, but I just hadn't gotten around to proposing something yet between all the reviewing and validator work. Likewise for lambdas by the way.

Longer-term, if we end up embracing the meta-type system that formed as part of my type derivation work, required enumeration arguments and type arguments could both be naturally expressed with a metavalue. You'd also get static boolean, integer, and string arguments for free, all usable within the type derivation system. They'd basically be the generalization of what most languages call template or generic arguments.

I'd then be inclined to just pull them into a different `repeated` as well, so you'd end up with something like

```proto
// A scalar function call.
message ScalarFunction {
  // Points to a function_anchor defined in this plan, which must refer
  // to a scalar function in the associated YAML file. Required; avoid
  // using anchor/reference zero.
  uint32 function_reference = 1;
  
  // Options to specify behavior for corner cases, or leave behavior unspecified
  // if not necessary.
  repeated FunctionOption options = 2;
  message FunctionOption {
    // Name of the option to set. If the consumer does not recognize the
    // option, it must reject the plan. The name is matched case-insensitively.
    string name = 1;
    
    // List of behavior options allowed by the producer. At least one must be
    // specified; to leave an option unspecified, simply don't add an entry to
    // `options`. The consumer must use the first option from the list that it
    // supports. If the consumer supports none of the specified options, it
    // must reject the plan. The name is matched case-insensitively.
    repeated string preference = 2;
  }
  
  // Generic parameters to configure the function statically. Must match the
  // signature of the function exactly.
  repeated Metavalue parameters = 3;
  
  // Runtime arguments passed to the function. The data types of the values
  // returned by the expressions must match the signature of the function
  // exactly. Constant arguments must be mapped to a Literal expression.
  repeated Expression arguments = 4;
  
  // Must be (redundantly) set to the return type of the function, exactly as
  // derived using the declaration in the extension.
  Type output_type = 5;
}
```

where `Metavalue` looks something like [this](https://github.com/substrait-io/substrait-validator/blob/a3a2fbc07761a61ec0756aec8470be005bee8e12/proto/substrait/validator/type_system.proto#L180-L203), but without the validator-specific `unresolved` option.

(note on case insensitivity: I dislike it personally, but dislike inconsistency more)

> Proto3 actually supports maps. Would we want to express the named options as a map then?

I'd prefer not to because A) I'm not sure how `prost` and other third-party protobuf implementations would deal with it (if at all) and B) it's essentially just sugar for

```proto
message MapFieldEntry {
  key_type key = 1;
  value_type value = 2;
}

repeated MapFieldEntry map_field = N;
```

anyway (as copypasted from the protobuf docs).

---

An important thing we need to change here still is that optional enumeration arguments need to be removed from the compound name signatures of functions, otherwise adding such arguments would *still* be a breaking change.

westonpace:
I've switched to the message format proposed by @jvanstraten and updated the spec around function signatures to specifically mention that options are not part of the signature.

jvanstraten (dismissed):
The implementation and semantics look good to me. I'm neutral about the documentation; I think it covers all the semantics but I think it might be easy to miss how different the various argument types are and behave, especially because the "argument type" concept doesn't translate well to other languages (at least none that I know). This is not necessarily a critique of this PR but more of the function docs in general though, and that can be improved later.

proto/substrait/algebra.proto
```diff
@@ -648,11 +647,27 @@ message Expression {
...
-    //  - Optional enum arguments must be bound using FunctionArgument.enum
-    //    followed by either Enum.specified or Enum.unspecified. If specified,
-    //    the string must case-insensitively match one of the allowed options.
     repeated FunctionArgument arguments = 4;
 
+    // Options to specify behavior for corner cases, or leave behavior
+    // unspecified if the consumer does not need specific behavior in these
+    // cases.
+    repeated FunctionOption options = 5;
+    message FunctionOption {
```
jacques-n, line 656: Why declare this in each function type? FunctionArgument is external and it seems like this should similarly be so.

jacques-n (changes requested):
My expectation here is this should also include a change to the yaml schema where optional arguments are not specified in the same way as other argument types. I don't see that identified here nor do I see any change to the yaml schema.

proto/substrait/algebra.proto
westonpace: Good point.  I've moved this so it is declared once, next to `FunctionArgument`.

westonpace:
> My expectation here is this should also include a change to the yaml schema where optional arguments are not specified in the same way as other argument types. I don't see that identified here nor do I see any change to the yaml schema.

@jacques-n That was a significantly more extensive change.  PTAL and let me know if this is what you were thinking.

site/docs/extensions/generate_function_docs.py
```diff
@@ -37,6 +37,8 @@ def write_markdown(file_obj: dict, file_name: str) -> None:
             options_list = []
 
             for count, impls in enumerate(implementations_list):
+                if "args" not in impls:
```
bkietz, line 40: AFAICT, this is not reading the "options" field of `impls` so we won't get any doc from that. Furthermore, I would expect `options_list` to be constructed from each `impls["options"]` entry since each impl may define its own options. Instead, since that variable is outside the loop it is accumulating the union of all options from anything in `implementation_list`- we haven't noticed because most impls of a single function have identical options.

It might be desirous to *require* that all impls have identical options, possibly by hoisting options to a property of the function rather than of the impl. That adds some validation overhead when multiple function declarations share a name- what should be the behavior if regular "add" and decimal "add" declare different options? I'd say keeping options as a property of `impls[n]` is probably preferable

bkietz (changes requested):

richtia:
@pdet

site/docs/extensions/generate_function_docs.py
westonpace: I'll fix this up so we generate docs for options.  Yes, we will want options to be per-impl I think.  There is some precedent for this already.  Integer arithmetic usually has overflow for example while fp arithmetic had rounding mode.

extensions/functions_aggregate_generic.yaml
```diff
@@ -5,22 +5,21 @@ aggregate_functions:
...
+            values: [SILENT, SATURATE, ERROR]
         nullability: DECLARED_OUTPUT
         decomposable: MANY
         intermediate: i64
         return: i64
   - name: "count"
     description: "Count a set of records (not field referenced)"
     impls:
-      - args:
+      - options:
```
jacques-n, line 20: did you consider making these maps instead of arrays? That way the name is more formally required purely by structure.

extensions/functions_datetime.yaml
```diff
@@ -6,30 +6,26 @@ scalar_functions:
...
-          - options: [ YEAR, MONTH, DAY, SECOND ]
+          - values: [ YEAR, MONTH, DAY, SECOND ]
             name: The part of the value to extract.
-            required: true
           - name: x
             value: timestamp
         return: i64
       - args:
-          - options: [ YEAR, MONTH, DAY, SECOND ]
+          - values: [ YEAR, MONTH, DAY, SECOND ]
```
jacques-n, line 15: I'm surprised by your change here. Are you renaming required enumeration argument types?

site/docs/expressions/scalar_functions.md
```diff
@@ -8,7 +8,7 @@ A function is a scalar function if that function takes in values from a single r
...
 | Session Dependent      | Whether this function is influenced by the session context it is invoked within. For example, a function may be influenced by a user who is invoking the function, the time zone of a session, or some other non-obvious parameter. This can inform caching systems on whether a particular function is cacheable. | Optional, defaults to false.        |
-| Variadic Behavior      | Whether the last argument of the function is variadic or a single argument.  If variadic, the argument can optionally have a lower bound (minimum number of instances) and an upper bound (maximum number of instances). | Optional, defaults to single value. |
+| Variadic Behavior      | Whether the last required argument of the function is variadic or a single argument.  If variadic, the argument can optionally have a lower bound (minimum number of instances) and an upper bound (maximum number of instances). | Optional, defaults to single value. |
```
jacques-n, line 11: this change seems unnecessary.

site/docs/expressions/scalar_functions.md
```diff
@@ -19,11 +19,14 @@ A function is a scalar function if that function takes in values from a single r
 ## Argument Types
 
 There are four main types of arguments: value arguments, type arguments, required enumerations, and optional enumerations.
+The first three types of arguments are considered positional arguments and must be specified in every invocation of the
```
jacques-n, line 22: Doesn't feel quite right with new pattern.

jacques-n (commented):
Good direction. Couple of couples...

extensions/functions_aggregate_generic.yaml
westonpace: That's a good idea and solves my other naming problem.  I'll give that a try.

extensions/functions_datetime.yaml
westonpace: My choices were to leave this alone and have it be out of sync with options (`values` vs `options`) or to have the word "options" doubly used in options `options::options`.  However, if `options` is a map then this is a moot point so I will revert this change.

westonpace: Actually I take that back.  The map change doesn't fix this.  So basically we can have approach 1:

```
args:
 - name: some_required_enum
   options: [ X, Y ]
options:
 some_optional_enum:
  options: [ A, B ]
```

Which is consistent but has some shadowing with the double options.  Or we can have approach 2:

```
args:
 - name: some_required_enum
   options: [ X, Y ]
options:
 some_optional_enum:
  values: [ A, B ]
```

This avoids the shadowing but is inconsistent (options vs. values).  Or we can have approach 3:

```
args:
 - name: some_required_enum
   values: [ X, Y ]
options:
 some_optional_enum:
   values: [ A, B ]
```

This avoids the shadowing and is consistent but is more of a breaking change.

I'm happy to go with any of the above choices.

site/docs/extensions/generate_function_docs.py
westonpace: Ok.  I restored it to what it used to be.  However, I left it per-impl.  This can go wrong as it does here:

image

However, that can be left for a future issue.

site/docs/expressions/scalar_functions.md
westonpace: Reverted.

site/docs/expressions/scalar_functions.md
westonpace: I reworded this entire section.  Please review.

extensions/functions_aggregate_generic.yaml
westonpace: Done.

extensions/functions_datetime.yaml
westonpace: I have currently implemented approach 2.

site/docs/expressions/scalar_functions.md
```diff
@@ -47,12 +46,20 @@ There are four main types of arguments: value arguments, type arguments, require
...
+
+### Option Precedence
+
+A producer may specify multiple vlaues for an option.  If the producer does so then the consumer must deliver the first behavior in the list of values that the consumer is capable of delivering.  For example, considering overflow as defined above, if a producer specified `[ERROR, SATURATE]` then the consumer must deliver `ERROR` if it is capable of doing so.  If it is not then it may deliver `SATURATE`.  If the consumer cannot deliver either behavior then it is an error and the consumer must reject the plan.
+
+#### Optional Properties
```
bkietz, line 57: There's some confusion here since it's not completely clear whether we're talking about the properties of an `algebra.proto::substrait.FunctionOption` message or about the properties of a `simple_extensions_schema.yaml::options` declaration (in general, I think this document would be improved by explicitly listing the substrait message to which these `Property, Description, Required` tables apply).

I would delete `### Optional Precedence``` and hoist this section above the paragraph explaining precedence. That paragraph can then serve as the explanation as to why an `algebra.proto::substrait.FunctionOption` has a list of values instead of a single value.

bkietz:
Other than these nits, LGTM

site/docs/expressions/scalar_functions.md
westonpace: > There's some confusion here since it's not completely clear whether we're talking about the properties of an algebra.proto::substrait.FunctionOption message or about the properties of a simple_extensions_schema.yaml::options declaration (in general, I think this document would be improved by explicitly listing the substrait message to which these Property, Description, Required tables apply).

I had the same confusion.  I think this entire section should probably be moved somewhere else.  This page is describing what belongs in a plan and this section is describing what belongs in extension files.  Those two blocks of documentation should be on separate pages.  The page describing what belongs in a plan can talk about precedence/preference.  The page describing what belongs in the extension yamls can describe the "set of possible values".  I would rather keep this as is and fix this in a future PR.  The proposed change isn't quite right either I think.

bkietz (approved):

proto/substrait/algebra.proto
```diff
@@ -490,11 +489,28 @@ message FunctionArgument {
   message Enum {
     oneof enum_kind {
       string specified = 1;
-      google.protobuf.Empty unspecified = 2;
     }
+    reserved 2;
   }
 }
```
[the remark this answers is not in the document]
westonpace: @jacques-n Any opinion on this suggestion?  This would get rid of `message Enum` entirely which seems a lot simpler.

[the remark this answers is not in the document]
westonpace: @cpcloud @jacques-n friendly ping.  I feel this PR has stalled a bit and I would like to get it through.  However, we should evaluate this carefully as I don't think we can easily revisit in the future.  Any second opinions on this suggestion?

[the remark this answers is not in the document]
westonpace: @cpcloud @jacques-n ping again.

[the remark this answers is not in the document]
cpcloud: This is a pretty big change, but I am in favor of it.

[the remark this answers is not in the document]
westonpace: I have made this change.

westonpace:
I was going to interpret [this comment](https://github.com/substrait-io/substrait/pull/342#discussion_r1015637716) as a second SMC approval and merge this :)

However, it seems I cannot merge while a review is left in the "change requested" state.  @jacques-n do you know if you will be able to look at this anytime soon?

westonpace:
> any chance someone on your team can pick up the downstream changes for validator and java?

I'll see what I can do.

PR #344, opened 2022-09-23
docs: clarify description of set relation

westonpace:


westonpace:
I removed "All inputs are ordinally matched and returned together." because I don't really know what that means and I think it is talking about rows.  Happy to clarify if someone can explain further.

I removed "record type" as this term does not appear to be defined or used elsewhere.  I assume it to be analogous for "output schema" but we don't say "output schema" much either.  I just used "fields" as that seems most consistent with "output order" terminology used elsewhere.

jvanstraten (approved):
LGTM, matches the interpretation I used for the validator.

PR #346, opened 2022-09-23
feat: add functions for splitting strings

richtia:
PR adds functions for splitting strings and fixed some spacing in the options of other functions.

### DIFF
--- a/extensions/functions_string.yaml
+++ b/extensions/functions_string.yaml
@@ -1288,8 +1288,10 @@ scalar_functions:
     name: regex_string_split
     description: >-
       Split a string into a list of strings, based on a regular expression pattern.  The
-      regular expression pattern should follow the International Components for Unicode
-      implementation (https://unicode-org.github.io/icu/userguide/strings/regexp.html).
+      substrings matched by the pattern will be used as the separators to split the input
+      string and will not be included in the resulting list. The regular expression
+      pattern should follow the International Components for Unicode implementation
+      (https://unicode-org.github.io/icu/userguide/strings/regexp.html).
 
       The `case_sensitivity` option specifies case-sensitive or case-insensitive matching.
       Enabling the `multiline` option will treat the input string as multiple lines. This makes
--- a/extensions/functions_string.yaml
+++ b/extensions/functions_string.yaml
@@ -101,13 +101,13 @@ scalar_functions:
     impls:
       - args:
           - name: case_sensitivity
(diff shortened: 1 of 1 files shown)


extensions/functions_string.yaml
```diff
@@ -1263,6 +1263,74 @@ scalar_functions:
...
+      Split a string into a list of strings, based on a specified `separator` character.
+    impls:
+      - args:
+          - value: "varchar<L1>"
+            name: "input"
+            description: The input string.
+          - value: "varchar<L2>"
+            name: "separator"
+            description: A character used for splitting the string.
+        return: "List<varchar<L1>>"
```
richtia, line 1278: Wasn't exactly sure how to put the return type.

jvanstraten: Looks good to me.

extensions/functions_string.yaml
```diff
@@ -1263,6 +1263,74 @@ scalar_functions:
...
+            name: "input"
+            description: The input string.
+          - value: "string"
+            name: "separator"
+            description: A character used for splitting the string.
+        return: "List<string>"
+  -
+    name: regex_string_split
+    description: >-
+      Split a string into a list of strings, based on a regular expression pattern.  The
```
jvanstraten, line 1290: I feel like this could use a bit more explanation. I guess the idea is that it works the same as a regular string split, i.e. removing the substrings matched by the regex from the resulting string list. However, I could also imagine someone interpreting it as the regex picking only the split point, such that every character from the original string ends up in one of the returned list elements. Both implementations would be useful, but the one that removes the matched string is more expressive, because you could wrap the regex in a positive lookahead to mimic the other implementation.

richtia: Good point.  I updated the description to be more detailed.  From what I've seen, the implementation that removes the matched substring is also how a bunch of different SQL dialects do it.

jvanstraten (approved):
LGTM (but I'll leave this open for a little while as I'm no longer effectively the only reviewer)

richtia:
@westonpace could you give this one a look?

westonpace (commented):
One question.  Do we want to support a `max_split` argument?  Most languages (Java, Python, JavaScript) support this but I notice it is not in some standard engines (e.g. PostgreSQL, SQL server, and DuckDb).

westonpace:
Per the sync meeting notes, since there are a significant # of engines that do not support max_split, that would be a second function (e.g. `bounded_string_split` or something).  So let's move forward with this as is.

Issue #274, opened 2022-08-03
Functions backwards compatibility


westonpace:
**Background:**

A few community syncs back we discussed backwards compatibility for functions (e.g. yaml elements) at some length.  I'd like to capture some of the key thoughts here:

 * Adding an option is unlikely to be backwards compatible, unless possibly if added at the very end, but given that most options are likely to be optional/enum, and those tend to be placed before value arguments, it seems unlikely.
 * Changes to properties (e.g. nullability) would not be backwards compatible
 * Adding new properties could be backwards compatible, provided the new property matched implicit assumptions that already existed.
 * Adding a kernel should be backwards compatible
 * Changes to the description could, in theory be backwards incompatible, even though no "technical content" changed.
 * We shouldn't try too hard to maintain backwards compatibility, just create a new function name at that point

We also discussed potentially versioning the URIs for functions:

 * Adding the version in the URI (or basing the URI on a commit hash) could allow for breaking changes but would but introduces the risk of artificial breaking changes.  For example, if we bump the version every release then it will look like all functions had a potentially backwards incompatible change / name change even if they didn't.  More significantly, plans written against v2 would not work against v3.
 * Removing the version from the URI means that any backwards incompatible change to a function would likely require a new name.

Finally, there was discussion about marking a function as "stable" / "sealed":

 * Before becoming stable, a function is liable to have backwards incompatible changes, without requiring a name change.
 * Once a function is stable, if any backwards incompatible change needs to be made, it would require a name change.
 * The stability could be implicit or it could be an explicit property of either the YAML or the name.

**Proposal:**

 * I imagine we will continue to learn things for a while (e.g. rounding modes) and we shouldn't consider any functions stable until at least 1.0, etc.
 * I think functions should be explicitly versioned.  Version 0 should be implicitly unstable.  Version 1 should generally be created as long as there are 2 consumers and 2 producers that support it.
 * Generally, higher version numbers are going to be "more precise behavior" and not "different behavior".  For example, an "add v3" shouldn't be a variation on add but should replace "add v2" with more precise information.
 * Producers should be able to specify a "minimum version" by asking for the older version.  For example, let's pretend we added rounding modes to floating arithmetic in version 2.  A producer could have a user facing option that defaults to "unspecified rounding mode" and, if the user chooses that, then version 1 is acceptable.  The producer would then supply the arguments for version 1.
[...]

PR #347, opened 2022-09-26
feat: attach Substrait version number to plans

jvanstraten:


proto/substrait/plan.proto
```diff
@@ -24,6 +24,10 @@ message PlanRel {
 // Describe a set of operations to complete.
 // For compactness sake, identifiers are normalized at the plan level.
 message Plan {
+  // Substrait version of the plan. Optional up to 0.16.0, required for later
```
jvanstraten, line 27: If this isn't merged this week, I'll probably have to update this to track the version that would introduce this.

mbrobbel (commented):
I guess we should also add Substrait versioning to extensions?

jvanstraten:
FWIW there's some discussion about extension (function) versioning here at #274, and there have been several verbal discussions about extension versioning and URI schemes during various sync meetings. I think a good summary of the situation is that no one's really sure how to do it :) This PR is just low-hanging fruit though.

proto/substrait/plan.proto
```diff
@@ -43,3 +47,20 @@ message Plan {
...
+
+  // If a particular version of Substrait is used that does not correspond to
+  // a version number exactly (for example when using an unofficial fork or
+  // using a version that is not yet released or is between versions), set this
+  // to the full git hash of the utilized commit of
+  // https://github.com/substrait-io/substrait (or fork thereof), represented
+  // using a lowercase hex ASCII string 40 characters in length. The version
+  // number above should be set to the most recent version tag in the history
+  // of that commit.
+  string git_hash = 4;
```
westonpace, line 65: Should the `major`/`minor`/`path` be left empty in this case (0.0.0)?  Or should they be set to the most recent version before the fork?  I would think empty but might be good to clarify.

jvanstraten: Doesn't the last sentence of the comment cover it? I guess I could make explicit that the tag I'm referring to must be an "official" substrait-io version tag.

westonpace: Ah, yeah, that is clear.  I think I just glossed over it reading, sorry

jacques-n, line 65: I think we need to introduce what tool name and version produced the plan. In the practical (annoying) world, there may be 3 different versions of a producer that all produce using the exact same Substrait plan version but one of them produces invalid plans that have been stored. Consumers may need to be conditional about what is seen to handle bugs in producers. We've had to do this a bunch of times with Parquet's producer field and I expect the same here (sadly).

jvanstraten: I don't necessarily disagree, but I fear we might end up with something like `Isthmus/5.0 (X11; Linux x86_64) Acero/537.36 (KHTML, like Gecko) Velox/105.0.0.0 Ibis/537.36` if we're not careful, especially if we just add a string field for this. For consumer-specific plans (i.e. those using extensions from some particular engine), shouldn't there then also be an optional intended consumer identifier?

Typing a bit more freely I came up with this...

```proto
message Plan {
  // Metadata for the plan. Optional up to 0.17.0, required for later versions.
  Metadata metadata = 6;
  
  // ...
}

message Metadata {
  // Optional name of the plan. No semantical meaning; should only be used for
  // debugging and tracing.
  string name = 1;

  // Substrait version of the plan. Optional up to 0.17.0, required for later
  // versions.
  Version version = 2;

  // Identification of the tool generating the plan. Optional (hand-written
  // plans, for example, don't logically have a producer), but strongly
  // recommended when a plan is
  Tool producer = 3;
  
  // If the plan was generated to be compatible with one or more specific
  // consumers, this/these consumer(s) can be identified here. Optional.
  repeated Tool consumer = 4;

  // A place to store custom metadata.
  google.protobuf.Any custom = 5;
}

message Version {
  // Substrait version number.
  uint32 major = 1;
  uint32 minor = 2;
  uint32 patch = 3;

  // If a particular version of Substrait is used that does not correspond to
  // a version number exactly (for example when using an unofficial fork or
  // using a version that is not yet released or is between versions), set this
  // to the full git hash of the utilized commit of
  // https://github.com/substrait-io/substrait (or fork thereof), represented
  // using a lowercase hex ASCII string 40 characters in length. The version
  // number above should be set to the most recent version tag in the history
  // of that commit.
  string git_hash = 4;
}
  
message Tool {
  // Name of the tool, e.g. Acero, Velox, DuckDB, etc. Required.
  string name = 1;

  // Version of the tool. Put the integer components of the version in
  // version_components, and anything that's needed to identify the version in
  // addition in version_suffix. For a release version of tools using semver,
  // version_components should have 3 entries, and version_suffix should be
  // empty.
  repeated uint32 version_components = 2;
  string version_suffix = 3;
  
  // An optional place to store tool-specific information, such as feature
  // flags.
  google.protobuf.Any custom = 4;
}
```

... but this is all pretty arbitrary and grows without bound.

IMO there's no need to add something like this already and we should postpone this, especially considering we've already had talks of more precise communication of capabilities and negotiation. An embedded Substrait version number, on the other hand, addresses a much more pressing need: if a proposal like #333 would have been accepted (shouldn't we close that PR, since even Weston changed his mind in the end...?) there would have been no way to tell without context what a project relation means, even without accounting for consumer/producer bugs.

jvanstraten:
Updated the version number in the comment for Plan.version, and also added a message that can be used to deserialize only the version of a plan even in the presence of breaking changes that would cause protobuf itself to throw.

proto/substrait/plan.proto
jacques-n: > `Isthmus/5.0 (X11; Linux x86_64) Acero/537.36 (KHTML, like Gecko) Velox/105.0.0.0 Ibis/537.36` if we're not careful, especially if we just add a string field for this

This is exactly what we want. Ideally, we never need to use the field. Practically, we will. We should add it now before we need it. This isn't structured information, it's a log entry to deal with unexpected bugs later.

jvanstraten: 1d6fa7c

jacques-n (approved):

PR #351, opened 2022-10-03
feat: add nested type constructor expressions

jvanstraten:
It's currently impossible to formulate expressions like `(field 0 - field 1, field 0 + field 1)` (using `(x, y)` as a two-field struct):

 - struct literals don't work because the fields aren't literals;
 - mask expressions don't work because the fields aren't pure field references;
 - extension functions don't work in general unless one is made specifically for structs with the desired number of fields, the desired nullability, and the desired type variation.

The same applies for lists and maps, though if we ignore type variations, the same could hypothetically be done for them with a finite number of function extensions in core.

This construct effectively makes struct/list/map literals and (I think) mask expressions redundant, but I didn't deprecate them because they at least feel different.

### DIFF
--- a/proto/substrait/algebra.proto
+++ b/proto/substrait/algebra.proto
@@ -507,6 +507,7 @@ message Expression {
     MultiOrList multi_or_list = 9;
     Cast cast = 11;
     Subquery subquery = 12;
+    Nested nested = 13;
 
     // deprecated: enum literals are only sensible in the context of
     // function arguments, for which FunctionArgument should now be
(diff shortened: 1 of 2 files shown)


jacques-n:
I'm generally supportive of this change. It feels cleaner as an expression than trying to introduce as functions. (As functions, type declaration would get nasty quickly.)

@westonpace and @cpcloud , are you supportive of this?

westonpace:
Took me a moment to figure this out but I think I get it now.  I'm +1.

jacques-n (approved):
Merging on lazy consensus

Issue #215, opened 2022-06-01
Documentation + Website


curino:
Per discussions at: https://groups.google.com/g/substrait/c/oXY2IRpKvRo/m/RfjjARRMAQAJ I am starting to inline document in the `.proto` files, and over time include in the website inlined examples from proto + enhanced discussion and documentation. (ongoing activity, will take few patches to avoid huge commits)

curino:
Thanks @jacques-n!!

curino:
Should we close this issue or leave it open as there is lots of room for further improvements?

aljazerzen:
As someone who is quite new to the project, I'd find useful an example on the first page (or README.md).

Something like how does `SELECT issued_at, total - promo + tax FROM orders WHERE promo > 0 LIMIT 10` translate to substrait? I know it cannot be displayed in ProtoBuf, but can you manually produce YAML (or JSON) that would be equivalent to the ProtoBuf representation?

jvanstraten:
+1, we need a proper list of known-correct examples. I guess the question is how to represent them though. Is the intention to postpone this until we have some kind of text format? OTOH even just listing JSON and hex string representation of simple messages would be useful as a starting point for newcomers. We could also run the validator when generating the website and include a link to the validator output.

westonpace:
Some commits have been merged and I'm not sure this issue needs to remain open.  Feel free to comment if anyone believes otherwise.

PR #354, opened 2022-10-06
docs: add tutorial for reading plans

wjones127:
The goal of the tutorial is to get someone to the point where they can read a serialized Substrait plan. SQL is used as a comparison since I assume most readers will have basic familiarity. JSON is used as the output format since it's used in some consumers as the input format for test.

### DIFF
--- a/site/docs/tutorial/expression_trees.svg
+++ b/site/docs/tutorial/expression_trees.svg
@@ -1,0 +1,92 @@
+<svg version="1.1" width="675" height="275" xmlns="http://www.w3.org/2000/svg">
+  <defs>
+    <marker id="arrowhead" markerWidth="10" markerHeight="7" 
+    refX="0" refY="3.5" orient="auto">
+      <polygon points="0 0, 10 3.5, 0 7" />
+    </marker>
+  </defs>
+
+  <rect width="100%" height="100%" fill="white" />
+
+  <g transform="translate(20, 20)">
+    <text x="305" y="15" font-size="20" text-anchor="middle" font-weight="bold">
+      <tspan fill="rebeccapurple">INDEX_IN(</tspan>
+      <tspan fill="steelblue">categories</tspan>,
+      <tspan fill="forestgreen">"Computers"</tspan>
+      <tspan fill="rebeccapurple">)</tspan>
+      <tspan fill="rebeccapurple">IS NULL</tspan>
+    </text>
+
+    <g transform="translate(225, 40)">
+      <rect width="160" height="30" x="0" y="0" class="node" fill="#eee" />
+      <rect width="80" height="30" x="0" y="0" class="node" fill="rebeccapurple" />
+      <text class="node light" x="40" y="20">Function</text>
+      <text class="node" x="120" y="20">is_null</text>
+      <path d="M 80 30 c 0 30, 0 20, 0 40" class="edge" fill="transparent"/>
+    </g>
+
+    <g transform="translate(225, 120)">
+      <rect width="160" height="30" x="0" y="0" class="node" fill="#eee" />
+      <rect width="80" height="30" x="0" y="0" class="node" fill="rebeccapurple" />
+      <text class="node light" x="40" y="20">Function</text>
+      <text class="node" x="120" y="20">index_in</text>
+      <path d="M 80 30 c 0 30, -100 20, -100 40" class="edge" fill="transparent"/>
+      <path d="M 80 30 c 0 30, 100 20, 100 40" class="edge" fill="transparent"/>
+    </g>
+
+    <g transform="translate(125, 200)">
+      <rect width="160" height="30" x="0" y="0" class="node" fill="#eee" />
+      <rect width="80" height="30" x="0" y="0" class="node" fill="steelblue" />
+      <text class="node light" x="40" y="20">Field Ref</text>
+      <text class="node" x="120" y="20">categories</text>
+    </g>
+
+    <g transform="translate(325, 200)">
+      <rect width="180" height="30" x="0" y="0" class="node" fill="#eee" />
+      <rect width="80" height="30" x="0" y="0" class="node" fill="forestgreen" />
+      <text class="node light" x="40" y="20">Literal</text>
+      <text class="node" x="130" y="20">"Computers"</text>
+    </g>
+
+  </g>
+
+  <style>
+    text {
+      font-family: apple-system, BlinkMacSystemFont, Segoe UI, Roboto, Helvetica Neue, Arial, Noto Sans, Liberation Sans, sans-serif, Apple Color Emoji, Segoe UI Emoji, Segoe UI Symbol, Noto Color Emoji;
+    }
+
+    path.edge {
+      stroke: black;
+      marker-end: url(#arrowhead);
+    }
+
+    rect.node {
+      /* fill: #eee; */
+      stroke: #333;
+    }
+
+    text.node {
+      fill: #333;
+      text-anchor: middle;
+      font-size: 14px;
+    }
+    
+    .node.light {
+      fill: #eee;
+    }
+
+    .label {
+      font-size: 14px;
+      font-weight: bold;
+      fill: #111;
+    }
+
+    .light {
+      fill: #eee;
+    }
+
+    #heading {
+      font-size: 24px;
+    }
+  </style>
+</svg>
--- a/site/docs/tutorial/final_plan.json
+++ b/site/docs/tutorial/final_plan.json
@@ -1,0 +1,366 @@
+{
+  "version": { "minorNumber": 20, "producer": "validator-test" },
+  "extensionUris": [
+    {
+      "extensionUriAnchor": 1,
+      "uri": "https://github.com/substrait-io/substrait/blob/main/extensions/functions_string.yaml"
+    },
+    {
+      "extensionUriAnchor": 2,
+      "uri": "https://github.com/substrait-io/substrait/blob/main/extensions/functions_comparison.yaml"
+    },
+    {
+      "extensionUriAnchor": 3,
+      "uri": "https://github.com/substrait-io/substrait/blob/main/extensions/functions_arithmetic_decimal.yaml"
+    }
+  ],
+  "extensions": [
+    {
+      "extensionFunction": {
+        "extensionUriReference": 1,
+        "functionAnchor": 1,
+        "name": "index_in"
+      }
+    },
+    {
+      "extensionFunction": {
+        "extensionUriReference": 2,
+        "functionAnchor": 2,
+        "name": "is_null"
+      }
+    },
+    {
+      "extensionFunction": {
+        "extensionUriReference": 2,
+        "functionAnchor": 3,
+        "name": "equal"
+      }
+    },
+    {
+      "extensionFunction": {
+        "extensionUriReference": 3,
+        "functionAnchor": 4,
+        "name": "sum"
+      }
+    },
+    {
+      "extensionFunction": {
+        "extensionUriReference": 3,
+        "functionAnchor": 5,
+        "name": "multiply"
+      }
+    }
+  ],
+  "relations": [
+    {
+      "root": {
+        "names": [
+          "product_name",
+          "product_id",
+          "sales"
+        ],
+        "input": {
+          "aggregate": {
+            "input": {
+              "join": {
+                "left": {
+                  "read": {
+                    "namedTable": {
+                      "names": [
+                        "orders"
+                      ]
+                    },
+                    "baseSchema": {
+                      "names": [
+                        "product_id",
+                        "quantity",
+                        "order_date",
+                        "price"
+                      ],
+                      "struct": {
+                        "types": [
+                          {
+                            "i64": {
+                              "nullability": "NULLABILITY_REQUIRED"
+                            }
+                          },
+                          {
+                            "i32": {
+                              "nullability": "NULLABILITY_REQUIRED"
+                            }
+                          },
+                          {
+                            "date": {
+                              "nullability": "NULLABILITY_REQUIRED"
+                            }
+                          },
+                          {
+                            "decimal": {
+                              "scale": 2,
+                              "precision": 10,
+                              "nullability": "NULLABILITY_NULLABLE"
+                            }
+                          }
+                        ],
+                        "nullability": "NULLABILITY_REQUIRED"
+                      }
+                    }
+                  }
+                },
+                "right": {
+                  "filter": {
+                    "input": {
+                      "read": {
+                        "namedTable": {
+                          "names": [
+                            "products"
+                          ]
+                        },
+                        "baseSchema": {
+                          "names": [
+                            "product_id",
+                            "categories",
+                            "details",
+                            "manufacturer",
+                            "year_created",
+                            "product_name"
+                          ],
+                          "struct": {
+                            "types": [
+                              {
+                                "i64": {
+                                  "nullability": "NULLABILITY_REQUIRED"
+                                }
+                              },
+                              {
+                                "list": {
+                                  "type": {
+                                    "string": {
+                                      "nullability": "NULLABILITY_REQUIRED"
+                                    }
+                                  },
+                                  "nullability": "NULLABILITY_REQUIRED"
+                                }
+                              },
+                              {
+                                "struct": {
+                                  "types": [
+                                    {
+                                      "string": {
+                                        "nullability": "NULLABILITY_NULLABLE"
+                                      }
+                                    },
+                                    {
+                                      "i32": {
+                                        "nullability": "NULLABILITY_NULLABLE"
+                                      }
+                                    }
+                                  ],
+                                  "nullability": "NULLABILITY_NULLABLE"
+                                }
+                              },
+                              {
+                                "string": {
+                                  "nullability": "NULLABILITY_NULLABLE"
+                                }
+                              }
+                            ],
+                            "nullability": "NULLABILITY_REQUIRED"
+                          }
+                        }
+                      }
+                    },
+                    "condition": {
+                      "scalarFunction": {
+                        "functionReference": 2,
+                        "outputType": {
+                          "bool": {
+                            "nullability": "NULLABILITY_REQUIRED"
+                          }
+                        },
+                        "arguments": [
+                          {
+                            "value": {
+                              "scalarFunction": {
+                                "functionReference": 1,
+                                "outputType": {
+                                  "i64": {
+                                    "nullability": "NULLABILITY_NULLABLE"
+                                  }
+                                },
+                                "arguments": [
+                                  {
+                                    "value": {
+                                      "literal": {
+                                        "string": "Computers"
+                                      }
+                                    }
+                                  },
+                                  {
+                                    "value": {
+                                      "selection": {
+                                        "directReference": {
+                                          "structField": {
+                                            "field": 1
+                                          }
+                                        },
+                                        "rootReference": {}
+                                      }
+                                    }
+                                  }
+                                ]
+                              }
+                            }
+                          }
+                        ]
+                      }
+                    }
+                  }
+                },
+                "type": "JOIN_TYPE_INNER",
+                "expression": {
+                  "scalarFunction": {
+                    "functionReference": 3,
+                    "outputType": {
+                      "bool": {
+                        "nullability": "NULLABILITY_NULLABLE"
+                      }
+                    },
+                    "arguments": [
+                      {
+                        "value": {
+                          "selection": {
+                            "directReference": {
+                              "structField": {
+                                "field": 0
+                              }
+                            },
+                            "rootReference": {}
+                          }
+                        }
+                      },
+                      {
+                        "value": {
+                          "selection": {
+                            "directReference": {
+                              "structField": {
+                                "field": 4
+                              }
+                            },
+                            "rootReference": {}
+                          }
+                        }
+                      }
+                    ]
+                  }
+                }
+              }
+            },
+            "groupings": [
+              {
+                "groupingExpressions": [
+                  {
+                    "selection": {
+                      "directReference": {
+                        "structField": {
+                          "field": 0
+                        }
+                      },
+                      "rootReference": {}
+                    }
+                  },
+                  {
+                    "selection": {
+                      "directReference": {
+                        "structField": {
+                          "field": 7
+                        }
+                      },
+                      "rootReference": {}
+                    }
+                  }
+                ]
+              }
+            ],
+            "measures": [
+              {
+                "measure": {
+                    "functionReference": 3,
+                    "outputType": {
+                      "decimal": {
+                        "scale": 2,
+                        "precision": 38,
+                        "nullability": "NULLABILITY_NULLABLE"
+                      }
+                    },
+                    "arguments": [
+                      {
+                        "value": {
+                          "scalarFunction": {
+                            "functionReference": 4,
+                            "outputType": {
+                              "decimal": {
+                                "scale": 2,
+                                "precision": 38,
+                                "nullability": "NULLABILITY_NULLABLE"
+                              }
+                            },
+                            "arguments": [
+                                {
+                                  "value": {
+                                    "cast": {
+                                      "type": {
+                                        "decimal": {
+                                          "scale": 2,
+                                          "precision": 10,
+                                          "nullability": "NULLABILITY_REQUIRED"
+                                        }
+                                      },
+                                      "input": {
+                                        "selection": {
+                                          "directReference": {
+                                            "structField": {
+                                              "field": 1
+                                            }
+                                          },
+                                          "rootReference": {}
+                                        }
+                                      }
+                                    }
+                                  }
+                                },
+                                {
+                                  "value": {
+                                    "selection": {
+                                      "directReference": {
+                                        "structField": {
+                                          "field": 3
+                                        }
+                                      },
+                                      "rootReference": {}
+                                    }
+                                  }
+                                }
+                            ]
+                          }
+                        }
+                      }
+                    ]
+                }
+              }
+            ],
+            "common": {
+              "emit": {
+                "outputMapping": [
+                  1,
+                  0,
+                  2
+                ]
+              }
+            }
+          }
+        }
+      }
+    }
+  ]
+}
--- a/site/docs/tutorial/sql_to_substrait.md
+++ b/site/docs/tutorial/sql_to_substrait.md
@@ -1,0 +1,968 @@
+# SQL to Substrait tutorial
+
+This is an introductory tutorial to learn the basics of Substrait for readers
+already familiar with SQL. We will look at how to construct a Substrait plan
+from an example query.
+
+We'll present the Substrait in JSON form to make it relatively readable to
+newcomers. Typically Substrait is exchanged as a protobuf message, but for
+debugging purposes it is often helpful to look at a serialized form. Plus,
+it's not uncommon for unit tests to represent plans as JSON strings. So
+if you are developing with Substrait, it's useful to have experience reading
+them.
+
+!!! note
+    Substrait is currently only defined with Protobuf. The JSON provided here is the
+    Protobuf JSON output, but it is *not* the official Substrait text format. Eventually,
+    Substrait will define it's own human-readable text format, but for now this
+    tutorial will make due with what Protobuf provides.
+
+Substrait is designed to communicate plans (mostly logical plans). Those plans
+contain types, schemas, expressions, extensions, and relations. We'll look at 
+them in that order, going from simplest to most complex until we can construct
+full plans.
+
+![](substrait_components.svg)
+
+This tutorial won't cover all the details of each piece, but it will give you an
+idea of how they connect together. For a detailed reference of each individual 
+field, the best place to look is reading the 
+[protobuf definitions](https://github.com/substrait-io/substrait/tree/main/proto/substrait).
+They represent the source-of-truth of the spec and are well-commented to address
+ambiguities.
+
+## Problem Set up
+
+To learn Substrait, we'll build up to a specific query. We'll be using the
+tables:
+
+```sql
+CREATE TABLE orders (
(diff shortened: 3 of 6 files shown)


wjones127:
Not sure if this addresses exactly what #215 is getting at. But as a newcomer, I found it hard to figure out where to start if I wanted to read plans. I started drafting this tutorial (still in very rough form) to try that out.

Does the JSON output seem correct and the right way to go? Do we feel like the serialized protobuf bytes should also be provided?

I based the JSON output on this snippet:

```python
import ibis
from ibis_substrait.compiler.core import SubstraitCompiler
from google.protobuf.json_format import MessageToJson

compiler = SubstraitCompiler()

t = ibis.table([("b", "string"), ("a", "int64")], name="my_database.my_table")
result = compiler.compile(t.filter(t.a > 10))
print(MessageToJson(result))

result_bytes = result.SerializeToString()
```

site/docs/tutorial/sql_to_substrait.md
```diff
@@ -0,0 +1,245 @@
+# SQL to Substrait tutorial
+
+To provide an introduction to Substrait for those already familiar with SQL,
+this tutorial walks through the basics of Substrait by translating basic SQL
+queries. We'll present the Substrait in JSON form so it is (somewhat) readable
+for newcomers.
+
+Substrait can be a type, an expression, a relation, or a plan. We'll look at them in
+those order, going from simplest to most complex.
```
westonpace, line 9:
```suggestion
that order, going from simplest to most complex.
```

site/docs/tutorial/sql_to_substrait.md
```diff
@@ -0,0 +1,245 @@
+# SQL to Substrait tutorial
+
+To provide an introduction to Substrait for those already familiar with SQL,
+this tutorial walks through the basics of Substrait by translating basic SQL
+queries. We'll present the Substrait in JSON form so it is (somewhat) readable
+for newcomers.
+
+Substrait can be a type, an expression, a relation, or a plan. We'll look at them in
```
westonpace, line 8: Substrait is a spec for plans.   A plan is made up of those building blocks but I'm not sure it is valid to say "Substrait can be a type, an expression, a relation, or a plan".

site/docs/tutorial/sql_to_substrait.md
```diff
@@ -0,0 +1,245 @@
...
+
+An expression can be one of several things. `a` itself will be translated to
+a field reference:
+
+```json
+{
+  "selection": {
+    "directReference": {
+      "structField": {
+        "field": 1
```
westonpace, line 32: Might be worth noting that Substrait refers to fields numerically based on some kind of schema / field order (depends on the operator).  In SQL field references are always named references.

site/docs/tutorial/sql_to_substrait.md
```diff
@@ -0,0 +1,245 @@
...
+    Protobuf may not serialize fields with integer type and value 0, since 0 is the default.
+    So if you instead saw `"structField": {}`, know that is is equivalent to 
+    `"structField": { "field": 0 }`.
+
+`10` will be translated to a literal expression:
+
+```json
+{
+  "literal": {
+    "i8": 10,
```
westonpace, line 50: Why is it an `i8`?

site/docs/tutorial/sql_to_substrait.md
```diff
@@ -0,0 +1,245 @@
...
+The name of the function is defined by the `functionReference`, which points to
+the index of the extension function to use. This is serialized separately, which
+you will see later in serialized plans.
+
+```json
+{
+  "extensionFunction": {
+    "extensionUriReference": 1,
+    "functionAnchor": 2,
+    "name": "gt"
```
westonpace, line 103: It might be a good idea to explicitly mention here that `gt` is equivalent to `>` in your above SQL expression.  Not everyone will be hard-wired to see "gt" and think "greater than".

site/docs/tutorial/sql_to_substrait.md
```diff
@@ -0,0 +1,245 @@
...
+  }
+}
+```
+
+## Relations
+
+An logical or physical plan is typically represented as a tree of nodes, such
```
westonpace, line 110:
```suggestion
A logical or physical plan is typically represented as a tree of nodes, such
```
Do we want to introduce or mention "logical" and "physical"?  Won't these terms confuse a reader at this point?

site/docs/tutorial/sql_to_substrait.md
```diff
@@ -0,0 +1,245 @@
...
+as filter, project, scan, or join. Each of these nodes is a **Relation** in 
+Substrait.
```
westonpace, line 112: I feel like, as a naive reader, I might be confused here.  What is a relation exactly?  A part of a tree?  Why is it a tree?  What does this tree represent?  How is a relation different from a function?

site/docs/tutorial/sql_to_substrait.md
```diff
@@ -0,0 +1,245 @@
...
+        "my_database",
+        "my_table"
+      ]
+    }
+  }
+}
+```
+
+Read relations provide a schema in `baseSchema` as discussed in the prior section.
+The type of read is a named table, so the `namedTable` field is present with
```
westonpace, line 151: The type of read?  There is more than one? <-- Something a naive reader might be thinking here.

site/docs/tutorial/sql_to_substrait.md
```diff
@@ -0,0 +1,245 @@
...
+  }
+}
+```
+
+Read relations provide a schema in `baseSchema` as discussed in the prior section.
+The type of read is a named table, so the `namedTable` field is present with 
+`names` containing the list of name segments (`my_database.my_table`).
+
+Read relations are leaf nodes, so other relations will have them nested within.
+For example, a filter relation of (`a > 10`) will look like:
```
westonpace, line 155: Maybe...

Read relations are leaf nodes.  Leaf nodes don't depend on any other node for data and usually represent a source of data in our plan.  Leaf nodes are then typically used as input for other nodes that manipulate the data.

site/docs/tutorial/sql_to_substrait.md
```diff
@@ -0,0 +1,245 @@
...
+      }
+    },
+    "condition": ...
+  }
+}
+```
+
+## Plans
+
+First, let's look at reading a table:
```
westonpace, line 171: I feel like the first paragraph should explain what a plan is and not be giving an example.

site/docs/tutorial/sql_to_substrait.md
```diff
@@ -0,0 +1,245 @@
...
+
+```sql
+SELECT * FROM my_table;
+```
+
+A plan has a list of relations, and each top one is a root relation. The root
+relation just has one input relation and then the names of the fields.
```
westonpace, line 178: These are facts, and they are true, but I'm not sure what the story you are trying to tell is.  Perhaps something like...

Many plans are represented by a single tree of relations.  The root of the tree is the output of the plan and the leaves in the tree are the sources of data.

Howevr, some plans actually contain multiple trees.  One reason for this is when a plan has a common subtree that occurs multiple times throughout the plan.  That subtree can be pulled out and described on its own and then can be referenced in other trees.  This is why a Substrait plan is actually a list of relations.  Only one of these is a "root relation" and it forms the root of the plan.

westonpace (commented):
Thanks!  I think this will be quite helpful and is a good start.  I left a few comments on the content itself.

site/docs/tutorial/sql_to_substrait.md
wjones127: Good point. I'll make sure to call that out. I'll need to look a little deeper to, since the natural question is "relative to *which* schema?"

site/docs/tutorial/sql_to_substrait.md
wjones127: :shrug: that's what Ibis emitted and it seemed fine to me. Though I suppose usually when people represent an integer it's an i32 or i64, so maybe it would be less weird if it was either of those.

I should probably explain this type to the reader (perhaps that's what you are getting at). One question I have: To what extent should a Substrait producer strive to keep the operands of the same type? If `a` in `a > 10` is an `i64` is it okay if I label `10` as a `i32`?

site/docs/tutorial/sql_to_substrait.md
wjones127: Good point. I'll also point out where the YAML definition is.

site/docs/tutorial/sql_to_substrait.md
wjones127: It's not clear here, but I was speaking of plans in general, not Substrait plans specifically. I assume that anyone coming from the SQL world has seen some form of logical / physical plan (at the very least, has run `EXPLAIN ...` at some point and tried to read the output). I'll make this more specific so that's clear.

site/docs/tutorial/sql_to_substrait.md
waynexia: Do you think we need to refer to the docs https://substrait.io/relations/logical_relations/ where has a table shows the input and output number of each relation? Like read relation here require 0 input so that's why it's a leaf relation.

waynexia:
Super thanks! It's a lifesaver 🥰 As a beginner to substrait I have a lot less doubts after reading this article. Can't wait to see it land!

site/docs/tutorial/expression_trees.svg
```diff
@@ -0,0 +1,115 @@
...
+      <text class="node light" x="40" y="20">Function</text>
+      <text class="node" x="120" y="20">IS NULL</text>
+      <path d="M 80 30 c 0 30, 20 20, 20 40" class="edge" fill="transparent"/>
+    </g>
+
+    <g transform="translate(400, 200)">
+      <rect width="160" height="30" x="0" y="0" class="node" fill="#eee" />
+      <rect width="80" height="30" x="0" y="0" class="node" fill="steelblue" />
+      <text class="node light" x="40" y="20">Field Ref</text>
+      <text class="node" x="120" y="20">quantity</text>
```
Z-Shang, line 67: shouldn't this field name to be discount?

site/docs/tutorial/sql_to_substrait.md
```diff
@@ -0,0 +1,1305 @@
...
+output columns?
+
+You might be tempted to add a [Project relation](/relations/logical_relations/#project-operation) 
+at the end to select a subset of columns. However, the project relation only
+adds columns; it is not responsible for subsetting or reordering columns.
+
+<!-- TODO: Is this true? -->
+Instead, any relation can reorder or subset columns through the `emit` property.
+By default, it is set to `direct`, which outputs all columns "as is". But it can
+also be specified as a sequence of field indices.
```
wjones127, line 796: Just want to make sure? Is this actually true? Is **any** relation able to subset or reorder columns?

site/docs/tutorial/sql_to_substrait.md
```diff
@@ -0,0 +1,1305 @@
...
+
+ * `index_in` (0), from `functions_string.yaml`,
+ * `is_null` (1), from `functions_comparison.yaml`,
+ * `equal` (2), from `functions_comparison.yaml`,
+ * `sum` (3), from `functions_arithmetic_decimal.yaml`,
+ * `multiply` (4), from `functions_arithmetic_decimal.yaml`.
+
+So first we can create the three extension uris:
+
+<!-- TODO: what is the URI format? -->
```
wjones127, line 868: All the plans I have seen from ibis-substrait omits the URI, but that seems... wrong? Have we established the pattern for these URIs?

site/docs/tutorial/sql_to_substrait.md
```diff
@@ -0,0 +1,1305 @@
...
+    "nullability": "NULLABILITY_REQUIRED"
+  }
+}
+```
+
+Here, `names` is the names of all fields. In nested schemas, this includes the
+names of subfields in depth-first order. So for the `products` table, the 
+`details` struct field will be included as well as the two subfields (`manufacturer`
+and `year_created`) right after. And because it's depth first, these subfields
+appear before `product_name`. The full schema looks like:
```
wjones127, line 179: Is this correct? There is no subfield for list or map types? (categories here is a list of strings)

site/docs/tutorial/sql_to_substrait.md
```diff
@@ -0,0 +1,1305 @@
...
+`details`.
+
+One complication is that whenever we reference by field index, those field
+indices are dependent on the context of the relation. The exact field indices
+can change from relation to relation. For example, a join relations field 
+indices are based on the combination of the left inputs fields and then the 
+right input fields. And, while we don't have one in our example, a Project relation
+adds new fields, which are visible to it's downstream relations. The diagram
+below shoes the mapping of field indices within each relation and how each of
+the field references show up in each relations properties.
```
wjones127, line 780: Have I described the rules adequately / accurately? Is there anything important I've missed?

wjones127 (commented):
I have rewritten the tutorial to focus on translating a single example SQL query. It's still very rough, but before I go any further I wanted to get some technical feedback. I've noted a few areas where I'm not 100% sure I got it correct, so would like some feedback before I polish this up. @westonpace

site/docs/tutorial/sql_to_substrait.md
westonpace: Yes.  This is quite different from some existing tooling in this area (e.g. Acero, dplyr, ibis) and so if you are familiar with those then it can be a bit odd.  It has been a consistent source of confusion :).

It is a very "logical" concept and allows for writing compact plans.  One way to think about it is to recognize that just about any relation might be the terminal point for a particular column:

 * You might load a column purely for the sake of applying a filter and can drop it after the filter
 * You might load a column purely to use as a join key and drop it after the join
 * Etc.

So, rather than be forced to write a plan that is full of project nodes you can instead just emit at any point to drop a column once you are finished with it.

A corollary confusion is that the project node, by itself, cannot actually drop or reorder columns.  The project node, in Substrait, only adds columns to the end.  The only way to drop columns in a project node is with an emit.

This means that `SELECT expr1, expr2, expr3` typically gets represented as:

`project(expr1, expr2, expr3):emit=[len(input)+0, len(input)+1, len(input)+2]`

site/docs/tutorial/sql_to_substrait.md
westonpace: > Have we established the pattern for these URIs?

Not quite, but the time is ripe to fix this.  All produers I've seen (Isthmus & DuckDb) operate the same as Ibis.  Originally the contention was between:

 * Brittle URIs that are tied to a specific version of an extension function but the content is fixed (e.g. https://github.com/substrait-io/substrait/blob/v0.18.0/extensions/functions_arithmetic.yaml)
 * Stable URIs that are not tied to a specific version and the content may vary (e.g. https://github.com/substrait-io/substrait/blob/main/extensions/functions_arithmetic.yaml)

Based on the conversation in https://github.com/substrait-io/substrait/issues/274 I believe we were leaning towards the latter (stable URI, potentially changing content).

Note that, in Acero, we treat the URI `/` (or the empty string) as a special case where we fallback to name-only lookup and look for a function with the given name, regardless of what URI it maps to (and if there are multiple I think we pick the first we find).  This allows us to consume plans from producers today that omit the URI.

site/docs/tutorial/sql_to_substrait.md
westonpace: I think so but I'm not certain.  I don't think a list column has names (even if it is a fixed size list) for its children.  In a map column the names are part of the data and not the field.

site/docs/tutorial/sql_to_substrait.md
westonpace: > those field indices are dependent on the context of the relation

> a join relations field indices

If I were to be pedantic I would say "the context of the expression".  A join relation does not have field indices.  A join relation has a "join expression" and the context of references in that expression is the output schema of the join relation.

An interesting case to consider (and I have a really old PR I need to get merged clarifying this :grimacing: https://github.com/substrait-io/substrait/pull/271/files) is the read relation.

A read relation has a filter expression and a projection (which isn't an expression exactly, but a masked expression).  A read relation effectively has _two_ schemas (maybe even more if we want to consider evolution but we won't go there :laughing: ).  There is the schema of the data on disk (the direct schema) and the output schema of the read relation (the direct schema after the projection mask has been applied to it).  So we have to be specific when talking about the filter expression which schema we are actually binding against.

site/docs/tutorial/sql_to_substrait.md
```diff
@@ -0,0 +1,1305 @@
...
+!!! note
+    Substrait is currently only defined with Protobuf. The JSON provided here is the
+    Protobuf JSON output, but it is *not* the official Substrait text format. Eventually,
+    Substrait will define it's own human-readable text format, but for now this
+    tutorial will make due with what Protobuf provides.
+
+Substrait is designed to communicate plans (mostly logical plans). Those plans
+contain types, schemas, expressions, extensions, and relations. We'll look at 
+them in that order, going from simplest to most complex until we can construct
+full plans.
```
vibhatha, line 17: nit: Suggestion

```suggestion
them in that order, going from basic to the most complex until we can construct
complete plans.
```

site/docs/tutorial/sql_to_substrait.md
```diff
@@ -0,0 +1,1305 @@
...
+
+![](substrait_components.svg)
+
+This tutorial won't cover all the details of each piece, but will give you an
```
vibhatha, line 21: nit: maybe? 

```suggestion
This tutorial won't cover all the details of each piece, but it will give you an
```

site/docs/tutorial/sql_to_substrait.md
```diff
@@ -0,0 +1,1305 @@
...
+In most SQL engines, a logical or physical plan is represented as a tree of 
+nodes, such as filter, project, scan, or join. The left diagram below may be a 
+familiar representation of our plan, where nodes feed data into each other
+moving from left to right. In Substrait, each of these nodes is a **Relation**.
+
+![](plan_tree_versus_expression.svg)
+
+A relation that takes another relation as input will contain (or refer to) that
+relation. This is usually a field called `input`, but for complex cases that
+have multiple inputs such as joins may have different keys. In JSON, the rough
```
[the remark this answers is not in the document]
vibhatha: Just want to know what you meant by this statement :)

site/docs/tutorial/sql_to_substrait.md
```diff
@@ -0,0 +1,1305 @@
...
+```
+
+The final aggregation requires two things, other than the input. First is the
+groupings. We'll use a single grouping expression containing the references to
+the fields `product_name` and `product_id`. (Multiple grouping expressions can
+be used to do cube aggregations.) These field references are indices, and we'll
+discuss later in [](#field-indices) where these come from.
+
+For `measures`, we'll need to define `sum(quantity * price) as sales`. Substrait
+is stricter about data types, and quantity is a integer while price is a
```
vibhatha, line 643:
```suggestion
is stricter about data types, and quantity is an integer while price is a
```

vibhatha:
@wjones127 Thanks a lot for this tutorial, it is much needed component to docs. 👍

site/docs/tutorial/sql_to_substrait.md
```diff
@@ -0,0 +1,1305 @@
...
+    Protobuf JSON output, but it is *not* the official Substrait text format. Eventually,
+    Substrait will define it's own human-readable text format, but for now this
+    tutorial will make due with what Protobuf provides.
+
+Substrait is designed to communicate plans (mostly logical plans). Those plans
+contain types, schemas, expressions, extensions, and relations. We'll look at 
+them in that order, going from simplest to most complex until we can construct
+full plans.
+
+![](substrait_components.svg)
```
westonpace, line 19: One minor nit about this image.  It shows four categories: types, relations, expressions, and functions.

I don't normally think of "functions" on the same level as the other three. Although, I'm also not sure exactly what "functions" is referring to, and maybe this is ok.

An expression can have a function invocation.  For example, a `ScalarFunction` message or a `WindowFunction` message.  But these are just a type of expression and so don't really deserve their own box.

On the other hand, a plan has a list of functions in the extension set (e.g. defines the URL/name/anchor).  If you're referring to this I suppose it's ok but in my mental picture I think of things like "read types" and "advanced extensions" at the same level of significance (e.g. not quite as significant as types/relations/expressions).

Anyways, I'm not sure anything here needs to change, it just seemed slightly off to me and I wanted to mention it.  If you're happy with it then we can stick with it as is.

site/docs/tutorial/sql_to_substrait.md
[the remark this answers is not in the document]
wjones127: Yeah that is complex. I just meant that it's usually called `input`, but in the join case it's called `left` and `right`, so it's not always the same field for the inputs.

site/docs/tutorial/sql_to_substrait.md
wjones127: Yeah, this diagram is adapted from one of Jacques presentations. It's not really meant to accurately describe the relationships between each entity (except for the fact that plans are the top-level thing). It just gives an idea of what kinds of things are in plans.

I think it's fine for now, but if someone wants to design a better alternative, I have no qualms with replacing this.

wjones127:
I have addressed the early PR feedback. I also validated the final query with substrait-validator (what an awesome tool! 🤩), and that let me know I was describing how field indices work incorrectly. I've revised the tutorial based on that feedback, and also added a "Next Steps" section pointing readers to that tool.

While there's a few TODOs left in there that would make this tutorial a bit better, I think this PR is ready for review.

site/docs/tutorial/sql_to_substrait.md
```diff
@@ -0,0 +1,993 @@
...
+indices are at each point in the plan. Try downloading the final plan JSON above
+and generating an HTML report on the plan with:
+
+```sh
+substrait-validator final_plan.json --out-file output.html
+```
+
+**Generate more Substrait plans from SQL using DuckDB's compiler**. See the 
+documentation on
+[the DuckDB Substrait extension](https://duckdb.org/docs/extensions/substrait).
```
wjones127, line 993: I'd love to point readers to one of the tools that currently can be used to generate Substrait from SQL.

DuckDB looked easy to install, but not sure if they want that attention yet. I noticed the instructions on the website had some typos and the provided example errors when run in Python. From the github repo, they aren't accepting bug reports except from those who have pre-arranged agreements.

I didn't see a user-facing way to install Isthmus (just compilation instructions), but correct me if I'm wrong.

site/docs/tutorial/final_plan.json
```diff
@@ -0,0 +1,381 @@
+{
+  "version": { "minor": 18, "producer": "validator-test" },
```
mbrobbel, line 2: The version field was renamed in 0.19.x via https://github.com/substrait-io/substrait/pull/362:
```suggestion
  "version": { "minorNumber": 19, "producer": "validator-test" },
```

site/docs/tutorial/sql_to_substrait.md
vibhatha: Frankly I also think the existing docs could be improved further. I am not an expert in Graal based builds, but I also think it would be great to document the build, install and usage of isthmus with little more detail. I guess non-Java developers would also use this.

site/docs/tutorial/sql_to_substrait.md
wjones127: Thanks. It is complex, but I found that `substrait-validator` provides a really helpful UI for showing those schemas and so I've referenced that tool in the tutorial.

site/docs/tutorial/sql_to_substrait.md
```diff
@@ -0,0 +1,995 @@
...
+  INDEX_IN("Computers", categories) IS NULL
+GROUP BY
+  product_name,
+  product_id
+```
+
+The query asked the question: For products that aren't in the `"Computer"`
+category, how much has each product generated in sales?
+
+However, Substrait doesn't correspond to SQL as much as it does to logical plans.
```
westonpace, line 85:
```suggestion
However, Substrait doesn't correspond to SQL as much as it does to query plans.
```

Minor nit: Substrait could be used to represent both logical and physical plans.  This sentence _could_ potentially lead someone to thinking it is only for logical plans.

site/docs/tutorial/sql_to_substrait.md
```diff
@@ -0,0 +1,995 @@
...
+{
+  "string": {
+    "nullability": "NULLABILITY_NULLABLE"
+  }
+}
+```
+
+Nullability is an enum not a boolean, since Substrait allows 
+`NULLABILITY_UNSPECIFIED` as an option, in addition to `NULLABILITY_NULLABLE`
+(nullable) and `NULLABILITY_REQUIRED` (not nullable).
```
westonpace, line 112: Eh...generally `_UNSPECIFIED` is an invalid value for an enum and will be rejected by all plans and consumers.  Pretty much every enum has an `_UNSPECIFIED` variant, even when it doesn't make sense (e.g. `JOIN_TYPE_UNSPECIFIED`).  This is done whenever an enum doesn't have a natural default value because otherwise there is no way in protobuf to distinguish between "use the default" and "this is missing".  So I'm not entirely sure why nullability is a boolean.  @jacques-n is `NULLABILITY_UNSPECIFIED` meant to be a legitimate thing?  I would think, if someone doesn't know, they would just pick `NULLABILITY_NULLABLE`.

site/docs/tutorial/sql_to_substrait.md
```diff
@@ -0,0 +1,995 @@
...
+```
+
+Both `IS NULL` and `INDEX_IN` will be scalar function expressions. Available 
+functions in Substrait are defined in extension YAML files contained in 
+https://github.com/substrait-io/substrait/tree/main/extensions. Additional
+extensions may be created elsewhere. `IS NULL` is defined as a `is_null` function
+in 
+[functions_comparison.yaml](https://github.com/substrait-io/substrait/blob/main/extensions/functions_comparison.yaml#:~:text=%2D-,name%3A%20%22is_null%22,-description%3A%20Whether)
+and `INDEX_IN` is defined as `index_in` function in
+[functions_set.yaml](https://github.com/substrait-io/substrait/blob/main/extensions/functions_set.yaml#:~:text=%2D-,name%3A%20%22index_in%22,-description%3A%20%3E).
```
westonpace, line 312:
```suggestion
[functions_comparison.yaml](https://github.com/substrait-io/substrait/blob/main/extensions/functions_comparison.yaml)
and `INDEX_IN` is defined as `index_in` function in
[functions_set.yaml](https://github.com/substrait-io/substrait/blob/main/extensions/functions_set.yaml).
```
Minor nit: I might be missing something but is the hash part of these URLs essential?

site/docs/tutorial/sql_to_substrait.md
```diff
@@ -0,0 +1,995 @@
...
+          }
+        }
+      }
+    ]
+  }
+}
+```
+
+`functionReference` will be explained later in the plans section. For now,
+understand that it's an index into a list of function definitions and since this
```
westonpace, line 358: Let's keep this unless you can come up with something better but this is technically not an "index into a list".  A functionReference of 7 could be the first item in the list or the 10th item.  It's more a key into a map.  However, I can't think of better wording without confusing a reader.

site/docs/tutorial/sql_to_substrait.md
```diff
@@ -0,0 +1,995 @@
...
+            "functionReference": 1,
+            "outputType": {
+              "i64": {
+                "nullability": "NULLABILITY_NULLABLE"
+              }
+            },
+            "arguments": [
+              {
+                "enum": {
+                  "specified": "NAN_IS_NOT_NAN"
```
westonpace, line 395: Sorry, but this is out of date (my fault for the breaking change).  Options are no longer treated as arguments but rather their own thing.  So the `NAN_IS_NOT_NAN` will not be set in this way.

site/docs/tutorial/sql_to_substrait.md
```diff
@@ -0,0 +1,995 @@
...
+are modified. For the purposes of our query, the relevant rules are:
+
+* For Read relations, their output schema is the schema of the table.
+* For Filter relations, the output schema is the same as in the input schema.
+* For Joins relations, the input schema is the concatenation of the left and then the
+  right schemas. The output schema is the same.
+* For Aggregate relations, the output schema is the group by fields followed by
+  the measures.
+
+<!-- TODO: do we document these rules somewhere? -->
```
westonpace, line 809: In the docs / spec this is referred to as the "direct output order"

image

westonpace (changes requested):
Once again, this is a great document.  A few more thoughts but most are minor.

site/docs/tutorial/sql_to_substrait.md
wjones127: That extra part is to highlight the relevant text. It means it will always scroll to the latest version of the function, regardless of where it is in the file.

site/docs/tutorial/sql_to_substrait.md
wjones127: I rephrased as an "ID corresponding to an entry in a list" which is hopefully clear and more accurate.

site/docs/tutorial/sql_to_substrait.md
wjones127: Yup. Removed those references since they aren't required.

site/docs/tutorial/sql_to_substrait.md
wjones127: Also seems like "emit order" is used. I've added mention of those two phrases to the doc so readers will recognize them.

westonpace (approved):
Thanks for creating this!

westonpace (changes requested):
At the moment, you will need to patch up the commit messages to ensure they all confirm to semantic commits.  Some of your commit messages don't quite match.  Either squash the invalid ones into valid ones or reword the invalid ones.

wjones127:
Thanks @westonpace. Just squashed all into one commit with what should be a conforming message.

westonpace (approved):

PR #359, opened 2022-10-19
feat: add chr() function to yaml

zhixingheyi-tian:
Hi @jvanstraten @jacques-n 
Please give a review, Thanks.

### DIFF
--- a/extensions/functions_string.yaml
+++ b/extensions/functions_string.yaml
@@ -20,7 +20,8 @@ scalar_functions:
   - name: chr
     description: >-
     Returns the ASCII character having the binary equivalent to input.
-    If n is larger than 256 the result is equivalent to chr(n % 256)
+    If n is larger than 256 the result is equivalent to chr(n % 256).
+    If n is negative value, the result is null.
     impls:
     - args:
       - value: i8
--- a/extensions/functions_string.yaml
+++ b/extensions/functions_string.yaml
@@ -17,6 +17,27 @@ scalar_functions:
         variadic:
           min: 1
         return: "string"
+  - name: chr
+    description: >-
+    Returns the ASCII character having the binary equivalent to input.
+    If n is larger than 256 the result is equivalent to chr(n % 256)
+    impls:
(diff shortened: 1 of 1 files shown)


extensions/functions_string.yaml
```diff
@@ -17,6 +17,27 @@ scalar_functions:
         variadic:
           min: 1
         return: "string"
+  - name: chr
+    description: >-
+    Returns the ASCII character having the binary equivalent to input.
+    If n is larger than 256 the result is equivalent to chr(n % 256)
```
richtia, line 23: what's the behavior for negative values of n?

zhixingheyi-tian: Add description for negative values.

zhixingheyi-tian:
Hi @richtia 
Can it be merged？

richtia (commented):
I'm thinking about how we should handle multi-byte character sets, such as unicode.

Some engines seem to provide an options for the chr function (oracle), while other have a separate nchr/nchar function (sql server).

@westonpace do you have any thoughts on this?

extensions/functions_string.yaml
```diff
@@ -17,6 +17,28 @@ scalar_functions:
...
+  - name: chr
+    description: >-
+    Returns the ASCII character having the binary equivalent to input.
+    If n is larger than 256 the result is equivalent to chr(n % 256).
+    If n is negative value, the result is null.
+    impls:
+    - args:
+      - value: i8
+        name: "input"
+      return: "fixedchar<L1>"
```
westonpace, line 29: What is `L1` here?  None of the input arguments are parameterized so I don't think the output can be parameterized.  Do you perhaps want `fixedchar<1>`?  It seems postgres, duckdb, datafusion, and presto all return a variable size string.  Only SQL server returns a fixed size string.  I think returning `string` here would be more consistent with existing implementations.

westonpace (changes requested):

jacques-n:

PR #360, opened 2022-10-20
feat: update ascii , char_length and strpos functions definitions.

lviiii:

Before you continue, please ensure that your commit messages follow [conventional commit syntax](1). This will save you a rebase later. Substrait uses an automated release process that, among other things, uses the commit messages to build a changelog, so the syntax and format matters!

The title of the PR should also be a valid commit header.

 - `feat: add feature X`
 - `fix: X in case of Y`

Furthermore, any commit that imposes a breaking change should end in a paragraph that starts with `BREAKING CHANGE: ...`, where `...` explains what changed. The automated release process uses this to determine how it should bump the version number. Anything that changes the behavior of a plan that was previously legal is considered a breaking change; note that this includes behavior specifications that only exist in Substrait in the form of behavior descriptions on the website or in comments.

Please also note breaking changes in the first comment of the PR, so we're less likely to miss them when we review or merge.

### DIFF
--- a/extensions/functions_string.yaml
+++ b/extensions/functions_string.yaml
@@ -503,76 +503,10 @@ scalar_functions:
           - name: case_sensitivity
             options: [ CASE_SENSITIVE, CASE_INSENSITIVE, CASE_INSENSITIVE_ASCII ]
             required: false
-          - value: "string"
-            name: "input"
-            description: The input string.
-          - value: "varchar<L1>"
-            name: "substring"
-            description: The substring to search for.
-        return: i32
-      - args:
-          - name: case_sensitivity
-            options: [ CASE_SENSITIVE, CASE_INSENSITIVE, CASE_INSENSITIVE_ASCII ]
-            required: false
-          - value: "string"
-            name: "input"
-            description: The input string.
-          - value: "fixedchar<L2>"
-            name: "substring"
-            description: The substring to search for.
-        return: i32
-      - args:
-          - name: case_sensitivity
-            options: [ CASE_SENSITIVE, CASE_INSENSITIVE, CASE_INSENSITIVE_ASCII ]
-            required: false
           - value: "varchar<L1>"
             name: "input"
             description: The input string.
-          - value: "string"
-            name: "substring"
-            description: The substring to search for.
-        return: i32
-      - args:
-          - name: case_sensitivity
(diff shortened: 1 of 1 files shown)


extensions/functions_string.yaml
```diff
@@ -17,6 +42,83 @@ scalar_functions:
...
+          - value: "string"
+            name: "substr"
+        return: i32
+      - args:
+          - value: "fixedchar<l1>"
+            name: "input"
+          - value: "fixedchar<l1>"
+            name: "substr"
+        return: i32
+  - name: length
```
richtia, line 104: https://github.com/substrait-io/substrait/blob/main/extensions/functions_string.yaml#L943

extensions/functions_string.yaml
```diff
@@ -17,6 +42,83 @@ scalar_functions:
         variadic:
           min: 1
         return: "string"
+  -
+    name: instr
```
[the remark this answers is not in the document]
lviiii: Thx, I updated strpos and char_length functions definitions to add i32 type.

extensions/functions_string.yaml
```diff
@@ -473,6 +498,138 @@ scalar_functions:
...
+      - args:
+          - name: case_sensitivity
+            options: [ CASE_SENSITIVE, CASE_INSENSITIVE, CASE_INSENSITIVE_ASCII ]
+            required: false
+          - value: "string"
+            name: "input"
+            description: The input string.
+          - value: "fixedchar<L2>"
+            name: "substring"
+            description: The substring to search for.
```
richtia, line 522: The original MR kept the types consistent. Is there a particular reason you want to introduce the interaction between different types?

Here's the original comment about this: https://github.com/substrait-io/substrait/issues/251#issuecomment-1191249269

lviiii: Thanks for your comments, and I kept the types consistent.

extensions/functions_string.yaml
```diff
@@ -1,6 +1,31 @@
 %YAML 1.2
 ---
 scalar_functions:
+  -
+    name: ascii
+    description: Returns the numeric value of the first character of `input`.
```
richtia, line 6:
```suggestion
    description: Returns the ASCII value of the first character of `input`.
```

richtia, line 6: Could you also give more details about the UTF8 option?  The first 128 characters match up with ascii, but what happens after that when the input is UTF8?

jacques-n:

Issue #361, opened 2022-10-20
Member names of the Version message conflict with some macros


bkietz:
For some compilers/C library combinations, the generated accessors `Version::major()` etc collide with a definition in `<sys/sysmacros.h>`

> /home/bkietz/arrow/cpp/debug-build/substrait_ep-generated/substrait/plan.pb.h:850:12: warning: In the GNU C Library, "major" is defined by <sys/sysmacros.h>. For historical compatibility, it is currently defined by <sys/types.h> as well, but we plan to remove this soon. To use "major", include <sys/sysmacros.h> directly. If you did not intend to use a system-defined macro "major", you should undefine it after including <sys/types.h>. [-W#pragma-messages]

This warning arose while compiling `plan.pb.cc`, so unfortunately the fix isn't as simple as an `#undef` before including generated code.

For projects whose CI builds with -Werror (all warnings are errors), this is problematic. Simply suppressing the warning will not work since the C preprocessor will replace instances of `major` with calls to the macro. Could those accessors could be renamed to `major_` and `minor_`?

westonpace:
Closed by #362

PR #362, opened 2022-10-20
fix: Rename Version fields which conflict with <sys/sysmacros.h>

bkietz:
BREAKING CHANGE: renamed `Version.{major -> major_number, minor -> minor_number}`

BREAKING CHANGE: renamed `Version.{patch -> patch_number}` for consistency

### DIFF
--- a/proto/substrait/plan.proto
+++ b/proto/substrait/plan.proto
@@ -59,9 +59,9 @@ message PlanVersion {
 
 message Version {
   // Substrait version number.
-  uint32 major = 1;
-  uint32 minor = 2;
-  uint32 patch = 3;
+  uint32 major_number = 1;
+  uint32 minor_number = 2;
+  uint32 patch_number = 3;
 
   // If a particular version of Substrait is used that does not correspond to
   // a version number exactly (for example when using an unofficial fork or


bkietz:
fixes #361

westonpace:
I don't think there has been widespread integration with the plan version feature yet so this is probably a good time to make this kind of change.

https://github.com/protocolbuffers/protobuf/issues/4654 is probably related but it isn't clear to me if they are saying they aren't going to fix it or if it's already been fixed.  What version of protoc are you using?  Does using an updated version fix it?

I'm +1 on this change.  I don't see much reason not to do it and, even if it can be avoided someway, it's probably best to reduce the number of things users may potentially have to work around.

bkietz:
I don't see any reference to this in CHANGES.txt which *does* list [similar fixes](https://github.com/protocolbuffers/protobuf/blob/main/CHANGES.txt#L1074).

From https://github.com/protocolbuffers/protobuf/issues/4654 there's one comment [saying we just need to rename things](https://github.com/protocolbuffers/protobuf/issues/4654#issuecomment-391457742). There's another comment claiming [this is fixed](https://github.com/protocolbuffers/protobuf/issues/4654#issuecomment-612929215), but that is from Apr 2020 -> circa version 3.12.0. I'm still seeing the issue

```
$ protoc --version
libprotoc 3.19.4
```

westonpace (approved):
Formally marking my +1 but since this is a breaking spec change we need 1 other approval.

cpcloud (approved):
Thanks!

PR #363, opened 2022-10-21
feat: add APPLY operator representation

### LINKED ISSUE #357: Add support for SQL APPLY operation
Related Substrait issues #119 #206 

`APPLY` [1], introduced in SQL Server in 2005, is frequently used in practice. In Calcite the operation is `LATERAL`. This operator looks similar to a `JOIN`, i.e. it executes an expression on the two input relations. However it is semantically different. In a `JOIN`, both sets to be joined must be self-sufficient. Otherwise it may result in errors like "multi-part identifier could not be bound". If dependencies between sets exists, a user can use correlated-subqueries. However,  nested subqueries result in recursive evaluation strategies which incur O(n2) runtime [3].

Example [4]:
Intent: For each row from `table1` select first `rowcount` rows from `table2`, ordered by `table2.id`

Query using `APPLY`
`SELECT  * FROM  table1, CROSS APPLY ( SELECT  TOP (table1.rowcount) * FROM    table2 ORDER BY id ) t2`

Queries without using `APPLY`
```
SELECT  * FROM 
table1 t1 JOIN ( SELECT  t2o.*, ROW_NUMBER() OVER (ORDER BY id) AS rn FROM    table2 t2o ) t2 
ON t2.rn <= t1.rowcount
```

As benchmarked in [4], execution of query with `APPLY` in SQL Server is 2 orders of magnitude faster than the query without apply. I.e. `APPLY` yields efficiency and is not replaceable with traditional `JOIN`s.

Currently APPLY operation cannot be represented in Substrait. Which means, optimizations in various engines cannot be leveraged if `APPLY` like operations are deconstructed and mapped into existing operations supported by `Substrait`. To efficiently support full SQL, operation for `APPLY` is needed in Substrait. De-correlation should not be enforced.

CC: @jcamachor

[1] https://www.mssqltips.com/sqlservertip/1958/sql-server-cross-apply-and-outer-apply/
[2] [CALCITE-1472](https://issues.apache.org/jira/browse/CALCITE-1472)
[3] https://15721.courses.cs.cmu.edu/spring2018/papers/16-optimizer2/hyperjoins-btw2017.pdf
[4] https://explainextended.com/2009/07/16/inner-join-vs-cross-apply/

ashvina:
This PR adds a new relation type for APPLY operation. APPLY is similar to JOIN
operator since it joins two table sources. However, in APPLY, the table source
on the right depends on the result of computation of the table source on left.
Since the right source cannot be computed without left, the right source is
represented by an Expression in Substrait and not a relation. The right source
will most frequently be a SubQuery.

Closes #357

### DIFF
--- a/proto/substrait/algebra.proto
+++ b/proto/substrait/algebra.proto
@@ -178,6 +178,29 @@ message JoinRel {
   substrait.extensions.AdvancedExtension advanced_extension = 10;
 }
 
+// Represents APPLY relational operator, implemented as APPLY in SQL Server and LATERAL in Calcite. It resembles a
+// JOIN operator, but has different semantics. To use a JOIN, there must be no dependency between the two input
+// relations to be joined. Whereas APPLY depends on a correlation predicate, and to perform APPLY, the table
+// source on the right is evaluated against every row of the table source on the left, and not just once.
+// APPLY is frequently used when the expression on the right is a Table-Valued-Function.
+// See: https://www.mssqltips.com/sqlservertip/1958/sql-server-cross-apply-and-outer-apply/
+// e.g. select * from store_sales CROSS APPLY (select * from item where item.i_item_sk = store_sales.ss_item_sk);
+message ApplyRel {
+  RelCommon common = 1;
+  Rel left = 2;
+  Expression right = 3;
+
+  ApplyType type = 4;
+
+  enum ApplyType {
+    Apply_TYPE_UNSPECIFIED = 0;
+    APPLY_TYPE_CROSS = 1;
+    APPLY_TYPE_OUTER = 2;
+  }
+
+  substrait.extensions.AdvancedExtension advanced_extension = 10;
+}
+
 // Cartesian product relational operator of two tables (left and right)
 message CrossRel {
   RelCommon common = 1;


jacques-n:
@ashvina , thanks for proposing this. It seems like we need an additional subquery type to support this pattern, no?

@hannes, I your team worked extensively on these patterns. Do you have a thought of how we should express this?

jacques-n:
Following up here @ashvina , did you see my question about the need for an additional subquery type?

ashvina:
Hi @jacques-n. Sorry, I missed your question earlier.

I agree, to properly support `APPLY`, additional subquery types will be needed. I have added examples queries in [substrait-java/pull/106](https://github.com/substrait-io/substrait-java/pull/106/files#diff-77e15f76bf10fa7f81f2ebd1112adeeb8d0f34bc67831afaaf2feab731f15563). In the examples, the correlated-right-subquery (`FROM` clause) produce a `rel`. A `subquery_type` for this rel is needed. I think the new type would be similar to `subquery:in_predicate`, as it returns a rel.

Although, I think extension of `subquery` expression is independent of this PR and is needed.

Assuming a new `subquery type` which produces a `rel` is defined, what do you think about specification of the `APPLY`?

Note, that the APPLY is generally used with `table-value functions`. IIUC, TVF spec in missing. Also, it is tricky to get java unit-test to work with TVF. Hence I have omitted TVF query examples. I believe the proposed APPLY spec should be able to support TVF  whenever its expression spec is added.

proto/substrait/algebra.proto
```diff
@@ -178,6 +178,29 @@ message JoinRel {
...
+// e.g. select * from store_sales CROSS APPLY (select * from item where item.i_item_sk = store_sales.ss_item_sk);
+message ApplyRel {
+  RelCommon common = 1;
+  Rel left = 2;
+  Expression right = 3;
+
+  ApplyType type = 4;
+
+  enum ApplyType {
+    Apply_TYPE_UNSPECIFIED = 0;
```
elharo, line 196: Why lower case Apply here and APPLY for the next two?

westonpace:
This has been on my "to do" list for a while but I finally got around to looking at it because some behavior like cross apply is being discussed in Arrow (https://github.com/apache/arrow/pull/14682).

@jacques-n I am curious if this really is a subquery.  @ashvina I am also uncertain about expressing the `right` side as an "expression".  That being said, I am very much a novice here :)

Correlated subqueries traditionally return a single value (maybe a single row if multiple aggregates) for each input row:

 * WHERE `X IN ...` (returns a single bool for each row)
 * WHERE `X < ANY(...)` (returns a single bool for each row)
 * SELECT X, `(SUM(...) FROM ...)` (returns a single int64 sum for each row)

However, the right side in a cross apply could return a complete table for each input row.  Or it could (I think) return nothing at all.

That being said, I may be significantly misunderstanding any of these points and would appreciate any clarification.

jacques-n:

PR #368, opened 2022-11-09
feat: add expand rel

JkSelf:
Adds an expand relation.  This relation can be used to create near-duplicate
copies of each input row based on templates describing how to create the
copies.  This is used within spark to implement certain operations like aggregate
rollup and pivot longer.

### DIFF
--- a/proto/substrait/algebra.proto
+++ b/proto/substrait/algebra.proto
@@ -399,7 +399,7 @@ message Rel {
     HashJoinRel hash_join = 13;
     MergeJoinRel merge_join = 14;
     ExchangeRel exchange = 15;
-    ExpandRel exchange = 16;
+    ExpandRel expand = 16;
   }
 }
 
--- a/proto/substrait/algebra.proto
+++ b/proto/substrait/algebra.proto
@@ -364,7 +364,8 @@ message ExpandRel {
   message SwitchingField {
     // Array that contains an expression to output per duplicate_id
     // each `switching_field` must have the same number of expressions
-    // all expressions must be of the same type a
+    // all expressions within a switching field be the same type class but can differ in nullability.
+    // this column will be nullable if any of the expressions are nullable.
     repeated Expression duplicates = 1;
   }
 }
--- a/proto/substrait/algebra.proto
+++ b/proto/substrait/algebra.proto
@@ -362,7 +362,7 @@ message ExpandRel {
   }
 
   message SwitchingField {
-    // Array that contains the expression to output per duplicate_id
+    // Array that contains an expression to output per duplicate_id
     // must have the same number of expressions as each input record
     // all expressions must be of the same type a
     repeated Expression duplicates = 1;
--- a/proto/substrait/algebra.proto
+++ b/proto/substrait/algebra.proto
@@ -342,6 +342,25 @@ message ExchangeRel {
   }
 }
 
+// The expand relation generates multiple rows per input row (denoted by
+// aggregate_expressions) which are then aggregated by groupings. No
+// ordering is applied to the aggregated results.
+message ExpandRel {
+  RelCommon common = 1;
+  Rel input = 2;
+  // The output of ExpandRel.
+  NamedStruct output = 3;
+
+  // A list of expression grouping that the aggregation measured should be calculated for.
+  repeated Grouping groupings = 4;
+
+  message Grouping {
+    repeated Expression projection_expression = 1;
+  }
+
+  substrait.extensions.AdvancedExtension advanced_extension = 10;
+}
+
 // A relation with output field names.
 //
 // This is for use at the root of a `Rel` tree.
@@ -367,11 +386,11 @@ message Rel {
     ExtensionMultiRel extension_multi = 10;
     ExtensionLeafRel extension_leaf = 11;
     CrossRel cross = 12;
-
     //Physical relations
     HashJoinRel hash_join = 13;
     MergeJoinRel merge_join = 14;
     ExchangeRel exchange = 15;
+    ExpandRel exchange = 16;
   }
 }
 
--- a/proto/substrait/algebra.proto
+++ b/proto/substrait/algebra.proto
@@ -343,8 +343,7 @@ message ExchangeRel {
 }
 
 // The expand relation generates multiple rows per input row (denoted by
-// aggregate_expressions) which are then aggregated by groupings. No
-// ordering is applied to the aggregated results.
+// groupings). No ordering is applied to the results.
 message ExpandRel {
   RelCommon common = 1;
   Rel input = 2;
--- a/proto/substrait/algebra.proto
+++ b/proto/substrait/algebra.proto
@@ -347,11 +347,9 @@ message ExchangeRel {
 message ExpandRel {
   RelCommon common = 1;
   Rel input = 2;
-  // The output of ExpandRel.
-  NamedStruct output = 3;
 
   // A list of expression grouping that the aggregation measured should be calculated for.
-  repeated Grouping groupings = 4;
+  repeated Grouping groupings = 3;
 
   message Grouping {
     repeated Expression projection_expression = 1;
--- a/proto/substrait/algebra.proto
+++ b/proto/substrait/algebra.proto
@@ -343,7 +343,7 @@ message ExchangeRel {
 }
 
 // Duplicates records, possibly switching output expressions between each duplicate.
-// Default output is all of the fields declared + one field that int64 type that returns the
+// Default output is all of the fields declared followed by one int64 field that contains the
 // duplicate_id which is a zero-index ordinal of which duplicate of the original record this
 // corresponds to.
 message ExpandRel {
--- a/proto/substrait/algebra.proto
+++ b/proto/substrait/algebra.proto
@@ -364,7 +364,7 @@ message ExpandRel {
   message SwitchingField {
     // Array that contains an expression to output per duplicate_id
     // each `switching_field` must have the same number of expressions
-    // all expressions within a switching field be the same type class but can differ in nullability.
+    // all expressions within a switching field must be the same type class but can differ in nullability.
     // this column will be nullable if any of the expressions are nullable.
     repeated Expression duplicates = 1;
   }
--- a/proto/substrait/algebra.proto
+++ b/proto/substrait/algebra.proto
@@ -342,20 +342,31 @@ message ExchangeRel {
   }
 }
 
-// The expand relation generates multiple rows per input row (denoted by
-// groupings). No ordering is applied to the results.
+// Duplicates records, possibly switching output expressions between each duplicate.
+// Default output is all of the fields declared + one field that int64 type that returns the
+// duplicate_id which is a zero-index ordinal of which duplicate of the original record this
+// corresponds to.
 message ExpandRel {
   RelCommon common = 1;
   Rel input = 2;
+  repeated ExpandField fields = 4;
 
-  // A list of expression grouping that the aggregation measured should be calculated for.
-  repeated Grouping groupings = 3;
+  message ExpandField {
+    oneof field_type {
+      // Field that switches output based on which duplicate_id we're outputting
+      SwitchingField switching_field = 2;
 
-  message Grouping {
-    repeated Expression projection_expression = 1;
+      // Field that outputs the same value no matter which duplicate_id we're on.
+      Expression consistent_field = 3;
+    }
   }
 
-  substrait.extensions.AdvancedExtension advanced_extension = 10;
+  message SwitchingField {
+    // Array that contains the expression to output per duplicate_id
+    // must have the same number of expressions as each input record
+    // all expressions must be of the same type a
+    repeated Expression duplicates = 1;
+  }
 }
 
 // A relation with output field names.
--- a/proto/substrait/algebra.proto
+++ b/proto/substrait/algebra.proto
@@ -363,7 +363,7 @@ message ExpandRel {
 
   message SwitchingField {
     // Array that contains an expression to output per duplicate_id
-    // must have the same number of expressions as each input record
+    // each `switching_field` must have the same number of expressions
     // all expressions must be of the same type a
     repeated Expression duplicates = 1;
   }
--- a/site/docs/relations/physical_relations.md
+++ b/site/docs/relations/physical_relations.md
@@ -224,7 +224,7 @@ similar to a project operation.
 | Property  | Description                          | Required |
 | --------- |--------------------------------------| -------- |
 | Input     | The relational input.                | Required |
-| ExpandField | One of SwitchingField of Expression. | Required |
+| Fields | Expressions describing the output fields.  These refer to the schema of the input.  A field can have a different expression for each duplicate.  | Required |
 
 ## Hashing Window Operation
 
--- a/site/docs/relations/physical_relations.md
+++ b/site/docs/relations/physical_relations.md
@@ -210,12 +210,12 @@ The streaming aggregate operation leverages data ordered by the grouping express
 
 The expand operation creates duplicates of input records based on ExpandFields. The ExpandField can be the SwitchingField or ConsistentField. The SwitchingField here means that switches output based on which duplicate_id we're outputting. And the ConsistentField means that outputs the same value no matter which duplicate_id we're on.
 
-| Signature            | Value                                |
-| -------------------- |--------------------------------------|
-| Inputs               | 1                                    |
-| Outputs              | 1                                    |
-| Property Maintenance | Distribution is maintained for consistent fields with direct references.  Ordering is maintained if all ordering fields are consistent fields with direct references. |
-| Direct Output Order  | The expand fields and duplicated id. |
+| Signature            | Value                                                                                                                                                                          |
+| -------------------- |--------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
+| Inputs               | 1                                                                                                                                                                              |
+| Outputs              | 1                                                                                                                                                                              |
+| Property Maintenance | Distribution is maintained if all the distribution fields are consistent fields with direct references. Ordering only matintaines the consistent fields not the switch fields. |
+| Direct Output Order  | The expand fields and duplicated id.                                                                                                                                           |
 
 ### Expand Properties
 
--- a/site/docs/relations/physical_relations.md
+++ b/site/docs/relations/physical_relations.md
@@ -206,6 +206,28 @@ The streaming aggregate operation leverages data ordered by the grouping express
 | Measures         | A list of one or more aggregate expressions. Aggregate expressions ordering requirements must be compatible with expected ordering. | Optional, required if no grouping sets. |
 
 
+## Expand Operation
+
+The expand operation creates duplicates of input records based on groupings.  For each grouping the
+expressions specific to that grouping are evaluated and the result is added to the input record in a manner
+similar to a project operation.  The expand operation can be used to calculate aggregate grouping sets.
+
+| Signature            | Value                                                                               |
+| -------------------- | ----------------------------------------------------------------------------------- |
+| Inputs               | 1                                                                                   |
+| Outputs              | 1                                                                                   |
+| Property Maintenance | Distribution and ordering are not maintained.                                       |
+| Direct Output Order  | Same as defined by the [Project](logical_relations.md#project-operation) operation. |
+
+### Expand Properties
+
+| Property  | Description                                                                             | Required |
+| --------- | --------------------------------------------------------------------------------------- | -------- |
+| Input     | The relational input.                                                                   | Required |
+| Groupings | Sets of expressions.  There will be one output row for each input row for each grouping | Required |
+
+Note: Each grouping must have the same number of expressions and the return types for each new output column must
+be consistent across all groupings.
 
 ## Hashing Window Operation
 
--- a/site/docs/relations/physical_relations.md
+++ b/site/docs/relations/physical_relations.md
@@ -208,7 +208,7 @@ The streaming aggregate operation leverages data ordered by the grouping express
 
 ## Expand Operation
 
-The expand operation creates duplicates of input records based on ExpandFields. The ExpandField can be the SwitchingField or ConsistentField. The SwitchingField here means that switches output based on which duplicate_id we're outputting. And the ConsistentField means that outputs the same value no matter which duplicate_id we're on.
+The expand operation creates duplicates of input records based on the Expand Fields. Each Expand Field can be a Switching Field or an expression. Switching Fields are described below.  If an Expand Field is an expression then its value is consistent across all duplicate rows.
 
 | Signature            | Value                                                                                                                                                                          |
 | -------------------- |--------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
--- a/site/docs/relations/physical_relations.md
+++ b/site/docs/relations/physical_relations.md
@@ -215,7 +215,7 @@ The expand operation creates duplicates of input records based on the Expand Fie
 | Inputs               | 1                                                                                                                                                                              |
 | Outputs              | 1                                                                                                                                                                              |
 | Property Maintenance | Distribution is maintained if all the distribution fields are consistent fields with direct references. Ordering can only be maintained down to the level of consistent fields that are kept.|
-| Direct Output Order  | The expand fields and duplicated id.                                                                                                                                           |
+| Direct Output Order  | The expand fields followed by an i32 column describing the index of the duplicate that the row is derived from.                                                                                                                                           |
 
 ### Expand Properties
 
--- a/site/docs/relations/physical_relations.md
+++ b/site/docs/relations/physical_relations.md
@@ -214,7 +214,7 @@ The expand operation creates duplicates of input records based on ExpandFields. 
 | -------------------- |--------------------------------------|
 | Inputs               | 1                                    |
 | Outputs              | 1                                    |
-| Property Maintenance | Duplicated columns are maintained.   |
+| Property Maintenance | Distribution is maintained for consistent fields with direct references.  Ordering is maintained if all ordering fields are consistent fields with direct references. |
 | Direct Output Order  | The expand fields and duplicated id. |
 
 ### Expand Properties
--- a/site/docs/relations/physical_relations.md
+++ b/site/docs/relations/physical_relations.md
@@ -208,16 +208,14 @@ The streaming aggregate operation leverages data ordered by the grouping express
 
 ## Expand Operation
 
-The expand operation creates duplicates of input records based on ExpandFields. For each ExpandField the
-expressions specific to that grouping are evaluated and the result is added to the input record in a manner
-similar to a project operation. 
+The expand operation creates duplicates of input records based on ExpandFields. The ExpandField can be the SwitchingField or ConsistentField. The SwitchingField here means that switches output based on which duplicate_id we're outputting. And the ConsistentField means that outputs the same value no matter which duplicate_id we're on.
 
-| Signature            | Value                                                                               |
-| -------------------- | ----------------------------------------------------------------------------------- |
-| Inputs               | 1                                                                                   |
-| Outputs              | 1                                                                                   |
-| Property Maintenance | Distribution and ordering are not maintained.                                       |
-| Direct Output Order  | Same as defined by the [Project](logical_relations.md#project-operation) operation. |
+| Signature            | Value                                |
+| -------------------- |--------------------------------------|
+| Inputs               | 1                                    |
+| Outputs              | 1                                    |
+| Property Maintenance | Duplicated columns are maintained.   |
+| Direct Output Order  | The expand fields and duplicated id. |
 
 ### Expand Properties
 
--- a/site/docs/relations/physical_relations.md
+++ b/site/docs/relations/physical_relations.md
@@ -208,9 +208,9 @@ The streaming aggregate operation leverages data ordered by the grouping express
 
 ## Expand Operation
 
-The expand operation creates duplicates of input records based on groupings.  For each grouping the
+The expand operation creates duplicates of input records based on ExpandFields. For each ExpandField the
 expressions specific to that grouping are evaluated and the result is added to the input record in a manner
-similar to a project operation.  The expand operation can be used to calculate aggregate grouping sets.
+similar to a project operation. 
 
 | Signature            | Value                                                                               |
 | -------------------- | ----------------------------------------------------------------------------------- |
@@ -221,13 +221,10 @@ similar to a project operation.  The expand operation can be used to calculate a
 
 ### Expand Properties
 
-| Property  | Description                                                                             | Required |
-| --------- | --------------------------------------------------------------------------------------- | -------- |
-| Input     | The relational input.                                                                   | Required |
-| Groupings | Sets of expressions.  There will be one output row for each input row for each grouping | Required |
-
-Note: Each grouping must have the same number of expressions and the return types for each new output column must
-be consistent across all groupings.
+| Property  | Description                          | Required |
+| --------- |--------------------------------------| -------- |
+| Input     | The relational input.                | Required |
+| ExpandField | One of SwitchingField of Expression. | Required |
 
 ## Hashing Window Operation
 
--- a/site/docs/relations/physical_relations.md
+++ b/site/docs/relations/physical_relations.md
@@ -222,7 +222,7 @@ The expand operation creates duplicates of input records based on the Expand Fie
 | Property  | Description                          | Required |
 | --------- |--------------------------------------| -------- |
 | Input     | The relational input.                | Required |
-| Direct Fields | Expressions describing the output fields.  These refer to the schema of the input.  Each Direct Field must be a Consistent Field or a Switching Field  | Required |
+| Direct Fields | Expressions describing the output fields.  These refer to the schema of the input.  Each Direct Field must be an expression or a Switching Field  | Required |
 
 ### Switching Field Properties
 
--- a/site/docs/relations/physical_relations.md
+++ b/site/docs/relations/physical_relations.md
@@ -222,7 +222,15 @@ The expand operation creates duplicates of input records based on the Expand Fie
 | Property  | Description                          | Required |
 | --------- |--------------------------------------| -------- |
 | Input     | The relational input.                | Required |
-| Fields | Expressions describing the output fields.  These refer to the schema of the input.  A field can have a different expression for each duplicate.  | Required |
+| Direct Fields | Expressions describing the output fields.  These refer to the schema of the input.  Each Direct Field must be a Consistent Field or a Switching Field  | Required |
+
+### Switching Field Properties
+
+A switching field is a field whose value is different in each duplicated row.  All switching fields in an Expand Operation must have the same number of duplicates.
+
+| Property  | Description                          | Required |
+| --------- |--------------------------------------| -------- |
+| Duplicates | List of one or more expressions.  The output will contain a row for each expression. | Required |
 
 ## Hashing Window Operation
 
--- a/site/docs/relations/physical_relations.md
+++ b/site/docs/relations/physical_relations.md
@@ -214,7 +214,7 @@ The expand operation creates duplicates of input records based on ExpandFields. 
 | -------------------- |--------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
 | Inputs               | 1                                                                                                                                                                              |
 | Outputs              | 1                                                                                                                                                                              |
-| Property Maintenance | Distribution is maintained if all the distribution fields are consistent fields with direct references. Ordering only matintaines the consistent fields not the switch fields. |
+| Property Maintenance | Distribution is maintained if all the distribution fields are consistent fields with direct references. Ordering can only be maintained down to the level of consistent fields that are kept.|
 | Direct Output Order  | The expand fields and duplicated id.                                                                                                                                           |
 
 ### Expand Properties


JkSelf:
@jacques-n @rui-mo Please help to review. Thanks.

jacques-n:
Can you add side content that describes what expand would do? I'm not familiar with the operator.

JkSelf:
@jacques-n Updated the description document. 
Currently the Aggregate Rel output is mainly the Grouping sets expression and Measure.  And we need the  Grouping sets expression, aggregate expressions and goup ids. So the Aggregate Rel may can't meet the needs of Expand very well.

JkSelf:
cc @FelixYBW @baibaichen

proto/substrait/algebra.proto
```diff
@@ -339,6 +339,23 @@ message ExchangeRel {
     }
   }
 }
+message ExpandRel {
```
EpsilonPrime, line 342: This is what I see the relation doing (based on my cursory glance at the referenced spark PR).  Please feel to adjust.

```suggestion

// The expand relation generates multiple rows per input row (denoted by 
// aggregate_expressions) which are then aggregated by groupings.  No
// ordering is applied to the aggregated results.
message ExpandRel {
```

JkSelf: Updated. Thanks for your review!

JkSelf:
Just for reference PR.
Spark add the Expand operator in order to support the `GROUPING SETS`, `ROLLUP`, `CUBE `expression in [PR#1567](https://github.com/apache/spark/pull/1567), which apply all the GroupExpressions to every input row and then get multiple output rows for an input row. When we add the Expand operator in gluten, we need adding ExpandRel support in proto.

baibaichen:
@jacques-n @westonpace 

`Expand` operator is introduced to support cube, rollup, and grouping sets(see PR in @JkSelf comments). To support these functions in a naive way, we can scan the source input for each group and then union the results. `Expand` is used to avoid such multiple scans by ouptuting multiple rows for each input row.

Furthermore, Group sets are also used to optimize multiple distinct aggregations. see [SPARK-9241](https://issues.apache.org/jira/browse/SPARK-9241) and [CALCITE-732](https://issues.apache.org/jira/browse/CALCITE-732)

I [did some research](https://github.com/oap-project/gluten/pull/1104/commits/1298d5678e97d986aa0f8056c13f4f32e9d2635d) on translating `Aggregate` over `Expand` to substrait style aggregate. It's ok for normal groupingsets, cube and rollup by its fixed pattern. However, `Expand` just outputs multiple rows,  it can be configured by different pattern, i.e. [SPARK-9241](https://issues.apache.org/jira/browse/SPARK-9241).  I failed to translate in such case.

westonpace:
I think I understand that `EXPAND` is a helper that is used by Spark to calculate an aggregate that contains grouping sets.  I'm not quite sure I understand how it works.  Does each copy get sent to a different output?  Can you share a Spark physical plan that makes use of EXPAND?

Does it look something like this?

```
SELECT Country, Region, SUM(Sales) AS TotalSales
FROM Sales
GROUP BY ROLLUP (Country, Region);
```

Translates to...

```
SCAN -> EXPAND --- AGG(keys={country,region}) --- UNION
               \-- AGG(keys={country}) ---------/
               \-- AGG(keys={}) ---------------/
```

baibaichen:
> I think I understand that `EXPAND` is a helper that is used by Spark to calculate an aggregate that contains grouping sets. I'm not quite sure I understand how it works. Does each copy get sent to a different output? Can you share a Spark physical plan that makes use of EXPAND?
> 
> Does it look something like this?
> 
> ```
> SELECT Country, Region, SUM(Sales) AS TotalSales
> FROM Sales
> GROUP BY ROLLUP (Country, Region);
> ```
> 
> Translates to...
> 
> ```
> SCAN -> EXPAND --- AGG(keys={country,region}) --- UNION
>                \-- AGG(keys={country}) ---------/
>                \-- AGG(keys={}) ---------------/
> ```

`Scan -> Expand(output, projections)-> Aggregate[groupby, measure]`

```
Expand:
  output = sales, Country, Region, grouping_id(Literal)
  projects =[
     [sales, Country, Region, 0],
     [sales, Country, Null,   1],
     [sales, Null, Null,      6]
  ]

Aggregate:
  groupby = Country, Region, grouping_id
  meausre = sum(sales)
```

westonpace:
Ok.  I think I understand now.  So, in the above example, if the input data was:

| Country | Region | Sales |
| --- | --- | --- |
| US | North America | 100 |
| US | North America | 200 |
| Canada | North America | 300 |
| France | Europe | 400 |

Then the output data would be:

| Country | Region | Sales |
| --- | --- | --- |
| US | North America | 100 |
| US | NULL | 100 |
| NULL | NULL | 100 |
| US | North America | 200 |
| US | NULL | 200 |
| NULL | NULL | 200 |
| Canada | North America | 300 |
| Canada | NULL | 300 |
| NULL | NULL | 300 |
| France | Europe | 400 |
| France | NULL | 400 |
| NULL | NULL | 400 |

JkSelf:
@westonpace  Yes. The output also contains the group_id cols.



<head>

<meta name=ProgId content=Excel.Sheet>
<meta name=Generator content="Microsoft Excel 15">
<link id=Main-File rel=Main-File
href="file:///C:/Users/kejia/AppData/Local/Temp/msohtmlclip1/01/clip.htm">
<link rel=File-List
href="file:///C:/Users/kejia/AppData/Local/Temp/msohtmlclip1/01/clip_filelist.xml">



</head>

<body link="#0563C1" vlink="#954F72">



Sales | Country | Region | grouping_id
-- | -- | -- | --
100 | US | North America | 0
100 | US | NULL | 1
100 | NULL | NULL | 3
200 | US | North America | 0
200 | US | NULL | 1
200 | NULL | NULL | 3
300 | Canada | North America | 0
300 | Canada | NULL | 1
300 | NULL | NULL | 3
400 | France | Europe | 0
400 | France | NULL | 1
400 | NULL | NULL | 3



</body>

</html>

baibaichen:
@JkSelf As we discussed offline, it's better to follow spark's semantics to define `ExpandRel`

proto/substrait/algebra.proto
```diff
@@ -340,6 +340,23 @@ message ExchangeRel {
...
+// aggregate_expressions) which are then aggregated by groupings. No
+// ordering is applied to the aggregated results.
+message ExpandRel {
+  RelCommon common = 1;
+  Rel input = 2;
+
+  // A list of expression grouping that the aggregation measured should be calculated for.
+  repeated Grouping groupings = 4;
+
+  message Grouping {
```
baibaichen, line 353: where is `ExpandRel` output?

JkSelf: Do we need to define the output of the physical operator? It seems the output is defined by the consumer.

baibaichen: How does `Aggregate` 's  operator reference the output of `Expand`

JkSelf: @baibaichen Added the `output ` in `ExpandRel`.

westonpace:
This will need a description in the markdown as well.  I have created a PR into your branch with a suggestion.

proto/substrait/algebra.proto
westonpace: I disagree with this.  Substrait relations do not typically include names for the output.  See the `ProjectRel` for an example.

baibaichen: > I disagree with this. Substrait relations do not typically include names for the output. See the `ProjectRel` for an example.

@westonpace, you are right, the name is unnecessry,  `Type.Struct` should be fine.  We actually need the index (for oder) and the type (for nullability) of `ExpandRel`'s ouput.

westonpace: I believe the order should be the same as `ProjectRel`.  By default the newly created columns will be at the end.  You can use an emit relation to reorder them where needed.  Is there some reason this wouldn't work?

The type should come from the expressions themselves.  Every expression is has a return type and so you should be able to determine the output type of the columns from the expressions.  Again, this is the same way the `ProjectRel` works.

baibaichen: IIUC, expression need refer to input from child plan. However, there is no such input for outupt of `ExpandRel`

westonpace: I'm not certain I understand.  Let us consider the example above:

```
The physical plan looks like Scan -> Expand(output, projections)-> Aggregate[groupby, measure]
```

The expressions in `groupings` will refer to the output schema of the scan in this case.  So, for example, to refer to the Country column:

```
{
  "selection": {
    "direct_reference": {
      "struct_field": 1 # Assuming 1 is the index of Country in the scan output
    },
    "root_reference": {}
  }
}
```

The type can be found by looking at the output schema of scan.  To insert a literal `NULL` the expression would be...

```
{
  "literal": {
    "null": {
      "string": {}
    }   
  }
}
```
All literals, including null literals, are typed (all expressions have a clearly defined type).

baibaichen: @westonpace 

> I'm not certain I understand. Let us consider the example above:
> ```
> The physical plan looks like Scan -> Expand(output, projections)-> Aggregate[groupby, measure]
> ```
> The expressions in groupings will refer to the output schema of the scan in this case. So, for example, to refer to the Country column:

1. `groupings` of `ExpandRel` will refer to its child's output, for example, it refers to scan in your example.
2.  `output` of `ExpandRel` is the input of  `AggregateRel` which is `ExpandRel`'s parent.

The reason that we can't refer to `groupings' is that there is a group of expressions, and each group with different nullable property.

Spark mixes measure column and group column into `Expand`'s `projections` (which @JkSelf named `groups` in substrait's ExpandRel), and we can't distinguish them easily.​

westonpace: > If this is only duplicating records (and nothing else

It isn't duplication and nothing else though.  Each copy of the input row will have a unique projection applied to it.  Consider the above example.  The groupings / projection expressions will be:

```
[
  [field("country"), field("region")]
  [field("country"), literal(null, string())]
  [literal(null, string()), literal(null, string())]
]
```

The grouping id (like in aggregate rel) can be implicit (not specified in the projections).  Given an input row:

| Country | Region | Sales |
| --- | --- | --- |
| US | North America | 100 |

There will be three output rows.  Before any emit they will be:

| Country | Region | Sales | Expr0 | Expr1 | Grouping Id |
| --- | --- | --- | --- | --- | --- |
| US | North America | 100 | US | North America | 0 |
| US | North America | 100 | US | NULL | 1 |
| US | North America | 100 | NULL | NULL | 3 |

Presumably, an emit will be added `[2, 3, 4, 5]` to yield:

| Sales | Expr0 | Expr1 | Grouping Id |
| --- | --- | --- | --- |
| 100 | US | North America | 0 |
| 100 | US | NULL | 1 |
| 100 | NULL | NULL | 3 |

westonpace: > I don't understand why the same behavior as aggregate wouldn't work here. It may not be how people have thought about it but it seems like it would make things consistent and would apply cleanly onto the concepts trying to be capture here.

I think that is fine.  This particular node operates more similarly to projection than to aggregation in my mind.  So I was suggesting we keep the output order similar to projection.

However, I don't know of any case where the original columns, once referenced, wouldn't be dropped.  So if we want the output order to be:

```
unreferenced columns | grouping expressions | grouping id
```

I think that should still work (though "unreferenced columns" is a bit of a new concept).

> My bad, you're right that it's duplication + nullification of specific columns per grouping, (same as the already specified behavior in aggregate).

Yes.  I think it would be useful if the `nullification` wasn't required though.  The operator would be simpler to understand and then this expand behavior is also useful for [`pivot_longer`](https://tidyr.tidyverse.org/reference/pivot_longer.html)

Given an input table:

| School | Grade 1 Avg Score | Grade 2 Avg Score |
| --- | --- | --- |
| Alpha Elementary | 95 | 92 |
| Beta Elementary | 97 | 93 |

and an expand spec:

```
[
  [literal("Grade 1", string()), field("Grade 1 Avg Score")],
  [literal("Grade 2", string()), field("Grade 2 Avg Score")]
]
```

You can get:

| School | Expr0 | Expr1 | Grouping Id |
| --- | --- | --- | --- |
| Alpha Elementary | Grade 1 | 95 | 0 |
| Alpha Elementary | Grade 2 | 92 | 1 |
| Beta Elementary | Grade 1 | 97 | 0 |
| Beta Elementary | Grade 2 | 93 | 1 |

Although in this case one would normally drop the `Grouping Id` column (but that could be done with an emit).

jacques-n: > pivot_longer

WRT to pivot_longer, I think keeping nullification and simply using a subsequent projection with coalesce would be the preferred way to physically implement.

> unreferenced columns/ "like" projection

Hmm. I think of projection as being fundamentally something that doesn't change the number of records so I don't really think of this as the same (fwiw). 

As I think more about it, one of the question I have: can grouping expressions be complex in the systems that use this? If not, this can be come simplified by simply being a int[] "direct projection" column ids + a int[][] of groupings. Then default output order would be direct projection + referenced grouping columns in their original order + group id. If they support arbitrary expressions, we'll need to come up with a more complicated logic around output order (similar to the way it is expressed in aggregate).

westonpace: > WRT to pivot_longer, I think keeping nullification and simply using a subsequent projection with coalesce would be the preferred way to physically implement.

This could work.  I'm against it since I don't think it would be more general, more efficient, or more obvious.  However, I don't want a discussion of pivot_longer to interfere with this PR so we can drop it for later.

> If not, this can be come simplified by simply being a int[] "direct projection" column ids + a int[][] of groupings. Then default output order would be direct projection + referenced grouping columns in their original order + group id

As far as I know, this should work.  I'd support this approach if it works for @chaojun-zhang

baibaichen: > I'd like to propose that we use the same pattern as we do in aggregate. The index of the groupings is an additional projection. If this is only duplicating records (and nothing else, that allows us to remove projection_expression and the type.struct output. People can use emit if they want to hide some columns. This is really a physical decomposition of the first half of the logical aggregate with grouping sets so let's be consistent.

@jacques-n  please say [this reply](https://github.com/substrait-io/substrait/pull/368#issuecomment-1469600604).  

This is actually Spark's implementation of the grouping set, spark compiles (optimizes) the grouping set into `Aggregation` on `Expand`.

It is difficult to revert this optimization on the producer side and re-optimizes it on the consumer side.  I did some research for  translating Expand. My research was a bit like disassembling binary codes, which doesn't guarantee that all cases will work. In some cases, because the pattern is fixed, I can work like this. In some cases(i.e. [Spark-9241](https://issues.apache.org/jira/browse/SPARK-9241)), there is no fixed pattern, so I can’t do such a revert.

baibaichen: @jacques-n @westonpace 

I am not clear about your discussion. What do we need to do to make progress for his PR now?

westonpace: > I am not clear about your discussion. What do we need to do to make progress for his PR now?

Currently we have:

```
message ExpandRel {
  RelCommon common = 1;
  Rel input = 2;
  Type.Struct output = 3;
  repeated Grouping groupings = 4;
  message Grouping {
    repeated Expression projection_expression = 1;
  }
}
```

I am proposing (remove `output`):

```
message ExpandRel {
  RelCommon common = 1;
  Rel input = 2;
  repeated Grouping groupings = 4;
  message Grouping {
    repeated Expression projection_expression = 1;
  }
}
```

I believe that `output` can be calculated from `input` and `groupings` and is redundant.

Jacques is proposing (convert `groupings` from `Expression[][]` to `int[][]`):

```
message ExpandRel {
  RelCommon common = 1;
  Rel input = 2;
  repeated Grouping groupings = 4;
  message Grouping {
    repeated int columns = 1;
  }
}
```

> As I think more about it, one of the question I have: can grouping expressions be complex in the systems that use this? If not, this can be come simplified by simply being a int[] "direct projection" column ids + a int[][] of groupings. Then default output order would be direct projection + referenced grouping columns in their original order + group id. If they support arbitrary expressions, we'll need to come up with a more complicated logic around output order (similar to the way it is expressed in aggregate).

@jacques-n I have slept on this and I think I disagree.  The way `Expand` is currently implemented in both Acero and Spark is a 2d array of expressions.  Since this is a physical relation I think we should aim to match closely the capability being provided more than the intent.

Also, I said:

> I don't want a discussion of pivot_longer to interfere with this PR so we can drop it for later.

However, I have since discovered that Spark also uses Expand for `pivot_longer` (Spark calls this `UNPIVOT`) so I think it is relevant for both engines.

```
-- !query
SELECT * FROM courseEarnings
UNPIVOT (
  earningsYear FOR year IN (`2012`, `2013`, `2014`)
)
-- !query analysis
Project [course#x, year#x, earningsYear#x]
+- Filter isnotnull(coalesce(earningsYear#x))
   +- Expand [[course#x, 2012, 2012#x], [course#x, 2013, 2013#x], [course#x, 2014, 2014#x]], [course#x, year#x, earningsYear#x]
      +- SubqueryAlias courseearnings
         +- View (`courseEarnings`, [course#x,2012#x,2013#x,2014#x])
            +- Project [cast(course#x as string) AS course#x, cast(2012#x as int) AS 2012#x, cast(2013#x as int) AS 2013#x, cast(2014#x as int) AS 2014#x]
               +- Project [course#x, 2012#x, 2013#x, 2014#x]
                  +- SubqueryAlias courseEarnings
                     +- LocalRelation [course#x, 2012#x, 2013#x, 2014#x]
```

In Acero we require each row (each `Grouping`) to have the same number of expressions.  So the output is fairly straightforward if you think of it like projection (perhaps it would help to think of it as a split into N paths where each path has a project rel and then all the paths are combined in a set union).

I believe Spark also requires each grouping to have the same number of expressions.

baibaichen: @westonpace Thank you for your detailed reply！

First,  "unreferenced columns" in your example is actually refered by Spark `Aggregate`'s `AggregateExpression`, so I called it " measure columns". The nullable attribute of the measure column comes from the child nodes of `Expand`, we can **assume** that the nullable attribute of the grouping expression is always true.

```
|unreferenced columns | grouping expressions | grouping id  => your example
|    measure columns  | grouping expressions | grouping id  => 
```

1. With `Type.Struct output`,  we can simply restore the `Expand` operator on the consumer side. 
2. Without `Type.Struct output`,  we need to distinguish between the measure columns and the grouping columns first. For the measure columns, we may need check whether all the projections are the same. For the grouping columns, we can directly set it to null **based on the assumption**.

My question is, adding a field can simplify the work of the consumer side, why do we need to save this field and make the consumer side more complicated? In addition, these assumptions may also be broken in the future.

JkSelf: One question: spark generates different `projections `in `Expand `with `grouping sets`, `rollup `and `cube ` expression. And how we handle these different cases if without the output?

westonpace: > My question is, adding a field can simplify the work of the consumer side, why do we need to save this field and make the consumer side more complicated? In addition, these assumptions may also be broken in the future.

I thought this would be very similar to how `ProjectRel` is handled.  For example:

```
ProjectRel {
  expressions = [some_function(selection(3))]
}
```

The output's nullability will be determined by the function `some_function`.  For example, if `some_function` is `abs` (absolute value) then the type and nullability will be the same as column 3.  If `some_function` is `is_null` then the type will be `bool` and the output will be non-null.

In `Expand` I think this is a little harder.  For example, if we have:

```
[
  [field("country"), field("region")]
  [field("country"), literal(null, string())]
  [literal(null, string()), literal(null, string())]
]
```

We have to make sure:

 * The output type must be the same for each position (e.g. `field("country")` and `literal(null, string())` must have the same output type.  `field("region") and `literal(null, string())` must have the same output type).
 * If any output type in a position is nullable then the output type must be nullable.

Putting this together the code might look like:

```
def calculate_output(groupings):
  output_columns=[]
  # Use the first grouping to determine output types
  for expression in groupings[0]:
    output_columns.append(expression.output_type)
  # Make sure the rest of the groupings have the same output types
  # and check to see if they are nullable
  for grouping in groupings[1:]:
    for col_idx in range(len(grouping)):
      # types must be same but nullability can differ
      if output_columns[col_idx].type != grouping[col_idx].output_type:
        raise Exception("All output types for a column must be the same")
      if grouping[col_idx].output_type.nullable:
        output_columns[col_idx].nullable = true
  return output_columns
```

Note: I think the above algorithm works for both "unreferenced columns" and "measure columns".  You don't need to rely on assumption because the the expression `literal(null, string())` will always have a nullable output type so the output for that column will always be nullable.

> why do we need to save this field and make the consumer side more complicated?

The downsides I see are:

 1. It is harder to produce the plan.  The producer now has to calculate the output.
 2. It is possible for a plan to be invalid if output doesn't match the expression (e.g. output is `int` but the expression return type is `bool`).
 3. The plan is larger than it needs to be
 4. This is different than ProjectRel and other relations where we expect the consumer to be able to calculate the output schema.

> One question: spark generates different projections in Expand with grouping sets, rollup and cube expression. And how we handle these different cases if without the output?

I think that is ok.  If you have `ROLLUP` then you have:

```
[
  [field("country"), field("region")],
  [field("country"), literal(null, string())],
  [literal(null, string()), literal(null, string())]
]
```

If you have `CUBE` then you have:

```
[
  [field("country"), field("region")],
  [field("country"), literal(null, string())],
  [literal(null, string()), field("region")],
  [literal(null, string()), literal(null, string())]
]
```

baibaichen: @westonpace thanks.
> ```
> def calculate_output(groupings):
> ....
> ```

1. We assume the type should be same at the same posisiton, it would be true now, but I don't think **it would always true**.
2. The cons is all consumers should have the similar codes.

> 4. This is different than ProjectRel and other relations where we expect the consumer to be able to calculate the output schema.

I agree with the 4th. So let's go on with `calculate_output` until we find it can't meet the requirements.

JkSelf: @westonpace Thanks for your detailed explanation. I remove the `output `type in `ExpandRel`. Please help to review again.

baibaichen: > @westonpace Thanks for your detailed explanation. I remove the `output `type in `ExpandRel`. Please help to review again.

@JkSelf Than you

baibaichen:
@westonpace ok to merge?

westonpace (approved):
I'm happy with where this is.  I'll give @jacques-n a chance to comment further before merging.

proto/substrait/algebra.proto
```diff
@@ -340,6 +340,22 @@ message ExchangeRel {
...
+// The expand relation generates multiple rows per input row (denoted by
+// groupings). No ordering is applied to the results.
+message ExpandRel {
+  RelCommon common = 1;
+  Rel input = 2;
+
+  // A list of expression grouping that the aggregation measured should be calculated for.
+  repeated Grouping groupings = 3;
+
+  message Grouping {
```
jacques-n, line 352: A few thoughts:
* I didn't see whether there was an answer to my question on whether arbitrary expressions are allowed inside Acero and Spark. Are they?
* I don't think grouping is a good name for what we're doing. We are not grouping anything.
* I don't understand by the comments whether this operator maintains order or not. The comment about order is vague to me.
* I think the pseudo-code put forth about how we calculate output type and output fields is problematic and suggests that we have the wrong representation. It also works opposite of the way aggregate works which makes it less desireable.
* It isn't clear what the conclusion is wrt emit of existing columns. 

As I think more about this I think the grouping pattern just seems wrong given the requirements around same data types, etc. I almost think that the array of values should be transposed. For example:

```
// Duplicates records, possibly switching output expressions between each duplicate.
// Default output is all of the fields declared + one field that int64 type that returns the duplicate_id which is a zero-index ordinal of which duplicate of the original record this corresponds to
message ExpandRel {
  RelCommon common = 1;
  Rel input = 2;
  // tells how many times output records there are for each input record.
  int duplicate_count = 3; // Not strictly necessary
  repeated ExpandField fields = 4;

  message ExapandField {
    Type type = 1;
    oneof field_type {
      // Field that switches output based on which duplicate_id we're outputting
      SwitchingField switching_field = 2;

      // Field that outputs the same value no matter which duplicate id we're on.
      Expression consistent_field = 3;
    } 
  }

  message SwitchingField {
    // array that contains the expression to output per expansion_id
    // must have the same number of expressions as each input record
    // all expressions must be of the same type
    repeated Expression duplicates = 1;
  }

}


```

I think this does a much clearer job of expressing what we're supporting. We're duplicating records. In some fields, each duplicate gets a different expression. In some cases, all duplicates get the same record.

Thoughts?

westonpace: > I didn't see whether there was an answer to my question on whether arbitrary expressions are allowed inside Acero and Spark. Are they?

For Acero, the answer is yes.  I don't know for Spark.  Given all the use cases I am aware of we could probably restrict this to "field refs or literals" but I don't know if that would be useful or not.

> I don't think grouping is a good name for what we're doing. We are not grouping anything.

If it helps, in Acero we use "templates".

> I don't understand by the comments whether this operator maintains order or not. The comment about order is vague to me.

There are a few different styles you can use here.  E.g. given ROW1, ROW2, ROW3 and two duplicates you could have `ROW1D1 ROW1D2 ROW2D1 ROW2D2 ROW3D1 ROW3D2` or you could have `ROW1D1 ROW2D1 ROW3D1 ROW1D2 ROW2D2 ROW3D2`.  Dplyr, for example, in pivot_longer, allows you to specify one or the other.  Acero however, does not actually support either of those for performance reasons (we implement this as a zero-copy operation).

So I interpreted the comment about order as saying that we are not guaranteeing any specific ordering of rows.

> As I think more about this I think the grouping pattern just seems wrong given the requirements around same data types, etc. I almost think that the array of values should be transposed

Your proposal (with `ExpandField`) looks good to me.  I can see how it will simplify the calculation of the output type.  I'd rather not include `duplicate_count` since it should be redundant (each `SwitchingField` should have the same number of `duplicates`).

westonpace: To clarify, I'd rather not include `duplicate_count` but I'd be fine with leaving it in.

jacques-n: > Arbitrary expressions allowed in acero

Ok, I figured as much which lead me to propose this with expressions as opposed to subset of items.

>  I'd rather not include duplicate_count 

Yeah, I actually commented it and then uncommmented it. I'm generally also inclined to exclude. If you have duplicate data then you have to specify the behavior when it is inconsistent :)

@baibaichen and @JkSelf , what are your thoughts on my inverted representation?

baibaichen: @jacques-n 
I don't understand how `ExpandField` works, it looks complicated to me. I am not clear about your concern, why we need duplcate_id? Back to `Expand`'s semantic：**Apply a number of projects to every input row,  hence we will get mulitple ouput rows for an input row**.  

> I don't think grouping is a good name for what we're doing. We are not grouping anything.

Yes, I prefer to `projections`,  I think grouping us confusing us. 

For other comments:

> I didn't see whether there was an answer to my question on whether arbitrary expressions are allowed inside Acero and Spark. Are they?

Yes, arbitrary expressions.


> I don't understand by the comments whether this operator maintains order or not. The comment about order is vague to me.

Spark uses expression id to refer input's(child plan）Field, while Calcite uses index. So When substrait plan is convertted back to Spark plan,  The field order is very important.

For `ProjectRel`, we **assume** output order is same as its project expression, which is right for Spark `Project`. But, I don't know whether this assumption is always ture for `Expand` or not.

> I think the pseudo-code put forth about how we calculate output type and output fields is problematic and suggests that we have the wrong representation. It also works opposite of the way aggregate works which makes it less desireable.

I think it's OK for now. It's better to add `output`, but we can add it latter once assumption is broken.

> It isn't clear what the conclusion is wrt emit of existing columns.

I don't understand.

JkSelf: Here the `ExpandField ` maybe not suite for spark. Because we only get the projection list in [ExpandExec ](https://github.com/apache/spark/blob/cb7d08280623794b238297d2f3de8abdc8b72bdb/sql/core/src/main/scala/org/apache/spark/sql/execution/ExpandExec.scala#L37)and it is hard to distinguish the consistent_field and switching_field.

jacques-n: > Here the `ExpandField ` maybe not suite for spark. Because we only get the projection list in [ExpandExec ](https://github.com/apache/spark/blob/cb7d08280623794b238297d2f3de8abdc8b72bdb/sql/core/src/main/scala/org/apache/spark/sql/execution/ExpandExec.scala#L37)and it is hard to distinguish the consistent_field and switching_field.

I think you're saying that it is a two dimensional array. In those cases, you'd create all items switching fields. I don't see a problem.

> I don't understand how ExpandField works, it looks complicated to me

What is unclear or complex? 

I believe this is actually a simplification that more accurately represents the semantics that you desire. In expand: you're producing N output rows for each input row. All output rows have M columns. For each column in M, you do one of two things when outputting a column/field: 
1. duplicate the value same expression N times (consistent_field)
2. use a different value for each of the N rows (switching_field)
 
ExpandField clearly distinguishes between these two behaviors.

This representation also makes a clear hierarchical relationship such that it is clear that all items within a list of switching field must have the same data type.

baibaichen: >I think you're saying that it is a two dimensional array. In those cases, you'd create all items as switching fields. I don't see a problem with mapping to/from this representation. Additionally, if you received something that had consistent fields and wanted to convert to Spark representation, it would be a simple matter of duplication the expression for the consistent field across all expansion items.

Yes,  as I talked to @JkSelf offline. we will create all items as switching fields. If you don't think it's a problem, then do it this way.

JkSelf: > > I think you're saying that it is a two dimensional array. In those cases, you'd create all items as switching fields. I don't see a problem with mapping to/from this representation. Additionally, if you received something that had consistent fields and wanted to convert to Spark representation, it would be a simple matter of duplication the expression for the consistent field across all expansion items.
> 
> Yes, as I talked to @JkSelf offline. we will create all items as switching fields. If you don't think it's a problem, then do it this way.
@jacques-n Updated the ExpandRel. Do you have any further suggestion? Thanks.

proto/substrait/algebra.proto
```diff
@@ -340,6 +340,33 @@ message ExchangeRel {
   }
 }
 
+// Duplicates records, possibly switching output expressions between each duplicate.
+// Default output is all of the fields declared + one field that int64 type that returns the
```
westonpace, line 344:
```suggestion
// Default output is all of the fields declared followed by one int64 field that contains the
```

proto/substrait/algebra.proto
```diff
@@ -340,6 +340,33 @@ message ExchangeRel {
...
+      // Field that switches output based on which duplicate_id we're outputting
+      SwitchingField switching_field = 2;
+
+      // Field that outputs the same value no matter which duplicate_id we're on.
+      Expression consistent_field = 3;
+    }
+  }
+
+  message SwitchingField {
+    // Array that contains the expression to output per duplicate_id
```
westonpace, line 363:
```suggestion
    // Array that contains an expression to output per duplicate_id
```

proto/substrait/algebra.proto
```diff
@@ -340,6 +340,33 @@ message ExchangeRel {
...
+    // must have the same number of expressions as each input record
```
westonpace, line 364:
```suggestion
    // each `switching_field` must have the same number of expressions
```

proto/substrait/algebra.proto
```diff
@@ -340,6 +340,33 @@ message ExchangeRel {
...
+    // all expressions must be of the same type a
```
westonpace, line 365:
```suggestion
    // all expressions within a switching field be the same type class but can differ in nullability.
    // this column will be nullable if any of the expressions are nullable.
```

This one is debatable.  If we want them to be exactly the same then the expressions can be something like:

```
cast(field_ref("country"), string(nullable: true))
literal(null, type: string(nullable: true))
```

However, I think it is easier to allow each expression to have different nullability.  Then we can have:

```
field_ref("country"), string(nullable: true)
literal(null, type: string(nullable: true))
```

I'm fine either way.

site/docs/relations/physical_relations.md
```diff
@@ -206,6 +206,28 @@ The streaming aggregate operation leverages data ordered by the grouping express
...
+
+### Expand Properties
+
+| Property  | Description                                                                             | Required |
+| --------- | --------------------------------------------------------------------------------------- | -------- |
+| Input     | The relational input.                                                                   | Required |
+| Groupings | Sets of expressions.  There will be one output row for each input row for each grouping | Required |
+
+Note: Each grouping must have the same number of expressions and the return types for each new output column must
+be consistent across all groupings.
```
westonpace, line 230: This needs to be updated.

westonpace (changes requested):
I'm content with this approach but I have some wording suggestions.

jacques-n:
Same comments as @westonpace. Impl looks good, docs need updating as he identified. I'm fine with the null behavior Weston suggested.

site/docs/relations/physical_relations.md
JkSelf: @westonpace Can you help to review again? Thanks.

baibaichen: @westonpace ok to merge

site/docs/relations/physical_relations.md
```diff
@@ -206,6 +206,25 @@ The streaming aggregate operation leverages data ordered by the grouping express
...
+| Outputs              | 1                                                                                   |
+| Property Maintenance | Distribution and ordering are not maintained.                                       |
+| Direct Output Order  | Same as defined by the [Project](logical_relations.md#project-operation) operation. |
+
+### Expand Properties
+
+| Property  | Description                          | Required |
+| --------- |--------------------------------------| -------- |
+| Input     | The relational input.                | Required |
+| ExpandField | One of SwitchingField of Expression. | Required |
```
westonpace, line 227:
```suggestion
| Fields | Expressions describing the output fields.  These refer to the schema of the input.  A field can have a different expression for each duplicate.  | Required |
```

westonpace (approved):
Last small change and then I approve and it sounds like we can merge.

site/docs/relations/physical_relations.md
```diff
@@ -206,6 +206,25 @@ The streaming aggregate operation leverages data ordered by the grouping express
...
+The expand operation creates duplicates of input records based on ExpandFields. For each ExpandField the
+expressions specific to that grouping are evaluated and the result is added to the input record in a manner
+similar to a project operation. 
+
+| Signature            | Value                                                                               |
+| -------------------- | ----------------------------------------------------------------------------------- |
+| Inputs               | 1                                                                                   |
+| Outputs              | 1                                                                                   |
+| Property Maintenance | Distribution and ordering are not maintained.                                       |
+| Direct Output Order  | Same as defined by the [Project](logical_relations.md#project-operation) operation. |
```
jacques-n, line 220: This is not correct with the latest representation. The only output is the fields in the expand plus the duplicate id. We should clearly state the order of these.

site/docs/relations/physical_relations.md
```diff
@@ -206,6 +206,25 @@ The streaming aggregate operation leverages data ordered by the grouping express
 | Measures         | A list of one or more aggregate expressions. Aggregate expressions ordering requirements must be compatible with expected ordering. | Optional, required if no grouping sets. |
 
 
+## Expand Operation
+
+The expand operation creates duplicates of input records based on ExpandFields. For each ExpandField the
+expressions specific to that grouping are evaluated and the result is added to the input record in a manner
```
jacques-n, line 212: grouping doesn't seem like the right word here. 

Additionally, as part of these docs we should be talking about the two behaviors "switching" versus "duplicating" (or whatever).

site/docs/relations/physical_relations.md
```diff
@@ -206,6 +206,25 @@ The streaming aggregate operation leverages data ordered by the grouping express
...
+similar to a project operation. 
+
+| Signature            | Value                                                                               |
+| -------------------- | ----------------------------------------------------------------------------------- |
+| Inputs               | 1                                                                                   |
+| Outputs              | 1                                                                                   |
+| Property Maintenance | Distribution and ordering are not maintained.                                       |
```
jacques-n, line 219: I would expect distribution to be maintained for columns that are duplicated (not switching)

JkSelf:
@[jacques-n](https://github.com/jacques-n) Thanks for your review. And I have updated the docs. Please help to review again. Thanks.

site/docs/relations/physical_relations.md
```diff
@@ -206,6 +206,23 @@ The streaming aggregate operation leverages data ordered by the grouping express
...
 
+## Expand Operation
+
+The expand operation creates duplicates of input records based on ExpandFields. The ExpandField can be the SwitchingField or ConsistentField. The SwitchingField here means that switches output based on which duplicate_id we're outputting. And the ConsistentField means that outputs the same value no matter which duplicate_id we're on.
+
+| Signature            | Value                                |
+| -------------------- |--------------------------------------|
+| Inputs               | 1                                    |
+| Outputs              | 1                                    |
+| Property Maintenance | Duplicated columns are maintained.   |
```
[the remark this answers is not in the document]
jacques-n: > is this description accurate?

Distribution is only maintained if all the distribution fields are consistent fields with direct references. For example, `distributed by a,b` does not imply `distributed by a`. As such, if `a` was a consistent field but `b` was not, distribution wouldn't be maintained.

Ordering can only be maintained down to the level of consistent fields that are kept. So if the original set is `ordered by a,b,c` and then we have consistent fields `a,b` and `switching field(c)`, only ordering of `a,b` would be maintained.

baibaichen:
@jacques-n would you please directly update documentation? I didn't catch your meaning.

site/docs/relations/physical_relations.md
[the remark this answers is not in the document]
JkSelf: @jacques-n @westonpace Update the doc. And please help to review again. Thanks.

westonpace (approved):
I think this one is ready to go too.  @jacques-n any last concerns?

JkSelf:
@jacques-n  Can you help to approve if no further comment?

baibaichen:
@jacques-n do you have any comment?

JkSelf:
@jacques-n Can you help to review again if you have any comment? Thanks for your help.

JkSelf:
@westonpace @jacques-n Can you help to review again if you have any comment? If not, please help to merge. Thanks for your help.

proto/substrait/algebra.proto
```diff
@@ -340,6 +340,34 @@ message ExchangeRel {
...
+
+      // Field that outputs the same value no matter which duplicate_id we're on.
+      Expression consistent_field = 3;
+    }
+  }
+
+  message SwitchingField {
+    // Array that contains an expression to output per duplicate_id
+    // each `switching_field` must have the same number of expressions
+    // all expressions within a switching field be the same type class but can differ in nullability.
```
westonpace, line 365:
```suggestion
    // all expressions within a switching field must be the same type class but can differ in nullability.
```

site/docs/relations/physical_relations.md
```diff
@@ -206,6 +206,23 @@ The streaming aggregate operation leverages data ordered by the grouping express
 | Measures         | A list of one or more aggregate expressions. Aggregate expressions ordering requirements must be compatible with expected ordering. | Optional, required if no grouping sets. |
 
 
+## Expand Operation
+
+The expand operation creates duplicates of input records based on ExpandFields. The ExpandField can be the SwitchingField or ConsistentField. The SwitchingField here means that switches output based on which duplicate_id we're outputting. And the ConsistentField means that outputs the same value no matter which duplicate_id we're on.
```
westonpace, line 211:
```suggestion
The expand operation creates duplicates of input records based on the Expand Fields. Each Expand Field can be a Switching Field or an expression. Switching Fields are described below.  If an Expand Field is an expression then its value is consistent across all duplicate rows.
```

site/docs/relations/physical_relations.md
```diff
@@ -206,6 +206,23 @@ The streaming aggregate operation leverages data ordered by the grouping express
...
+| -------------------- |--------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
+| Inputs               | 1                                                                                                                                                                              |
+| Outputs              | 1                                                                                                                                                                              |
+| Property Maintenance | Distribution is maintained if all the distribution fields are consistent fields with direct references. Ordering can only be maintained down to the level of consistent fields that are kept.|
+| Direct Output Order  | The expand fields and duplicated id.                                                                                                                                           |
```
westonpace, line 218:
```suggestion
| Direct Output Order  | The expand fields followed by an i32 column describing the index of the duplicate that the row is derived from.                                                                                                                                           |
```

site/docs/relations/physical_relations.md
```diff
@@ -206,6 +206,23 @@ The streaming aggregate operation leverages data ordered by the grouping express
...
+
+### Expand Properties
+
+| Property  | Description                          | Required |
+| --------- |--------------------------------------| -------- |
+| Input     | The relational input.                | Required |
+| Fields | Expressions describing the output fields.  These refer to the schema of the input.  A field can have a different expression for each duplicate.  | Required |
```
westonpace, line 225:
```suggestion
| Direct Fields | Expressions describing the output fields.  These refer to the schema of the input.  Each Direct Field must be a Consistent Field or a Switching Field  | Required |

### Switching Field Properties

A switching field is a field whose value is different in each duplicated row.  All switching fields in an Expand Operation must have the same number of duplicates.

| Property  | Description                          | Required |
| --------- |--------------------------------------| -------- |
| Duplicates | List of one or more expressions.  The output will contain a row for each expression. | Required |
```

westonpace (changes requested):
This was discussed in today's community meeting.  It was agreed that the protobuf is correct but the wording in the specification does not accurately reflect the protobuf.  I have made some suggestions on how we can correct that.

westonpace:
Also, this PR will need a rebase.

JkSelf:
@westonpace Thanks for your review. I have updated and can you help to review again?

site/docs/relations/physical_relations.md
```diff
@@ -206,6 +206,31 @@ The streaming aggregate operation leverages data ordered by the grouping express
...
+| Outputs              | 1                                                                                                                                                                              |
+| Property Maintenance | Distribution is maintained if all the distribution fields are consistent fields with direct references. Ordering can only be maintained down to the level of consistent fields that are kept.|
+| Direct Output Order  | The expand fields followed by an i32 column describing the index of the duplicate that the row is derived from.                                                                                                                                           |
+
+### Expand Properties
+
+| Property  | Description                          | Required |
+| --------- |--------------------------------------| -------- |
+| Input     | The relational input.                | Required |
+| Direct Fields | Expressions describing the output fields.  These refer to the schema of the input.  Each Direct Field must be a Consistent Field or a Switching Field  | Required |
```
westonpace, line 225:
```suggestion
| Direct Fields | Expressions describing the output fields.  These refer to the schema of the input.  Each Direct Field must be an expression or a Switching Field  | Required |
```

westonpace (approved):
One minor suggestion.  @jacques-n can you take another look at this one now? I think the markdown / spec is consistent with the proto files now.

jacques-n (approved):
looks good to me.

proto/substrait/algebra.proto
```diff
@@ -342,6 +342,34 @@ message ExchangeRel {
...
+      Expression consistent_field = 3;
+    }
+  }
+
+  message SwitchingField {
+    // Array that contains an expression to output per duplicate_id
+    // each `switching_field` must have the same number of expressions
+    // all expressions within a switching field must be the same type class but can differ in nullability.
+    // this column will be nullable if any of the expressions are nullable.
+    repeated Expression duplicates = 1;
```
EpsilonPrime, line 369: Is there any value in having a repeated inside a repeated?  Seems like this should just be a single item (to be consistent with consistent_field above).

westonpace: The outer repeated (`fields`) is columns and the inner repeated (`duplicates`) is rows.

So if our goal is to take input:

| X | Y | Z |
| --- | --- | --- |
| 1 | 2 | 3 |

and generate:

| X | Y | Z |
| --- | --- | --- |
| 1 | 2 | 3 |
| 1 | NULL | 3 |
| NULL | NULL | 3 |

Then we need something like...

```
fields: [
  { duplicates: [ field(x), field(x), NULL ] },
  { duplicates: [ field(y), NULL, NULL ] }
]
```

EpsilonPrime: Thanks for the clarification.  I've sent #546 to expand the documentation.

PR #388, opened 2022-11-25
feat: add skip_nulls option to scalar aggregate functions

thisisnic:
This PR adds a `skip_nulls` option to each of the scalar aggregate functions, with a default value of `TRUE` (which would just replicate existing behaviour) for all of these functions except for `quantile()` which had the previous behaviour of returning `NULL` if any input values are `NULL`.

Although some backends remove NULL values by default in computations, others allow specifying this option to return an NULL value if any of the input values are NULL.

Now I've done this, I'm wondering if this is something relevant to the `NULLABILITY HANDLING` field, but if I'm completely honest I still don't understand it after reading the docs (that said, I'm happy to submit another PR updating the docs for that section if someone reviewing this explains it to me!)  I don't *think* that's relevant here, though I'm not sure.

### DIFF
--- a/extensions/functions_arithmetic.yaml
+++ b/extensions/functions_arithmetic.yaml
@@ -1404,7 +1404,7 @@ aggregate_functions:
 
       The function fails and returns null in the following cases:
         - `n` is null or less than one;
-        - any value in `distribution` is null.
+        - option `skip_nulls` is `FALSE` (the default) and any value in `distribution` is null.
 
       The function returns an empty list if `n` equals 1 and `boundaries` is
       set to `NEITHER`.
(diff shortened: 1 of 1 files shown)


extensions/functions_arithmetic.yaml
```diff
@@ -756,6 +756,8 @@ aggregate_functions:
         options:
           overflow:
             values: [ SILENT, SATURATE, ERROR ]
+          skip_nulls:
```
jacques-n, line 759: I'm wondering if we should add this property to the aggregate function invocation as opposed to every aggregate function. It feels kind of like aggregate input sorting or some other function-impl independent concept. Thoughts @westonpace @gforsyth ?

gforsyth: I agree that if we're specifying an option for every function definition that we should move it up to the invocation

thisisnic: That makes sense, will close this PR and have opened https://github.com/substrait-io/substrait/issues/401 instead.

PR #399, opened 2022-12-01
feat: add windowrel support  in proto

JkSelf:


JkSelf:
@jacques-n Please help to review if you have available time. Thanks for your help!

JkSelf:
cc @FelixYBW @baibaichen

proto/substrait/algebra.proto
```diff
@@ -833,6 +853,12 @@ message Expression {
     }
   }
 
+  enum WindowType {
+    WINDOW_TYPE_UNSPECIFIED = 0;
+    WINDOW_TYPE_ROWS = 1;
+    WINDOW_TYPE_RANGE = 2;
+  }
+
```
PHILO-HE, line 861: Just a small suggestion. I think using `window frame` is more proper and more clear to user than using `window type`. Right?
```
  enum WindowFrame {
    WINDOW_FRAME_UNSPECIFIED = 0;
    WINDOW_FRAME_ROWS = 1;
    WINDOW_FRAME_RANGE = 2;
  }
```

JkSelf: Thanks for you suggestion. And very sorry that I did not reply in time. I have already changed to WindowFrame.

JkSelf:
@jacques-n Can you help to review again?

proto/substrait/algebra.proto
```diff
@@ -228,6 +228,19 @@ message AggregateRel {
   }
 }
 
+message WindowRel {
```
EpsilonPrime, line 231: Probably useful to add comments for the various parts.  I'm especially curious how this differs from the FilterRel so that might be useful information to provide here.

proto/substrait/algebra.proto
```diff
@@ -891,6 +911,12 @@ message Expression {
     }
   }
 
+  enum WindowFrame {
+    WINDOW_FRAME_UNSPECIFIED = 0;
+    WINDOW_FRAME_ROWS = 1;
+    WINDOW_FRAME_RANGE = 2;
+  }
```
westonpace, line 918: How does this define a window frame?  How many rows?  How big is the range?

For example, given the query:

`SELECT max(col) OVER (ROWS BETWEEN 1 PRECEDING AND 1 FOLLOWING) FROM my_table`

It's not clear where the `1 PRECEDING` and `1 FOLLOWING` would go.

Also, I don't see any support for `CURRENT ROW` or `UNBOUNDED`.

proto/substrait/algebra.proto
```diff
@@ -891,6 +911,12 @@ message Expression {
     }
   }
 
+  enum WindowFrame {
+    WINDOW_FRAME_UNSPECIFIED = 0;
+    WINDOW_FRAME_ROWS = 1;
+    WINDOW_FRAME_RANGE = 2;
```
westonpace, line 917: Maybe add `WINDOW_FRAME_GROUPS` for peer groups?

proto/substrait/algebra.proto
```diff
@@ -228,6 +228,20 @@ message AggregateRel {
   }
 }
 
+// The Window relational operate on a group of rows, referred to as a window, and calculate a return value for each row based on the group of rows.
+message WindowRel {
+  RelCommon common = 1;
+  Rel input = 2;
+  repeated Measure measures = 3;
+  repeated Expression partition_expressions = 4;
+  repeated SortField sorts = 5;
```
westonpace, line 237: Should a window relation specify sorts?  Or should it just use whatever orderedness is currently established?  For example, the `FetchRel` does not have "sorts".

proto/substrait/algebra.proto
```diff
@@ -843,6 +857,12 @@ message Expression {
     // Optional; defaults to the end of the partition.
     Bound upper_bound = 4;
 
+    // Defines the window frame: ROWS, RANGE
+    WindowFrame window_frame = 12;
+
+    // Defines window column name for current window function.
+    string window_column_name = 13;
```
westonpace, line 864: What is this?

westonpace (changes requested):
I have a few questions but this seems like a good start.

proto/substrait/algebra.proto
JkSelf: The [Bound ](https://github.com/substrait-io/substrait/blob/9c350c953c4901aba4525fa6f198c98cb8fc4590/proto/substrait/algebra.proto#L850) already define the specification about which row to start  and where to end it in the Window.

proto/substrait/algebra.proto
JkSelf: The definition of `WindowFrame ` refers to the definition of Window in [Spark ](https://spark.apache.org/docs/latest/sql-ref-syntax-qry-select-window.html). We can also add additional options if necessary.

proto/substrait/algebra.proto
JkSelf: Window relation need to control the way that rows in a partition are ordered, determining the position of the given row in its partition. Refers to the definition of Window in [Spark ](https://spark.apache.org/docs/latest/sql-ref-syntax-qry-select-window.html)and [VELOX](https://facebookincubator.github.io/velox/develop/operators.html#windownode)

proto/substrait/algebra.proto
JkSelf: This is only to pass the column name with window operator. And we can also store it in AdvancedExtension node.

proto/substrait/algebra.proto
westonpace: Ok, I was looking at postgres for reference.  Since this would be an additive change (should be backwards compatible to add a new WINDOW_FRAME_...) I think we can leave this as it is.  We can add a new option later when we have a producer or consumer that supports it.

proto/substrait/algebra.proto
westonpace: Yes, I agree that a window relation needs an established ordering.  I don't know that the window relation itself has to establish the ordering.

A `FetchRel` also relies on an established ordering.  However, it does not establish the ordering itself.

For example, if you wanted to do a fetch to limit the rows to 10Mi and then run a window operation on that you should be able to do:

```
OrderBy -> Fetch -> Window (no sorting)
```

Curious for @jacques-n or @cpcloud 's input on this as they have more experience with multiple implementations.

However, I will concede that the two established systems you mentioned (spark and velox) would not support that and so maybe we should do something more traditional here.

proto/substrait/algebra.proto
westonpace: Substrait doesn't normally name columns in the middle of a plan so I think we can remove this.

proto/substrait/algebra.proto
westonpace: I missed that.  Thank you for pointing it out.

proto/substrait/algebra.proto
westonpace: Also, how is this related to the `sorts` property on `WindowFunction`?

https://github.com/substrait-io/substrait/blob/main/proto/substrait/algebra.proto#L819

JkSelf: It may be more reasonable to put the `sorts `property in WindowRel, because the sort requires all window functions to be sorted according to the specified sort order when processing. And we may can remove this sorts property in WindowFunction.

proto/substrait/algebra.proto
JkSelf: Yes. I already removed it.

proto/substrait/algebra.proto
westonpace: SQL allows for each window function to have their own sort:

```
SELECT RANK() OVER (ORDER BY val ASC), RANK() OVER (ORDER BY val DESC) FROM my_table;
```

Would you envision this as two different `WindowRel`s?

cpcloud: What is the motivation for a separate window relation? When would I use a `WindowRel` versus a projection with `WindownFunction` expressions?

westonpace: Physically I think there is some advantage in having a a window rel which has one or more window functions that all share the same sorting & partitioning.  I'm not sure how best to mark this as a physical node instead of a logical node (it's also not how I read this originally).  E.g. this matches Velox's interpretation: https://github.com/facebookincubator/velox/blob/main/velox/core/PlanNode.h#L1687-L1689

> /// PlanNode used for evaluating Sql window functions.
> /// All window functions evaluated in the operator have the same
> /// window spec (partition keys + order columns).

It also seems to match Spark's interpretation.

I agree, however, that there is no need for this in a purely logical plan.

cpcloud: Then I think I'm not understanding how the logical plan of

```
SELECT
  SUM(x) OVER (PARTITION BY a),
  AVG(y) OVER (PARTITION BY b)
FROM t
```

turns into a physical plan containing `WindowRel`s.

westonpace: Since the partition is different there would be two `WindowRel` nodes.  They could run in parallel or sequentially.  Ideally an engine could find some way to do something to avoid double-accumulating all the data.  Perhaps an order by `a,b` for example.

However, naively, it seems spark SQL gives us:

```
AdaptiveSparkPlan isFinalPlan=false
+- Project [sum(x) OVER (PARTITION BY a ROWS BETWEEN UNBOUNDED PRECEDING AND UNBOUNDED FOLLOWING)#63L, sum(y) OVER (PARTITION BY b ROWS BETWEEN UNBOUNDED PRECEDING AND UNBOUNDED FOLLOWING)#64L]
   +- Window [sum(y#43) windowspecdefinition(b#45, specifiedwindowframe(RowFrame, unboundedpreceding$(), unboundedfollowing$())) AS sum(y) OVER (PARTITION BY b ROWS BETWEEN UNBOUNDED PRECEDING AND UNBOUNDED FOLLOWING)#64L], [b#45]
      +- Sort [b#45 ASC NULLS FIRST], false, 0
         +- Exchange hashpartitioning(b#45, 200), ENSURE_REQUIREMENTS, [plan_id=42]
            +- Project [y#43, b#45, sum(x) OVER (PARTITION BY a ROWS BETWEEN UNBOUNDED PRECEDING AND UNBOUNDED FOLLOWING)#63L]
               +- Window [sum(x#42) windowspecdefinition(a#44, specifiedwindowframe(RowFrame, unboundedpreceding$(), unboundedfollowing$())) AS sum(x) OVER (PARTITION BY a ROWS BETWEEN UNBOUNDED PRECEDING AND UNBOUNDED FOLLOWING)#63L], [a#44]
                  +- Sort [a#44 ASC NULLS FIRST], false, 0
                     +- Exchange hashpartitioning(a#44, 200), ENSURE_REQUIREMENTS, [plan_id=37]
                        +- Project [_1#33 AS x#42, _3#35 AS a#44, _2#34 AS y#43, _4#36 AS b#45]
                           +- SerializeFromObject [knownnotnull(assertnotnull(input[0, scala.Tuple4, true]))._1 AS _1#33, knownnotnull(assertnotnull(input[0, scala.Tuple4, true]))._2 AS _2#34, staticinvoke(class org.apache.spark.unsafe.types.UTF8String, StringType, fromString, knownnotnull(assertnotnull(input[0, scala.Tuple4, true]))._3, true, false, true) AS _3#35, staticinvoke(class org.apache.spark.unsafe.types.UTF8String, StringType, fromString, knownnotnull(assertnotnull(input[0, scala.Tuple4, true]))._4, true, false, true) AS _4#36]
                              +- Scan[obj#32]
```

Stripping away the noise:

```
AdaptiveSparkPlan isFinalPlan=false
+- Project [...]
   +- Window [...],
      +- Sort [...],
         +- Exchange,
            +- Project [...],
               +- Window [...],
                  +- Sort [...],
                     +- Exchange,
                        +- Project [...],
                           +- SerializeFromObject [...]
                              +- Scan[..]
```

westonpace: SQL server does something considerably more complicated:

image

cpcloud: I see, so it's a big chain of projections in the spark case. Thanks, this makes sense to me now.

westonpace: And for completeness, DuckDb:

```
┌───────────────────────────┐
│         PROJECTION        │
│   ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─   │
│   sum(read_csv.x) OVER    │
│ (PARTITION BY read_csv.a) │
│   avg(read_csv.y) OVER    │
│ (PARTITION BY read_csv.b) │
└─────────────┬─────────────┘                             
┌─────────────┴─────────────┐
│         PROJECTION        │
│   ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─   │
│             #0            │
│             #1            │
│             #2            │
│             #3            │
│             #4            │
│             #5            │
└─────────────┬─────────────┘                             
┌─────────────┴─────────────┐
│           WINDOW          │
│   ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─   │
│avg(y) OVER (PARTITION BY b│
│             )             │
└─────────────┬─────────────┘                             
┌─────────────┴─────────────┐
│           WINDOW          │
│   ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─   │
│sum(x) OVER (PARTITION BY a│
│             )             │
└─────────────┬─────────────┘                             
┌─────────────┴─────────────┐
│         PROJECTION        │
│   ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─   │
│             #0            │
│             #2            │
│             #1            │
│             #3            │
└─────────────┬─────────────┘                             
┌─────────────┴─────────────┐
│          READ_CSV         │
│   ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─   │
│           EC: 42          │
└───────────────────────────┘                             

```

JkSelf: > SQL allows for each window function to have their own sort:
> 
> ```
> SELECT RANK() OVER (ORDER BY val ASC), RANK() OVER (ORDER BY val DESC) FROM my_table;
> ```
> 
> Would you envision this as two different `WindowRel`s?

Yes. Spark will create two different Window operator with different sort spec.

image

JkSelf: @westonpace @cpcloud Do you have any further comment about Window Rel?

JkSelf:
@westonpace @cpcloud Can you help to review again? Thanks for your help!

westonpace:
@JkSelf thanks for the update.  Based on previous discussion I think this is a physical counterpart to WindowFunction.  In other words:

A logical plan will only need `WindowFunction`
A physical (or less logical plan) may instead use `WindowRel`.

If you agree I think we should make a few changes:

 * `WindowRel` should not use `WindowFunction`.  Instead, copy the fields that you need (I think it will be similar to `ScalarFunction`) into a separate message that is local to `WindowRel`.
   * For example, `WindowRel::Function` will not have `sorts`.  This way there is no confusion which instance of `sorts` should be used.
 * Similar to above, but should `lower_bound`/`upper_bound` be added to `WindowRel`?
 * We should update `physical_relations.md` to reflect this new operator.  Currently, there is "Hashing Window Operator" and "Streaming Window Operator".  Does this replace one of these?  Or is it a third kind of node?  Or does "Hashing" and "Streaming" not make sense and we should combine them into this?
 * Does this operator handle "scalar window functions" (e.g. `SELECT RANK() OVER (ORDER BY time)`) or "window aggregates" (e.g. `SELECT SUM(val) OVER (PARTITION BY month)`) or both?  I think it is best if it only does one or the other.  That way the number of output rows is clear.

JkSelf:
@westonpace Thanks for your reply.
> WindowRel should not use WindowFunction. Instead, copy the fields that you need (I think it will be similar to ScalarFunction) into a separate message that is local to WindowRel. For example, WindowRel::Function will not have sorts. This way there is no confusion which instance of sorts should be used.

Yes. We can define a `WindowRel::Function` with not `sorts`. But this will bring ambiguity in understanding, because `WindowRel `is originally designed to perform a series of Window functions on a specific partition. And it will also bring redundancy in the code, because both functions need to define Bound. Can we remove the sort in `WindowFunction` here? If not, I will add a new `Function `later.

> We should update physical_relations.md to reflect this new operator. Currently, there is "Hashing Window Operator" and "Streaming Window Operator". Does this replace one of these? Or is it a third kind of node? Or does "Hashing" and "Streaming" not make sense and we should combine them into this?

The difference between `Hashing `and `Steaming `is whether it is necessary to obtain the entire partition before execution described in [WindowType](https://github.com/substrait-io/substrait/blob/96b686128e57eb66c55ab04e4dcbae64c9250dca/site/docs/expressions/window_functions.md). For example, `SUM `does not need it, but `NTILE `does. The `WindowRel `we support here includes these two situations.

> Does this operator handle "scalar window functions" (e.g. SELECT RANK() OVER (ORDER BY time)) or "window aggregates" (e.g. SELECT SUM(val) OVER (PARTITION BY month)) or both? I think it is best if it only does one or the other. That way the number of output rows is clear.

If we use `WindowFunction ` without sort here, we need to support both here. Because the current WindowFunction supports both [here](https://github.com/substrait-io/substrait/blob/96b686128e57eb66c55ab04e4dcbae64c9250dca/site/docs/expressions/window_functions.md).

westonpace:
> Yes. We can define a WindowRel::Function with not sorts. But this will bring ambiguity in understanding, because WindowRel is originally designed to perform a series of Window functions on a specific partition. And it will also bring redundancy in the code, because both functions need to define Bound.

I think this redundancy is expected when we have logical and physical operators.  For example, the physical `HashJoinRel::left_keys` and `HashJoinRel::right_keys` are narrower, but redundant, when compared with the logical `JoinRel::expression`.

> Can we remove the sort in WindowFunction here?

No, because the logical representation needs to be able to stand alone.  It should be legal to have a "purely logical plan" that has `WindowFunction` but no `WindowRel` (the window functions would be part of a `ProjectRel`).

> The difference between Hashing and Steaming is whether it is necessary to obtain the entire partition before execution described in [WindowType](https://github.com/substrait-io/substrait/blob/96b686128e57eb66c55ab04e4dcbae64c9250dca/site/docs/expressions/window_functions.md). For example, SUM does not need it, but NTILE does.

Yes.  Also, streaming could make sense if the data was also sorted by the partition field.

> The WindowRel we support here includes these two situations.

Ok, then I think we need to add a new relation to `physical_relations.md` which is `WindowOperation`.  Perhaps:

```
## Window Operation

A window operation is a special type of project operation where every function
is a window function and all of the window functions share the same sorting and
partitioning.  This allows for the sort and partition to be calculated once and
shared between the various function evaluations.

| Signature | Value |
...

### Window Operation Properties

| Property | Description | Required |
...

## Hashing Window Operation
...
```

westonpace:
> If we use WindowFunction without sort [here](https://github.com/substrait-io/substrait/blob/96b686128e57eb66c55ab04e4dcbae64c9250dca/site/docs/expressions/window_functions.md), we need to support both here. Because the current WindowFunction supports both here.

Ok, I agree this should support both.  However, the output of an aggregate window function is scalar.  In other words:

Given `x=[1,2,3,4]` then `SUM(x) OVER ()` is `[10, 10, 10, 10]` and not `[10]`.

JkSelf:
> > If we use WindowFunction without sort [here](https://github.com/substrait-io/substrait/blob/96b686128e57eb66c55ab04e4dcbae64c9250dca/site/docs/expressions/window_functions.md), we need to support both here. Because the current WindowFunction supports both here.
> 
> Ok, I agree this should support both. However, the output of an aggregate window function is scalar. In other words:
> 
> Given `x=[1,2,3,4]` then `SUM(x) OVER ()` is `[10, 10, 10, 10]` and not `[10]`.

Yes. the `sum(x) over()` return `[10, 10, 10, 10]` and also the `rank() over()` return `[1, 2, 3, 4]`. They all return a single value for each row. So I think the output of rows is same both "scalar window functions" and  "window aggregates".

JkSelf:
@westonpace Thanks for your detailed explanation. I will define a new function without sort to replace the Window function. And also update the `physical_relations.md` based on your suggestions later.

baibaichen:
@westonpace is this ok to merge?

JkSelf:
@westonpace Can you help to review again? Thanks for your help.

proto/substrait/algebra.proto
```diff
@@ -891,6 +906,120 @@ message Expression {
     }
   }
 
+  // A window scalar function call.
+  message WindowScalarFunction {
```
westonpace, line 910: Instead of `WindowScalarFunction` maybe `WindowRelFunction`?  Also, can we put this message inside of `WindowRel`?

proto/substrait/algebra.proto
```diff
@@ -228,6 +228,20 @@ message AggregateRel {
...
+  RelCommon common = 1;
+  Rel input = 2;
+  repeated Measure measures = 3;
+  repeated Expression partition_expressions = 4;
+  repeated SortField sorts = 5;
+  substrait.extensions.AdvancedExtension advanced_extension = 10;
+
+  message Measure {
+    Expression.WindowScalarFunction measure = 1;
+  }
```
westonpace, line 242: I don't think a dedicated message type is needed here. Also, can we call this `expressions` instead of `measures` since this is more similar to "project" than it is to "aggregate"?

In other words.  Can we change:

```
repeated Measure measures = 3;
...
message Measure {
  Expression.WindowScalarFunction measure = 1;
}
```

to...

```
repeated Expression.WindowScalarFunction expressions = 3;
```

proto/substrait/algebra.proto
```diff
@@ -891,6 +906,120 @@ message Expression {
...
+    repeated FunctionOption options = 11;
+
+    // Must be set to the return type of the function, exactly as derived
+    // using the declaration in the extension.
+    Type output_type = 7;
+
+    // Describes which part of the window function to perform within the
+    // context of distributed algorithms. Required. Must be set to
+    // INITIAL_TO_RESULT for window functions that are not decomposable.
+    AggregationPhase phase = 6;
```
westonpace, line 942: Do we need this here?  Isn't distributed execution more a concern for higher level operators?

proto/substrait/algebra.proto
```diff
@@ -891,6 +906,120 @@ message Expression {
...
+    // Optional, defaults to AGGREGATION_INVOCATION_ALL.
+    AggregateFunction.AggregationInvocation invocation = 10;
+
+    // When one or more partition expressions are specified, two records are
+    // considered to be in the same partition if and only if these expressions
+    // yield an equal tuple of values for both. When computing the window
+    // function, only the subset of records within the bounds that are also in
+    // the same partition as the current record are aggregated.
+    repeated Expression partitions = 2;
+
```
westonpace, line 954:
```suggestion
```
partitions are already specified in the window rel

proto/substrait/algebra.proto
```diff
@@ -891,6 +906,120 @@ message Expression {
...
+    // Defines the record relative to the current record from which the window
+    // extends. The bound is inclusive. If the lower bound indexes a record
+    // greater than the upper bound, TODO (null range/no records passed?
+    // wrapping around as if lower/upper were swapped? error? null?).
+    // Optional; defaults to the start of the partition.
+    Bound lower_bound = 5;
```
westonpace, line 960: Can the bounds be different for different window functions in the same window rel?

proto/substrait/algebra.proto
```diff
@@ -891,6 +906,120 @@ message Expression {
...
+    // less than the lower bound, TODO (null range/no records passed?
+    // wrapping around as if lower/upper were swapped? error? null?).
+    // Optional; defaults to the end of the partition.
+    Bound upper_bound = 4;
+
+    // Defines the window frame: ROWS, RANGE
+    WindowFrame window_frame = 12;
+
+    // Defines one of the two boundaries for the window of a window function.
+    message Bound {
```
westonpace, line 973: Maybe we can reuse `WindowFunction.Bound` since it is identical?

westonpace (commented):
A few more questions

proto/substrait/algebra.proto
JkSelf: Done.

proto/substrait/algebra.proto
JkSelf: Yes. A window function corresponds to a bound.

proto/substrait/algebra.proto
JkSelf: Yes. It is not used in the operator of spark now. It is only reserved for future use.  And already removed.

proto/substrait/algebra.proto
JkSelf: Updated.

proto/substrait/algebra.proto
JkSelf: Updated.

JkSelf:
@westonpace Resolved based on your suggestions. Can you help to review again? Thanks.

JkSelf:
@westonpace  Do you have any further comment? Thanks.

JkSelf:
@westonpace Please help to review again. Thanks for your help.

proto/substrait/algebra.proto
```diff
@@ -228,6 +228,71 @@ message AggregateRel {
   }
 }
 
+// The Window relational operate on a group of rows, referred to as a window, and calculate a return value for each row based on the group of rows.
+message WindowRel {
+  RelCommon common = 1;
+  Rel input = 2;
+  repeated WindowRelFunction window_functions = 3;
```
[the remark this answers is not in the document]
westonpace: Both Spark & Velox have an operator which receives a single sort/partition and a collection of functions to apply given that ordering/grouping.

This is is a physical relation targeting those systems so it is ok if it is not general enough to cover all systems.

There is no logical equivalent of `WindowRel` because a logical plan will use `ProjectRel` with `WindowFunction`.  This will allow specifying partition/sort on each function.

So it sounds like Calcite could use the logical representation directly.  Though if we want to introduce a physical "window operator" (maybe calcite needs scalar and window functions to be separated?) then I would think that Calcite is welcome to do so.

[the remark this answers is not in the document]
JkSelf: Yes. It seems the Calcite can use the logical representation. What do your think? @jacques-n 
@westonpace  Do you have any further comment for this PR? Thanks.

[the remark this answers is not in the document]
JkSelf: @jacques-n @westonpace Can you help to review again? Thanks.

proto/substrait/algebra.proto
```diff
@@ -228,6 +228,71 @@ message AggregateRel {
...
+  repeated Expression partition_expressions = 4;
+  repeated SortField sorts = 5;
+
+  substrait.extensions.AdvancedExtension advanced_extension = 10;
+
+  // A window rel function call.
```
jacques-n, line 241: We need text here which clearly describes how this is different from a Expression.WindowFunction and why.

proto/substrait/algebra.proto
```diff
@@ -228,6 +228,71 @@ message AggregateRel {
...
+  message WindowRelFunction {
```
jacques-n, line 242: This function should be inside the WindowRel message. (missed feedback that Weston already gave).

proto/substrait/algebra.proto
```diff
@@ -228,6 +228,71 @@ message AggregateRel {
...
+
+    // Defines the record relative to the current record up to which the window
+    // extends. The bound is inclusive. If the upper bound indexes a record
+    // less than the lower bound, TODO (null range/no records passed?
+    // wrapping around as if lower/upper were swapped? error? null?).
+    // Optional; defaults to the end of the partition.
+    Expression.WindowFunction.Bound upper_bound = 4;
+
+    // Defines the window frame: ROWS, RANGE
+    WindowFrame window_frame = 12;
```
jacques-n, line 286: What is this? Why is this in the function? If it is in a function here, why wouldn't it need to be in the logical function as well? If I can guess at what it is (doing either a sort based op or a hash based op, I would expect it at the operator level since I the purpose of the operator is to do a single thing (single partitioning schema, single sorting scheme).

If I'm an implementer, I have no idea what this means or how to use it.

proto/substrait/algebra.proto
```diff
@@ -228,6 +228,71 @@ message AggregateRel {
   }
 }
 
+// The Window relational operate on a group of rows, referred to as a window, and calculate a return value for each row based on the group of rows.
+message WindowRel {
```
jacques-n, line 232: I don't think should be called WindowRel. How about ConsistentPartitionWindowRel or something else that clearly describes the intention.

proto/substrait/algebra.proto
```diff
@@ -228,6 +228,71 @@ message AggregateRel {
   }
 }
 
+// The Window relational operate on a group of rows, referred to as a window, and calculate a return value for each row based on the group of rows.
```
jacques-n, line 231: description doesn't effectively describe the purpose.

site/docs/relations/physical_relations.md
```diff
@@ -205,6 +205,23 @@ The streaming aggregate operation leverages data ordered by the grouping express
 | Per Grouping Set | A list of expression grouping that the aggregation measured should be calculated for. | Optional, defaults to 0.                |
 | Measures         | A list of one or more aggregate expressions. Aggregate expressions ordering requirements must be compatible with expected ordering. | Optional, required if no grouping sets. |
 
+## Window Operation
```
jacques-n, line 208: Should clearly distiguish itself (with references) to the other window operation (and be much more descriptive).

site/docs/relations/physical_relations.md
```diff
@@ -205,6 +205,23 @@ The streaming aggregate operation leverages data ordered by the grouping express
...
+### Window Properties
+
+| Property           | Description                     | Required               |
+| ------------------ | ------------------------------- | ---------------------- |
+| Input              | The relational input.           | Required               |
+| Window Expressions | One or more window expressions. | At least one required. |
+
 
 
 ## Hashing Window Operation
```
jacques-n, line 227: Is this replaced by the above? Seems like maybe it is.

proto/substrait/algebra.proto
```diff
@@ -228,6 +228,71 @@ message AggregateRel {
...
+  substrait.extensions.AdvancedExtension advanced_extension = 10;
+
+  // A window rel function call.
+  message WindowRelFunction {
+    // Points to a function_anchor defined in this plan, which must refer
+    // to a window function in the associated YAML file. Required; 0 is
+    // considered to be a valid anchor/reference.
+    uint32 function_reference = 1;
+
+    // The arguments to be bound to the function. This must have exactly the
```
jacques-n, line 248: Why have we removed the all/distinct property from this function?

proto/substrait/algebra.proto
```diff
@@ -228,6 +228,71 @@ message AggregateRel {
...
+    // number of arguments specified in the function definition, and the
```
jacques-n, line 249: I can't come up with a good reason why aggregation phase was removed. The reason for a distinct expression-like thing is to move partition and sort out of the object. Let's not extra differences that will confuse people unnecessarily. (one of the reasons that this kind of duplication is something I'd prefer to work harder to avoid).

proto/substrait/algebra.proto
JkSelf: The window_frame is used to define which row to start the window on and where to end it both in [spark ](https://spark.apache.org/docs/latest/sql-ref-syntax-qry-select-window.html)and [postgreSQL](https://www.postgresql.org/docs/current/sql-expressions.html#SYNTAX-WINDOW-FUNCTIONS).

proto/substrait/algebra.proto
JkSelf: It seems these properties are not used in Spark. And we can also remain here for the convenience of other systems if necessary.

proto/substrait/algebra.proto
JkSelf: I don't understand what you're saying about `all/distinct property` here.

site/docs/relations/physical_relations.md
JkSelf: Maybe not. It seems the `Window Operator` doesn't need to build hash table.

proto/substrait/algebra.proto
JkSelf: `ConsistentPartitionWindowRel ` is ok to me.

JkSelf:
@jacques-n Thanks for your review. I have resolved your comments. Please help to review again. Thanks.

proto/substrait/algebra.proto
westonpace: Perhaps call this `BoundsType` or something?  A "window frame" (in Spark) is the entire `{ RANGE | ROWS } { frame_start | BETWEEN frame_start AND frame_end }`.

proto/substrait/algebra.proto
```diff
@@ -228,6 +228,82 @@ message AggregateRel {
...
+    // Optional; defaults to the end of the partition.
+    Expression.WindowFunction.Bound upper_bound = 4;
+
+    // Defines the window frame: ROWS, RANGE
+    WindowFrame window_frame = 12;
+  }
+
+  enum WindowFrame {
+    WINDOW_FRAME_UNSPECIFIED = 0;
+    WINDOW_FRAME_ROWS = 1;
```
westonpace, line 302:
```suggestion
    // The lower and upper bound specify how many rows before and after the current row
    // the window should extend.
    WINDOW_FRAME_ROWS = 1;
```

proto/substrait/algebra.proto
```diff
@@ -228,6 +228,82 @@ message AggregateRel {
...
+    WINDOW_FRAME_RANGE = 2;
```
westonpace, line 303:
```suggestion
    // The lower and upper bound describe a range of values.  The window should include all rows
    // where the value of the ordering column is greater than or equal to (current_value - lower bound)
    // and less than or equal to (current_value + upper bound).  This bounds type is only valid if there
    // is a single ordering column.
    WINDOW_FRAME_RANGE = 2;
```

proto/substrait/algebra.proto
```diff
@@ -228,6 +228,82 @@ message AggregateRel {
...
+    // less than the lower bound, TODO (null range/no records passed?
+    // wrapping around as if lower/upper were swapped? error? null?).
+    // Optional; defaults to the end of the partition.
+    Expression.WindowFunction.Bound upper_bound = 4;
+
+    // Defines the window frame: ROWS, RANGE
+    WindowFrame window_frame = 12;
+  }
+
+  enum WindowFrame {
```
westonpace, line 300: This will need to be added to the logical `WindowFunction` as well.

proto/substrait/algebra.proto
westonpace: This is below at:

> AggregateFunction.AggregationInvocation invocation = 10;

I don't think it is missing.

proto/substrait/algebra.proto
westonpace: Please keep it.

site/docs/relations/physical_relations.md
westonpace: Why not?  Do implementations just sort everything by [ Partition Cols | Order Cols ] and then use a streaming approach?  I suppose this probably makes sense but I haven't implemented window functions so I don't know.

JkSelf: Yes. The newly added `Window Operator` will sort by the order columns and then use a streaming approach to calculate. So I think it is different with `Hashing Window Operator ` and `Streaming Window Operator`.

site/docs/relations/physical_relations.md
westonpace, line 208:
```suggestion
## Consistent Partition Window Operation
```

site/docs/relations/physical_relations.md
```diff
@@ -205,6 +205,23 @@ The streaming aggregate operation leverages data ordered by the grouping express
 | Per Grouping Set | A list of expression grouping that the aggregation measured should be calculated for. | Optional, defaults to 0.                |
 | Measures         | A list of one or more aggregate expressions. Aggregate expressions ordering requirements must be compatible with expected ordering. | Optional, required if no grouping sets. |
 
+## Window Operation
+A window operation is a special type of project operation where every function is a window function and all of the window functions share the same sorting and partitioning. This allows for the sort and partition to be calculated once and shared between the various function evaluations. This operator doesn't need to build hash tables as Hashing Window and rely on the sorted input as Streaming Window.
```
westonpace, line 209:
```suggestion
A consistent partition window operation is a special type of project operation where every function is a window function and all of the window functions share the same sorting and partitioning. This allows for the sort and partition to be calculated once and shared between the various function evaluations. This operator doesn't need to build hash tables as Hashing Window and rely on the sorted input as Streaming Window.
```

site/docs/relations/physical_relations.md
```diff
@@ -205,6 +205,23 @@ The streaming aggregate operation leverages data ordered by the grouping express
...
+| Outputs              | 1                                                            |
+| Property Maintenance | Maintains distribution. Eliminates ordering.                 |
+| Direct Output Order  | Same as Project operator (input followed by each window expression). |
+
+### Window Properties
+
+| Property           | Description                     | Required               |
+| ------------------ | ------------------------------- | ---------------------- |
+| Input              | The relational input.           | Required               |
+| Window Expressions | One or more window expressions. | At least one required. |
```
westonpace, line 223:
```suggestion
| Window Functions | One or more window functions. | At least one required. |
```

site/docs/relations/physical_relations.md
```diff
@@ -205,6 +205,23 @@ The streaming aggregate operation leverages data ordered by the grouping express
...
 | Measures         | A list of one or more aggregate expressions. Aggregate expressions ordering requirements must be compatible with expected ordering. | Optional, required if no grouping sets. |
 
+## Window Operation
+A window operation is a special type of project operation where every function is a window function and all of the window functions share the same sorting and partitioning. This allows for the sort and partition to be calculated once and shared between the various function evaluations. This operator doesn't need to build hash tables as Hashing Window and rely on the sorted input as Streaming Window.
+
+| Signature            | Value                                                        |
+| -------------------- | ------------------------------------------------------------ |
+| Inputs               | 1                                                            |
+| Outputs              | 1                                                            |
+| Property Maintenance | Maintains distribution. Eliminates ordering.                 |
```
westonpace, line 215: Should it establish a new ordering based on `sorts`?

westonpace (commented):
A few minor comments.  I'm not sure I understand the comparison to hashing and streaming yet.

@jacques-n is there any engine out there that uses hashing to calculate window functions?  If not, then we should get rid of the hashing and streaming operators.

site/docs/relations/physical_relations.md
JkSelf: Yes. Already updated.

JkSelf:
@jacques-n @westonpace Please help to review again. Thanks.

JkSelf:
@jacques-n @westonpace Can you help to review again? Thanks for your help.

site/docs/relations/physical_relations.md
```diff
@@ -205,6 +205,23 @@ The streaming aggregate operation leverages data ordered by the grouping express
...
 | Measures         | A list of one or more aggregate expressions. Aggregate expressions ordering requirements must be compatible with expected ordering. | Optional, required if no grouping sets. |
 
+## Consistent Partition Window Operation
+A consistent partition window operation is a special type of project operation where every function is a window function and all of the window functions share the same sorting and partitioning. This allows for the sort and partition to be calculated once and shared between the various function evaluations. This operator doesn't need to build hash tables as Hashing Window and rely on the sorted input as Streaming Window.
+
+| Signature            | Value                                                                |
+| -------------------- |----------------------------------------------------------------------|
+| Inputs               | 1                                                                    |
+| Outputs              | 1                                                                    |
+| Property Maintenance | Maintains distribution and ordering.                                   |
```
westonpace, line 215:
```suggestion
| Property Maintenance | Maintains distribution and ordering.                                 |
```
Minor nit: whitespace change to align formatting.

westonpace (approved):
Two minor suggestions but I don't see anything major.  I'm in favor of merging this and we can tweak if needed as implementations start to support this.

westonpace:
This is a spec modification so we will need one more approval (CC @jacques-n / @cpcloud )

JkSelf:
@westonpace Thanks for your review again. @jacques-n / @cpcloud Can you help to approve?

JkSelf:
@westonpace @jacques-n Can you help to review again if you have any comment? If not, please help to merge. Thanks for your help.

site/docs/relations/physical_relations.md
westonpace: This PR was discussed during the community meeting today.  @jacques-n explained the behavior of the "hashing window operator" further.  "hashing" is probably not the correct word to use here.  There is a variant of the window operator that can be used for some window functions (e.g. count) where the function does not need to represent the entire window in memory.

I propose we do not try and solve this problem in this PR.  Perhaps we can replace this whole section:

```
## Hashing Window Operation

A window aggregate operation that will build hash tables for each distinct partition expression.

| Signature            | Value                                                        |
| -------------------- | ------------------------------------------------------------ |
| Inputs               | 1                                                            |
| Property Maintenance | Maintains distribution. Eliminates ordering.                 |
| Direct Output Order  | Same as Project operator (input followed by each window expression). |

### Hashing Window Properties

| Property           | Description                     | Required               |
| ------------------ | ------------------------------- | ---------------------- |
| Input              | The relational input.           | Required               |
| Window Expressions | One or more window expressions. | At least one required. |
```

with:

```
## Hashing Window Operation

A window aggregate operation that does not need to represent the entire window in memory.  TBD.
```

Or, we can just remove the section entirely.

@jacques-n , thoughts?

proto/substrait/algebra.proto
```diff
@@ -228,6 +228,90 @@ message AggregateRel {
...
+    // less than the lower bound, TODO (null range/no records passed?
+    // wrapping around as if lower/upper were swapped? error? null?).
+    // Optional; defaults to the end of the partition.
+    Expression.WindowFunction.Bound upper_bound = 4;
+
+    // Defines the bounds type: ROWS, RANGE
+    BoundsType bounds_type = 12;
+  }
+
+  enum BoundsType {
```
vbarua, line 302: minor: can we make this enum a top-level enum like `AggregationPhase`.

This will make it easier/nicer to re-use in the `WindowFunction` expression, where it is also needed.

proto/substrait/algebra.proto
```diff
@@ -228,6 +228,90 @@ message AggregateRel {
...
+    BOUNDS_TYPE_UNSPECIFIED = 0;
+    // The lower and upper bound specify how many rows before and after the current row
+    // the window should extend.
+    BOUNDS_TYPE_ROWS = 1;
+    // The lower and upper bound describe a range of values.  The window should include all rows
+    // where the value of the ordering column is greater than or equal to (current_value - lower bound)
+    // and less than or equal to (current_value + upper bound).  This bounds type is only valid if there
+    // is a single ordering column.
+    BOUNDS_TYPE_RANGE = 2;
+  }
```
vbarua, line 312: Can we add `BOUNDS_TYPE_GROUP` while we're adding these?

* https://trino.io/docs/current/sql/pattern-recognition-in-window.html
* https://www.postgresql.org/docs/current/sql-expressions.html#SYNTAX-WINDOW-FUNCTIONS

I'm also happy to do this as a later change to reduce the scope of this.

proto/substrait/algebra.proto
```diff
@@ -829,6 +914,8 @@ message Expression {
     // the same partition as the current record are aggregated.
     repeated Expression partitions = 2;
 
+    ConsistentPartitionWindowRel.BoundsType bounds_type = 12;
```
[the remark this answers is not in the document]
westonpace: I agree this would be a good idea.

proto/substrait/algebra.proto
vbarua: This can be done separately.

proto/substrait/algebra.proto
vbarua: I've added `bounds_type` to `WindowFunction`

proto/substrait/algebra.proto
vbarua: This has been moved into `WindowFunction` as having all the references in `WindowRelFunction` use the same prefix seemed the most consistent. 
```proto
    ...
    Expression.WindowFunction.Bound lower_bound = 5;
    ...
    Expression.WindowFunction.Bound upper_bound = 4;
    ...
    Expression.WindowFunction.BoundsType bounds_type = 12;
```

vbarua:
Per discussion in the sync this morning, @westonpace and I have merged in main and applied some small changes to move this along.

proto/substrait/algebra.proto
```diff
@@ -230,6 +230,78 @@ message AggregateRel {
...
+  Rel input = 2;
+  repeated WindowRelFunction window_functions = 3;
+  repeated Expression partition_expressions = 4;
+  repeated SortField sorts = 5;
+
+  substrait.extensions.AdvancedExtension advanced_extension = 10;
+
+  // A WindowFunction can define itself WindowSpec(sorts columns and partitions keys)
+  // But WindowRelFunction remove the WindowSpec property, because all the WindowRelFunctions
+  // share the same WindowSpec() in WindowRel.
```
westonpace, line 247:
```suggestion
  // This message mirrors the `WindowFunction` message but removes the fields defining the partition,
  // sorts, and bounds, since those must be consistent across the various functions in this rel.  Refer
  // to the `WindowFunction` message for a description of these fields.
```

proto/substrait/algebra.proto
```diff
@@ -230,6 +230,78 @@ message AggregateRel {
...
+  message WindowRelFunction {
+    // Points to a function_anchor defined in this plan, which must refer
+    // to a window function in the associated YAML file. Required; 0 is
+    // considered to be a valid anchor/reference.
```
westonpace, line 251:
```suggestion
```
Remove to avoid double-documenting this field.

proto/substrait/algebra.proto
```diff
@@ -230,6 +230,78 @@ message AggregateRel {
...
+    // argument types must also match exactly:
+    //
+    //  - Value arguments must be bound using FunctionArgument.value, and
+    //    the expression in that must yield a value of a type that a function
+    //    overload is defined for.
+    //  - Type arguments must be bound using FunctionArgument.type, and a
+    //    function overload must be defined for that type.
+    //  - Enum arguments must be bound using FunctionArgument.enum
+    //    followed by Enum.specified, with a string that case-insensitively
+    //    matches one of the allowed options.
```
westonpace, line 265:
```suggestion
```

Remove to avoid double-documenting this field.

proto/substrait/algebra.proto
```diff
@@ -230,6 +230,78 @@ message AggregateRel {
...
+    repeated FunctionArgument arguments = 9;
+
+    // Options to specify behavior for corner cases, or leave behavior
+    // unspecified if the consumer does not need specific behavior in these
+    // cases.
```
westonpace, line 270:
```suggestion
```

proto/substrait/algebra.proto
```diff
@@ -230,6 +230,78 @@ message AggregateRel {
...
+    repeated FunctionOption options = 11;
+
+    // Must be set to the return type of the function, exactly as derived
+    // using the declaration in the extension.
```
westonpace, line 274:
```suggestion
```

proto/substrait/algebra.proto
```diff
@@ -230,6 +230,78 @@ message AggregateRel {
...
+    Type output_type = 7;
+
+    // Describes which part of the window function to perform within the
+    // context of distributed algorithms. Required. Must be set to
+    // INITIAL_TO_RESULT for window functions that are not decomposable.
```
westonpace, line 279:
```suggestion
```

proto/substrait/algebra.proto
```diff
@@ -230,6 +230,78 @@ message AggregateRel {
...
+    AggregationPhase phase = 6;
+
+    // Specifies whether equivalent records are merged before being aggregated.
+    // Optional, defaults to AGGREGATION_INVOCATION_ALL.
```
westonpace, line 283:
```suggestion
```

proto/substrait/algebra.proto
```diff
@@ -230,6 +230,78 @@ message AggregateRel {
...
+    AggregateFunction.AggregationInvocation invocation = 10;
+
+    // Defines the record relative to the current record from which the window
+    // extends. The bound is inclusive. If the lower bound indexes a record
+    // greater than the upper bound, TODO (null range/no records passed?
+    // wrapping around as if lower/upper were swapped? error? null?).
+    // Optional; defaults to the start of the partition.
```
westonpace, line 290:
```suggestion
```

proto/substrait/algebra.proto
```diff
@@ -230,6 +230,78 @@ message AggregateRel {
...
+    Expression.WindowFunction.Bound lower_bound = 5;
+
+    // Defines the record relative to the current record up to which the window
+    // extends. The bound is inclusive. If the upper bound indexes a record
+    // less than the lower bound, TODO (null range/no records passed?
+    // wrapping around as if lower/upper were swapped? error? null?).
+    // Optional; defaults to the end of the partition.
```
westonpace, line 297:
```suggestion
```

proto/substrait/algebra.proto
```diff
@@ -230,6 +230,78 @@ message AggregateRel {
...
+    Expression.WindowFunction.Bound upper_bound = 4;
+
+    // Defines the bounds type: ROWS, RANGE
```
westonpace, line 300:
```suggestion
```

site/docs/relations/physical_relations.md
```diff
@@ -205,6 +205,23 @@ The streaming aggregate operation leverages data ordered by the grouping express
 | Per Grouping Set | A list of expression grouping that the aggregation measured should be calculated for. | Optional, defaults to 0.                |
 | Measures         | A list of one or more aggregate expressions. Aggregate expressions ordering requirements must be compatible with expected ordering. | Optional, required if no grouping sets. |
 
+## Consistent Partition Window Operation
+A consistent partition window operation is a special type of project operation where every function is a window function and all of the window functions share the same sorting and partitioning. This allows for the sort and partition to be calculated once and shared between the various function evaluations. This operator doesn't need to build hash tables as Hashing Window and rely on the sorted input as Streaming Window.
```
westonpace, line 209:
```suggestion
A consistent partition window operation is a special type of project operation where every function is a window function and all of the window functions share the same sorting and partitioning. This allows for the sort and partition to be calculated once and shared between the various function evaluations.
```

I don't know that this sentence is needed.  The consumer is free to implement this however they see fit.

westonpace (approved):
Structurally I think we've reached the correct point for this PR.  I have some minor suggestions regarding how we document this.  We recently updated (or are updating) the wording of `WindowFunction.function_reference` to state that we allow both window and aggregate functions.  However, that change is not mirrored here.

Rather than try and keep the comments in sync between the two definitions I think we should only have one set of definitions and only document the parts that are different here.

I'm marking this as approved because this PR has been around for a while.  If these changes are not addressed then I propose we merge it as is and clean up the docs in a follow-up PR.

JkSelf:
@vbarua @westonpace @jacques-n 
Sorry for the late response. I have updated the changes. Can you help to review again?

westonpace (approved):

jacques-n (approved):
lgtm

PR #410, opened 2023-01-02
chore: update commit msg checks to look at PR title/desc

westonpace:
This replaces the old check which looked at the commit messages.


westonpace:
After this merges I will update the repository settings to always use the PR title & description as the commit message.

CC @mbrobbel as I borrowed some of your code and wouldn't mind a review (note I am checking both the PR title & description so we can check for BREAKING CHANGE in the description)

.github/workflows/pr_title.yml
```diff
@@ -0,0 +1,109 @@
...
+  pull_request:
+    types: [opened, edited, synchronize, reopened]
+jobs:
+  commitlint:
+    name: PR title / description conforms to semantic-release
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+        with:
+          fetch-depth: 0
```
mbrobbel, line 13: Checkout is unused here.
```suggestion
```

.github/workflows/pr_title.yml
```diff
@@ -0,0 +1,109 @@
+name: PR Title Check
+
+on:
+  pull_request:
```
mbrobbel, line 4: With `pull_request` there are no write permissions to post a comment to the PR. So this should be `pull_request_target` to enable that.
```suggestion
  pull_request_target:
```
We should however consider the consequences: https://securitylab.github.com/research/github-actions-preventing-pwn-requests/

.github/workflows/pr_title.yml
```diff
@@ -0,0 +1,109 @@
...
+            breaking="true"
+          else
+            breaking="false"
+          fi
+
+          echo "::set-output name=breaking::${breaking}"
+      - name: check whether the PR description includes a breaking change footer
+        if: ${{ fromJson(steps.check-breaking.outputs.breaking) }}
+        run: |
+          set -euo pipefail
```
mbrobbel, line 74: I guess we don't need this here: https://docs.github.com/en/enterprise-cloud@latest/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idstepsshell
```suggestion
```

.github/workflows/pr_title.yml
```diff
@@ -0,0 +1,109 @@
...
+              repo: context.repo.repo,
+              issue_number: context.issue.number,
+              body: message
+            })
+            core.setFailed('Invalid PR title')
+  breaking:
+    name: Ensure breaking changes are labeled in description
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
```
mbrobbel, line 57: Because this job requires a checkout we should probably move this back to the other PR workflow (with a normal `pull_request` trigger) and skip the PR comment (so it doesn't require access to the secrets).
https://securitylab.github.com/research/github-actions-preventing-pwn-requests/

.github/workflows/pr_title.yml
```diff
@@ -0,0 +1,109 @@
...
+      - uses: bufbuild/buf-setup-action@v1.9.0
+      - name: check for breaking changes
+        id: check-breaking
+        run: |
+          set -euo pipefail
```
mbrobbel, line 62:
```suggestion
```

.github/workflows/pr_title.yml
```diff
@@ -0,0 +1,109 @@
...
+
+          if ! buf breaking --against 'https://github.com/substrait-io/substrait.git#branch=main'; then
+            breaking="true"
+          else
+            breaking="false"
+          fi
+
+          echo "::set-output name=breaking::${breaking}"
```
mbrobbel, line 70: The method to define outputs here is deprecated (https://docs.github.com/en/actions/using-workflows/workflow-commands-for-github-actions#setting-an-output-parameter)
```suggestion
          echo "breaking=${breaking}" >> $GITHUB_OUTPUT
```

mbrobbel (changes requested):
Great, this makes it easier to contribute.

.github/workflows/pr_title.yml
westonpace: Got it.  All my testing was between branches in the same fork so I didn't notice.  So, assuming we remove the checkout and the `breaking` workflow the only user input would be the PR title and description which are used in:

```
npx commitlint --verbose <<< $COMMIT_MSG
```

I think that is a reasonable risk.

.github/workflows/pr_title.yml
westonpace: I moved it to a separate workflow (using the `pull_request` target) because I want it to be triggered on a PR edit.

mbrobbel (approved):

.github/pull_request_template.md
```diff
@@ -1,8 +1,8 @@
 Thank you for submitting a PR!
 
-Before you continue, please ensure that your commit messages follow [conventional commit syntax](1). This will save you a rebase later. Substrait uses an automated release process that, among other things, uses the commit messages to build a changelog, so the syntax and format matters!
+Before you continue, please ensure that your PR title and description (this message!) follow [conventional commit syntax](1).  Substrait uses an automated release process that, among other things, uses PR titles & descriptions to build a changelog, so the syntax and format matter!
```
vibhatha, line 3:
```suggestion
Before you continue, please ensure that your PR title and description (this message!) follow [conventional commit syntax](1). Substrait uses an automated release process that, among other things, uses PR titles & descriptions to build a changelog, so the syntax and format matter!
```

westonpace:
Demo: https://github.com/westonpace/substrait/pull/3

vibhatha (approved):

PR #423, opened 2023-01-21
fix: update extension yaml files to match type-syntax spec

chaojun-zhang:
update extension yaml files to match type-syntax spec


westonpace (approved):

westonpace:
Thanks!

PR #427, opened 2023-01-26
docs: update DataFusion entry in Powered By page

ianmcook:
The datafusion-substrait code recently moved into the main datafusion repo.
This updates the "Powered By" page accordingly.


vibhatha (approved):
@ianmcook I think the commit message needs to be formatted. Once fixed it should pass the CIs.

westonpace (approved):

westonpace:
The problem was "body lines cannot be longer than 100 characters".  "body lines" in this case refers to the PR description.  I manually link-broke your description for now but it appears that GH also does some kind of line breaking when creating commit messages so maybe we need to relax that constraint.

PR #435, opened 2023-02-02
docs: remove mention of the options in the tutorial

### LINKED ISSUE #433: Confused section for expressions arguments
In [link](https://github.com/substrait-io/substrait/blob/07e4feb5983478e7d0d95dc1d9b5e176685dbdc3/site/docs/tutorial/sql_to_substrait.md?plain=1#L359) 

> `arguments` defines the arguments being passed into the function, which are all
done positionally based on the function definition in the YAML file. The first 
argument is an option that defines how `NaN` equality is handled, and since
that's not relevant here we just pass `"NAN_IS_NOT_NAN"`. The next two options
will be familiar as the literal and the field reference we constructed earlier.

I do not see the argument for the `how NaN equality is handled` in the `INDEX_IN` and `IS NULL`, nor the `NAN_IS_NOT_NAN`. Do I miss something?

wjones127:
Fixes #433

### DIFF
--- a/site/docs/tutorial/sql_to_substrait.md
+++ b/site/docs/tutorial/sql_to_substrait.md
@@ -355,10 +355,9 @@ definitions that we will create later.
 `i64` type since that is what the function definition declares in the YAML file.
 
 `arguments` defines the arguments being passed into the function, which are all
-done positionally based on the function definition in the YAML file. The first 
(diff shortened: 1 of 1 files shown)


wjones127:
I updated `substrait-validator` to version `0.0.11` and verified the example plan is still valid. It checked out 👍

westonpace (approved):

PR #438, opened 2023-02-06
feat: add isPartial flag in fetchRel and SortRel

ZJie1:
In velox topn/orderby/limit node, there need a boolean indicating 
whether the node generates partial results on local workers or finalizes the partial result.

### DIFF
--- a/proto/substrait/algebra.proto
+++ b/proto/substrait/algebra.proto
@@ -195,6 +195,10 @@ message FetchRel {
   int64 offset = 3;
   // the amount of records to return
   int64 count = 4;
+  // A Boolean indicating whether the node generates partial
+  // results on local workers or finalizes the partial results.
+  bool is_partial = 5;
+
   substrait.extensions.AdvancedExtension advanced_extension = 10;
 }
 
@@ -233,6 +237,10 @@ message SortRel {
   RelCommon common = 1;
   Rel input = 2;
   repeated SortField sorts = 3;
+
+  // True if this node only sorts a portion of the final result. If it is true,
+  // a local merge or merge exchange is required to merge the sorted runs.
+  bool is_partial = 4;
   substrait.extensions.AdvancedExtension advanced_extension = 10;
 }
 


proto/substrait/algebra.proto
```diff
@@ -233,6 +237,10 @@ message SortRel {
   RelCommon common = 1;
   Rel input = 2;
   repeated SortField sorts = 3;
+
+  // True if this node only sorts a portion of the final result. If it is true,
+  // a local merge or merge exchange is required to merge the sorted runs.
```
westonpace, line 242: This needs more explanation.  Also, I don't like the term "partial".  A "partial sort", as defined by [Wikipedia](https://en.wikipedia.org/wiki/Partial_sorting) is what might be called "top N".  It sorts only part of a range.  This is also what we get from the [C++ standard library](https://en.cppreference.com/w/cpp/algorithm/partial_sort).  These are the top 2 google results for "partial sort".

To be honest, I'm not really sure what is intended here.  I assume this is based on velox.  However, the documentation on this parameter is rather sparse.  From scanning the code the only place I see this get used is to determine if the driver can use more than 1 thread.  The orderby operator itself does not actually appear to use this flag.

My educated guess is that this is related to a distributed plan (in Velox I think a "multi-threaded" plan is treated the same as "locally distributed").  So if this is a partial sort and there are multiple splits then each split can be given its own thread and sorted independently.  If it is not a partial sort and there are multiple splits then the splits must be processed one at a time and the data from all the splits needs to be accumulated and sorted.

We have had very little discussion on how Substrait will look in a distributed environment but I think it is something we very much want to solve.  However, in my mind, the "partialness" of a sort would not be something that had to be expressed in the plan.  If, for example, a plan is to be distributed to 10 workers, then I think you would create a unique substrait plan for each of the workers (and each worker would see the sort as a non-partial sort because the worker only knows about the data it is processing and there is no concept of a partial sort).

The velox approach seems more complicated, but I also don't want to reject it.  @mbasmanova can you verify some of my guessing above?  Do you have any thoughts or opinions on how we might want to tackle distributed plans in Substrait?

westonpace (changes requested):
Thanks for the ping @vibhatha .  I think this would be good to get sorted out (pun absolutely intended :) ).

proto/substrait/algebra.proto
jacques-n: To clarify, I'm also super supportive on making sure Substrait supports Velox well, just want to make sure we're not mapping things too literally.

mbasmanova:
@westonpace 

> if this is a partial sort and there are multiple splits then each split can be given its own thread and sorted independently. If it is not a partial sort and there are multiple splits then the splits must be processed one at a time and the data from all the splits needs to be accumulated and sorted.

This is how we are thinking about it in Presto and Velox. The concept of "partial" execution applies to aggregations as well. In this case, partial aggregation is different from final aggregation in that it is allowed to flush results if memory gets tight which may result in multiple results with the same group-by keys.

westonpace:
> This is how we are thinking about it in Presto and Velox. The concept of "partial" execution applies to aggregations as well. In this case, partial aggregation is different from final aggregation in that it is allowed to flush results if memory gets tight which may result in multiple results with the same group-by keys.

This aggregation case actually sounds more compelling since the `partial` flag modifies how the operation itself behaves (it may release info early and isn't simply a marker for the pipeline.

For sort & fetch can we do this as a "hint" for now?

mbasmanova:
> For sort & fetch can we do this as a "hint" for now?

@westonpace I'm not familiar with the concept of "hint". Would you point me to some reading material and explain what it is?

westonpace:
We haven't used hints much.  They are an extension point described indirectly [here](https://substrait.io/extensions/#advanced-extensions).  Specifically I'm thinking of the row `Relation Modification (optimization)` which is described as:

> Extensions to an existing relation that can improve the efficiency of a plan consumer but don’t fundamentally change the behavior of the operation. An example might be an estimated amount of memory the relation is expected to use or a particular algorithmic pattern that is perceived to be optimal.

I am suggesting an extension point because I believe this field is unique to Velox.  Other engines appear to have solved this in different ways.  Since `SortRel` and `FetchRel` are very common (and very logical) relations used by most engines today I think we should aim to keep them generic.  Another approach that we could take is to define physical `SortRel` and `FetchRel` equivalents that are more specifically tailored to Velox's implementation.  I'd be fine with that too but I think it would be more work.

In the plan itself I am thinking about _this message_ (which can be applied to any relation):

```
  // Changes to the operation that can influence efficiency/performance but
  // should not impact correctness.
  message Hint {
    Stats stats = 1;
    RuntimeConstraint constraint = 2;

    // The statistics related to a hint (physical properties of records)
    message Stats {
      double row_count = 1;
      double record_size = 2;
    }

    message RuntimeConstraint {
      // TODO: nodes, cpu threads/%, memory, iops, etc.

    }
  }
```

It sounds (to me at least) like this is a sort of runtime constraint defined as "if a pipeline contains this operator then that pipeline must use a single-threaded driver".  It doesn't really sound like it needs to be unique to sort and fetch.  Using a hint therefore seems like an easier solution (add it in one spot instead of two spots).  However, if Velox would instead prefer to mirror the current design then I think physical variations of sort and fetch are the way to go.

mbasmanova:
> Other engines appear to have solved this in different ways.

I'm curious which engines you are thinking about and how do thy use sort and limit. I know that Presto use sort and limit in the same way as Velox.

westonpace:
> I'm curious which engines you are thinking about and how do thy use sort and limit. I know that Presto use sort and limit in the same way as Velox.

I don't know the details of their threading models and schedulers.  All I know is that there is no "is_partial" flag that is part of their sort relation.

* DuckDb

DuckDb currently only interacts with Substrait at a logical plan level.  However, I can inspect their physical implementation a bit.  DuckDb does not use exchange for intra-query parallelism.  They use morsel driven parallelism (more about it [here](file:///home/pace/Downloads/duckdb-parallelism.pdf)).  Instead of static parallelism based on splits they have dynamic parallelism based on tasks (e.g. every 100k rows read from disk starts a new thread task to feed the pipeline).  Scanning through their physical plan I don't see any concept of a order-preserving merge node.  Since they aren't using exchange for their own purposes this would only be needed if someone was trying to map a distributed query onto DuckDb and it doesn't seem that anyone has done distributed sort with DuckDb yet.  If DuckDb were used in a distributed fashion then my guess is that each instance of DuckDb would get their own plan and so there would be no need to indicate a difference between first and final sorts.
 
* Datafusion

Datafusion is more similar to Velox.  They also use exchange for parallelism and they have "streams" which represent the same concept as splits.  They have distributed sort implemented and it uses an order preserving merge step similar to Velox.  However, the sort operator is the same regardless of whether it is the first (partial) sort or the final sort.  Instead, Datafusion creates a separate execution plan for each split.  There is a diagram [here](https://docs.rs/datafusion/21.1.0/datafusion/physical_plan/sorts/sort_preserving_merge/struct.SortPreservingMergeExec.html) and a video overview [here](https://www.youtube.com/watch?v=2jkWU3_w6z0).

* Acero

Acero uses a threading model similar to DuckDb.  It does not rely on splits or static partitioning.  There is no merging exchange at the moment and no concept of a partial sort.  I did a rough sketch of this at one point.  It assumed that each worker has its own plan.

So I think the unique thing about Velox is that it represents parallelism implicitly and not explicitly in the shape of the plan itself.  In other words, there is only one `SortRel` even if there are 20 different workers:

image

> I know that Presto use sort and limit in the same way as Velox.

That's very possible.  I don't know anything about Presto really.  I can see pros & cons to the way Velox represents plans but don't know that we need to go into more detail here.

### Bigger picture

I think there is a larger issue at play if we zoom out.  Velox's representation is a physical representation.  DuckDb, Ibis, Datafusion, and Calcite have all been interacting with Substrait at the logical level.  The relations we have today are mean to be logical relations.  This doesn't mean that Substrait is only designed for logical plans.  It's simply that the emphasis so far has been on logical plans.

The approach for consuming Substrait that both Velox and Acero have taken is to consume Substrait's logical relations (since they exist and are well defined) and still produce a physical plan.  This is only one spot where this is starting to cause friction.  I'm more familiar with the pain points in Acero but, for example:

 * Acero only supports hash equi-join and so, when it parses Substrait's logical join relation, it has to try and convert an arbitrary join expression into equi-join criteria (and rejects plans that aren't equijoins).  Velox simply didn't add early support for the logical join operator and waited for the physical hash equijoin operator to be finalized.
 * Acero expects aggregate keys to be field references and not arbitrary expressions.  So it rejects substrait plans that have arbitrary expressions for join keys (I think Velox may have a similar constraint)
 * Etc.

I think we both agree that the "partial" flag for sort is not a part of the logical plan.  This is why my earlier suggestion was "either create a physical sort or add a velox-specific hint".  It wasn't meant to be pejorative and it doesn't mean that Velox is doing anything wrong.  Physical plans are always going to be unique to the engines that consume them.  I think trying to unify the three pictures above is probably a lost cause.

mbasmanova:
> I think trying to unify the three pictures above is probably a lost cause.

I somewhat feel similar. It is not clear to me what the value is in creating Velox-specific hints or extensions as this would yield plans that can only be processed by Velox.

westonpace:
> I somewhat feel similar. It is not clear to me what the value is in creating Velox-specific hints or extensions as this would yield plans that can only be processed by Velox.

At the moment I agree there isn't much value.  If an planner wanted to, for example, write plans for both Velox and Acero, then that planner would need to know the details of both physical representations.  There is some small advantage in the fact that they at least consistently use protobuf, there is some attempt at consistent terminology, etc.  However, it is not a magic bullet by any means.

Longer term there is a possibility for development of tools that work for both logical and physical plans.  For example, plan visualization, tracing, opt-gen like plan transformation languages, etc.

westonpace:
I think we have a good understanding of how Velox interprets this field now.  Perhaps the question is now for @ZJie1 around what the goals are.  I'm not sure if this originated from Gluten or not.  In that case I think the goal is to go from Spark's physical plan to Velox.

How does Spark model sorts?  If it is not the same as Velox then there will need to be a mapping.

* We could use Spark's model and create Substrait relations that map to Spark and then do this mapping in Velox's substrait consumer (this means Velox's substrait consumer accepts "spark physical substrait")
* We could model Velox's sort and then the mapping would need to happen in Gluten when creating the Substrait
* We could model both Spark and Velox's sorts and the mapping could happen in some tool in between (there is a long term goal of Substrait to make it easier to create these kinds of tools but I don't know how far along they are)

There are other options too (e.g. velox could accept logical substrait and then apply a planning stage to convert to the physical sort so then we go Spark logical -> Spark physical -> Substrait logical -> Velox logical -> Velox physical)

ianmcook:
FYI, Hive's SQL dialect includes both `ORDER BY` and `SORT BY`: https://cwiki.apache.org/confluence/display/Hive/LanguageManual+SortBy#LanguageManualSortBy-DifferencebetweenSortByandOrderBy

My impression here is that:
- Hive `ORDER BY` maps to `is_partial = false`
- Hive `SORT BY` maps to `is_partial = true`

ianmcook:
> I think we both agree that the "partial" flag for sort is not a part of the logical plan. 

I disagree. I think this is flag could absolutely be a part of a logical plan.

westonpace:
So it sounds like this wouldn't be a relation that is only consumable by Velox then.  It would be usable by any system that has a plan representation for distributed queries that have not yet been split up (and, in the case of Velox, may never be split up).  For lack of a better term I'm going to call it a "compact distributed plan"

It sounds like, given a pipeline in this representation, each node may or may not be allowed to run in parallel (e.g. worker/thread per partition/split)

 * Some nodes, like scan and project, can always run in parallel
 * Some nodes, like merging capture (e.g. merge exchange, sort-preserving-merge), can never be run in parallel
 * Some nodes, like sort may or may not be able to run in parallel, and we need a flag to tell us which case it is

The previous point still stands though that it would help to know the purpose of this PR.

jacques-n:

PR #439, opened 2023-02-09
chore: add gha token to buf to avoid rate limit failures in ci

westonpace:
The buf action in our CI scripts will, without configuration, make
anonymous calls to download artifacts from github (previous /
golden versions of the proto files that it will compare against). 
These calls may fail due to rate limiting.

By passing in the runner's github token (the one that is automatically
created).  We can bypass the strict anonymous rate limits.

### DIFF
--- a/.github/workflows/pr.yml
+++ b/.github/workflows/pr.yml
@@ -28,6 +28,8 @@ jobs:
     steps:
       - uses: actions/checkout@v3
       - uses: bufbuild/buf-setup-action@v1.9.0
+        with:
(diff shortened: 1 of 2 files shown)


vibhatha:
@westonpace sort of got an idea from the initial commit message. Shall we add a description?

westonpace:
> @westonpace sort of got an idea from the initial commit message. Shall we add a description?

Good idea.  Done.

vibhatha (approved):

PR #442, opened 2023-02-14
feat: add null aware anti join

### LINKED ISSUE #325: What are the semantics of the ANTI join
Folks,

I'm looking at the join descriptions and in particular at ANTI join.

https://substrait.io/relations/logical_relations/#join-types

> Return records from the left input. These are returned only if the records do not have a join partner on the right side.

I believe there are 2 kinds of ANTI join. One is used for NOT IN queries, the other for NOT EXISTS. They differ in how they handle nulls.

- Null aware anti join (NOT IN) returns no rows if build side has null in the join key. 
- Null aware anti join (NOT IN) returns rows with null in the probe side join key only if build side empty.

vs. 

- Anti join (NOT EXISTS) ignores nulls on the build side.
- Anti join (NOT EXISTS) always returns rows with null in the probe side join key.

Here are some examples to illustrate the differences.

Null aware anti join (NOT IN):

NULL on build side:

```
with a as (select * from unnest(array[null, 1, 2]) as _t(id)),
b as (select * from unnest(array[null, 2, 3]) as _t(id))
select * from a where a.id not in (select id from b)

 id
----
(0 rows)
```

No nulls on build side:

```
with a as (select * from unnest(array[null, 1, 2]) as _t(id)),
b as (select * from unnest(array[2, 3]) as _t(id))
select * from a where a.id not in (select id from b)

 id
----
  1
(1 row)
```

EMPTY build side

```
with a as (select * from unnest(array[null, 1, 2]) as _t(id)),
b as (select * from unnest(array[]) as _t(id))
select * from a where a.id not in (select id from b)

  id
------
    1
    2
 NULL
(3 rows)
```

Anti join (NOT EXISTS):

NULL on build side:

```
with a as (select * from unnest(array[null, 1, 2]) as _t(id)),
b as (select * from unnest(array[null, 2, 3]) as _t(id))
select * from a where not exists (select * from b where b.id = a.id)

  id
------
 NULL
    1
(2 rows)
```

No nulls on build side

```
with a as (select * from unnest(array[null, 1, 2]) as _t(id)),
b as (select * from unnest(array[2, 3]) as _t(id))
select * from a where not exists (select * from b where b.id = a.id)

  id
------
    1
 NULL
(2 rows)
```

EMPTY build side

```
with a as (select * from unnest(array[null, 1, 2]) as _t(id)),
b as (select * from unnest(array[]) as _t(id))
select * from a where not exists (select * from b where b.id = a.id)

  id
------
    2
    1
 NULL
(3 rows)
```

I'm wondering which of these semantics are assumed for Substrait ANTI join? Would it make sense to define two anti joins? Null-aware anti join and regular anti-join?

CC: @rui-mo @jacques-n

westonpace:
an anti join is typically used because it is semantically equivalent
to a NOT IN or NOT EXISTS correlated subquery.  However, the
semantics of these two operations differ in their treatment of nulls.
This is discussed in greater detail in the linked issue.  By adding a
second type of anti join we make it possible to reflect both semantics.

closes #325

### DIFF
--- a/proto/substrait/algebra.proto
+++ b/proto/substrait/algebra.proto
@@ -170,6 +170,7 @@ message JoinRel {
     JOIN_TYPE_RIGHT = 4;
     JOIN_TYPE_SEMI = 5;
     JOIN_TYPE_ANTI = 6;
+	JOIN_TYPE_NULL_AWARE_ANTI = 8;
     // This join is useful for nested sub-queries where we need exactly one tuple in output (or throw exception)
     // See Section 3.2 of https://15721.courses.cs.cmu.edu/spring2018/papers/16-optimizer2/hyperjoins-btw2017.pdf
     JOIN_TYPE_SINGLE = 7;
@@ -505,6 +506,8 @@ message HashJoinRel {
     JOIN_TYPE_RIGHT_SEMI = 6;
     JOIN_TYPE_LEFT_ANTI = 7;
     JOIN_TYPE_RIGHT_ANTI = 8;
+	JOIN_TYPE_LEFT_NULL_AWARE_ANTI = 9;
+	JOIN_TYPE_RIGHT_NULL_AWARE_ANTI = 10;
   }
 
   substrait.extensions.AdvancedExtension advanced_extension = 10;
@@ -532,6 +535,8 @@ message MergeJoinRel {
     JOIN_TYPE_RIGHT_SEMI = 6;
     JOIN_TYPE_LEFT_ANTI = 7;
     JOIN_TYPE_RIGHT_ANTI = 8;
+	JOIN_TYPE_LEFT_NULL_AWARE_ANTI = 9;
+	JOIN_TYPE_RIGHT_NULL_AWARE_ANTI = 10;
   }
 
   substrait.extensions.AdvancedExtension advanced_extension = 10;
--- a/proto/substrait/algebra.proto
+++ b/proto/substrait/algebra.proto
@@ -170,7 +170,7 @@ message JoinRel {
     JOIN_TYPE_RIGHT = 4;
     JOIN_TYPE_SEMI = 5;
     JOIN_TYPE_ANTI = 6;
-	JOIN_TYPE_NULL_AWARE_ANTI = 8;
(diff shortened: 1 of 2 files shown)


westonpace:
As a related note, I am not sure if there is any reason to have both LEFT_ANTI_JOIN and RIGHT_ANTI_JOIN in the hash join and merge join physical relations.

EpsilonPrime:
An alternative would be to implement null-awareness as an option.  Not being familiar with the history of this particular project do you know of anything that might help decide whether this should be a relation or an option?

westonpace:
A fair question.  When you say `option` I think "function option" and I'm not sure it belongs at that level.  In SQL `op(null, x)` (e.g. `equal(left.id, right.id)` when `left.id` is `null`) usually returns `null`.  This is pretty well defined and I don't think we want to open up a new option for that and there are functions like `is_null` or `is_valid` which can be used.

However, it does mean producers could choose to distinguish between the two cases by saying:

```
ANTI JOIN ON or(equal(left.id, right.id), is_null(equal(left.id, right.id)))
```

I think this would be a little more confusing for producers to write.

On the other hand, perhaps by `option`, you are referring to an option on the join relation.  In other words we could do something like...

```
message JoinRel {
  ...
  Expression post_join_filter = 5;
  JoinType type = 6;
  bool null_awareness = 7;
  ...
```

Then we would need to clarify some of the documentation but I think that is doable.  This is how both Acero and Velox handle this.  Acero calls the property `key_cmp` and Velox calls the property `nullAware`.  One potential downside is that I'm not aware of the usefulness of this property for other join types (though I may simply be uninformed).

Another approach we could take would be to define it as an option on the relation (e.g as just described above) but only on the physical HashJoinRel and not on the more general JoinRel.  This would work I think since (as far as I know) producers don't typically specify an anti-join anyways.  A producer would normally specify a subquery and an optimizer would be the tool that realizes the subquery could be satisfied by an anti-join or semi-join.  In fact, one could probably reasonable argue that semi-join and anti-join are not valid options for the logical join message at all (e.g. it is not a valid join type in SQL server IIUC).

westonpace:
Closing this approach in favor of an approach discussed in the community sync meeting (I'll open a new PR soon)

PR #446, opened 2023-02-22
docs: add tools page

richtia:
Add a new page linking to existing tools that support substrait.

### DIFF
--- a/site/docs/tools/_config
+++ b/site/docs/tools/_config
@@ -1,0 +1,4 @@
+arrange:
+  - producer_tools.md
+  - substrait_validator.md
+  - third_party_tools.md
--- a/site/docs/tools/producer_tools.md
+++ b/site/docs/tools/producer_tools.md
@@ -1,0 +1,7 @@
+# Producer Tools
+
+## Isthmus
+
+[Substrait Isthmus](https://github.com/substrait-io/substrait-java/tree/main/isthmus) is a Java 
+library which enables serializing SQL to [Substrait Protobuf](https://substrait.io/serialization/binary_serialization/) 
+via the Calcite SQL compiler.
--- a/site/docs/tools/producer_tools.md
+++ b/site/docs/tools/producer_tools.md
@@ -2,6 +2,6 @@
 
 ## Isthmus
 
-[Substrait Isthmus](https://github.com/substrait-io/substrait-java/tree/main/isthmus) is a Java 
-library which enables serializing SQL to [Substrait Protobuf](https://substrait.io/serialization/binary_serialization/) 
+[Isthmus](https://github.com/substrait-io/substrait-java/tree/main/isthmus) is an application
+that serializes SQL to [Substrait Protobuf](https://substrait.io/serialization/binary_serialization/) 
 via the Calcite SQL compiler.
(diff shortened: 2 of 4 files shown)


site/docs/tools/_config
```diff
@@ -0,0 +1,3 @@
+arrange:
+  - producer_tools.md
+  - substrait_validator.md
```
westonpace, line 3: Missing third_party_tools.md?

site/docs/tools/producer_tools.md
```diff
@@ -0,0 +1,7 @@
+# Producer Tools
```
westonpace, line 1: Are these linked into the main site automatically somehow?

site/docs/tools/_config
richtia: Added

site/docs/tools/producer_tools.md
richtia: Yes, mkdocs should handle everything.

westonpace:
Can you rebase/merge (I'm pretty sure I can do it but don't want to mess up your branch if you're working on it):

image

richtia:
Done!

site/docs/tools/producer_tools.md
```diff
@@ -0,0 +1,7 @@
...
+
+## Isthmus
+
+[Substrait Isthmus](https://github.com/substrait-io/substrait-java/tree/main/isthmus) is a Java 
+library which enables serializing SQL to [Substrait Protobuf](https://substrait.io/serialization/binary_serialization/) 
+via the Calcite SQL compiler.
```
westonpace, line 7:
```suggestion
[Isthmus](https://github.com/substrait-io/substrait-java/tree/main/isthmus) is an application 
that serializes SQL to [Substrait Protobuf](https://substrait.io/serialization/binary_serialization/) 
via the Calcite SQL compiler.
```
I think the fact that it uses Java under the hood is irrelevant.  It tries (via graal) to produce a simple executable so the user would never have to know that they were using Java.  Now, it's not easily built and requires cloning substrait-java but I think those details are irrelevant in a tool description.

site/docs/tools/third_party_tools.md
```diff
@@ -0,0 +1,6 @@
+# Third Party Tools
+
+## Substrait-tools
+The [substrait-tools](https://pypi.org/project/substrait-tools/) python package provides 
+a command line interface for producing/consuming substrait plans by leveraging the APIs
+from different producers and consumers. Source code can be found [here](https://github.com/richtia/substrait-tools).
```
westonpace, line 6:
```suggestion
from different producers and consumers.
```

Let's leave the "source code can be found here" for pypi (as an aside, you can configure pypi to show a github link by using project links https://stackoverflow.com/questions/61156290/how-to-set-project-links-in-pypi).

westonpace (changes requested):
Ok, I built locally.  The pages seem a little sparse but that is probably fine for now.  Maybe in the future we can add a few usage examples for each one?

site/docs/tools/third_party_tools.md
richtia: Cool, just configured it.

westonpace (approved):

PR #449, opened 2023-02-23
feat: explicitly reference IEEE 754 and mandate precision as well as range

### LINKED ISSUE #447: Point to IEEE 754
The definition of the fb64 and fp32 types should be defined as and reference the actual IEEE-754 2008 spec, not to mutable Wikipedia articles about the spec. 

https://substrait.io/types/type_classes/#simple-types

westonpace:
Closes #447 

BREAKING CHANGE: The minimum precision for floating point numbers is
now mandated.


westonpace:
This is a breaking change only because we did not explicitly mandate precision before.  However, range for a floating point number is rather useless without some notion of precision.  I doubt this change will affect anyone since everyone is probably using IEEE 754 compliant types under the hood.

I'm also a little discouraged by the fact that there is no freely available link to the standard itself.  However, I imagine that most people will already know what they need to about IEEE 754 and, if they don't, they can discover the information pretty easily.

westonpace:
If I'm going to be strict this would be a breaking change to the spec and so it will require @jacques-n or @cpcloud 's approval

westonpace:
Ping @cpcloud @jacques-n this one should be fairly simple but we do need a second approval.

jacques-n (approved):

PR #454, opened 2023-03-01
chore: add a CODEOWNERS file

westonpace:


westonpace:
The proposed configuration will automatically request myself, @jacques-n, and @cpcloud for review whenever a new PR is created.  This seemed like a valid list given that the governance currently requires 1 of those three to review whenever there is a non-breaking change to the spec and requires 2 to review when the change is a breaking change.

Alternatively, we could narrow the scope so that it only requested reviews from all three when the change was in the `/proto` or `/site` directory.

Others with write permission and interest in reviewing would be free to add their own names to this list as they see fit.

westonpace:
I'm going to interpret silence as lazy consensus in this case.

westonpace:
Ah, looks like I can't approve my own PR.  Foiled.

mbrobbel (approved):

PR #461, opened 2023-03-03
docs: fetch operation has one input

vbarua:


westonpace (approved):
Thanks for the catch.  Looks like a copy-paste error.  @jacques-n / @cpcloud can one of you provide a quick sanity check since this is (technically) a backwards incompatible change to the spec (arguable since this is probably a typo).

EpsilonPrime:
Given that the proto is correct and the documentation is wrong, I'd argue that this is just a documentation error and not a backwards incompatible change which should make things easier.

westonpace:
True, but for the moment I'm going to err on the side of being pedantic here.  The site text (markdown) is "the spec" and takes precedence over proto files (this has come up once or twice in the past).

cpcloud (approved):

jacques-n (approved):
Thanks for the fetch fix

Issue #401, opened 2022-12-05
Add "skip nulls" property to aggregate function invocation


thisisnic:
Some backends (e.g. DuckDB) remove NULL values by default in computations involving scalar aggregate functions, others (e.g. Acero) allow specifying this option to return an NULL value if any of the input values are NULL.

In #388 (closed, not merged), I proposed adding this as an option to each of the scalar aggregate functions, but in the review, there was discussion recommending instead adding this as a property to the aggregate function invocation.

westonpace:
What should the default behavior be?  I think I'd prefer skipping nulls by default but will every function "skip".  I don't have any good counter-example at the moment (postgres' `array_agg` does not skip nulls but it's a bit of an odd aggregate function and doesn't exist in Substrait).

Default - No special behavior.  Nulls are skipped unless otherwise specified by the function.
Alternate - If any input value is null then the output value is null.

Also, I'm not super familiar with the rationale behind the null handling logic in Acero.  Do you know why this "emit null if any input is null" behavior was desired in the first place?  I think Acero might be a bit of an oddball here.

westonpace:
> If it's exceptionally niche and not something we'd want to support here, there are workarounds I can implement in the R Substrait producer (i.e. wrap the bindings to the scalar aggregate functions in further calls to other functions which first check if any results are NULL, and return either NULL or the calculated value depending on the outcome).

That would work.   We could also consider this a "physical optimization" in Acero and, if we recognize this pattern (collapse-to-null followed by aggregate function) we could collapse it into a single aggregate operator with skip_nulls=false.

That being said, I'd consider R/dplyr to be a separate "engine" and so now I suppose there are two engines that support this feature.  I'd be curious to hear what others think.

ianmcook:
The default should definitely be to skip nulls. That's what most engines do.

IMO it does seem worthwhile to expose an option for aggregates to emit null if the input contains any nulls. Besides R working this way by default (which is peculiar), the aggregate functions in pandas also offer this as an option (`skipna=False`). So it seems worthwhile to allow producers to easily represent this intended behavior in Substrait plans. It seems like a win in terms of expressiveness and explicitness.

ianmcook:
The default should definitely be to skip nulls. That's what most engines do.

[...]

PR #465, opened 2023-03-07
feat: adding SUM0 definition for aggregate functions

vibhatha:
This PR includes a resolution to https://github.com/substrait-io/substrait/issues/259

### DIFF
--- a/extensions/functions_arithmetic.yaml
+++ b/extensions/functions_arithmetic.yaml
@@ -923,7 +923,10 @@ aggregate_functions:
         intermediate: fp64?
         return: fp64?
   - name: "sum0"
-    description: Sum a set of values. The sum of zero elements yields zero.
+    description: >
+      Sum a set of values. The sum of zero elements yields zero.
+
+      Null values are ignored.
     impls:
       - args:
           - name: x
--- a/extensions/functions_arithmetic.yaml
+++ b/extensions/functions_arithmetic.yaml
@@ -936,7 +936,7 @@ aggregate_functions:
             values: [ SILENT, SATURATE, ERROR ]
         nullability: DECLARED_OUTPUT
         decomposable: MANY
-        intermediate: i64?
+        intermediate: i64
         return: i64
       - args:
           - name: x
@@ -946,7 +946,7 @@ aggregate_functions:
             values: [ SILENT, SATURATE, ERROR ]
         nullability: DECLARED_OUTPUT
         decomposable: MANY
-        intermediate: i64?
+        intermediate: i64
         return: i64
       - args:
           - name: x
@@ -956,7 +956,7 @@ aggregate_functions:
             values: [ SILENT, SATURATE, ERROR ]
         nullability: DECLARED_OUTPUT
         decomposable: MANY
-        intermediate: i64?
+        intermediate: i64
         return: i64
       - args:
           - name: x
@@ -966,7 +966,7 @@ aggregate_functions:
             values: [ SILENT, SATURATE, ERROR ]
         nullability: DECLARED_OUTPUT
         decomposable: MANY
-        intermediate: i64?
+        intermediate: i64
         return: i64
       - args:
           - name: x
@@ -976,7 +976,7 @@ aggregate_functions:
             values: [ SILENT, SATURATE, ERROR ]
         nullability: DECLARED_OUTPUT
         decomposable: MANY
-        intermediate: fp64?
+        intermediate: fp64
         return: fp64
       - args:
           - name: x
@@ -986,7 +986,7 @@ aggregate_functions:
             values: [ SILENT, SATURATE, ERROR ]
         nullability: DECLARED_OUTPUT
         decomposable: MANY
-        intermediate: fp64?
+        intermediate: fp64
         return: fp64
   - name: "avg"
     description: Average a set of values. For integral types, this truncates partial values.
--- a/extensions/functions_arithmetic.yaml
+++ b/extensions/functions_arithmetic.yaml
@@ -922,6 +922,69 @@ aggregate_functions:
         decomposable: MANY
         intermediate: fp64?
         return: fp64?
+  - name: "sum0"
+    description: Sum a set of values. The sum of zero elements yields zero.
+    impls:
+      - args:
+          - name: x
+            value: i8
+        options:
+          overflow:
+            values: [ SILENT, SATURATE, ERROR ]
+        nullability: DECLARED_OUTPUT
+        decomposable: MANY
+        intermediate: i64?
+        return: i64
+      - args:
+          - name: x
+            value: i16
+        options:
+          overflow:
+            values: [ SILENT, SATURATE, ERROR ]
+        nullability: DECLARED_OUTPUT
+        decomposable: MANY
+        intermediate: i64?
+        return: i64
+      - args:
+          - name: x
+            value: i32
+        options:
+          overflow:
+            values: [ SILENT, SATURATE, ERROR ]
+        nullability: DECLARED_OUTPUT
+        decomposable: MANY
+        intermediate: i64?
+        return: i64
+      - args:
+          - name: x
+            value: i64
+        options:
+          overflow:
+            values: [ SILENT, SATURATE, ERROR ]
+        nullability: DECLARED_OUTPUT
+        decomposable: MANY
+        intermediate: i64?
+        return: i64
+      - args:
+          - name: x
+            value: fp32
+        options:
+          overflow:
+            values: [ SILENT, SATURATE, ERROR ]
+        nullability: DECLARED_OUTPUT
+        decomposable: MANY
+        intermediate: fp64?
+        return: fp64
+      - args:
+          - name: x
+            value: fp64
+        options:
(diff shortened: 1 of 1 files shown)


extensions/functions_arithmetic.yaml
```diff
@@ -922,6 +922,69 @@ aggregate_functions:
...
+      - args:
+          - name: x
+            value: fp64
+        options:
+          overflow:
+            values: [ SILENT, SATURATE, ERROR ]
+        nullability: DECLARED_OUTPUT
+        decomposable: MANY
+        intermediate: fp64?
+        return: fp64
```
EpsilonPrime, line 987: Do we also want to support decimal?

extensions/functions_arithmetic.yaml
```diff
@@ -922,6 +922,69 @@ aggregate_functions:
         decomposable: MANY
         intermediate: fp64?
         return: fp64?
+  - name: "sum0"
```
EpsilonPrime, line 925: An analog could be the multiply function returning one for a set of zero elements.  Do we want to make a similar change?

extensions/functions_arithmetic.yaml
```diff
@@ -922,6 +922,69 @@ aggregate_functions:
...
+    description: Sum a set of values. The sum of zero elements yields zero.
```
EpsilonPrime, line 926: An alternative is to have one sum function with an option for how to handle the set of zero inputs.  Potential complications of course arise if we want to add an option after the fact but is that something we are interested in?

extensions/functions_arithmetic.yaml
vibhatha: Mmm… sum did support, and the functionality difference is the no input returning zero. I assumed decimals should also be there. 

cc @westonpace

extensions/functions_arithmetic.yaml
vibhatha: I think you’re correct. We should probably add that in another PR?

EpsilonPrime: Agreed, no need to make this PR do too much.

extensions/functions_arithmetic.yaml
vibhatha: I see your point. Let’s discuss further. 
Options is interesting and that could simplify things, but not sure how to represent nullability in return statements. 

cc @westonpace @jacques-n

extensions/functions_arithmetic.yaml
vibhatha: 👍

extensions/functions_arithmetic.yaml
westonpace: It's also pretty easy to get the behavior already with `COALESCE(SUM(x), 0)` (though this is not exactly the same because it would return 0 for the sum of all null input where this function would hopefully return null).  However, I don't think we need to shy away from functions that can be represented elsewhere.  So I think adding a new function or option is still a good idea.

I'm not sure where exactly to draw the line between "new function" and "option to an existing function".  I don't have a really strong opinion either way.

extensions/functions_arithmetic.yaml
westonpace: Yes.  We want to support decimal.

westonpace (changes requested):
Given:

```
sum0([1, 2, 3]) = 6
sum0([]) = 0
```

What does this result in:
```
sum0([null, null, null]) = ?
```

Is it `null` or `0`?  I think it should be `null`.  If you agree we should note that in the description.  In this case, I think the `return` attribute still needs to be `i64?`/`fp64?`.

westonpace:
sqlite has `sum` and `total`.  `total` is the sqlite equivalent of `sum0`.  Curiously, in sqlite, `total([null, null, null]) = 0` so perhaps my intuition is off :)

EpsilonPrime:
> sqlite has `sum` and `total`. `total` is the sqlite equivalent of `sum0`. Curiously, in sqlite, `total([null, null, null]) = 0` so perhaps my intuition is off :)

Ah, so total is more of "add all the values that are present" where sum is treating null as unknown.  (So for sum adding a known value to an unknown one is still unknown.)

westonpace:
Note that https://github.com/substrait-io/substrait/issues/401 is probably related

extensions/functions_arithmetic.yaml
wmalpica: Actually I think it would behave the same as `COALESCE(SUM(x), 0)` . It is my understanding that if you had a set of all null values and you do a SUM0() on them, you would get 0, not null.

This is a description I found for SUM0 that I found here https://www.tabnine.com/code/java/classes/org.apache.calcite.sql.fun.SqlSumEmptyIsZeroAggFunction
```
Sum0 is an aggregator which returns the sum of the values which go into it like Sum. It differs in that when no non null values are applied zero is returned instead of null. 
```

wmalpica:
> sqlite has `sum` and `total`. `total` is the sqlite equivalent of `sum0`. Curiously, in sqlite, `total([null, null, null]) = 0` so perhaps my intuition is off :)

Yes I think SUM0 would be like `total`

extensions/functions_arithmetic.yaml
westonpace: Then let's update the description to something like...

```
Sum a set of values. The sum of zero elements yields zero.  Null values are ignored.
```

We should probably clarify this for `sum` as well but that can be left for a future PR.

vibhatha: @westonpace And I think we should probably leave the return type without nulls. Is that okay?

westonpace: Yes, good catch.  Given this updated understanding I think the return type should be `DECALRED_OUTPUT` nullability and a non-nullable return type.

vibhatha: Got it. I will update the PR. Thanks @wmalpica for your feedback as well.

extensions/functions_arithmetic.yaml
```diff
@@ -922,6 +922,72 @@ aggregate_functions:
...
+    impls:
+      - args:
+          - name: x
+            value: i8
+        options:
+          overflow:
+            values: [ SILENT, SATURATE, ERROR ]
+        nullability: DECLARED_OUTPUT
+        decomposable: MANY
+        intermediate: i64?
```
westonpace, line 939:
```suggestion
        intermediate: i64
```
The intermediate is also non-nullable.

extensions/functions_arithmetic.yaml
```diff
@@ -922,6 +922,72 @@ aggregate_functions:
...
+        return: i64
+      - args:
+          - name: x
+            value: i16
+        options:
+          overflow:
+            values: [ SILENT, SATURATE, ERROR ]
+        nullability: DECLARED_OUTPUT
+        decomposable: MANY
+        intermediate: i64?
```
westonpace, line 949:
```suggestion
        intermediate: i64
```

extensions/functions_arithmetic.yaml
```diff
@@ -922,6 +922,72 @@ aggregate_functions:
...
+        return: i64
+      - args:
+          - name: x
+            value: i32
+        options:
+          overflow:
+            values: [ SILENT, SATURATE, ERROR ]
+        nullability: DECLARED_OUTPUT
+        decomposable: MANY
+        intermediate: i64?
```
westonpace, line 959:
```suggestion
        intermediate: i64
```

extensions/functions_arithmetic.yaml
```diff
@@ -922,6 +922,72 @@ aggregate_functions:
...
+        return: i64
+      - args:
+          - name: x
+            value: i64
+        options:
+          overflow:
+            values: [ SILENT, SATURATE, ERROR ]
+        nullability: DECLARED_OUTPUT
+        decomposable: MANY
+        intermediate: i64?
```
westonpace, line 969:
```suggestion
        intermediate: i64
```

extensions/functions_arithmetic.yaml
```diff
@@ -922,6 +922,72 @@ aggregate_functions:
...
+        return: i64
+      - args:
+          - name: x
+            value: fp32
+        options:
+          overflow:
+            values: [ SILENT, SATURATE, ERROR ]
+        nullability: DECLARED_OUTPUT
+        decomposable: MANY
+        intermediate: fp64?
```
westonpace, line 979:
```suggestion
        intermediate: fp64
```

extensions/functions_arithmetic.yaml
```diff
@@ -922,6 +922,72 @@ aggregate_functions:
...
+        return: fp64
+      - args:
+          - name: x
+            value: fp64
+        options:
+          overflow:
+            values: [ SILENT, SATURATE, ERROR ]
+        nullability: DECLARED_OUTPUT
+        decomposable: MANY
+        intermediate: fp64?
```
westonpace, line 989:
```suggestion
        intermediate: fp64
```

westonpace (commented):
One small change.

Issue #466, opened 2023-03-08
Add `regexp_match_substring_all` function


ianmcook:
The Substrait functions YAML defines the function `regexp_match_substring` which returns a scalar string value containing a single occurrence of the matched substring. Many systems implement this function. Those systems often name it `regexp_extract`.

Some systems also implement a variation of this function that extracts an array/list of _all_ the matched substrings. For example [Presto](https://prestodb.io/docs/current/functions/regexp.html) and [Spark](https://spark.apache.org/docs/latest/api/sql/index.html#regexp_extract_all) both implement `regexp_extract_all`. We should define a function `regexp_match_substring_all` in the Substrait functions YAML that we can map onto these.

richtia:
> @pdet @cpcloud @westonpace @richtia

I can work on adding this.

westonpace:
> We should define a function regexp_match_substring_all in the Substrait functions YAML that we can map onto these.

Do you mean `regexp_extract_all`?

ianmcook:
> Do you mean `regexp_extract_all`?

The function that is named `regexp_extract` in many SQL systems is named [`regexp_match_substring`](https://github.com/substrait-io/substrait/blob/7d7ddf11f3ce0b5f69a9d32ef10a699888f18a61/extensions/functions_string.yaml#L80) in Substrait.

So following that naming convention, `regexp_extract_all` would be named `regexp_match_substring_all` in Substrait, I think?

westonpace:
Thanks for the clarification. I agree.

PR #469, opened 2023-03-09
feat: add regexp_match_substring_all function to yaml

richtia:
BREAKING CHANGE: `group` argument added to `regexp_match_substring` function

Add regexp_match_substring_all function

Resolves https://github.com/substrait-io/substrait/issues/466

### DIFF
--- a/extensions/functions_string.yaml
+++ b/extensions/functions_string.yaml
@@ -163,8 +163,8 @@ scalar_functions:
       beginning and end of the input string. Enabling the `dotall` option makes the `.` character
       match line terminator characters in a string.
 
-      Behavior is undefined if the regex fails to compile, the occurrence value is out of range,
-      the position value is out of range, or the group value is out of range.
+      Behavior is undefined if the regex fails to compile, the position value is out of range,
+      or the group value is out of range.
     impls:
       - args:
           - value: "varchar<L1>"
--- a/extensions/functions_string.yaml
+++ b/extensions/functions_string.yaml
@@ -88,7 +88,11 @@ scalar_functions:
       from the beginning of the string to begin starting to search for pattern matches can be
       specified using the `position` argument. Specifying `1` means to search for matches
       starting at the first character of the input string, `2` means the second character, and so
-      on. The `position` argument should be a positive non-zero integer.
+      on. The `position` argument should be a positive non-zero integer. The regular
+      expression capture group can be specified using the `group` argument. Specifying `0`
+      will return the substring matching the full regular expression. Specifying `1` will
+      return the substring matching only the first capture group, and so on. The `group`
+      argument should be a non-negative integer.
 
       The `case_sensitivity` option specifies case-sensitive or case-insensitive matching.
       Enabling the `multiline` option will treat the input string as multiple lines. This makes
@@ -96,8 +100,8 @@ scalar_functions:
       beginning and end of the input string. Enabling the `dotall` option makes the `.` character
       match line terminator characters in a string.
 
-      Behavior is undefined if the regex fails to compile, the occurrence value is out of range, or
-      the position value is out of range.
+      Behavior is undefined if the regex fails to compile, the occurrence value is out of range,
+      the position value is out of range, or the group value is out of range.
     impls:
       - args:
           - value: "varchar<L1>"
@@ -108,6 +112,8 @@ scalar_functions:
             name: "position"
           - value: i64
             name: "occurrence"
+          - value: i64
+            name: "group"
         options:
           case_sensitivity:
             values: [ CASE_SENSITIVE, CASE_INSENSITIVE, CASE_INSENSITIVE_ASCII ]
@@ -125,6 +131,8 @@ scalar_functions:
             name: "position"
           - value: i64
             name: "occurrence"
+          - value: i64
+            name: "group"
         options:
           case_sensitivity:
             values: [ CASE_SENSITIVE, CASE_INSENSITIVE, CASE_INSENSITIVE_ASCII ]
@@ -133,6 +141,65 @@ scalar_functions:
           dotall:
             values: [ DOTALL_DISABLED, DOTALL_ENABLED ]
         return: "string"
+  -
+    name: regexp_match_substring_all
+    description: >-
+      Extract all substrings that match the given regular expression pattern. This will return a
+      list of extracted strings with one value for each occurrence of a match. The regular expression
+      pattern should follow the International Components for Unicode implementation
+      (https://unicode-org.github.io/icu/userguide/strings/regexp.html). The number of characters
+      from the beginning of the string to begin starting to search for pattern matches can be
+      specified using the `position` argument. Specifying `1` means to search for matches
+      starting at the first character of the input string, `2` means the second character, and so
+      on. The `position` argument should be a positive non-zero integer. The regular
+      expression capture group can be specified using the `group` argument. Specifying `0`
+      will return substrings matching the full regular expression. Specifying `1` will return
+      substrings matching only the first capture group, and so on. The `group` argument should
+      be a non-negative integer.
+
+      The `case_sensitivity` option specifies case-sensitive or case-insensitive matching.
+      Enabling the `multiline` option will treat the input string as multiple lines. This makes
+      the `^` and `$` characters match at the beginning and end of any line, instead of just the
+      beginning and end of the input string. Enabling the `dotall` option makes the `.` character
+      match line terminator characters in a string.
+
+      Behavior is undefined if the regex fails to compile, the occurrence value is out of range,
+      the position value is out of range, or the group value is out of range.
+    impls:
+      - args:
+          - value: "varchar<L1>"
+            name: "input"
+          - value: "varchar<L2>"
+            name: "pattern"
+          - value: i64
+            name: "position"
+          - value: i64
+            name: "group"
+        options:
+          case_sensitivity:
+            values: [ CASE_SENSITIVE, CASE_INSENSITIVE, CASE_INSENSITIVE_ASCII ]
+          multiline:
+            values: [ MULTILINE_DISABLED, MULTILINE_ENABLED ]
+          dotall:
+            values: [ DOTALL_DISABLED, DOTALL_ENABLED ]
+        return: "List<varchar<L1>>"
+      - args:
+          - value: "string"
+            name: "input"
+          - value: "string"
+            name: "pattern"
+          - value: i64
+            name: "position"
+          - value: i64
+            name: "group"
+        options:
+          case_sensitivity:
+            values: [ CASE_SENSITIVE, CASE_INSENSITIVE, CASE_INSENSITIVE_ASCII ]
+          multiline:
+            values: [ MULTILINE_DISABLED, MULTILINE_ENABLED ]
+          dotall:
+            values: [ DOTALL_DISABLED, DOTALL_ENABLED ]
+        return: "List<string>"
   -
     name: starts_with
     description: >-


richtia:
@ianmcook

extensions/functions_string.yaml
```diff
@@ -133,6 +133,55 @@ scalar_functions:
           dotall:
             values: [ DOTALL_DISABLED, DOTALL_ENABLED ]
         return: "string"
+  -
+    name: regexp_match_substring_all
+    description: >-
+      Extract all substrings that match the given regular expression pattern. The regular expression
+      pattern should follow the International Components for Unicode implementation
```
EpsilonPrime, line 140: This standard includes negative lookahead assertions which is known to be quite slow.  To combat this Google wrote https://github.com/google/re2.  It's not quite an option but it definitely is a behavior that a dialect would have so I'm not sure we can code it but we could add something to the description so that users might be aware of the possibility.

ianmcook:
The existing implementation of [`regexp_match_substring`](https://github.com/substrait-io/substrait/blob/7d7ddf11f3ce0b5f69a9d32ef10a699888f18a61/extensions/functions_string.yaml#L80) takes the arguments `input`, `pattern`, `position`, `occurrence`.

The implementation of `regexp_match_substring_all` proposed here takes the arguments `input`, `pattern`, `group`.

I understand why there is no `occurrence` argument here. (Because it returns a list of all occurrences.)

But I do not understand:
- Why is there no `group` argument in `regexp_match_substring`?
- Why is there no `position` argument in `regexp_match_substring_all`?

richtia:
> The existing implementation of [`regexp_match_substring`](https://github.com/substrait-io/substrait/blob/7d7ddf11f3ce0b5f69a9d32ef10a699888f18a61/extensions/functions_string.yaml#L80) takes the arguments `input`, `pattern`, `position`, `occurrence`.
> 
> The implementation of `regexp_match_substring_all` proposed here takes the arguments `input`, `pattern`, `group`.
> 
> I understand why there is no `occurrence` argument here. (Because it returns a list of all occurrences.)
> 
> But I do not understand:
> 
> * Why is there no `group` argument in `regexp_match_substring`?

I was wondering the same thing when I submitted this PR.  I think this was the case because I was looking at other functions named `REGEXP_SUBSTR` and those didn't have the group argument (MySQL, Oracle).  I can add the group parameter to the function in this PR though.

> * Why is there no `position` argument in `regexp_match_substring_all`?

I wasn't sure whether or not to put the `position` argument for this function, since I felt like it defeated the purpose of the `all` part.  I don't have a strong opinion either way though.  I'm fine with adding it back into this function as well.

ianmcook:
> I wasn't sure whether or not to put the `position` argument for this function, since I felt like it defeated the purpose of the `all` part. I don't have a strong opinion either way though. I'm fine with adding it back into this function as well.

IIUC, the `position` argument is still relevant here, even with `all`. The `occurrence` argument is the one that is not needed with `all`.

I think we should aim for these two functions `regexp_match_substring` and ` regexp_match_substring_all` and their documentation to be identical in all respects except that the former returns one occurrence and the later returns all occurrences.

richtia:
> IIUC, the `position` argument is still relevant here, even with `all`. The `occurrence` argument is the one that is not needed with `all`.
> 
> I think we should aim for these two functions `regexp_match_substring` and ` regexp_match_substring_all` and their documentation to be identical in all respects except that the former returns one occurrence and the later returns all occurrences.

Sounds good, I've updated both.

extensions/functions_string.yaml
```diff
@@ -88,7 +88,10 @@ scalar_functions:
       from the beginning of the string to begin starting to search for pattern matches can be
       specified using the `position` argument. Specifying `1` means to search for matches
       starting at the first character of the input string, `2` means the second character, and so
-      on. The `position` argument should be a positive non-zero integer.
+      on. The `position` argument should be a positive non-zero integer. The regular
+      expression capture group can be specified using the `group` argument. Specifying `0`
+      will match the entire regular expression.  The default group value is `1`, which means the
+      first capture group will be matched. The `group` argument should be a non-negative integer.
```
ianmcook, line 94: @richtia thank you for adding this `group` argument to the existing `regexp_match_substring` function. This creates better correspondence between this existing function and the new `regexp_match_substring_all` function added here. But to be clear, it's not only for the sake of completeness. Multiple real-world functions that extract a single regex match—including R's [`stringr::str_extract()`](https://stringr.tidyverse.org/reference/str_extract.html) and Snowflake's with [`REGEXP_SUBSTR`](https://docs.snowflake.com/en/sql-reference/functions/regexp_substr)—accept an optional group number argument which allows the user to specify which capture group to return. So this is a practical addition.

Is it considered a breaking change when we add a new argument like this? I forget whether or not Substrait technically allows optional arguments / arguments with default values.

extensions/functions_string.yaml
```diff
@@ -133,6 +140,63 @@ scalar_functions:
...
+      Extract all substrings that match the given regular expression pattern. The regular expression
+      pattern should follow the International Components for Unicode implementation
+      (https://unicode-org.github.io/icu/userguide/strings/regexp.html). The number of characters
+      from the beginning of the string to begin starting to search for pattern matches can be
+      specified using the `position` argument. Specifying `1` means to search for matches
+      starting at the first character of the input string, `2` means the second character, and so
+      on. The `position` argument should be a positive non-zero integer. The regular
+      expression capture group can be specified using the `group` argument. Specifying `0`
+      will match the entire regular expression.  The default group value is `1`, which means the
+      first capture group will be matched. The `group` argument should be a non-negative integer.
```
ianmcook, line 155: This is clearer, I think. It also removes the mention of a "default" value for `group`, which was confusing.
```suggestion
      will return substrings matching the full regular expression. Specifying `1` will return
      substrings matching only the first capture group, and so on. The `group` argument
      should be a non-negative integer.
```

extensions/functions_string.yaml
ianmcook, line 94:
```suggestion
      will return the substring matching the full regular expression. Specifying `1` will
      return the substring matching only the first capture group, and so on. The `group`
      argument should be a non-negative integer.
```

richtia: I think it would be considered a breaking change.  I'm not 100% sure though.  @westonpace

extensions/functions_string.yaml
```diff
@@ -88,7 +88,11 @@ scalar_functions:
       from the beginning of the string to begin starting to search for pattern matches can be
       specified using the `position` argument. Specifying `1` means to search for matches
       starting at the first character of the input string, `2` means the second character, and so
-      on. The `position` argument should be a positive non-zero integer.
+      on. The `position` argument should be a positive non-zero integer. The regular
+      expression capture group can be specified using the `group` argument. Specifying `0`
```
westonpace, line 92: I'm confused (never surprising :smile:  ).  What is the difference is between `group` and `occurrence`?

extensions/functions_string.yaml
```diff
@@ -88,7 +88,11 @@ scalar_functions:
       from the beginning of the string to begin starting to search for pattern matches can be
       specified using the `position` argument. Specifying `1` means to search for matches
       starting at the first character of the input string, `2` means the second character, and so
-      on. The `position` argument should be a positive non-zero integer.
+      on. The `position` argument should be a positive non-zero integer. The regular
```
westonpace, line 91: Not your fault but why does `position` use "1-based" indices?

extensions/functions_string.yaml
westonpace: Why wouldn't regex dialect be an option?

extensions/functions_string.yaml
```diff
@@ -133,6 +141,64 @@ scalar_functions:
...
+      be a non-negative integer.
+
+      The `case_sensitivity` option specifies case-sensitive or case-insensitive matching.
+      Enabling the `multiline` option will treat the input string as multiple lines. This makes
+      the `^` and `$` characters match at the beginning and end of any line, instead of just the
+      beginning and end of the input string. Enabling the `dotall` option makes the `.` character
+      match line terminator characters in a string.
+
+      Behavior is undefined if the regex fails to compile, the occurrence value is out of range, or
+      the position value is out of range.
```
westonpace, line 166: Why are these things undefined behaviors and not errors?

extensions/functions_string.yaml
```diff
@@ -133,6 +141,64 @@ scalar_functions:
...
+      pattern should follow the International Components for Unicode implementation
+      (https://unicode-org.github.io/icu/userguide/strings/regexp.html). The number of characters
+      from the beginning of the string to begin starting to search for pattern matches can be
+      specified using the `position` argument. Specifying `1` means to search for matches
+      starting at the first character of the input string, `2` means the second character, and so
+      on. The `position` argument should be a positive non-zero integer. The regular
+      expression capture group can be specified using the `group` argument. Specifying `0`
+      will return substrings matching the full regular expression. Specifying `1` will return
+      substrings matching only the first capture group, and so on. The `group` argument should
+      be a non-negative integer.
```
westonpace, line 157: Why do I need to specify a `group` if it is `match_..._all`?  Wouldn't that return a value for each capture group?

westonpace (commented):
I'm not sure I understand yet but I haven't looked at the regex functions before so it's likely just ignorance on my part.

extensions/functions_string.yaml
westonpace: Yes, adding a new positional argument is a breaking change (adding a new option is not).

extensions/functions_string.yaml
westonpace: Ok.  I think I see now.  If I match the string `(0-50),(0-70)` with the pattern `(\d+)-(\d+)` then there are two occurrences `0-50` and `0-70` and two capture groups (the first matches `0` and `0` and the second matches `50` and `70`).

extensions/functions_string.yaml
westonpace: I understand the answer now.  However, can we clarify in the description that the "all" is referring to "occurrences" and not "groups"?  Maybe something like:

Extract all substrings that match the given regular expression pattern.  This will return a list of extracted strings with one value for each occurrence of a match.

extensions/functions_string.yaml
ianmcook: SQL systems tend to use 1-based indices for positions in strings, [as explained here](https://dba.stackexchange.com/questions/125950/why-do-string-functions-on-sql-based-database-servers-start-at-position-1-instea). Looking at all the other functions already defined in [functions_string.yaml](https://github.com/substrait-io/substrait/blob/main/extensions/functions_string.yaml), it looks like we've chosen to follow this convention. I'm not sure whether this was an explicit choice, but regardless, it's probably beyond the scope of this PR to discuss it.

extensions/functions_string.yaml
richtia: International Components for Unicode (ICU) has a regex dialect standard that a some other SQL dialects follow.  The description includes a link to the standard.  I originally went with that in [this PR](https://github.com/substrait-io/substrait/pull/281).  I think MySQL and IBM follow it.  I supposed the regex dialect may not matter so long as the producer is able to generate a substrait plan though?

richtia: I can open up a separate issue to discuss this.  The other existing regex functions state this as well, and I was just continuing with that.

extensions/functions_string.yaml
```diff
@@ -88,7 +88,11 @@ scalar_functions:
       from the beginning of the string to begin starting to search for pattern matches can be
       specified using the `position` argument. Specifying `1` means to search for matches
       starting at the first character of the input string, `2` means the second character, and so
-      on. The `position` argument should be a positive non-zero integer.
+      on. The `position` argument should be a positive non-zero integer. The regular
+      expression capture group can be specified using the `group` argument. Specifying `0`
+      will return the substring matching the full regular expression. Specifying `1` will
+      return the substring matching only the first capture group, and so on. The `group`
+      argument should be a non-negative integer.
```
westonpace, line 95: Does 0 for `group` still mean "the whole expression" here?

westonpace (changes requested):
One new question and there is one outstanding (related to undefined behaviors).  I've resolved the rest of my comments for clarity.

extensions/functions_string.yaml
ianmcook: FWIW the existing regexp functions in [functions_string.yaml](https://github.com/substrait-io/substrait/blob/main/extensions/functions_string.yaml) use this same kind of boilerplate "Behavior is undefined" language.

extensions/functions_string.yaml
```diff
@@ -133,6 +141,65 @@ scalar_functions:
...
+      be a non-negative integer.
+
+      The `case_sensitivity` option specifies case-sensitive or case-insensitive matching.
+      Enabling the `multiline` option will treat the input string as multiple lines. This makes
+      the `^` and `$` characters match at the beginning and end of any line, instead of just the
+      beginning and end of the input string. Enabling the `dotall` option makes the `.` character
+      match line terminator characters in a string.
+
+      Behavior is undefined if the regex fails to compile, the occurrence value is out of range, or
+      the position value is out of range.
```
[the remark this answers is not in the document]
ianmcook: @richtia could you please also change the "Behavior is undefined" sentence for the `regexp_match_substring` function to mention the group value?

extensions/functions_string.yaml
richtia: Yeah, I think that's just the decision we made during the original PR.  But if erroring out is preferred, we can go that route too.

extensions/functions_string.yaml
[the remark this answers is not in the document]
richtia: Done!

ianmcook:
> BREAKING CHANGE: adding a new positional argument

@richtia can you please change the PR description to say:

BREAKING CHANGE: `group` argument added to `regexp_match_substring` function

extensions/functions_string.yaml
```diff
@@ -133,6 +141,65 @@ scalar_functions:
...
+      be a non-negative integer.
+
+      The `case_sensitivity` option specifies case-sensitive or case-insensitive matching.
+      Enabling the `multiline` option will treat the input string as multiple lines. This makes
+      the `^` and `$` characters match at the beginning and end of any line, instead of just the
+      beginning and end of the input string. Enabling the `dotall` option makes the `.` character
+      match line terminator characters in a string.
+
+      Behavior is undefined if the regex fails to compile, the occurrence value is out of range,
+      the position value is out of range, or the group value is out of range.
```
ianmcook, line 167: Oops, there is no occurrence value for `regexp_match_substring_all`
```suggestion
      Behavior is undefined if the regex fails to compile, the position value is out of range,
      or the group value is out of range.
```

richtia: Oops.  Fixed!

westonpace (approved):
I'm good.  @ianmcook let me know when you're content.

ianmcook:
It looks good to me

westonpace:
+1

Since this is now a breaking extension function change (adding new positional arguments is always a breaking change) it requires two SMC votes.

@cpcloud / @jacques-n ?

cpcloud (approved):
LGTM!

Issue #487, opened 2023-04-04
strptime function definition violates spec


vbarua:
# Context
While upgrading substrait-java to the latest version of the spec, tests failed due to multiple scalar functions being added w/ the same type signature.

As defined, `strptime` has the following signature.
https://github.com/substrait-io/substrait/blob/e8ced2a8df89a2fc635b3e13b6b5cd84ef9306af/extensions/functions_datetime.yaml#L405-L442

```
* strptime:str_str_str
* strptime:str_str
* strptime:str_str
* strptime:str_str
```

# Problem
The addition of the the `strptime` function in https://github.com/substrait-io/substrait/pull/272 violates the spec. 

Specifically

>It is required that two function implementations with the same simple name must resolve to different compound names using types. If two function implementations in a YAML file resolve to the same compound name, the YAML file is invalid and behavior is undefined.

https://substrait.io/extensions/#function-signature-compound-names

rok:
Oh I missed this. How about: `strptime` -> `strptime`, `strptime_date`, `strptime_time`?

richtia:
> Oh I missed this. How about: `strptime` -> `strptime_timestamp`, `strptime_date`, `strptime_time`?

This is what I was thinking as well...separating the function into 3 separate ones.  I'm curious how producers/consumers will handle this with regards to mapping though.

rok:
> I'm curious how producers/consumers will handle this with regards to mapping though.

What would you expect to be an issue there?

richtia:
@rok I went ahead and submitted a PR for this since I have a few other PRs waiting for the fix.

rok:
Thanks @richtia !

richtia:
> Thanks @richtia !

Np! Do you mind giving it a quick look? https://github.com/substrait-io/substrait/pull/493

PR #272, opened 2022-07-29
feat: add temporal functions

rok:
This adds new scalar functions to work with timestamp/date/time types.
See #222 for previous discussion.

PR #493, opened 2023-04-10
fix: separate strptime to fix spec violation

richtia:
Fix spec violation described in https://github.com/substrait-io/substrait/issues/487 by
separating `strptime` into `strptime_timestamp`, `strptime_date`, and `strptime_time`

Here's the original PR adding this in case anyone wants to see the discussion
surrounding this addition: https://github.com/substrait-io/substrait/pull/272

### DIFF
--- a/extensions/functions_datetime.yaml
+++ b/extensions/functions_datetime.yaml
@@ -402,9 +402,33 @@ scalar_functions:
             value: string
         return: timestamp
   -
-    name: "strptime"
+    name: "strptime_time"
     description: >-
-      Parse string into timestamp/date/time using provided format,
+      Parse string into time using provided format,
+      see https://man7.org/linux/man-pages/man3/strptime.3.html for reference.
+    impls:
+      - args:
+          - name: time_string
+            value: string
+          - name: format
+            value: string
+        return: time
+  -
+    name: "strptime_date"
+    description: >-
+      Parse string into date using provided format,
+      see https://man7.org/linux/man-pages/man3/strptime.3.html for reference.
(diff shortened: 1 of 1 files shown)


richtia:
> **ACTION NEEDED**
> 
> Substrait follows the [Conventional Commits specification](https://www.conventionalcommits.org/en/v1.0.0/) for release automation.
> 
> The PR title and description are used as the merge commit message. Please update your PR title and description to match the specification.

Hmm...not actually sure why i'm seeing this?

rok:
> Hmm...not actually sure why i'm seeing this?

Maybe try `fix(extension): separate stptime to fix spec violation`?

rok (commented):
What about when timezone is present when parsing time and date? Would we want to return those in UTC or local time?

richtia:
> What about when timezone is present when parsing time and date? Would we want to return those in UTC or local time?

Good question.  Do you think it would make sense to add another impl to `strptime_time` and `strptime_date` that has timezone as a parameter and behave similarly to `strptime_timestamp`?

westonpace (approved):
I agree this is a good change and that impls cannot differ only by return type.  We were violating this rule:

> It is required that two function implementations with the same simple name must resolve to different compound names using types. If two function implementations in a YAML file resolve to the same compound name, the YAML file is invalid and behavior is undefined.


Given the previous version was not valid I don't think this will count as a backwards incompatible change and we should merge it so we can get moving forwards.

I'll CC @jacques-n and @cpcloud for a posthumous review if/when they have time.

PR #494, opened 2023-04-13
fix: referenced simple extension in tutorial (set instead of string)

mbrobbel:
I noticed that the tutorial has `functions_string.yaml` and `functions_set.yaml`
mixed up for the `index_in` extension function. This PR fixes that.


westonpace (approved):
Good catch

PR #495, opened 2023-04-13
fix(text)!: mark `name` property of `type` extension item as required

mbrobbel:
The spec on simple extensions
[states](https://substrait.io/extensions/#simple-extensions):
```
In the places where these entities are referenced, they will be referenced using a URI + name
reference.
```
And for type:
```
Type  | The name as defined on the type object.
```

However this `name` field is not marked as required on the type object
definition in the schema. This PR marks this field as required to make sure it
can always be referenced.

### DIFF
--- a/text/simple_extensions_schema.yaml
+++ b/text/simple_extensions_schema.yaml
@@ -10,7 +10,7 @@ properties:
     items:
       type: object
       additionalProperties: false
-      required: [name, structure]
+      required: [name]
       properties:
         name:
           type: string
--- a/text/simple_extensions_schema.yaml
+++ b/text/simple_extensions_schema.yaml
@@ -10,6 +10,7 @@ properties:
     items:
       type: object
       additionalProperties: false
+      required: [name]
       properties:
         name:
           type: string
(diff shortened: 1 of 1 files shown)


text/simple_extensions_schema.yaml
```diff
@@ -10,6 +10,7 @@ properties:
     items:
       type: object
       additionalProperties: false
+      required: [name]
```
westonpace, line 13: Wouldn't `structure` also be required?

westonpace: Got it.  I found this paragraph on the [site](https://substrait.io/types/type_classes/#user-defined-types):

> The structure field is optional. If not specified, the type class is considered to be fully opaque. This implies that a systems without built-in support for the type cannot manipulate values in any way, including moving and cloning. This may be useful for exotic, context-sensitive types, such as raw pointers or identifiers that cannot be cloned.

Sorry for the goose chase

text/simple_extensions_schema.yaml
```diff
@@ -10,6 +10,7 @@ properties:
     items:
       type: object
       additionalProperties: false
+      required: [name, structure]
```
westonpace:
```suggestion
      required: [name]
```

westonpace (approved):

PR #498, opened 2023-05-08
feat: add optional ordering requirement for aggregate relation

mustafasrepo:
BREAKING CHANGE: Optional Ordering field is added

### DIFF
--- a/proto/substrait/algebra.proto
+++ b/proto/substrait/algebra.proto
@@ -227,7 +227,7 @@ message AggregateRel {
     Expression filter = 2;
     // An optional sort expression that acts to define required ordering for the aggregator.
     // Helps to support ARRAY_AGG(<c> ORDER BY <C2>) syntax without masking opportunities for optimization
-    Expression order_by = 3;
+    repeated SortField order_by = 3;
   }
 }
 
--- a/proto/substrait/algebra.proto
+++ b/proto/substrait/algebra.proto
@@ -225,7 +225,8 @@ message AggregateRel {
     // within the measure.
     // Helps to support SUM(<c>) FILTER(WHERE...) syntax without masking opportunities for optimization
     Expression filter = 2;
-    // An optional sort expression that acts to define required ordering for the aggregator.
+    // An optional sort expression that acts to define required ordering
+    // (e.g the order that the aggregator expects its input to have) for the aggregator.
     // Helps to support ARRAY_AGG(<c> ORDER BY <C2>) syntax without masking opportunities for optimization
     repeated SortField order_by = 3;
   }
--- a/proto/substrait/algebra.proto
+++ b/proto/substrait/algebra.proto
@@ -208,7 +208,7 @@ message AggregateRel {
   // A list of expression grouping that the aggregation measured should be calculated for.
   repeated Grouping groupings = 3;
 
-  // A list of one or more aggregate expressions along with an optional filter.
+  // A list of one or more aggregate expressions along with an optional filter, and optional ordering.
   repeated Measure measures = 4;
 
   substrait.extensions.AdvancedExtension advanced_extension = 10;
@@ -225,6 +225,9 @@ message AggregateRel {
     // within the measure.
     // Helps to support SUM(<c>) FILTER(WHERE...) syntax without masking opportunities for optimization
     Expression filter = 2;
+    // An optional sort expression that acts to define required ordering for the aggregator.
+    // Helps to support ARRAY_AGG(<c> ORDER BY <C2>) syntax without masking opportunities for optimization
+    Expression order_by = 3;
   }
 }
 
--- a/site/docs/relations/logical_relations.md
+++ b/site/docs/relations/logical_relations.md
@@ -323,12 +323,12 @@ If at least one grouping expression is present, the aggregation is allowed to no
 
 ### Aggregate Properties
 
-| Property         | Description                                                  | Required                                |
-| ---------------- | ------------------------------------------------------------ | --------------------------------------- |
-| Input            | The relational input.                                        | Required                                |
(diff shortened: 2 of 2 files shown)


proto/substrait/algebra.proto
```diff
@@ -225,6 +225,9 @@ message AggregateRel {
     // within the measure.
     // Helps to support SUM(<c>) FILTER(WHERE...) syntax without masking opportunities for optimization
     Expression filter = 2;
+    // An optional sort expression that acts to define required ordering for the aggregator.
+    // Helps to support ARRAY_AGG(<c> ORDER BY <C2>) syntax without masking opportunities for optimization
+    Expression order_by = 3;
```
westonpace, line 230: We don't use a plain `Expression` to define ordering elsewhere.  For example, see the `SortRel` which uses `repeated SortField`.  Is there a particular reason you are using `Expression` here?

westonpace (changes requested):
This makes sense to me.  Could you also update https://github.com/substrait-io/substrait/blob/main/site/docs/relations/logical_relations.md ?

I think this also covers syntax like `WITHIN GROUP (ORDER BY ...)` that some SQL engines seem to prefer.  Support seems a bit mixed across engines but I find general enough support to consider this a real feature and not something specific to a single db.

@jacques-n / @cpcloud thoughts?

proto/substrait/algebra.proto
mustafasrepo: This makes sense to me, also it is similar to the pattern in `WindowFunction`. Also using `repeated` enables us to support lexicographical requirement. I have converted `Expression order_by` to `repeated SortField order_by`. Thanks for the suggestion.

mustafasrepo:
By the way, I have updated document also, as you suggested.

proto/substrait/algebra.proto
```diff
@@ -208,7 +208,7 @@ message AggregateRel {
   // A list of expression grouping that the aggregation measured should be calculated for.
   repeated Grouping groupings = 3;
 
-  // A list of one or more aggregate expressions along with an optional filter.
+  // A list of one or more aggregate expressions along with an optional filter, and optional ordering.
```
EpsilonPrime, line 211: The ordering isn't part of the measures is it?

proto/substrait/algebra.proto
```diff
@@ -225,6 +225,9 @@ message AggregateRel {
     // within the measure.
     // Helps to support SUM(<c>) FILTER(WHERE...) syntax without masking opportunities for optimization
     Expression filter = 2;
+    // An optional sort expression that acts to define required ordering for the aggregator.
```
EpsilonPrime, line 228: Required ordering?  Does that mean the order that the aggregator expects its input to have or how it is expected to provide its output?

site/docs/relations/logical_relations.md
```diff
@@ -323,12 +323,12 @@ If at least one grouping expression is present, the aggregation is allowed to no
...
-| Measures         | A list of one or more aggregate expressions along with an optional filter. | Optional, required if no grouping sets. |
+| Property         | Description                                                                                                   | Required                                |
+| ---------------- |---------------------------------------------------------------------------------------------------------------| --------------------------------------- |
+| Input            | The relational input.                                                                                         | Required                                |
+| Grouping Sets    | One or more grouping sets.                                                                                    | Optional, required if no measures.      |
+| Per Grouping Set | A list of expression grouping that the aggregation measured should be calculated for.                         | Optional.      |
+| Measures         | A list of one or more aggregate expressions along with an optional filter, and optional ordering requirement. | Optional, required if no grouping sets. |
```
EpsilonPrime, line 331: As above, the ordering isn't part of the measures.

proto/substrait/algebra.proto
mustafasrepo: Actually, it is part of the measure similar to the filter.

mustafasrepo: Since `AggregateRel` stores `repeated Measure measures` for each aggregate expression. Adding a new measure type would require us to store another `repeated Measure2 measures2`. I think, it is a bit confusing. However, it is an option also. I am not sure which approach is the right one. What do you think?

mustafasrepo:
By the way @westonpace I have done the changes you have requested.

proto/substrait/algebra.proto
westonpace: I think `measures` is the correct place for this.  There is one measure per output column.  Each output column could expect a different ordering for its inputs.  This is consistent for how this is expressed in SQL.

westonpace (approved):
This looks correct to me.  @jacques-n  / @cpcloud can we get a second approval here?

ozankabak:
Any progress on this? It'd be great if someone can give a quick look so we can get this in.

westonpace:
@mustafasrepo @ozankabak we discussed this today in the community meeting.  It was brought to my attention that we already have ordering specified in the AggregateFunction message itself.  Will that work for you?  Or is there some reason this is different?

The part I am thinking of is here:

https://github.com/substrait-io/substrait/blob/v0.30.0/proto/substrait/algebra.proto#L1265-L1269

mustafasrepo:
> @mustafasrepo @ozankabak we discussed this today in the community meeting. It was brought to my attention that we already have ordering specified in the AggregateFunction message itself. Will that work for you? Or is there some reason this is different?
> 
> The part I am thinking of is here:
> 
> https://github.com/substrait-io/substrait/blob/v0.30.0/proto/substrait/algebra.proto#L1265-L1269

I wasn't aware of this support. I think this will work, thanks for pointing this out. I will let you know, either way whether it is sufficient or not for our use case.

EpsilonPrime:
Yay!  Should we close this PR and reopen if needed?

PR #499, opened 2023-05-10
build: update buf plugin references in buf.gen.yaml

mfridman:
Update buf plugin references in the generation template.

### DIFF
--- a/buf.gen.yaml
+++ b/buf.gen.yaml
@@ -1,14 +1,14 @@
 version: v1
 plugins:
-  - remote: buf.build/protocolbuffers/plugins/cpp:v3.20.0-1
+  - plugin: buf.build/protocolbuffers/cpp:v23.0
     out: gen/proto/cpp
-  - remote: buf.build/protocolbuffers/plugins/csharp:v3.20.0-1
+  - plugin: buf.build/protocolbuffers/csharp:v23.0
     out: gen/proto/csharp
-  - remote: buf.build/protocolbuffers/plugins/java:v3.20.0-1
+  - plugin: buf.build/protocolbuffers/java:v23.0
     out: gen/proto/java
-  - remote: buf.build/protocolbuffers/plugins/python:v3.20.0-1
+  - plugin: buf.build/protocolbuffers/python:v23.0
     out: gen/proto/python
-  - remote: buf.build/protocolbuffers/plugins/go:v1.28.0-1
+  - plugin: buf.build/protocolbuffers/go:v1.30.0
     out: gen/proto/go
     opt:
       - paths=source_relative


buf.gen.yaml
```diff
@@ -1,14 +1,14 @@
 version: v1
 plugins:
-  - remote: buf.build/protocolbuffers/plugins/cpp:v3.20.0-1
+  - plugin: buf.build/protocolbuffers/cpp:v23.0
```
mfridman, line 3: Newer versions of the `buf` CLI use the `plugin` key for both local and remote plugin references.

mfridman:
Hey there, I noticed you were referencing old Remote Plugins in the `buf.gen.yaml` file. There are some upcoming [Remote Plugin changes](https://buf.build/blog/remote-packages-remote-plugins-approaching-v1) and this change ensures you're not affected.

In summary, update the plugin references to the ones managed by @bufbuild.

I've also bumped the versions to the latest available upstream, although that may not be what you want.

All public plugins are now available at [buf.build/plugins](https://buf.build/plugins)

mbrobbel (approved):

gforsyth:
Thanks @mfridman !  We do want to make sure that `buf` keeps working -- I don't think that we want to make changes to the plugin versions, though.

westonpace:
> I don't think that we want to make changes to the plugin versions, though.

@gforsyth 

Why not?  Any particular reason other than the "change one thing at a time" concern?

mfridman:
The plugin version has parity with the upstream version. E.g., here is the raw source for the dockerized Python plugins:

https://github.com/bufbuild/plugins/tree/main/plugins/protocolbuffers/python

Which get displayed here: https://buf.build/protocolbuffers/python

We did not package all upstream plugins because there was no indication users needed really old versions and have been rolling forward with automating new ones.

gforsyth:
Thanks for clarifying that @mfridman !

I think that because we use `buf` in this repository to check whether the protobuf CAN build, but the actual code generation happens in the various language-specific repos, we should probably just ship this as is (or drop the plugin versions to the oldest supported version published on `buf.build`).  Better to have _any_ checks than none.

We should probably get the PMC to weigh in, though, since eventually this will translate to the various language-specific repositories bumping the lower bound of `protobuf` that they use to generate code.  (Although if we want to "just fix" the CI, we can merge this as is and then always lower the plugin versions later as part of a larger conversation)

westonpace (approved):
Thank you all for the investigation and discussion.  It is good we think about things thoroughly.  I did a little bit of research but won't claim to be an expert.

I don't think we need to worry too much about the version used by `buf` for CI matching the version used by `protoc` in the actual implementation.  There may be some risk for false negative / positive but we can cross that bridge when we come to it.

Regarding the broader topic of "can a message written by protobuf version X be read by protobuf version Y, assuming the .proto files are identical on both sides?" my understanding is that the answer is always yes, regardless of the versions (I can't find concrete proof of this and am happy for someone to prove me wrong)

There is yet another topic which is "can files generated by protoc version X be linked/compiled with libprotobuf version Y?".  The answer there is ["only if X is less than or equal to Y or, in the case of C++, X must be exactly equal to Y".](https://protobuf.dev/support/cross-version-runtime-guarantee/)  However, I do not think this is relevant to this PR.  So I'm going to go ahead and merge.

mfridman:
Thanks, let me know if I can help to get ya'll unblocked or answer any questions.

I tested this locally from the root of the repository with the buf version being used in CI (v1.16.0) and all the files were generated successfully.

```
$ buf --version                                                                           
1.16.0

$ buf generate

$ tree -L 3 gen/proto
gen/proto
├── cpp
│   └── substrait
│       ├── algebra.pb.cc
│       ├── algebra.pb.h
│       ├── capabilities.pb.cc
│       ├── capabilities.pb.h
│       ├── extended_expression.pb.cc
│       ├── extended_expression.pb.h
│       ├── extensions
│       ├── function.pb.cc
│       ├── function.pb.h
│       ├── parameterized_types.pb.cc
│       ├── parameterized_types.pb.h
│       ├── plan.pb.cc
│       ├── plan.pb.h
│       ├── type.pb.cc
│       ├── type.pb.h
│       ├── type_expressions.pb.cc
│       └── type_expressions.pb.h
├── csharp
│   ├── Algebra.cs
│   ├── Capabilities.cs
│   ├── ExtendedExpression.cs
│   ├── Extensions.cs
│   ├── Function.cs
│   ├── ParameterizedTypes.cs
│   ├── Plan.cs
│   ├── Type.cs
│   └── TypeExpressions.cs
├── go
│   └── substrait
│       ├── algebra.pb.go
│       ├── capabilities.pb.go
│       ├── extended_expression.pb.go
│       ├── extensions
│       ├── function.pb.go
│       ├── parameterized_types.pb.go
│       ├── plan.pb.go
│       ├── type.pb.go
│       └── type_expressions.pb.go
├── java
│   └── io
│       └── substrait
└── python
    └── substrait
        ├── algebra_pb2.py
        ├── capabilities_pb2.py
        ├── extended_expression_pb2.py
        ├── extensions
        ├── function_pb2.py
        ├── parameterized_types_pb2.py
        ├── plan_pb2.py
        ├── type_expressions_pb2.py
        └── type_pb2.py

14 directories, 41 files
```

PR #503, opened 2023-05-21
feat: improve ddl rel message

moshe-yakovson:
BREAKING CHANGE: make Plan as a root of DdlRel 
by adding DdlRel as one of relation in Rel (Plan::PlanRel::Rel::DdlRel) 
The reasons to do it:
- currently DdlRel is a separate object
and It is missing important fields  from Plan object like:
`Version, SimpleExtensionURI, SimpleExtensionDeclaration `
these fields appear in Plan so when DdlRel is part of Plan it solves this issue
- DdlRel is also a type of Rel

BREAKING CHANGE:  add extension field to DdlRel 
needed for extend or introduce any target specific metadata about the table, 
like replication, partition, etc.

BREAKING CHANGE: add primary_keys field to DdlRel message

### DIFF
--- a/proto/substrait/algebra.proto
+++ b/proto/substrait/algebra.proto
@@ -369,6 +369,8 @@ message Rel {
     //Physical relations
     HashJoinRel hash_join = 13;
     MergeJoinRel merge_join = 14;
+
+    DdlRel ddl = 15;
   }
 }
 
@@ -409,6 +411,13 @@ message DdlRel {
 
   // The body of the CREATE VIEW
   Rel view_definition = 7;
+
+  // Field that could be used to extend or introduce any target specific metadata about the table,
+  // like replication, partition, etc.
+  ExtensionObject table_metadata = 8;
+
+  // Indices of all primary keys in the table
+  repeated int32 primary_keys = 9;
 
   enum DdlObject {
     DDL_OBJECT_UNSPECIFIED = 0;
--- a/proto/substrait/algebra.proto
+++ b/proto/substrait/algebra.proto
@@ -370,7 +370,9 @@ message Rel {
     HashJoinRel hash_join = 13;
     MergeJoinRel merge_join = 14;
 
+    //Write relations
     DdlRel ddl = 15;
+    WriteRel write = 16;
   }
 }
 
--- a/proto/substrait/algebra.proto
+++ b/proto/substrait/algebra.proto
@@ -6,6 +6,7 @@ package substrait;
 import "google/protobuf/any.proto";
 import "substrait/extensions/extensions.proto";
 import "substrait/type.proto";
+import "substrait/table.proto";
 
 option csharp_namespace = "Substrait.Protobuf";
 option go_package = "github.com/substrait-io/substrait-go/proto";
@@ -369,10 +370,6 @@ message Rel {
     //Physical relations
     HashJoinRel hash_join = 13;
     MergeJoinRel merge_join = 14;
-
-    //Write relations
-    DdlRel ddl = 15;
-    WriteRel write = 16;
   }
 }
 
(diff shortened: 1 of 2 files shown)


proto/substrait/algebra.proto
```diff
@@ -369,6 +369,8 @@ message Rel {
     //Physical relations
     HashJoinRel hash_join = 13;
     MergeJoinRel merge_join = 14;
+
+    DdlRel ddl = 15;
```
EpsilonPrime, line 373: This is a new category of relations so it probably requires a comment defining that category, perhaps "write relations".

proto/substrait/algebra.proto
```diff
@@ -410,6 +412,13 @@ message DdlRel {
   // The body of the CREATE VIEW
   Rel view_definition = 7;
 
+  // Field that could be used to extend or introduce any target specific metadata about the table,
+  // like replication, partition, etc.
+  ExtensionObject table_metadata = 8;
```
EpsilonPrime, line 417: There's an ExtensionObject above which deals with what we're operating on.  Is that different than what would be used here?  Do you see this useful in conjunction with a named_object?

proto/substrait/algebra.proto
```diff
@@ -410,6 +412,13 @@ message DdlRel {
...
+
+  // Indices of all primary keys in the table
+  repeated int32 primary_keys = 9;
```
EpsilonPrime, line 420: What source do the int32 values refer to?

EpsilonPrime: Also, update the comment below saying to add primary keys to the message.

proto/substrait/algebra.proto
moshe-yakovson: It looks like that the ExtensionObject inside oneof write_type field
intended for a somewhat different purpose
not for general information about table definitions / metadata
but indeed this information can also be transmitted via ExtensionObject in oneof write_type

proto/substrait/algebra.proto
moshe-yakovson: Assuming that the table columns are listed in table_schema 
(table_schema.struct.types)
as a list of columns types 
so the int32 value refer to the index of the column (type) in the list

proto/substrait/algebra.proto
moshe-yakovson: added "write relations" comment
in addition WriteRel object is also added
because if we add the DdlRel there is no reason not to add also WriteRel

proto/substrait/algebra.proto
```diff
@@ -369,10 +370,6 @@ message Rel {
     //Physical relations
     HashJoinRel hash_join = 13;
     MergeJoinRel merge_join = 14;
-
```
EpsilonPrime: With these removed DDL and Write relations won't be in the plan.

proto/substrait/algebra.proto
```diff
@@ -390,36 +387,39 @@ message ExtensionObject {
   google.protobuf.Any detail = 1;
 }
 
+
 message DdlRel {
+  //Substrait version of the DdlRel.
+  Version version = 1;
```
EpsilonPrime, line 393: Given that the plan has a version do we need another version here?

proto/substrait/algebra.proto
```diff
@@ -390,36 +387,39 @@ message ExtensionObject {
...
+  repeated substrait.extensions.extensions.SimpleExtensionURI extension_uris = 2;
+
+  // A list of extentions this DDL may depend on
+  repeated substrait.extensions.extensions.SimpleExtensionDeclaration extentions = 3;
+
   // Definition of which type of object we are operating on
   oneof write_type {
-    NamedObjectWrite named_object = 1;
-    ExtensionObject extension_object = 2;
+    NamedObjectWrite named_object = 4;
```
EpsilonPrime, line 403: Please avoid renumbering existing values as they mess with already constructed objects.

proto/substrait/table.proto
```diff
@@ -0,0 +1,24 @@
...
+option java_package = "io.substrait.proto";
+
+message Table {
+  string table_name = 1;
+  NamedStruct table_schema = 2;
+  ExtentionObject table_metadata = 3;
+  repeated string primary_keys = 4;
+}
+
+message Tables {
```
EpsilonPrime, line 22: This appears unused.

proto/substrait/table.proto
```diff
@@ -0,0 +1,24 @@
...
+import "substrait/type.proto";
+
+option csharp_namespace = "Substrait.Protobuf";
+option go_package = "github.com/substrait-io/substrait-go/proto";
+option java_multiple_files = true;
+option java_package = "io.substrait.proto";
+
+message Table {
+  string table_name = 1;
+  NamedStruct table_schema = 2;
```
EpsilonPrime, line 17: As this field is in a table object do we need to prefix table?

proto/substrait/table.proto
```diff
@@ -0,0 +1,24 @@
...
+  ExtentionObject table_metadata = 3;
```
EpsilonPrime, line 18: Extension?

jacques-n:

PR #504, opened 2023-05-22
feat: names should be compared case-insensitively

vbarua:
BREAKING CHANGE: type and function name comparisons should be done case-insensitively

### DIFF
--- a/site/docs/extensions/index.md
+++ b/site/docs/extensions/index.md
@@ -1,6 +1,6 @@
 # Extensions
 
-In many cases, the existing objects in Substrait will be sufficient to accomplish a particular use case. However, it is sometimes helpful to create a new data type, scalar function signature or some other custom representation within a system. For that, Substrait provides a number of extension points. 
+In many cases, the existing objects in Substrait will be sufficient to accomplish a particular use case. However, it is sometimes helpful to create a new data type, scalar function signature or some other custom representation within a system. For that, Substrait provides a number of extension points.
 
 ## Simple Extensions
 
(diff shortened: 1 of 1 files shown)


EpsilonPrime (commented):
FYI:  The text parser currently is case sensitive.  It will need to be updated to match this change.

vbarua:
# Context
The substrait-validator currently compares the names of types and function case-insensitively. 
https://github.com/substrait-io/substrait-validator/blob/3c934798c609a45b886fa805c694b5ad983acac3/proto/substrait/validator/simple_extensions.proto

This overall seems like reasonable behaviour, as it would be somewhat odd to have `i8` and `I8` be different types, and `add` and `ADD` be different functions.

This PR codifies the behaviour into the spec.

This change is technically breaking, as any user defining types or functions with names that were identical except for capitalization will need to modify them (assuming they wish them to be different entities).

westonpace:
Another solution would be to mandate lowercase names for functions and types.  It's slightly more challenging for producers perhaps (which must translate customer input) but simpler for consumers (which could then be case sensitive or insensitive).  But I don't mind adding case sensitivity either (Acero is case sensitive currently, but it would not be difficult to change).

Are we limiting identifiers to ASCII characters?  Because, if we don't, case insensitive comparison gets challenging.

mbrobbel:
+1 for case-sensitivity and restricting to ASCII chars.

vbarua:
Apologies. I have not had (and likely won't have) time to follow up on this specifically.

PR #508, opened 2023-06-14
feat: add options to substring for start parameter being negative

richtia:
The substring function `start` parameter has different behavior between different 
backends when it's value negative.  Sqlite will wrap around the end of the input string,
whereas postgres will start from the left of the input string.

### DIFF
--- a/extensions/functions_string.yaml
+++ b/extensions/functions_string.yaml
@@ -55,8 +55,9 @@ scalar_functions:
       from position `start` and ending at the end of the string.
 
       The `negative_start` option applies to the `start` parameter. `WRAP_FROM_END` means
-      the returned substring will start from the end of the `input`. The last character
-      has an index of -1.  `LEFT_OF_BEGINNING` means the returned substring will start from
+      the returned substring will start from the end of the `input` and move backwards.
+      The last character has an index of -1, the second to last character has an index of -2,
+      and so on. `LEFT_OF_BEGINNING` means the returned substring will start from
       the left of the first character.  A `start` of -1 will begin 2 characters left of the
       the `input`, while a `start` of 0 begins 1 character left of the `input`.
     impls:
@@ -69,7 +70,7 @@ scalar_functions:
             name: "length"
         options:
           negative_start:
-            values: [ WRAP_FROM_END, LEFT_OF_BEGINNING ]
+            values: [ WRAP_FROM_END, LEFT_OF_BEGINNING, ERROR ]
         return: "varchar<L1>"
       - args:
           - value: "string"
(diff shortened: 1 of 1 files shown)


extensions/functions_string.yaml
```diff
@@ -51,6 +51,12 @@ scalar_functions:
     description: >-
       Extract a substring of a specified `length` starting from position `start`.
       A `start` value of 1 refers to the first characters of the string.
+
+      The `negative_start` option applies to the `start` parameter. `WRAP_FROM_END` means
+      the returned substring will start from the end of the `input`. The last character
+      has an index of -1.  `LEFT_OF_BEGINNING` means the returned substring will start from
```
westonpace, line 57: This description of `WRAP_FROM_END` seems a little confusing.

> the returned substring will start from the end of `input`.

The returned substring won't start at the end right?  For example, substr("hello", -3, 1) returns "l" which doesn't start at the end of the "hello"

> The last character has an index of -1

This is correct, but it is not clear to me that we count backwards, in other words, it is not clear that -2 is the second from last character.

westonpace (changes requested):
Do we want to add an option "ERROR"?  Are there any implementations that simply reject negative indices?

richtia:
> Do we want to add an option "ERROR"? Are there any implementations that simply reject negative indices?

That makes sense, just added that option.

extensions/functions_string.yaml
```diff
@@ -51,6 +51,13 @@ scalar_functions:
     description: >-
       Extract a substring of a specified `length` starting from position `start`.
       A `start` value of 1 refers to the first characters of the string.
+
+      The `negative_start` option applies to the `start` parameter. `WRAP_FROM_END` means
+      the returned substring will start from the end of the `input` and move backwards.
```
westonpace, line 56: Maybe "means the index will start from the end..."?  It still kind of sounds like the string itself is moving backwards.

richtia: Thanks for the catch.  updated!

westonpace (dismissed):
LGTM.  However, I do not believe this is a breaking change.  There is no such thing as required options in Substrait (there are required enum arguments but this is not specified that way, and I don't believe it should be).

I believe that adding new optional arguments is a non-breaking change.  If the argument is not specified then consumers are free to continue doing whatever they are doing today (e.g. any behavior is valid).

westonpace:
@richtia can you rebase this?  Do you agree it is a non-breaking change?

richtia:
> @richtia can you rebase this? Do you agree it is a non-breaking change?

Done!  Yeah, that makes sense that it shouldn't be a breaking change.

westonpace (approved):

PR #511, opened 2023-06-26
feat: add timestamp in max/min function

icexelloss:
This PR add supports for timestamp types in max/min function.

This is currently supported in Acero and I believe most databases support this as well.

### DIFF
--- a/extensions/functions_arithmetic.yaml
+++ b/extensions/functions_arithmetic.yaml
@@ -1096,6 +1096,20 @@ aggregate_functions:
         decomposable: MANY
         intermediate: fp64?
         return: fp64?
+      - args:
+          - name: x
+            value: timestamp
+        nullability: DECLARED_OUTPUT
+        decomposable: MANY
+        intermediate: timestamp?
+        return: timestamp?
+      - args:
+          - name: x
+            value: timestamp_tz
+        nullability: DECLARED_OUTPUT
+        decomposable: MANY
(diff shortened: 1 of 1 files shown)


gforsyth (approved):
This is supported in Acero and also in DuckDB (and likely others).  Not a breaking change, so I think this can go in.

westonpace (approved):
The comparison functions (e.g. lt, gt, etc.) are [defined with `any`](https://github.com/substrait-io/substrait/blob/main/extensions/functions_comparison.yaml#L52-L65) so it seems we are already making an implicit assumption that all types are comparable amongst themselves.  It seems, if comparison functions are defined, then max/min must exist as well.  Perhaps it would make more sense to change `min`/`max` to use `any` instead of explicitly listing every type?

Looking at existing implementations we have:

| Implementation | Description of supported types |
| --- | --- |
| Postgres | any numeric, string, date/time, network, or enum type, or arrays of these types |
| SQL Server | MAX can be used with numeric, character, uniqueidentifier, and datetime columns, but not with bit columns. |
| Acero | Non-nested types |
| MariaDb, Sqlite, Spark, Presto, DuckDb, Datafusion | Does not list allowed types in documentation |

That being said, this addition is pretty straightforward, so we can also defer this discussion for another day.

icexelloss:
Thanks @westonpace and @gforsyth !

PR #513, opened 2023-06-29
feat: add a two-arg variant of substring

westonpace:
When only two arguments are specified then the method will return everything
from `start` to the end of the string.  This variant is implemented in pretty much
all major engines today.

### DIFF
--- a/extensions/functions_string.yaml
+++ b/extensions/functions_string.yaml
@@ -50,7 +50,9 @@ scalar_functions:
     name: substring
     description: >-
       Extract a substring of a specified `length` starting from position `start`.
-      A `start` value of 1 refers to the first characters of the string.
+      A `start` value of 1 refers to the first characters of the string.  When
+      `length` is not specified the function will extract a substrait starting
+      from position `start` and ending at the end of the string.
     impls:
       - args:
           - value: "varchar<L1>"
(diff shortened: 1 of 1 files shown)


westonpace:
This will conflict with #508 so let's just be cognizant not to merge them both back-to-back without a rebase.

richtia:
> When only two arguments are specified then the method will return everything from `start` to the end of the string. This variant is implemented in pretty much all major engines today.

Should we update the description too?

westonpace:
> Should we update the description too?

Great idea.  Done.

richtia (approved):
LGTM

extensions/functions_string.yaml
```diff
@@ -50,7 +50,9 @@ scalar_functions:
     name: substring
     description: >-
       Extract a substring of a specified `length` starting from position `start`.
-      A `start` value of 1 refers to the first characters of the string.
+      A `start` value of 1 refers to the first characters of the string.  When
+      `length` is not specified the function will extract a substrait starting
```
gforsyth, line 54:
```suggestion
      `length` is not specified the function will extract a substring starting
```
that is my favorite typo this week.

gforsyth (changes requested):
just the hilarious typo, then this looks good to go.

extensions/functions_string.yaml
westonpace: That was not the plan I had in mind.  At least it wasn't substrate

gforsyth (approved):
:shipit:

PR #518, opened 2023-07-12
feat: add ExchangeRel as a type in Rel

whutjs:
The `ExchangeRel` is already in the spec, but it is not a type of `Rel` yet.

### DIFF
--- a/proto/substrait/algebra.proto
+++ b/proto/substrait/algebra.proto
@@ -369,6 +369,7 @@ message Rel {
     //Physical relations
     HashJoinRel hash_join = 13;
     MergeJoinRel merge_join = 14;
+    ExchangeRel exchange = 15;
   }
 }
 


proto/substrait/algebra.proto
```diff
@@ -369,6 +369,7 @@ message Rel {
     //Physical relations
     HashJoinRel hash_join = 13;
     MergeJoinRel merge_join = 14;
+    ExchangeRel exchange = 15;
```
EpsilonPrime, line 372: There are three write relations that should be in a section (they are not physical relations).  These are ExchangeRel, DdlRel, and WriteRel.  It should look something like:

`
    // Write relations
    ExchangeRel exchange = 100;
    DdlRel ddl = 101;
    WriteRel write = 102;
`

westonpace: It is indeed listed as a physical relation in the docs so I think this placement is fine.  There are arguments to be made that "exchange" is a different sort of physical than "hash_join" (exchange is less about engine-specific optimization and more about distribution) but we'd need a lot more nuance than the docs currently have.  Right now I think our working definition of "logical" is "is this something an end-user would put in a query?" and our definition of "physical" is "everything else".

This structure is largely meaningless anyways (as in, it's just a comment and grouping for readability, it makes no difference to the programs themselves since they see this whole list of options as one thing).

EpsilonPrime: Agreed, we can go with this PR as is.

EpsilonPrime (approved):

westonpace:
Since this is a specification modification we will need two SMC votes.  @jacques-n @cpcloud can one of you chime in here?

proto/substrait/algebra.proto
jacques-n: LGTM +1

PR #520, opened 2023-07-13
docs: add example for field_references

MasseGuillaume:
related PR: https://github.com/substrait-io/substrait-java/pull/150


site/docs/expressions/field_references.md
```diff
@@ -11,6 +11,59 @@ In Substrait, all fields are dealt with on a positional basis. Field names are o
...
+a: struct
+  b: list
+    c: map string
+      x: i32
+
+SQL expression:
+a.b[2].c['my_map_key'].x
+```
+
+```
```
MasseGuillaume, line 26: I couldn't find a way to set this block as protobuf text. Maybe we could show the example as JSON.

https://github.com/github-linguist/linguist/blob/master/lib/linguist/languages.yml#L5370

westonpace: JSON would be good I think.  In many places we are hoping to add examples using the text serialization but I don't think that would be helpful here since we want to demonstrate a nuance of how to specify this in protobuf.

site/docs/expressions/field_references.md
```diff
@@ -11,6 +11,59 @@ In Substrait, all fields are dealt with on a positional basis. Field names are o
...
+              }
+            }
+          }
+        }
+      }
+    }
+  }
+  root_reference {
+  }
+}
```
westonpace, line 64: I'll repeat my concern from Slack.  I thought, given a reference X.Y.Z that the outer/top-most `struct_field` would be X and the inner/bottom-most `struct_field` would be Z.  However, this example provides the opposite.  The outer-most is Z and the inner-most is X.

If we agree on this definition I will need to update Acero.  I am curious if any other implementations are currently relying on one behavior or another.  Is there a particular reason to prefer this structure?

vbarua: This also feels backwards to me actually. Given `a.b[2].c['my_map_key'].x` I would expect the structure of this to be reversed. That is the:
1. Struct field reference `a`
2. Struct field `b`
3. List offset `2`
4. Struct field `c`
5. Map key `my_map_key`
6. Struct field `x`

I think of this as the series of steps needed to unpack the desired field out of the original input.

site/docs/expressions/field_references.md
```diff
@@ -11,6 +11,59 @@ In Substrait, all fields are dealt with on a positional basis. Field names are o
 | Map KeyExpression         | A wildcard string that is matched against a simplified form of regular expressions. Requires the key type of the map to be a character type. [Format detail needed, intention to include basic regex concepts such as greedy/non-greedy.] | map                | List of map value type     |
 | Masked Complex Expression | An expression that provides a mask over a schema declaring which portions of the schema should be presented. This allows a user to select a portion of a complex object but mask certain subsections of that same object. | any                | any                        |
 
+
+```
+Schema:
+a: struct
+  b: list
+    c: map string
+      x: i32
```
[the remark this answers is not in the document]
MasseGuillaume: this is the spark SQL type:

```
struct<
  a: list<
    struct<
      c: map<string, 
        struct<
          x: i32>>>>>
```

site/docs/expressions/field_references.md
MasseGuillaume: The downside of JSON is the lack of comments. I kept Protobuf text for now.

westonpace (dismissed):
This example is currently just dropped in without any explanation or motivation. Can we add a paragraph before it explaining why there is an example here?

image

The masked complex expression section is a good example

image

EpsilonPrime:
@westonpace It looks like changes were made based on your feedback.  Could you take a second look please?  Thanks!

jacques-n (approved):

EpsilonPrime (approved):

PR #521, opened 2023-07-13
docs: clarify map key behavior

EpsilonPrime:
BREAKING CHANGE: 
  * Map keys may be repeated.
   * Map keys must not be NULL.
   * The map key type may be nullable.

This is based on the current restrictions found in the wild.

DuckDB, Velox, Spark, and Acero all reject attempts to provide NULL as a key.

Despite DuckDB specifically calling out that keys must be unique in its
implementation other implementations such as Velox and Acero do not
require the key to be unique so we cannot require the map key to be 1:1 with map values.


jacques-n:
I'd like to see a more thorough analysis of other tools before agreeing with this. I feel like other systems don't have these constraints (but also admin that I haven't researched it recently.)

EpsilonPrime:
I did some investigation and found that a few implementations do allow multiple values for the same key so I've updated accordingly.  I was unable to find a system that did not reject NULL as a key so I've left that in as a new restriction.

jacques-n (approved):
Let's do a follow-up around field reference behavior. (Pick first? Pick last?)

EpsilonPrime:
Added #579 to track the additional information required here.

PR #524, opened 2023-07-17
feat: add current_date and current_timestamp functions

MasseGuillaume:


westonpace (changes requested):
We need some more details around how this should behave.  There was an old discussion on these functions a while back (I don't recall exactly where) and one of the concerns was that many query engines will go out of their way to ensure that this function gives a consistent result if invoked multiple times within a query.

For example, a query might be a streaming query (and thus the function is called on each batch as the data passes through) or a plan might have multiple project nodes (and each one might call this function) or maybe the plan includes distribution and this function is called both on the workers and on the master node.

Given that `current_date` and `current_timestamp` can be replaced by literals on the producer is it essential that this be part of the plan? If so, can you explain how it should behave in the above scenarios?

MasseGuillaume:
Ah got it, it makes sense. In my use case, I use substrait to serialize the SQL so a rich UI can display a query plan to the user. In this case, the query is not yet executed, so it would not make sense to constant fold the current_time/current_date. What's missing is substrait is an optional function result for non-deterministic functions.

EpsilonPrime:
Similar functionality was added in #528 so I believe we can close this PR.

westonpace:
@EpsilonPrime did you perhaps mean #548 ?

EpsilonPrime:
Ah, yes, thanks for the correction!

jacques-n:
Closing in preference to #548

PR #525, opened 2023-07-17
docs: clarify requirements for groupings and measures in AggregateRel

vbarua:


vbarua:
The spec has some additional wording around the optionality of grouping sets and measures in the AggregateRel.

https://substrait.io/relations/logical_relations/#aggregate-properties

I was initially confused by how to represent a query like
```sql
SELECT a, b FROM foo GROUP BY a, b
```

because AggregateRel indicates
```proto
// A list of one or more aggregate expressions along with an optional filter.
  repeated Measure measures = 4;
```

However from the full spec the intent appears to be to represent this by not populating the measures at all.

This commits adds some additional language around the requirement for measures and groupings to the proto file.

proto/substrait/algebra.proto
```diff
@@ -206,9 +206,11 @@ message AggregateRel {
   Rel input = 2;
 
   // A list of expression grouping that the aggregation measured should be calculated for.
+  // Required if there are no measures
   repeated Grouping groupings = 3;
 
   // A list of one or more aggregate expressions along with an optional filter.
```
westonpace, line 212: Should this change to zero or more?

westonpace (changes requested):
Just one question

proto/substrait/algebra.proto
vbarua: Per the [Aggregate Properties](https://substrait.io/relations/logical_relations/#aggregate-properties) portion of the spec, both groupings and measures are optional lists of one or more elements.

As written, it appears that the intent is that the empty list should be represented as null. It might be worth loosening this requirement, but that becomes a spec change beyond just documentation.

westonpace: Ok, that makes sense to me.

westonpace (approved):

PR #526, opened 2023-07-18
docs: add a try the tutorial call to action

EpsilonPrime:
This adds a "Check out the Tutorial" call to action to the Substrait front page.
The call to action can be dismissed and will not return on the same browser (the
dismissal is stored in the client's local storage).  The design of the call to action
allows for multiple call to actions which are differentiated by the admonition's title
(which is not shown in the call to action behavior).


westonpace (approved):
I've been overdue in adding this review for a bit :)

I think this approach is ok.  Excess JS can be hard to maintain but I understand why it is there.  The call to action does it's job I believe.

I think, in general, I approve.  I'm not much of a graphic design / UX expert so let's briefly ask for feedback in the community meeting and, if there is nothing significant, go ahead and merge.

EpsilonPrime:
Based on the discussion at the most recent Community Meeting we are inclined to modify the website to make it easier to find the content that is important than to add more elements to make it more visible.  The suggested approach is to remove extraneous items so the important stuff is more clear.  I'll address that in a different PR.

PR #530, opened 2023-07-21
docs: document IfThen proto

vbarua:


vbarua:
The proto comments are based on
https://substrait.io/expressions/specialized_record_expressions/#if-expression

westonpace (approved):

PR #532, opened 2023-07-26
feat: add field-id resolution to parquet reads

westonpace:
This PR clarifies how field resolution should happen when the base schema
does not perfectly match the file schema (this is a very common case in some
environments where there is no catalog or the catalog does not include the
full parquet schema).

This should not be a breaking change as all existing implementations are
operating by the default (name-based resolution) as far as I know.

### DIFF
--- a/site/docs/relations/logical_relations.md
+++ b/site/docs/relations/logical_relations.md
@@ -65,11 +65,35 @@ provide the consumer with more information on how to retrieve the data.
 
 #### Files Type
 
+A file read lists one or more files or directories that should be scanned.
+
 | Property                    | Description                                                       | Required |
 | --------------------------- | ----------------------------------------------------------------- | -------- |
 | Items                       | An array of Items (path or path glob) associated with the read.   | Required |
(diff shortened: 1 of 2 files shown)


westonpace:
I am leaving this in a draft PR until I have prototyped support for this in Acero to confirm this should work.

site/docs/relations/logical_relations.md
```diff
@@ -65,12 +65,36 @@ provide the consumer with more information on how to retrieve the data.
...
+file formats.  When specifying a file format there may be format-specific options that should be specified
+as well.
+
+###### Parquet Format Options
+
+The following options are specific to the parquet file format.
+
+| Property  | Description                                                                                        | Required |
+| --------- | -------------------------------------------------------------------------------------------------- | -------- |
+| Field Ids | The field ids, if specified, define an alterantive way to resolve columns in the file (see below). | Optional |
```
mbrobbel, line 88:
```suggestion
| Field Ids | The field ids, if specified, define an alternative way to resolve columns in the file (see below). | Optional |
```

PR #534, opened 2023-08-01
feat: add more window functions

richtia:
PR to add more window functions:
* first_value
* last_value
* nth_value
* lead
* lag

### DIFF
--- a/extensions/functions_arithmetic.yaml
+++ b/extensions/functions_arithmetic.yaml
@@ -1641,12 +1641,18 @@ window_functions:
       Returns a value from the nth row based on the `window_offset`. `window_offset` should
       be a positive integer. If the value of the `window_offset` is outside the range
       of the window, `null` is returned.
+
+      The `on_domain_error` option governs behavior in cases where `window_offset` is not
+      a positive integer or `null`.
     impls:
       - args:
           - value: any1
             name: expression
           - value: i32
             name: window_offset
+        options:
+          on_domain_error:
+            values: [ NULL, ERROR ]
         nullability: DECLARED_OUTPUT
         decomposable: NONE
         return: any1?
@@ -1658,11 +1664,14 @@ window_functions:
 
       The `expression` is evaluated against a row that comes after the current row based
       on the `window_offset`.  The `window_offset` should be a positive integer and is set to
-      1 if not specified explicitly. The function returns the `default` input value if
-      `window_offset` goes beyond the scope of the partition. If a `default` value is not
-      specified, it is set to `null`
+      1 if not specified explicitly. If the `window_offset` is negative, the expression
+      will be evaluated against a row coming before the current row, similar to the `lag`
+      function. A `window_offset` of `null` will return `nulls`. The function returns the
+      `default` input value if `window_offset` goes beyond the scope of the partition.
+      If a `default` value is not specified, it is set to `null`
 
       Example comparing the sales of the current year to the following year.
+      `window_offset` of 1.
       | year | sales  | next_year_sales |
       | 2019 | 20.50  | 30.00           |
       | 2020 | 30.00  | 45.99           |
@@ -1703,11 +1712,14 @@ window_functions:
       The `expression` is evaluated against a row that comes before the current row based
       on the `window_offset`.  The `expression` can be a column, expression or subquery that
       evaluates to a single value. The `window_offset` should be a positive integer and is set to
-      1 if not specified explicitly. The function returns the `default` input value if
-      `window_offset` goes beyond the scope of the partition. If a `default` value is not
-      specified, it is set to `null`
+      1 if not specified explicitly. If the `window_offset` is negative, the expression will
+      be evaluated against a row coming after the current row, similar to the `lead` function.
+      A `window_offset` of `null` will return `nulls`. The function returns the `default`
+      input value if `window_offset` goes beyond the scope of the partition. If a `default`
+      value is not specified, it is set to `null`
 
       Example comparing the sales of the current year to the previous year.
+      `window_offset` of 1.
       | year | sales  | previous_year_sales |
       | 2019 | 20.50  | null                |
       | 2020 | 30.00  | 20.50               |
--- a/extensions/functions_arithmetic.yaml
+++ b/extensions/functions_arithmetic.yaml
@@ -1653,7 +1653,7 @@ window_functions:
         options:
           on_domain_error:
             values: [ NAN, ERROR ]
-        nullability: DECLARED_OUTPU
+        nullability: DECLARED_OUTPUT
         decomposable: NONE
         return: any1?
         window_type: PARTITION
--- a/extensions/functions_arithmetic.yaml
+++ b/extensions/functions_arithmetic.yaml
@@ -1689,7 +1689,7 @@ window_functions:
             name: default
         nullability: DECLARED_OUTPUT
         decomposable: NONE
-        return: i32?
+        return: any1?
         window_type: PARTITION
       - args:
           - value: any1
@@ -1700,7 +1700,7 @@ window_functions:
             name: default
         nullability: DECLARED_OUTPUT
         decomposable: NONE
-        return: i64?
+        return: any1?
         window_type: PARTITION
   - name: "lag"
     description: >
--- a/extensions/functions_arithmetic.yaml
+++ b/extensions/functions_arithmetic.yaml
@@ -1673,6 +1673,12 @@ window_functions:
       1 if not specified explicitly. The function returns the `default` input value if
       `offset` goes beyond the scope of the partition. If a `default` value is not
       specified, it is set to `null`
+
+      Example comparing the sales of the current year to the following year.
+      | year | sales  | next_year_sales |
+      | 2019 | 20.50  | 30.00           |
+      | 2020 | 30.00  | 45.99           |
+      | 2021 | 45.99  | 50.00           |
     impls:
       - args:
           - value: any1
@@ -1707,6 +1713,12 @@ window_functions:
       1 if not specified explicitly. The function returns the `default` input value if
       `offset` goes beyond the scope of the partition. If a `default` value is not
       specified, it is set to `null`
+
+      Example comparing the sales of the current year to the previous year.
+      | year | sales  | previous_year_sales |
+      | 2019 | 20.50  | 10.00               |
+      | 2020 | 30.00  | 20.50               |
+      | 2021 | 45.99  | 30.00               |
     impls:
       - args:
           - value: any1
--- a/extensions/functions_arithmetic.yaml
+++ b/extensions/functions_arithmetic.yaml
@@ -1616,8 +1616,7 @@ window_functions:
         window_type: PARTITION
   - name: "first_value"
     description: >
-      The `expression` is evaluated against the first row.  The `expression` can be a
-      column, expression or subquery that evaluates to a single value.
+      Returns the first value in the window.
     impls:
       - args:
           - value: any1
@@ -1628,8 +1627,7 @@ window_functions:
         window_type: PARTITION
   - name: "last_value"
     description: >
-      The `expression` is evaluated against the last row.  The `expression` can be a
-      column, expression, or subquery that evaluates to a single value.
+      Returns the last value in the window.
     impls:
       - args:
           - value: any1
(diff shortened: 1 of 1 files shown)


extensions/functions_arithmetic.yaml
```diff
@@ -1614,3 +1614,104 @@ window_functions:
...
+        nullability: DECLARED_OUTPUT
+        decomposable: NONE
+        return: i64?
+        window_type: PARTITION
+  - name: "lag"
+    description: >
+      Return a value from a previous row based on a specified physical offset.
+
+      The `offset` is 1 if not specified explicitly.  The function returns
+      the `default` input value if `offset` goes beyond the scope of te partition.
```
EpsilonPrime, line 1693: te -> the

extensions/functions_arithmetic.yaml
```diff
@@ -1614,3 +1614,104 @@ window_functions:
...
+            name: offset
+          - value: any1
+            name: default
+        nullability: DECLARED_OUTPUT
+        decomposable: NONE
+        return: i64?
+        window_type: PARTITION
+  - name: "lag"
+    description: >
+      Return a value from a previous row based on a specified physical offset.
```
EpsilonPrime, line 1690: Without looking at the other functions being added here it's not clear what is happening.  Could you add a brief example please?  One point of confusion is which value is being returned (seems like you could choose any column from the description).

extensions/functions_arithmetic.yaml
```diff
@@ -1614,3 +1614,104 @@ window_functions:
...
+        nullability: DECLARED_OUTPUT
+        decomposable: NONE
+        return: any1?
+        window_type: PARTITION
+  - name: "lead"
+    description: >
+      Return a value from a following row based on a specified physical offset.
+
+      The `offset` is 1 if not specified explicitly.  The function returns
+      the `default` input value if `offset` goes beyond the scope of te partition.
```
EpsilonPrime, line 1663: te -> the

extensions/functions_arithmetic.yaml
richtia: Thanks @EpsilonPrime! Added more details to the description as well as an example.  Lmk if there's anything else that would make it easier to understand.

EpsilonPrime: Should one of the values have a NULL?  Otherwise lgtm.

richtia: Yeah, that would make more sense in this example.  Just updated!

EpsilonPrime (dismissed):

extensions/functions_arithmetic.yaml
```diff
@@ -1614,3 +1614,131 @@ window_functions:
...
+      - args:
+          - value: any1
+            name: expression
+          - value: i32
+            name: offset
+          - value: any1
+            name: default
+        nullability: DECLARED_OUTPUT
+        decomposable: NONE
+        return: i32?
```
jacques-n, line 1692: looks wrong

extensions/functions_arithmetic.yaml
```diff
@@ -1614,3 +1614,131 @@ window_functions:
...
+      - args:
+          - value: any1
+            name: expression
+          - value: i64
+            name: offset
+          - value: any1
+            name: default
+        nullability: DECLARED_OUTPUT
+        decomposable: NONE
+        return: i64?
```
jacques-n, line 1703: looks wrong

jacques-n (commented):
mostly looks good but a couple of return types seem weird.

extensions/functions_arithmetic.yaml
richtia: updated!

extensions/functions_arithmetic.yaml
```diff
@@ -1614,3 +1614,131 @@ window_functions:
         decomposable: NONE
         return: i64?
         window_type: PARTITION
+  - name: "first_value"
+    description: >
+      The `expression` is evaluated against the first row.  The `expression` can be a
+      column, expression or subquery that evaluates to a single value.
```
westonpace, line 1620: The wording here seems overly verbose to me.  Technically every argument in every function is an expression.

> The `expression` can be a column, expression or subquery that evaluates to a single value.

A column is not a thing in the spec. I suspect you mean a field reference which is an expression.

> or subquery that evaluates to a single value.

This is also an expression.

> The `expression` is evaluated against the first row.

This is not necessarily true in many implementations.  Many naive implementations of `first(x*x)` will first calculate `x*x` on the entire input and then take the first item.  It's a good optimization but I don't think we should imply that it is always done this way.

extensions/functions_arithmetic.yaml
```diff
@@ -1614,3 +1614,131 @@ window_functions:
...
+            name: expression
+        nullability: DECLARED_OUTPUT
+        decomposable: NONE
+        return: any1?
+        window_type: PARTITION
+  - name: "nth_value"
+    description: >
+      The `expression` is evaluated against a row that comes after the current row based
+      on the `offset`.  `offset` should be a positive integer.  The `expression` can be a
+      column, expression or subquery that evaluates to a single value.
```
westonpace, line 1645: This is incorrect.  Also, we shouldn't use the term `offset` as it is ambiguous.  Perhaps `window_offset`.  Given a window of `[A, B, C, D, E, F, G]` and the call `nth_value(col, 2)` then I would expect, based on this description, that the return value would depend on the current row.  In fact, The description here reads very much like `lead`.  In other words, I would expect the return would be `[C, D, E, F, G, NULL, NULL]` but that is incorrect.

It's more correct to say that it grabs "the nth row in the window".  The return in the above example should be `[B, B, B, B, B, B, B]`

extensions/functions_arithmetic.yaml
```diff
@@ -1614,3 +1614,131 @@ window_functions:
...
+        window_type: PARTITION
+  - name: "lead"
+    description: >
+      Return a value from a following row based on a specified physical offset.
+      This allows you to compare a value in the current row against a following row.
+
+      The `expression` is evaluated against a row that comes after the current row based
+      on the `offset`.  The `expression` can be a column, expression or subquery that
+      evaluates to a single value. The `offset` should be a positive integer and is set to
+      1 if not specified explicitly. The function returns the `default` input value if
```
westonpace, line 1673: I don't see any `impls` that allow me to omit the offset (and I don't think there should be any).  Would it be more accurate to say the `offset` will be set to 1 if it is negative or null?  Is this how current engines work?

extensions/functions_arithmetic.yaml
```diff
@@ -1614,3 +1614,131 @@ window_functions:
...
+      - args:
+          - value: any1
+            name: expression
+        nullability: DECLARED_OUTPUT
+        decomposable: NONE
+        return: any1?
+        window_type: PARTITION
+  - name: "nth_value"
+    description: >
+      The `expression` is evaluated against a row that comes after the current row based
```
westonpace, line 1643: What happens if n is outside the range of the window?  E.g. a window of size 10 and n is 100.

extensions/functions_arithmetic.yaml
```diff
@@ -1614,3 +1614,131 @@ window_functions:
...
+        decomposable: NONE
+        return: any1?
+        window_type: PARTITION
+      - args:
+          - value: any1
+            name: expression
+          - value: i64
+            name: offset
+          - value: any1
+            name: default
```
westonpace, line 1700: I'm not certain there is much value in an `i64` variant.  I don't think there are any realistic scenarios where a window has more than 2^32 rows.  We can always add that later if needed.  I suppose this same logic applies to lag and lead.

westonpace (changes requested):
I have a few requests.  Also, I'm not sure we need to say things like "expression is evaluated against".  We don't really use that terminology elsewhere.  For example, we could phrase addition as:

> Two different expressions are evaluated.  The results of these expressions are added together.

But I think that is more confusing.  In other words, I think:

> Returns the first value in the window

is more clear than:

> The `expression` is evaluated against the first row.

extensions/functions_arithmetic.yaml
richtia: Current engines will default to 1 if it's not set.  So it actually might be best to have an `impl` for that.

westonpace: Ok.  I think that will work.  We should still clarify what the behavior will be if the offset is -1 or null.

richtia: Updated!  Also updated nth_value function to specify negative or null `window_offset` behavior.

extensions/functions_arithmetic.yaml
```diff
@@ -1614,3 +1614,141 @@ window_functions:
...
+      a positive integer or `null`.
+    impls:
+      - args:
+          - value: any1
+            name: expression
+          - value: i32
+            name: window_offset
+        options:
+          on_domain_error:
+            values: [ NULL, ERROR ]
```
westonpace, line 1655: Do we have evidence that engines do both of these things? In other words, is there at least one existing engine that does `NULL` and at least one existing engine that does `ERROR`?

extensions/functions_arithmetic.yaml
```diff
@@ -1614,3 +1614,141 @@ window_functions:
...
+  - name: "lead"
+    description: >
+      Return a value from a following row based on a specified physical offset.
+      This allows you to compare a value in the current row against a following row.
+
+      The `expression` is evaluated against a row that comes after the current row based
+      on the `window_offset`.  The `window_offset` should be a positive integer and is set to
+      1 if not specified explicitly. If the `window_offset` is negative, the expression
+      will be evaluated against a row coming before the current row, similar to the `lag`
+      function. A `window_offset` of `null` will return `nulls`. The function returns the
```
westonpace, line 1669:
```suggestion
      function. A `window_offset` of `null` will return `null`. The function returns the
```
`nulls` in backticks seems to imply there is some special variable `nulls`.  I think using the singular form is ok here.

extensions/functions_arithmetic.yaml
```diff
@@ -1614,3 +1614,141 @@ window_functions:
...
+      `default` input value if `window_offset` goes beyond the scope of the partition.
```
westonpace, line 1670:
```suggestion
      `default` input value if `window_offset` goes beyond the scope of the window.
```
The window is not always a partition.

extensions/functions_arithmetic.yaml
```diff
@@ -1614,3 +1614,141 @@ window_functions:
...
+        return: any1?
+        window_type: PARTITION
+  - name: "lead"
+    description: >
+      Return a value from a following row based on a specified physical offset.
+      This allows you to compare a value in the current row against a following row.
+
+      The `expression` is evaluated against a row that comes after the current row based
+      on the `window_offset`.  The `window_offset` should be a positive integer and is set to
+      1 if not specified explicitly. If the `window_offset` is negative, the expression
```
westonpace, line 1667: This is my fault for not being clear but I think `window_offset` is incorrect for `lag` and `lead`.  In other words...

| Function | Variable Name |
| --- | --- |
| nth_value | window_offset |
| lag | row_offset |
| lead | row_offset |

This is because, in `nth_value` the offset is from the start of the window.  In `lag` and `lead` the offset is from the current row.

westonpace (changes requested):
A few more nitpicks

extensions/functions_arithmetic.yaml
richtia: Yes, sqlite accept null for window offset, but postgres will error out.

extensions/functions_arithmetic.yaml
richtia: updated!  `null` is returned in this scenario

extensions/functions_arithmetic.yaml
richtia: updated! thanks

extensions/functions_arithmetic.yaml
richtia: updated! thanks

extensions/functions_arithmetic.yaml
richtia: updated!

EpsilonPrime:
Looks like there's one failing check to address before we can proceed here.

richtia:
> Looks like there's one failing check to address before we can proceed here.

Thanks! Had a typo!

EpsilonPrime (approved):

EpsilonPrime:
@westonpace Anything else you'd like to see?

extensions/functions_arithmetic.yaml
```diff
@@ -1654,3 +1654,141 @@ window_functions:
...
+  - name: "first_value"
+    description: >
+      Returns the first value in the window.
+    impls:
+      - args:
+          - value: any1
+            name: expression
+        nullability: DECLARED_OUTPUT
+        decomposable: NONE
+        return: any1?
```
westonpace, line 1666:
```suggestion
        return: any1
```

extensions/functions_arithmetic.yaml
```diff
@@ -1654,3 +1654,141 @@ window_functions:
...
+  - name: "last_value"
+    description: >
+      Returns the last value in the window.
+    impls:
+      - args:
+          - value: any1
+            name: expression
+        nullability: DECLARED_OUTPUT
+        decomposable: NONE
+        return: any1?
```
westonpace, line 1677:
```suggestion
        return: any1
```

westonpace (approved):
The return of `first_value` and `last_value` is only nullable if the input is nullable.  I don't remember if we have syntax for that or not.  But this is minor and we can always adjust later if needed.

extensions/functions_arithmetic.yaml
richtia: @westonpace Updated! Thanks!

EpsilonPrime (approved):

PR #536, opened 2023-08-04
feat: add extra option for on domain errors in log functions

richtia:
BREAKING CHANGE:  Adding a NULL option to the on_domain_errors.

SQLite returns null for some inputs such as negative infinity

### DIFF
--- a/extensions/functions_logarithmic.yaml
+++ b/extensions/functions_logarithmic.yaml
@@ -12,7 +12,7 @@ scalar_functions:
           rounding:
             values: [ TIE_TO_EVEN, TIE_AWAY_FROM_ZERO, TRUNCATE, CEILING, FLOOR ]
           on_domain_error:
-            values: [ NAN, NONE, ERROR ]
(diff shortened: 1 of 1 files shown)


EpsilonPrime (dismissed):
This appears to be a breaking change as a new producer could send this new option to an old consumer which wouldn't know what to do with it.

Otherwise I'm good to go with this PR.

jacques-n:
Should we use NULL instead of NONE if the behavior is to return a NULL if a domain error occurs?

EpsilonPrime (approved):

richtia:
> I do like NULL. You are right it is a reserved word in YAML but you should be able to make it work with quotes: `[ NAN, "NULL", ERROR ]`

This is what I originally wanted to do as well...but putting NULL around quotes seemed a little different than the convention we were following.  I'm okay with doing this moving forward so long as everyone else is too. Just made the update

westonpace (approved):

PR #537, opened 2023-08-07
feat!: require compound functions names in extension references

vbarua:
BREAKING CHANGE: plans referencing functions using simple
names (e.g. not vs not:bool) will no longer be valid.

### DIFF
--- a/site/docs/extensions/index.md
+++ b/site/docs/extensions/index.md
@@ -1,6 +1,6 @@
 # Extensions
 
-In many cases, the existing objects in Substrait will be sufficient to accomplish a particular use case. However, it is sometimes helpful to create a new data type, scalar function signature or some other custom representation within a system. For that, Substrait provides a number of extension points. 
+In many cases, the existing objects in Substrait will be sufficient to accomplish a particular use case. However, it is sometimes helpful to create a new data type, scalar function signature or some other custom representation within a system. For that, Substrait provides a number of extension points.
 
 ## Simple Extensions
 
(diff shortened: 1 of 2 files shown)


vbarua:
As discusses in Slack, this PR removes the ability to use simple names in function references. This change makes function lookup easier from an implementation perspective.

It also helps avoid a potential pitfall where, for a function with only a single implementation, adding an implementation becomes a breaking change as an plan that referenced that function using its simple name is no longer valid (as the function reference is now ambiguous).

mbrobbel (dismissed):

site/docs/extensions/index.md
```diff
@@ -21,21 +21,21 @@ A Substrait plan can reference one or more YAML files via URI for extension. In
...
+| Function Signature | A function signature compound name as described below.       |
 
 ### Function Signature Compound Names
 
-A YAML file may contain one or more functions by the same name. When only a single function is declared within the file, it can be referenced using the name of that function or a compound name. When more than one function of the same name is declared within a YAML file, the key used in the function extension declaration is a combination of the name of the function along with a list of the required input argument types. Optional arguments are not included in the signature.  The format is as follows:
+A YAML file may contain one or more functions by the same name. The key used in the function extension declaration to reference a function is a combination of the name of the function along with a list of the required input argument types. Optional arguments are not included in the signature. The format is as follows:
```
westonpace, line 28:
```suggestion
A YAML file may contain one or more functions by the same name. The key used in the function extension declaration to reference a function is a combination of the name of the function along with a list of the required input argument types. The format is as follows:
```

I believe we removed optional arguments.

site/docs/extensions/index.md
```diff
@@ -21,21 +21,21 @@ A Substrait plan can reference one or more YAML files via URI for extension. In
...
 <function name>:<short_arg_type0>_<short_arg_type1>_..._<short_arg_typeN>
 ```
 
-Rather than using a full data type representation, the input argument types (`short_arg_type`) are mapped to single-level short name. The mappings are listed in the table below. 
+Rather than using a full data type representation, the input argument types (`short_arg_type`) are mapped to single-level short name. The mappings are listed in the table below.
 
 !!! note
 
-It is required that two function implementations with the same simple name must resolve to different compound names using types. If two function implementations in a YAML file resolve to the same compound name, the YAML file is invalid and behavior is undefined.
+It is required that two functions with the same name must resolve to different compound names using types. If two functions in a YAML file resolve to the same compound name, the YAML file is invalid and behavior is undefined.
```
westonpace, line 38:
```suggestion
Every compound function signature must be unique.  If two function implementations in a YAML file would generate the same compound function signature, then the YAML file is invalid and behavior is undefined.
```

I'm not sure this new wording is better than the old wording because I think we are talking about implementations here (and not functions).  Though I don't like the old wording much either.  How about this as a compromise?

westonpace (changes requested):
Some minor wording suggestions but I agree with the overall change.

jacques-n:
I'm good with this change

westonpace (changes requested):
This is "technically" a breaking change.  Although it's not actually a proto change so I doubt it will affect too many people.  However, it would be good to label it as a breaking change.  Can we do that real quick?  Then I approve and we can merge.

@mbrobbel we might need to update the validator to be more strict now (and can probably simplify some things in the process).  Let me know if I can help.

vbarua:
@westonpace 
Added a BREAKING CHANGE to the commit message. The `feat!` should also mark this as a breaking change. Is that what you had in mind, or is there something else I need to do?

mbrobbel:
> @mbrobbel we might need to update the validator to be more strict now (and can probably simplify some things in the process). Let me know if I can help.

Yes, this being more strict simplifies the validation logic. I haven't had much time lately to make progress on the parser, but I do plan on continuing the effort. I already have [a TODO](https://github.com/substrait-io/substrait-rs/pull/89/files#diff-cee11e25bbd78f2a4a6dc90846a6e6084923c8ec4d572693290dca3c37797c18R185) for this, because I didn't implement compound name parsing yet.

westonpace (approved):

westonpace:
Thanks for writing up this change!

Issue #496, opened 2023-04-26
Allow Function & Type References to Between Extensions


vbarua:
# Problem
It is difficult to re-use existing functions and types from one extension in another extension as there is no mechanism to reference functions and types outside of the current extension.

# Use Cases
* When defining fallback expressions for functions, a fallback expression might wish to use a function that already exists, but in a different extension file.
* A system might wish to add functions or types that depend on types that are defined in a different extension

nielspardon:
It looks to me like this may have been fixed in https://github.com/substrait-io/substrait/pull/265 which was merged on Nov 22, 2023. @vbarua can you verify whether this issue is addressed by #265?

nielspardon:
Took another look at this against `main` (v0.99.0). Confirming my earlier suspicion: the type half is done, and the function half is moot as written.

## Types: addressed by #265

The `dependencies` mechanism is fully in place and has been maintained since — it survived the URI→URN migration in #859/#971:

- **Schema:** the `dependencies` property in `text/simple_extensions_schema.yaml` maps identifier-safe aliases to extension URNs.
- **Grammar:** `grammar/SubstraitType.g4` has `(dependencyAlias=Identifier Dot)? UserDefined Identifier ...`, so `ext.u!point` parses.
- **Docs + example:** [Referencing User-Defined Types](https://substrait.io/extensions/#referencing-user-defined-types) documents it with an ABNF grammar note, and `site/examples/extensions/distance_functions.yaml` is a worked cross-extension example that CI validates against the schema.

That covers the second use case in the description — functions and types that depend on types defined in a different extension.

## Functions: no call site exists to reference them from

The first use case is predicated on fallback expressions. The term appears nowhere in `text/`, `proto/`, or `site/`, and a `git log -S fallback` across all history shows it never did — the feature was never specified. There is no construct anywhere in a simple extension YAML that references a function, so there is currently no place a cross-extension function reference could be written. The `dependencies` mechanism is scoped accordingly.

The adjacent needs are already covered:

- Plans reference functions by extension URN + signature, so using functions from several extensions in one plan was never blocked.
- The one case that does need another extension's function — lambda bodies in function test files — has `### SUBSTRAIT_DEPENDENCY:` (see `tests/README.md`).

## Closing

Closing as addressed by #265. If fallback expressions, or any other construct that needs to name a function from a different extension, is ever proposed, that should carry its own issue with the concrete syntax — reviving this one would just re-litigate a premise that no longer applies.

[...]

PR #543, opened 2023-08-16
feat: add geometric data types and functions

richtia:
PR to add new geometric data types and functions.

### DIFF
--- a/extensions/extension_types.yaml
+++ b/extensions/extension_types.yaml
@@ -8,3 +8,35 @@ types:
     structure:
       start: point
       end: point
+  - name: coordinate
+    structure:
+      x: fp64
+      y: fp64
+  - name: crs_point
+    structure:
+      x: fp64
+      y: fp64
+      crs: string
+  - name: linestring
+    structure:
+      T: List<coordinate>
+  - name: polygon
+    structure:
+      T: List<List<coordinate>>
+  - name: multipoint
+    structure:
+      T: List<coordinate>
+  - name: multilinestring
+    structure:
+      T: List<linestring>
+  - name: multipolygon
+    structure:
+      T: List<polygon>
+  - name: any_simple_geometry
+    parameters:
+      - name: T
+        type: union<crs_point, linestring, polygon, multipoint, multilinestring, multipolygon>
+  - name: geometry
+    parameters:
+      - name: T
+        type: List<any_simple_geometry>
--- a/extensions/extension_types.yaml
+++ b/extensions/extension_types.yaml
@@ -8,35 +8,3 @@ types:
     structure:
       start: point
       end: point
-  - name: coordinate
-    structure:
-      x: fp64
-      y: fp64
-  - name: crs_point
-    structure:
-      x: fp64
-      y: fp64
-      crs: string
-  - name: linestring
-    structure:
-      T: List<coordinate>
-  - name: polygon
-    structure:
-      T: List<List<coordinate>>
-  - name: multipoint
-    structure:
-      T: List<coordinate>
-  - name: multilinestring
-    structure:
-      T: List<linestring>
-  - name: multipolygon
-    structure:
-      T: List<polygon>
-  - name: any_simple_geometry
-    parameters:
-      - name: T
-        type: union<crs_point, linestring, polygon, multipoint, multilinestring, multipolygon>
-  - name: geometry
-    parameters:
-      - name: T
-        type: List<any_simple_geometry>
--- a/extensions/functions_geometry.yaml
+++ b/extensions/functions_geometry.yaml
@@ -3,9 +3,9 @@
 types:
   - name: geometry
     structure: "BINARY"
-    description: |
-      An opaque type that can represent one or many points, lines, or shapes encompassing
-      2, 3 or 4 dimension.
+#    description: |
+#      An opaque type that can represent one or many points, lines, or shapes encompassing
+#      2, 3 or 4 dimension.
 scalar_functions:
   -
     name: "point"
--- a/extensions/functions_geometry.yaml
+++ b/extensions/functions_geometry.yaml
@@ -3,9 +3,9 @@
 types:
   - name: geometry
     structure: "BINARY"
-#    description: |
-#      An opaque type that can represent one or many points, lines, or shapes encompassing
-#      2, 3 or 4 dimension.
+    description: |
+      An opaque type that can represent one or many points, lines, or shapes encompassing
+      2, 3 or 4 dimension.
 scalar_functions:
   -
     name: "point"
--- a/extensions/functions_geometry.yaml
+++ b/extensions/functions_geometry.yaml
@@ -1,0 +1,29 @@
+%YAML 1.2
+---
+scalar_functions:
+  -
+    name: "point"
+    description: >
+      Return a 2D point with the given `x` and `y` coordinate values.
+    impls:
+      - args:
+          - name: x
+            value: fp64
+          - name: y
+            value: fp64
+        return: geometry
+  -
+    name: "makeline"
+    description: >
+      Return a linestring containing the points of the two input geometries.
+      Repeated points at the beginning of input geometries are collapsed to a single point.
+
+      A linestring can be closed or simple.  A closed linestring starts and ends on the same
+      point. A simple linestring does not cross or touch itself.
+    impls:
+      - args:
+          - name: x
+            value: geometry
+          - name: y
+            value: geometry
+        return: geometry
--- a/extensions/functions_geometry.yaml
+++ b/extensions/functions_geometry.yaml
@@ -1,29 +1,35 @@
 %YAML 1.2
 ---
+types:
+  - name: geometry
+    structure: "BINARY"
+#    description: |
+#      An opaque type that can represent one or many points, lines, or shapes encompassing
+#      2, 3 or 4 dimension.
 scalar_functions:
   -
     name: "point"
     description: >
-      Return a 2D point with the given `x` and `y` coordinate values.
+      Returns a 2D point with the given `x` and `y` coordinate values.
     impls:
       - args:
           - name: x
             value: fp64
           - name: y
             value: fp64
-        return: geometry
+        return: u!geometry
   -
     name: "makeline"
     description: >
-      Return a linestring containing the points of the two input geometries.
-      Repeated points at the beginning of input geometries are collapsed to a single point.
+      Returns a linestring connecting the endpoint of geometry `x` to the begin point of
+      geometry `y`. Repeated points at the beginning of input geometries are collapsed to a single point.
 
       A linestring can be closed or simple.  A closed linestring starts and ends on the same
       point. A simple linestring does not cross or touch itself.
     impls:
       - args:
           - name: x
-            value: geometry
+            value: u!geometry
           - name: y
-            value: geometry
-        return: geometry
+            value: u!geometry
+        return: u!geometry
--- a/site/docs/extensions/generate_function_docs.py
+++ b/site/docs/extensions/generate_function_docs.py
@@ -15,7 +15,7 @@ from mdutils.mdutils import MdUtils
 def write_markdown(file_obj: dict, file_name: str) -> None:
     if "types" in file_obj:
         custom_types = file_obj.pop("types")
-        mdFile.new_header(level=2, title=f"Data Types")
+        mdFile.new_header(level=2, title="Data Types")
         for type in custom_types:
             for key, value in type.items():
                 mdFile.new_line(f"{key}: {value}")
--- a/site/docs/extensions/generate_function_docs.py
+++ b/site/docs/extensions/generate_function_docs.py
@@ -1,20 +1,20 @@
 #!/usr/bin/python3
 # SPDX-License-Identifier: Apache-2.0
 
+import filecmp
 import os
-import mkdocs_gen_files
+import tempfile
 from itertools import cycle
 from pathlib import Path
-import tempfile
-import filecmp
 
+import mkdocs_gen_files
 import oyaml as yaml
 from mdutils.mdutils import MdUtils
 
 
 def write_markdown(file_obj: dict, file_name: str) -> None:
-    if 'types' in file_obj:
-        custom_types = file_obj.pop('types')
+    if "types" in file_obj:
+        custom_types = file_obj.pop("types")
         mdFile.new_header(level=2, title=f"Data Types")
         for type in custom_types:
             for key, value in type.items():
--- a/site/docs/extensions/generate_function_docs.py
+++ b/site/docs/extensions/generate_function_docs.py
@@ -13,6 +13,13 @@ from mdutils.mdutils import MdUtils
 
 
 def write_markdown(file_obj: dict, file_name: str) -> None:
+    if 'types' in file_obj:
+        custom_types = file_obj.pop('types')
+        mdFile.new_header(level=2, title=f"Data Types")
+        for type in custom_types:
+            for key, value in type.items():
+                mdFile.new_line(f"{key}: {value}")
+
     for function_classification, value in file_obj.items():
         function_classification_str = function_classification.replace("_", " ").title()
         mdFile.new_header(level=2, title=f"{function_classification_str}")


extensions/functions_geometry.yaml
```diff
@@ -0,0 +1,39 @@
...
+    impls:
+      - args:
+          - name: x
+            value: point
+          - name: y
+            value: point
+        return: linestring
+      - args:
+          - name: x
+            value: multipoint
```
EpsilonPrime, line 30: Is this the same as a list<point>?

extensions/functions_geometry.yaml
```diff
@@ -0,0 +1,39 @@
...
+          - name: y
+            value: multipoint
+        return: linestring
+      - args:
+          - name: x
+            value: linestring
```
EpsilonPrime, line 36: What are the semantics here?  Do we append the two lines assuming there is a join point or does it create a line between x's endpoint and y's startpoint?

extensions/functions_geometry.yaml
```diff
@@ -0,0 +1,39 @@
...
+      Return a point with the given x and y coordinate values.
+    impls:
+      - args:
+          - name: x
+            value: fp32
+          - name: y
+            value: fp32
+        return: point
+  -
+    name: "st_makeline"
```
EpsilonPrime, line 16: why is the point creation function name st_point but this function is st_makeline?

extensions/functions_geometry.yaml
```diff
@@ -0,0 +1,39 @@
...
+---
+scalar_functions:
+  -
+    name: "st_point"
+    description: >
+      Return a point with the given x and y coordinate values.
+    impls:
+      - args:
+          - name: x
+            value: fp32
```
EpsilonPrime, line 11: Could you create a point with integers or fp64?  How about decimal?  Note that the point structure currently is made up of i32.

richtia: oops, those should be integers actually

extensions/functions_geometry.yaml
richtia: yeah, it's referred to as a collection in postgis, but for our purposes I think list would be fine.

extensions/functions_geometry.yaml
richtia: It doesn't necessarily have to join.  We can form an open (or simple) linestring, with 2 linestrings that don't connect.

http://postgis.net/workshops/postgis-intro/geometries.html#linestrings

EpsilonPrime: The semantics would be useful to have in the description.

EpsilonPrime: That's helpful.  The part I was seeking clarification on was whether call st_makeline with two linestrings has any special behavior.  Are the endpoints connected or not as a result of this operation?   It appears that the two liststrings (effectively list<list<point>>) are just concatenated.  To connect the two list strings one would need to add the first new list<point> to the last list<point> and append the rest.

paleolimbot (commented):
Thank you for getting this started!

For now, I wonder if what you want is just:

```
types:
  - name: geometry
    # ... is constraining the storage type mandatory?
```

Getting into the details of exactly how coordinates are stored and which dimensions are stored (e.g., sometimes there's a Z coordinate for elevation or an M coordinate with time) is a rather large can of worms that geospatial people have a lot of opinions about. If you *must* specify some constraint on the type, you might want to rename the type after the encoding (e.g., call it wkb and specify that the encoding is well-known binary).

Having just one type simplifies the function definitions: many are "just" geometry in, geometry out. I don't know if you want to get into the weeds on `list` and `structure` with `point`, `linestring`, `polygon`, `multipoint`, `multielinestring`, and `multipolygon`...every function treats those differently and I'm not sure that any two GIS engine people would give you the same answer if you asked them to come up with that list. Even then you still need a generic `geometry` (e.g., `st_intersection(polygon, polygon)` might give you any combination of points, lines, and/or polygons). You can see how the [GeoArrow memory layout spec](https://github.com/geoarrow/geoarrow/blob/main/format.md) deals with this if you really want to be specific about types.

One choice you may have to make is whether to add a string parameter "crs" to the type. This is sort of like "timezone"...geospatial people care quite a lot about it. PostGIS handles this by making it datum specific (every value has its own uint32 SRID, which refers to a row in an SRID table); the [GeoArrow metadata specification](https://github.com/geoarrow/geoarrow/blob/main/extension-types.md) makes it part of the type. You could also punt and say that if it's missing, the CRS is "OGC:CRS84", which is GIS speak for "longitude/latitude". This is what the [GeoParquet specification](https://github.com/opengeospatial/geoparquet/blob/main/format-specs/geoparquet.md#crs) does.

Another term you may see is "GEOGRAPHY"...PostGIS has this type and so does BigQuery and Snowflake. GeoArrow handles this as a type parameter (`edges: spherical | planar`, with spherical being the one that corresponds to GEOGRAPHY). Geography is always longitude/latitude and never has Z or M values.

extensions/extension_types.yaml
```diff
@@ -4,7 +4,21 @@ types:
...
+      x: point
+      y: point
   - name: line
     structure:
       start: point
       end: point
+  - name: linestring
+    structure:
+      coordinate: point
+    variadic:
```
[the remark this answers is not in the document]
richtia: latest commit removes this data type

extensions/functions_geometry.yaml
EpsilonPrime: Could we correct the naming here to avoid perpetuating their mistake?  st_makeline seems to be the problematic one as it's actually performing a join on two arbitrary geometries instead of two points.  Or is it making a line between the two geometries?

extensions/extension_types.yaml
```diff
@@ -8,3 +8,6 @@ types:
     structure:
       start: point
       end: point
+  - name: geometry
```
EpsilonPrime, line 11: So this represents a series of connected points.  Is there a structure that represents holes such as a donut?

richtia:
@EpsilonPrime i'll update the description for `st_makeline`, the geometries shouldn't be entirely arbitrary, they should both either be a point, a multipoint (collection of points), or a linestring (points that are already connected to each other)

extensions/extension_types.yaml
richtia: I don't think there is a singular structure that can represent a donut, but you could have 2 sets of connected points that each form a circle and one could be entirely within the other.  Does that answer your question?

extensions/functions_geometry.yaml
EpsilonPrime: The ST_ (Spatial Type) prefix isn't really necessary in Substrait as we're already associating all of the geometry functions within the geometry YAML file.  I believe we should probably use MakePoint and MakeLine to be compatible with what PostGIS supports.  Some Substrait implementations will only support the 2 argument version of MakePoint which would be the same as Point.

extensions/extension_types.yaml
EpsilonPrime: I found ST_MakePolygon which can represent a polygon within a geometry which indeed is constructed from 2 sets of connected points.  It looks like everything is defined as a geometry type in POSTGIS.

extensions/functions_geometry.yaml
```diff
@@ -0,0 +1,29 @@
+%YAML 1.2
+---
+scalar_functions:
+  -
+    name: "st_point"
+    description: >
+      Return a point with the given x and y coordinate values.
+    impls:
+      - args:
```
EpsilonPrime, line 9: PostGIS has at least three functions that create points but they all return geometry.  Do we want to also return a geometry?  Do we want to support 3D and 4D points?

extensions/functions_geometry.yaml
paleolimbot: FWIW, `ST_Point()` is in the simple features access standard (which is probably why it is not `ST_MakePoint()`). If you want to correct the name it would probably be `ST_Line()`but PostGIS names are probably the option that is most likely to get this spec taken seriously by anybody who will actually implement it. You could also just omit any functions you're not sure about.

paleolimbot (commented):
A few more comments!

You may also be interested in the Simple Features Access standard for SQL which defines these functions and their return types (link and a few relevant sections pasted below). The functions that never made it to PostGIS are probably not worth including.

Link: https://portal.ogc.org/files/?artifact_id=25354

<details>

Methods on abstract geometry class:

```
METHOD ST_Dimension()
RETURNS SMALLINT
LANGUAGE SQL
DETERMINISTIC
CONTAINS SQL
RETURNS NULL ON NULL INPUT,

METHOD ST_GeometryType()
RETURNS CHARACTER VARYING(ST_MaxTypeNameLength)
LANGUAGE SQL
DETERMINISTIC
CONTAINS SQL
RETURNS NULL ON NULL INPUT,

METHOD ST_AsText()
RETURNS CHARACTER LARGE OBJECT(ST_MaxGeometryAsText)
LANGUAGE SQL
DETERMINISTIC
CONTAINS SQL
RETURNS NULL ON NULL INPUT
RETURNS NULL ON NULL INPUT,

METHOD ST_AsBinary()
RETURNS BINARY LARGE OBJECT(ST_MaxGeometryAsBinary)
LANGUAGE SQL
DETERMINISTIC
CONTAINS SQL
RETURNS NULL ON NULL INPUT,

METHOD ST_IsEmpty()
RETURNS INTEGER
LANGUAGE SQL
DETERMINISTIC
CONTAINS SQL
RETURNS NULL ON NULL INPUT,

METHOD ST_IsSimple()
RETURNS INTEGER
LANGUAGE SQL
DETERMINISTIC
CONTAINS SQL
RETURNS NULL ON NULL INPUT,

METHOD ST_Boundary()
RETURNS ST_Geometry
LANGUAGE SQL
DETERMINISTIC
CONTAINS SQL
RETURNS NULL ON NULL INPUT,
METHOD ST_Envelope()
RETURNS ST_Polygon
LANGUAGE SQL
DETERMINISTIC
CONTAINS SQL
RETURNS NULL ON NULL INPUT,

METHOD ST_Equals(ageometry ST_Geometry)
RETURNS INTEGER
LANGUAGE SQL
DETERMINISTIC
CONTAINS SQL
RETURNS NULL ON NULL INPUT,

METHOD ST_Disjoint(ageometry ST_Geometry)
RETURNS INTEGER
LANGUAGE SQL
DETERMINISTIC
CONTAINS SQL
RETURNS NULL ON NULL INPUT,

METHOD ST_Intersects (ageometry ST_Geometry)
RETURNS INTEGER
LANGUAGE SQL
DETERMINISTIC
CONTAINS SQL
RETURNS NULL ON NULL INPUT,

METHOD ST_Touches(ageometry ST_Geometry)
RETURNS INTEGER
LANGUAGE SQL
DETERMINISTIC
CONTAINS SQL
RETURNS NULL ON NULL INPUT,

METHOD ST_Crosses(ageometry ST_Geometry)
RETURNS INTEGER
LANGUAGE SQL
DETERMINISTIC
CONTAINS SQL
RETURNS NULL ON NULL INPUT,

METHOD ST_Within (ageometry ST_Geometry)
RETURNS INTEGER
LANGUAGE SQL
DETERMINISTIC
CONTAINS SQL
RETURNS NULL ON NULL INPUT,

METHOD ST_Contains(ageometry ST_Geometry)
RETURNS INTEGER
LANGUAGE SQL
DETERMINISTIC
CONTAINS SQL
RETURNS NULL ON NULL INPUT,

METHOD ST_Overlaps(ageometry ST_Geometry)
RETURNS INTEGER
LANGUAGE SQL
DETERMINISTIC
CONTAINS SQL
RETURNS NULL ON NULL INPUT,

METHOD ST_Relate(ageometry ST_Geometry, amatrix CHARACTER(9))
RETURNS INTEGER
LANGUAGE SQL
DETERMINISTIC
CONTAINS SQL
RETURNS NULL ON NULL INPUT,

METHOD ST_Distance(ageometry ST_Geometry)
RETURNS DOUBLE PRECISION
LANGUAGE SQL
DETERMINISTIC
CONTAINS SQL
RETURNS NULL ON NULL INPUT,

METHOD ST_Intersection(ageometry ST_Geometry)
RETURNS ST_Geometry
LANGUAGE SQL
DETERMINISTIC
CONTAINS SQL
RETURNS NULL ON NULL INPUT,

METHOD ST_Difference(ageometry ST_Geometry)
RETURNS ST_Geometry
LANGUAGE SQL
DETERMINISTIC
CONTAINS SQL
RETURNS NULL ON NULL INPUT,

METHOD ST_Union(ageometry ST_Geometry)
RETURNS ST_Geometry
LANGUAGE SQL
DETERMINISTIC
CONTAINS SQL
RETURNS NULL ON NULL INPUT,

METHOD ST_SymDifference (ageometry ST_Geometry)
RETURNS ST_Geometry
LANGUAGE SQL
DETERMINISTIC
CONTAINS SQL
RETURNS NULL ON NULL INPUT,

METHOD ST_Buffer (adistance DOUBLE PRECISION)
RETURNS ST_Geometry
LANGUAGE SQL
DETERMINISTIC
CONTAINS SQL
RETURNS NULL ON NULL INPUT,

METHOD ST_Buffer ( adistance DOUBLE PRECISION,
aunit CHARACTER VARYING(ST_MaxUnitNameLength))
RETURNS ST_Geometry
LANGUAGE SQL
DETERMINISTIC
CONTAINS SQL
RETURNS NULL ON NULL INPUT,

METHOD ST_ConvexHull()
RETURNS ST_Geometry
LANGUAGE SQL
DETERMINISTIC
CONTAINS SQL
RETURNS NULL ON NULL INPUT
```

...there are more but you probably don't need all of them?

</details>

extensions/extension_types.yaml
```diff
@@ -8,6 +8,27 @@ types:
...
+  - name: polygon
+    structure:
+      T: List<List<coordinate>>
+  - name: multipoint
     structure:
-      coordinates: List<point>
+      T: List<coordinate>
+  - name: multilinestring
+    structure:
+      T: List<List<coordinate>>
```
EpsilonPrime, line 26: This could be written in terms of LineString.

extensions/extension_types.yaml
```diff
@@ -8,6 +8,27 @@ types:
...
+  - name: multipolygon
+    structure:
+      T: List<List<List<coordinate>>>
+  - name: geometry
+    parameters:
+      - name: T
+        type: union<point, linestring, polygon, multipoint, multilinestring, multipolygon>
+    variadic: true
```
EpsilonPrime, line 34: I believe we only want to have one argument so variadic should be false.

extensions/extension_types.yaml
```diff
@@ -8,3 +8,26 @@ types:
...
+  - name: multilinestring
+    structure:
+      T: List<linestring>
+  - name: multipolygon
+    structure:
+      T: List<polygon>
+  - name: geometry
+    parameters:
+      - name: T
+        type: union<point, linestring, polygon, multipoint, multilinestring, multipolygon>
```
paleolimbot, line 33: In the spec we're currently negotiating the language, but right now it looks like what you have as `geometry` here is what we are currently calling `mixed` and `geometry` would be `List<mixed>` ( https://github.com/geoarrow/geoarrow/pull/43 ). This is needed to handle a situation where you have (e.g.) a polygon that intersects another polygon in two places: one that just touches one vertex and in another place that might intersect along an edge (in which case you'd have the result of `st_intersects()` that contains one point and one linestring and you need a single output type).

extensions/extension_types.yaml
```diff
@@ -8,3 +8,26 @@ types:
     structure:
       start: point
       end: point
+  - name: coordinate
+    structure:
+      x: fp64
+      y: fp64
```
paleolimbot, line 14: I know it seems redundant, but in negotiating the language of the spec it was decided that "point" is different than "coordinate" ( https://github.com/geoarrow/geoarrow/pull/26#discussion_r983610927 ). A "point" is basically "coordinate" + a string "crs" parameter..."point" can also be empty/null but "coordinate" cannot.

extensions/extension_types.yaml
```diff
@@ -27,7 +32,11 @@ types:
   - name: multipolygon
     structure:
       T: List<polygon>
+  - name: mixed
```
EpsilonPrime, line 35: Mixed could apply to any type.  Perhaps make this the any_geometry or any_simple_geometry type?  Fortunately this is internal only so folks shouldn't need to reference it if it has a complicated name.

extensions/extension_types.yaml
```diff
@@ -27,7 +32,11 @@ types:
...
+    parameters:
+      - name: T
+        type: union<crs_point, linestring, polygon, multipoint, multilinestring, multipolygon>
   - name: geometry
     parameters:
       - name: T
-        type: union<point, linestring, polygon, multipoint, multilinestring, multipolygon>
+        type: List<mixed>
```
EpsilonPrime, line 42: Geometry should allow for individual items as well based on the usage we're seeing.   Perhaps we need both mixed and List<mixed> here?

richtia: hmm....so 2 geometry types? one list<mixed> and one just mixed?  Or is there a way for it being one or the other?

EpsilonPrime: A union would allow it to contain both lists and non-lists.  I'm now second guessing if we want an intermediate type though.  Might just be cleaner for one level of union.

extensions/extension_types.yaml
EpsilonPrime: Odd that a coordinate is the pair x,y, tuple x,y,z, and whatever the 4d version is.  That means you can't refer to the X coordinate without confusion (it is just the one fp64 value here).

paleolimbot: I'm not sure I follow (this yaml syntax is new to me so I may have suggested something incorrect?). It is true that there are two possible *three*-dimensional dimension types (XYZ and XYM) and so referring to the third dimension requires some context if you want to do it by name. M geometries don't come up very often (although it can happen during a coordinate transform).

extensions/extension_types.yaml
paleolimbot: I'm not sure I love the name 'mixed' either but it *is* what we're using in GeoArrow at the moment. It's mostly just a way to get to the `geometry` type.

extensions/extension_types.yaml
paleolimbot: I am almost positive that `list<mixed>` is what you want here (if Substrait insists on taking a stance on storage types, which it seems like it has to?).

extensions/extension_types.yaml
paleolimbot: If Substrait has to take a stand on storage AND you want to support 3D geometry (and/or 4D geometry?) you may need a lot of type definitions (`linestring_z`, `polygon_z`, `coordinate_z`, etc.). It would probably be cleaner (and more realistic?) if the generic "geometry" type definition is allowed to be opaque.

EpsilonPrime: I've been convinced.  I also agree we should remove the other types in favor of one opaque geometry type.  Where we would take in a point, polygon, or whatever we can take in geometry and always return a geometry.

extensions/functions_geometry.yaml
```diff
@@ -0,0 +1,29 @@
...
+    name: "point"
+    description: >
+      Return a 2D point with the given `x` and `y` coordinate values.
+    impls:
+      - args:
+          - name: x
+            value: fp64
+          - name: y
+            value: fp64
+        return: geometry
```
vbarua, line 14: `geometry` as a type would only be valid if it was a core substrait type (e.g `i64`). To reference a user-defined type, the syntax would be `u!geometry`.

Currently though, there's a limitation where types from one extension file can't be referenced from a different extension file: https://github.com/substrait-io/substrait/issues/496

This means that you can't reference the `geometry` type from `extension_types.yaml` (or really any of the types) in this file.

One way to work around this for now would be to define `geometry` and the other types we need directly in this file.

extensions/functions_geometry.yaml
EpsilonPrime: Agreed, sticking to two 2D for now is fine.

EpsilonPrime (approved):
Looks like this should work nicely.  Thanks for handling all the iterations.

extensions/functions_geometry.yaml
richtia: Thanks!  Made some updates!

extensions/functions_geometry.yaml
```diff
@@ -0,0 +1,35 @@
...
+      geometry `y`. Repeated points at the beginning of input geometries are collapsed to a single point.
+
+      A linestring can be closed or simple.  A closed linestring starts and ends on the same
+      point. A simple linestring does not cross or touch itself.
+    impls:
+      - args:
+          - name: x
+            value: u!geometry
+          - name: y
+            value: u!geometry
```
jorisvandenbossche, line 34: Small nitpick (and I also don't know if substrait has something like positional args where the actual name doesn't matter, in which case it is only for documentation): you might want to use different argument names than `x` and `y`, since naively that would easily be interpreted as x and y coordinates (like in `point()`), while here it's rather "point 1" and "point 2". 
So I would rather go with something like `geom1`/`geom2`, or if preferring something generic, eg `a`/`b`

richtia: Ahh...I think I missed this when we decided to change to using the opaque geometry type.  Thanks for catching this!  I'm addressing it in the PR: https://github.com/substrait-io/substrait/pull/554/files

PR #544, opened 2023-08-17
feat: add radians and degrees functions

richtia:
PR to add radians and degrees functions

### DIFF
--- a/extensions/functions_arithmetic.yaml
+++ b/extensions/functions_arithmetic.yaml
@@ -644,6 +644,34 @@ scalar_functions:
             values: [ NAN, ERROR ]
         return: fp64
   -
+    name: "radians"
+    description: >
+      Converts angle `x` in degrees to radians.
+
+    impls:
+      - args:
+          - name: x
+            value: fp32
+        return: fp64
+      - args:
+          - name: x
+            value: fp64
+        return: fp64
+  -
+    name: "degrees"
+    description: >
+      Converts angle `x` in radians to degrees.
+
+    impls:
+      - args:
+          - name: x
+            value: fp32
+        return: fp64
+      - args:
+          - name: x
+            value: fp64
+        return: fp64
+  -
     name: "abs"
     description: >
       Calculate the absolute value of the argument.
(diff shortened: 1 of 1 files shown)


EpsilonPrime (dismissed):
Is it worth putting these functions in a new file (extensions/geometric.yaml perhaps) to keep the file size down?  Two small functions aren't going to make much of a difference though.

richtia:
> Is it worth putting these functions in a new file (extensions/geometric.yaml perhaps) to keep the file size down? Two small functions aren't going to make much of a difference though.

I thought these made more sense going next to the trig functions.  I was planning on adding mostly geospatial functions from postgis to the geometric yaml.

EpsilonPrime:
Works for me.

extensions/functions_arithmetic.yaml
```diff
@@ -643,6 +643,34 @@ scalar_functions:
...
+  -
+    name: "radians"
+    description: >
+      Converts angle `x` in degrees to radians.
+
+    impls:
+      - args:
+          - name: x
+            value: fp32
+        return: fp64
```
gforsyth, line 655: Most extension functions avoid this kind of implicit upcasting -- is there a particular reason we want this to return `fp64` for a `fp32` input?

gforsyth (commented):
Small question about the output types -- and should we also add in `rounding` behavior for the floating point args?

extensions/functions_arithmetic.yaml
richtia: oh yeah...we shouldn't be implicit upcasting here.  I'll update it!  I'll also add the rounding options.

gforsyth (approved):
thanks @richtia!  this looks good to me!

@EpsilonPrime -- does this need PMC approval since it's not changing existing functionality?

EpsilonPrime:
> thanks @richtia! this looks good to me!
> 
> @EpsilonPrime -- does this need PMC approval since it's not changing existing functionality?

Technically probably since it adds functionality.  But given that this isn't controversial I don't think anyone would mind.

EpsilonPrime:
> > thanks @richtia! this looks good to me!
> > @EpsilonPrime -- does this need PMC approval since it's not changing existing functionality?
> 
> Technically probably since it adds functionality. But given that this isn't controversial I don't think anyone would mind.

Hmm, but a reread of substrait.io/governance does explicitly state that addition of a non-breaking function merely requires one committer to approve.  So we should be good to merge this one.

PR #545, opened 2023-08-21
docs: clarify compound extension signature naming for bools

gforsyth:
#537 clarified that compound extension signatures are now required.
I noticed that the table specifying how to type those signatures didn't 
include booleans -- I _think_ it's probably `bool` but either way it 
should be included.

Also in this PR, a quick indentation fix so that the `Note:` admonitions
 render correctly (there needs to be blank line and then an indented block)


gforsyth:
Current website




Fixed admonition

EpsilonPrime (approved):

PR #552, opened 2023-09-12
feat: initial commit for adding geometry accessor functions

richtia:
PR to geometry accessor geospatial functions

### DIFF
--- a/extensions/functions_geometry.yaml
+++ b/extensions/functions_geometry.yaml
@@ -123,7 +123,8 @@ scalar_functions:
     name: "dimension"
     description: >
       Return the dimension of the input geometry.  If the input is a collection of geometries,
-      return the largest dimension from the collection.
+      return the largest dimension from the collection. Dimensionality is determined by
+      the complexity of the input and not the coordinate system being used.
 
       POINT   - 0
       LINE    - 1
--- a/extensions/functions_geometry.yaml
+++ b/extensions/functions_geometry.yaml
@@ -33,3 +33,103 @@ scalar_functions:
           - name: y
             value: u!geometry
         return: u!geometry
+  -
+    name: "x_coordinate"
+    description: >
+      Return the x coordinate of the point.  Return null if not available.
+    impls:
+      - args:
+          - name: point
+            value: u!geometry
+        return: fp64
+  -
+    name: "y_coodinate"
+    description: >
+      Return the y coordinate of the point.  Return null if not available.
+    impls:
+      - args:
+          - name: point
+            value: u!geometry
+        return: fp64
+  -
+    name: "numpoints"
+    description: >
+      Return the number of points in the geometry.  The geometry should be an linestring
+      or circularstring.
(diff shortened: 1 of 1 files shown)


EpsilonPrime (commented):
As this is an initial set of functions there are obviously going to be some missing such as the ability to access the third and fourth dimension of a point.  This is fine.

I'm still wondering if we want to absorb the legacy names including deprecated behaviors or want to create a cleaner version (that should still have a one to one correspondence to the non-deprecated postgis implementation).

extensions/functions_geometry.yaml
```diff
@@ -0,0 +1,103 @@
...
+    name: "st_issimple"
+    description: >
+      Return true if the geometry does not self intersect.
+    impls:
+      - args:
+          - value: geometry
+            name: geom
+        return: boolean
+  -
+    name: "st_isring"
```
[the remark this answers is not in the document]
richtia: I also agree that the prefix isn't useful here, but I figured i'd just follow the same names as used in PostGIS.  I think @paleolimbot mentioned that it might be better to keep the same names in order to be taken seriously?  I'm open to either though.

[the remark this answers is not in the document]
paleolimbot: I think either is ok...when I was referring to the name similarity I meant the part that comes after the st_. I might perhaps even lean slightly towards dropping the prefix.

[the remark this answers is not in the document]
richtia: ahh gotcha...I'll go ahead and drop the prefix in both the open PRs then!

EpsilonPrime (approved):

EpsilonPrime (dismissed):
Looks good once the remaining typo is fixed.

extensions/functions_geometry.yaml
```diff
@@ -33,3 +33,103 @@ scalar_functions:
...
+      input geometry is a point or a line, the returned geometry can also be a point or line.
+    impls:
+      - args:
+          - name: geom
+            value: u!geometry
+        return: u!geometry
+  -
+    name: "dimension"
+    description: >
+      Return the dimension of the input geometry.  If the input is a collection of geometries,
```
EpsilonPrime, line 125: The examples below don't necessarily imply dimension.  Could we clarify if a line in 3D space considered one dimensional or three?

richtia: I believe it would still be one dimensional.  I don't think this function takes into consideration the dimensionality of the space, only the geometry itself.  Maybe @paleolimbot would have a better idea on this.

EpsilonPrime: I think all we need to do is add a line that it refers to the complexity of the figures in the geometry and not the coordinate system being used.

richtia: Sounds good.  Just updated.

EpsilonPrime (approved):

PR #554, opened 2023-09-14
feat: add geometry editor functions

richtia:
PR to add geometry editor functions

### DIFF
--- a/extensions/functions_geometry.yaml
+++ b/extensions/functions_geometry.yaml
@@ -19,7 +19,7 @@ scalar_functions:
             value: fp64
         return: u!geometry
   -
-    name: "makeline"
+    name: "make_line"
     description: >
       Returns a linestring connecting the endpoint of geometry `geom1` to the begin point of
       geometry `geom2`. Repeated points at the beginning of input geometries are collapsed to a single point.
@@ -43,7 +43,7 @@ scalar_functions:
             value: u!geometry
         return: fp64
   -
-    name: "y_coodinate"
+    name: "y_coordinate"
     description: >
       Return the y coordinate of the point.  Return null if not available.
     impls:
@@ -52,7 +52,7 @@ scalar_functions:
             value: u!geometry
         return: fp64
   -
-    name: "numpoints"
+    name: "num_points"
     description: >
       Return the number of points in the geometry.  The geometry should be an linestring
       or circularstring.
@@ -62,7 +62,7 @@ scalar_functions:
             value: u!geometry
         return: i64
   -
-    name: "isempty"
+    name: "is_empty"
     description: >
       Return true is the geometry is an empty geometry.
     impls:
@@ -71,7 +71,7 @@ scalar_functions:
             value: u!geometry
         return: boolean
   -
-    name: "isclosed"
+    name: "is_closed"
     description: >
       Return true if the geometry's start and end points are the same.
     impls:
@@ -80,7 +80,7 @@ scalar_functions:
             value: geometry
         return: boolean
   -
-    name: "issimple"
+    name: "is_simple"
     description: >
       Return true if the geometry does not self intersect.
     impls:
@@ -89,7 +89,7 @@ scalar_functions:
             value: u!geometry
         return: boolean
   -
-    name: "isring"
+    name: "is_ring"
     description: >
       Return true if the geometry's start and end points are the same and it does not self
       intersect.
@@ -99,7 +99,7 @@ scalar_functions:
             value: u!geometry
         return: boolean
   -
-    name: "geometrytype"
+    name: "geometry_type"
     description: >
       Return the type of geometry as a string.
     impls:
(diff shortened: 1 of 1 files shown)


extensions/functions_geometry.yaml
```diff
@@ -0,0 +1,68 @@
+%YAML 1.2
+---
+scalar_functions:
+  -
+    name: "isvalid"
+    description: >
+      Return true if the input geometry is a valid 2D geometry.
+
+      For 3 dimensional and 4 dimensional geometries, the validity is still only tested
```
EpsilonPrime, line 9: If we're going to have this restriction it'd be nice to either have 2d in the name or an option that specifies the number of dimensions to check against.

extensions/functions_geometry.yaml
```diff
@@ -0,0 +1,68 @@
...
+      - args:
+          - value: geometry
+            name: geom_collection
+          - value: i8
+            name: type
+        return: geometry
+  -
+    name: "flip_coordinates"
+    description: >
+      Return a version of the input geometry with the X and Y axis flipped.
```
EpsilonPrime, line 45: This appears to be another 2D operation.  We should probably state what happens to the other dimensions and/or rename the function to clarify its 2D nature.

extensions/functions_geometry.yaml
```diff
@@ -0,0 +1,68 @@
...
+      - args:
+          - value: geometry
+            name: geom_collection
+        return: geometry
+  -
+    name: "remove_repeated_points"
+    description: >
+      Return a version of the input geometry with duplicate consecutive points removed.
+
+      If the `tolerance` argument is provided, consecutive points within the tolerance
```
EpsilonPrime, line 56: What happens if a 2D point and a 3D point near the zero plane are nearby?  Are those eligible to be considered duplicates or does the fact that they have different dimensionalities exclude them from being considered?

extensions/functions_geometry.yaml
```diff
@@ -0,0 +1,68 @@
...
+      distance of one another are considered to be duplicates.
+    impls:
+      - args:
+          - value: geometry
+            name: geom
+        return: geometry
+      - args:
+          - value: geometry
+            name: geom
+          - value: fp32
```
EpsilonPrime, line 66: Is fp32 sufficient or do we want to support fp64 as well?

extensions/functions_geometry.yaml
richtia: That's a good point.  I also noticed that even though postGIS sets this restriction (https://postgis.net/docs/ST_IsValid.html), some other backends don't specify the dimensionality.  For example: https://docs.snowflake.com/en/sql-reference/functions/st_isvalid

@paleolimbot, do you have any thoughts on this?

paleolimbot: Snowflake only supports 2D, so they probably don't need to specify that.

As with all of these PRs, I would urge that Substrait copies PostGIS' (or simple features access) naming + functionality identically. Many of those decisions are a result of years of negotiation and there is a long history of database engineers coming in with a new hot take on how it should be done (e.g., the built-in Postgres geometry types that aren't part of PostGIS and SQLite's geo module support, neither of which are in common use).

That's a long way of saying I agree with what you have in this PR 🙂

extensions/functions_geometry.yaml
paleolimbot: This seems undocumented pretty much everywhere but it seems to be just 2D in the PostGIS implementation ( https://github.com/libgeos/geos/blob/main/src/operation/valid/RepeatedPointRemover.cpp#L50-L56 ). This function isn't in simple features access and was only added to a relatively new version of GEOS (so pretty low priority if it needs to be dropped).

extensions/functions_geometry.yaml
richtia: Updated to support fp64

EpsilonPrime (dismissed):

EpsilonPrime (approved):

extensions/functions_geometry.yaml
```diff
@@ -21,16 +21,16 @@ scalar_functions:
   -
     name: "makeline"
     description: >
-      Returns a linestring connecting the endpoint of geometry `x` to the begin point of
-      geometry `y`. Repeated points at the beginning of input geometries are collapsed to a single point.
+      Returns a linestring connecting the endpoint of geometry `geom1` to the begin point of
```
richtia, line 24: @jorisvandenbossche

EpsilonPrime (dismissed):

extensions/functions_geometry.yaml
jorisvandenbossche: Thanks!

extensions/functions_geometry.yaml
```diff
@@ -134,3 +134,71 @@ scalar_functions:
           - name: geom
             value: u!geometry
         return: i8
+  -
+    name: "isvalid"
```
jorisvandenbossche, line 138:
```suggestion
    name: "is_valid"
```

?

EpsilonPrime (approved):

paleolimbot (approved):

PR #555, opened 2023-09-18
fix!: specify nullability for is_not_distinct_from

vbarua:
BREAKING CHANGE: nullability of is_not_distinct_from has changed


mbrobbel (approved):
I think this should be marked as a breaking change.

EpsilonPrime (approved):

PR #556, opened 2023-09-19
feat: geometry processing functions

richtia:
PR to add geometry processing functions

### DIFF
--- a/extensions/functions_geometry.yaml
+++ b/extensions/functions_geometry.yaml
@@ -227,7 +227,7 @@ scalar_functions:
       - args:
           - value: geometry
             name: geom
-        return: crs_point
+        return: geometry
   -
     name: "minimum_bounding_circle"
     description: >
--- a/extensions/functions_geometry.yaml
+++ b/extensions/functions_geometry.yaml
@@ -214,26 +214,26 @@ scalar_functions:
       a negative `buffer_radius` will always return an emtpy geometry.
     impls:
       - args:
-          - value: geometry
-            name: geom
-          - value: fp64
-            name: buffer_radius
-        return: geometry
+          - name: geom
+            value: u!geometry
+          - name: buffer_radius
+            value: fp64
+        return: u!geometry
   -
     name: "centroid"
     description: >
       Return a point which is the geometric center of mass of the input geometry.
     impls:
       - args:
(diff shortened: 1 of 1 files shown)


extensions/functions_geometry.yaml
```diff
@@ -0,0 +1,37 @@
...
+        return: geometry
+  -
+    name: "centroid"
+    description: >
+      Return a point which is the geometric center of mass of the input geometry.
+    impls:
+      - args:
+          - value: geometry
+            name: geom
+        return: crs_point
```
EpsilonPrime, line 28: Based on other feedback it seems we should always take in geometry and return geometry types.  Then we don't need to care about the internal behavior of the type at all.

richtia: sounds good! updated!

EpsilonPrime (dismissed):

extensions/functions_geometry.yaml
```diff
@@ -0,0 +1,37 @@
...
+      shrink instead of expand.  A negative `buffer_radius` may shrink the geometry completely,
+      in which case an empty geometry is returned. For input the geometries of points or lines,
+      a negative `buffer_radius` will always return an emtpy geometry.
+    impls:
+      - args:
+          - value: geometry
+            name: geom
+          - value: fp64
+            name: buffer_radius
+        return: geometry
```
vbarua, line 19: What geometry type is this referencing?

richtia: the `geometry` type is supposed to be a list of simple geometries defined in this PR: https://github.com/substrait-io/substrait/pull/543/files#diff-130b5d19a0015ea70d9adfa93d28ab07d70d340f02085bbeefcf1b875b154d84

vbarua: Ah, will comment on that PR.

vbarua: https://github.com/substrait-io/substrait/pull/543/files#r1332285773

EpsilonPrime (dismissed):

EpsilonPrime (dismissed):

EpsilonPrime (approved):

paleolimbot (approved):

Issue #456, opened 2023-03-01
Add a proto definition for nested loop join


westonpace:
This node is [described](https://substrait.io/relations/physical_relations/#nlj-operator) in the physical relations section of the site.  However, there is no proto spec for it.

PR #561, opened 2023-10-02
feat: add NestedLoopJoinRel definition

danepitkin:
This PR adds the NestedLoopJoinRel proto definition.

Closes https://github.com/substrait-io/substrait/issues/456

### DIFF
--- a/proto/substrait/algebra.proto
+++ b/proto/substrait/algebra.proto
@@ -578,6 +578,31 @@ message HashJoinRel {
   substrait.extensions.AdvancedExtension advanced_extension = 10;
 }
 
+// The nested loop join (NLJ) operator will hold the entire right input and iterate over it using the
+// left input, evaluating the join expression on the Cartesian product of all rows.
+message NestedLoopJoinRel {
+  RelCommon common = 1;
+  Rel left = 2;
+  Rel right = 3;
+  Expression join_filter = 4;
+
+  JoinType type = 5;
+
+  enum JoinType {
+    JOIN_TYPE_UNSPECIFIED = 0;
+    JOIN_TYPE_INNER = 1;
+    JOIN_TYPE_OUTER = 2;
+    JOIN_TYPE_LEFT = 3;
+    JOIN_TYPE_RIGHT = 4;
+    JOIN_TYPE_LEFT_SEMI = 5;
+    JOIN_TYPE_RIGHT_SEMI = 6;
+    JOIN_TYPE_LEFT_ANTI = 7;
+    JOIN_TYPE_RIGHT_ANTI = 8;
+  }
+
+  substrait.extensions.AdvancedExtension advanced_extension = 10;
+}
+
 // The merge equijoin does a join by taking advantage of two sets that are sorted on the join keys.
 // This allows the join operation to be done in a streaming fashion.
 message MergeJoinRel {
--- a/proto/substrait/algebra.proto
+++ b/proto/substrait/algebra.proto
@@ -585,7 +585,7 @@ message NestedLoopJoinRel {
   RelCommon common = 1;
   Rel left = 2;
   Rel right = 3;
-  Expression join_filter = 4;
+  Expression expression = 4;
 
   JoinType type = 5;
 
--- a/proto/substrait/algebra.proto
+++ b/proto/substrait/algebra.proto
@@ -433,8 +433,8 @@ message Rel {
     CrossRel cross = 12;
     //Physical relations
     HashJoinRel hash_join = 13;
-    NestedLoopJoinRel nested_loop_join = 18;
     MergeJoinRel merge_join = 14;
+    NestedLoopJoinRel nested_loop_join = 18;
     ConsistentPartitionWindowRel window = 17;
     ExchangeRel exchange = 15;
     ExpandRel expand = 16;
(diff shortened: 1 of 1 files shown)


proto/substrait/algebra.proto
```diff
@@ -433,6 +433,7 @@ message Rel {
     CrossRel cross = 12;
     //Physical relations
     HashJoinRel hash_join = 13;
+    NestedLoopJoinRel nested_loop_join = 18;
```
EpsilonPrime, line 436: Given that this is the third defined join relation perhaps it should come after MergeJoinRel (here in Rel and as defined later)?

EpsilonPrime (approved):
Thanks!

proto/substrait/algebra.proto
```diff
@@ -605,6 +606,31 @@ message MergeJoinRel {
...
 }
 
+// The nested loop join (NLJ) operator will hold the entire right input and iterate over it using the
+// left input, evaluating the join expression on the Cartesian product of all rows.
+message NestedLoopJoinRel {
+  RelCommon common = 1;
+  Rel left = 2;
+  Rel right = 3;
+  Expression expression = 4;
+
```
westonpace, line 616: Should you add post_join_filter?

proto/substrait/algebra.proto
```diff
@@ -605,6 +606,31 @@ message MergeJoinRel {
...
+  enum JoinType {
+    JOIN_TYPE_UNSPECIFIED = 0;
+    JOIN_TYPE_INNER = 1;
+    JOIN_TYPE_OUTER = 2;
+    JOIN_TYPE_LEFT = 3;
+    JOIN_TYPE_RIGHT = 4;
+    JOIN_TYPE_LEFT_SEMI = 5;
+    JOIN_TYPE_RIGHT_SEMI = 6;
+    JOIN_TYPE_LEFT_ANTI = 7;
+    JOIN_TYPE_RIGHT_ANTI = 8;
```
westonpace, line 628: I [historically](https://github.com/substrait-io/substrait/pull/467) don't like anti joins because they are not well defined.

For example, a semi join can be implemented as:

```
for l_row in left:
  for r_row in right:
    result = expression(l_row, r_row):
    if result == True:
      yield l_row
```

However, when we look at the anti join:

```
for l_row in left:
  for r_row in right:
    result = expression(l_row, r_row)
    if ???:
      yield l_row
```

Should `???` be `result == False` or should it be `result != True`?  Specifically the question is around how we handle the case where `result` is `null`.

Do you want to take a stab at either picking one of these and clearly documenting this or adding a new proto field allowing the user to choose between these two behaviors?

That being said, I won't hold this issue up over this, so if you'd rather punt on the question then that is fine too.

westonpace (changes requested):
Do you want to also add this to the markdown?  It can go on the physical relations page next to the [hash equijoin](https://substrait.io/relations/physical_relations/#hash-equijoin-operator).

proto/substrait/algebra.proto
danepitkin: According to the existing NLJ definition, it doesn't require one. I think that makes sense because anything that goes in a post_join_filter could also just be included in the expression IIUC.

proto/substrait/algebra.proto
danepitkin: Very illustrative point, thank you. I am in favor of allowing the user to define NULL behavior for the anti-join. However, I would prefer to push this to a separate PR where JoinType is refactored out for all of NLJ/Hash/SM operators (I assume this should be applied to all three). I can create a separate issue for that if you think it's appropriate?

danepitkin (commented):
Hey @westonpace! Thanks for the review 😁 

The NLJ operator is already defined in the markdown: https://substrait.io/relations/physical_relations/#nlj-operator

Is there anything you'd change about it?

proto/substrait/algebra.proto
danepitkin: I created https://github.com/substrait-io/substrait/issues/563

proto/substrait/algebra.proto
westonpace: Good point.  I believe you are correct.

proto/substrait/algebra.proto
```diff
@@ -605,6 +606,31 @@ message MergeJoinRel {
   substrait.extensions.AdvancedExtension advanced_extension = 10;
 }
 
+// The nested loop join (NLJ) operator will hold the entire right input and iterate over it using the
+// left input, evaluating the join expression on the Cartesian product of all rows.
+message NestedLoopJoinRel {
+  RelCommon common = 1;
+  Rel left = 2;
+  Rel right = 3;
+  Expression expression = 4;
```
westonpace, line 615:
```suggestion
  // optional, defaults to true (a cartesian join)
  Expression expression = 4;
```

westonpace (approved):
One last suggestion and then I'm satisfied.

proto/substrait/algebra.proto
westonpace: Although this same point would apply to the logical operator as well :thinking:

danepitkin: Haha, interesting... 😅

EpsilonPrime (approved):

westonpace (approved):

vbarua (approved):
Changes looks good to me!

vbarua:
This PR has 3 SMC votes. Will merge it.

Issue #563, opened 2023-10-13
Refactor JoinType for physical operators


danepitkin:
HashEquiJoin, SortMergeJoin, and NestedLoopJoin all duplicate the JoinType:

```
  enum JoinType {
    JOIN_TYPE_UNSPECIFIED = 0;
    JOIN_TYPE_INNER = 1;
    JOIN_TYPE_OUTER = 2;
    JOIN_TYPE_LEFT = 3;
    JOIN_TYPE_RIGHT = 4;
    JOIN_TYPE_LEFT_SEMI = 5;
    JOIN_TYPE_RIGHT_SEMI = 6;
    JOIN_TYPE_LEFT_ANTI = 7;
    JOIN_TYPE_RIGHT_ANTI = 8;
  }
```

Let's create a shared JoinType for physical operators.

Let's also add JOIN_TYPE_RIGHT_ANTI_WITH_NULL and JOIN_TYPE_RIGHT_ANIT_WITHOUT_NULL options for left/right anti joins. See this [discussion](https://github.com/substrait-io/substrait/pull/561/files#r1358361888) for more detail.

PR #572, opened 2023-11-08
feat: add PhysicalJoinType

danepitkin:
BREAKING CHANGE: the JoinType for physical operators HashJoin, MergeJoin, and
NestedLoopJoin has been refactored into a shared enum called PhysicalJoinType.
It is wire compatible and a new is_null_aware field is added to the physical join rels.

Closes https://github.com/substrait-io/substrait/issues/563


danepitkin:
Question: Is it better for the original ANTI_JOIN enums to default to NO_NULLS or WITH_NULLs?

proto/substrait/algebra.proto
```diff
@@ -562,19 +576,7 @@ message HashJoinRel {
...
-    JOIN_TYPE_INNER = 1;
-    JOIN_TYPE_OUTER = 2;
-    JOIN_TYPE_LEFT = 3;
-    JOIN_TYPE_RIGHT = 4;
-    JOIN_TYPE_LEFT_SEMI = 5;
-    JOIN_TYPE_RIGHT_SEMI = 6;
-    JOIN_TYPE_LEFT_ANTI = 7;
-    JOIN_TYPE_RIGHT_ANTI = 8;
-  }
+  PhysicalJoinType type = 7;
```
EpsilonPrime, line 579: For protobuf compatibility it is better to make this 8 and mark 7 as deprecated.

EpsilonPrime (commented):
Is null handling part of the type or is it a separate setting/option?

proto/substrait/algebra.proto
EpsilonPrime: I suppose if we keep all of the values the same as you've done here it would be wire compatible which mean we wouldn't need to renumber the field for the "type" change.

danepitkin: Yep, it's still technically wire compatible! The Anti Join will have to default to one of the two options though.

danepitkin:
> Is null handling part of the type or is it a separate setting/option?

This Velox documentation makes me think that these are considered different types of anti joins (rather than a separate config): https://facebookincubator.github.io/velox/develop/anti-join.html

danepitkin:
Hmm, although they did implement it slightly differently, using `JoinType::kAnti` and `JoinType::kNullAwareAnti`, presumably with the NULL config separate.

proto/substrait/algebra.proto
danepitkin: Maybe I change this to be 

```
  PHYSICAL_JOIN_TYPE_LEFT_ANTI = 7;
  PHYSICAL_JOIN_TYPE_RIGHT_ANTI = 8;
  PHYSICAL_JOIN_TYPE_LEFT_ANTI_NULL_AWARE = 9;
  PHYSICAL_JOIN_TYPE_RIGHT_ANTI_NULL_AWARE = 10;
```

?

danepitkin: Updated to this ^

danepitkin:
I see @westonpace 's old PR here does use a `null_is_match` bool instead of expanding the jointype enum https://github.com/substrait-io/substrait/pull/467

danepitkin:
Hmm, maybe it would be better to have `bool null_aware` in each physical join so that other join types can use it.

jacques-n:
This shouldn't be a breaking binary change, right?

proto/substrait/algebra.proto
danepitkin: Updated again to use a separate `is_null_aware` field in the rels

danepitkin:
> This shouldn't be a breaking binary change, right?

Yes, good call!

EpsilonPrime (approved):

proto/substrait/algebra.proto
```diff
@@ -568,19 +580,8 @@ message HashJoinRel {
...
+  bool is_null_aware = 8;
```
vbarua, line 584: What exactly does this boolean enable? It would be good to document it.

westonpace:
I've created a new PR specifically for dealing with the `null_is_match` / `null_aware` stuff at https://github.com/substrait-io/substrait/pull/585

Do you want to rescope this PR to just dealing with the join type?

jacques-n:

PR #578, opened 2023-11-17
docs: faq section + post-join filter question

vbarua:


EpsilonPrime (approved):

jacques-n (commented):
I like it but I also think we need to avoid adding more top-level stuff. We already have too many. Is there way we can have either drop-down navs or left hand navs with this theme? I can't remember but I think there is.

EpsilonPrime:
I agree that we have too many top level items (11).  I was looking to spend more time on the website but I already have 4 pending documentation updates.  I'd lean towards getting this in and fixing the navigation later.

jacques-n (approved):
I'm okay with this going in but we should also work on cleaning things up at the nav level.

PR #580, opened 2023-11-29
feat: add interval multiplication

arkanovicz:
Add multiplication of an interval by an integral number `multiply_interval(x,y)`

+ x can be i8, i16, i32 or i64
+ y can be interval_day or interval_year

### DIFF
--- a/extensions/functions_datetime.yaml
+++ b/extensions/functions_datetime.yaml
@@ -219,7 +219,7 @@ scalar_functions:
             value: interval_day
         return: timestamp
   -
-    name: "multiply_interval"
+    name: "multiply"
     description: Multiply an interval by an integral number.
     impls:
       - args:
--- a/extensions/functions_datetime.yaml
+++ b/extensions/functions_datetime.yaml
@@ -219,6 +219,58 @@ scalar_functions:
             value: interval_day
         return: timestamp
   -
+    name: "multiply_interval"
+    description: Multiply an interval by an integral number.
+    impls:
+      - args:
+          - name: x
+            value: i8
+          - name: y
+            value: interval_day
+        return: interval_day
+      - args:
+          - name: x
+            value: i16
+          - name: y
+            value: interval_day
+        return: interval_day
+      - args:
+          - name: x
+            value: i32
+          - name: y
+            value: interval_day
+        return: interval_day
+      - args:
+          - name: x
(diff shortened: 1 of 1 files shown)


extensions/functions_datetime.yaml
```diff
@@ -218,6 +218,58 @@ scalar_functions:
           - name: y
             value: interval_day
         return: timestamp
+  -
+    name: "multiply_interval"
```
EpsilonPrime, line 222: Should this be just another multiply specialization as add is above?

extensions/functions_datetime.yaml
```diff
@@ -218,6 +218,58 @@ scalar_functions:
...
+        return: interval_day
+      - args:
+          - name: x
+            value: i16
+          - name: y
+            value: interval_day
+        return: interval_day
+      - args:
+          - name: x
+            value: i32
```
EpsilonPrime, line 239: Postgres supports double as well.

EpsilonPrime (commented):
Postgres additionally supports dividing an interval by a double.

Given the commutative nature of multiply we could consider supporting the reverse order for the arguments but I don't see support for that in any existing documentation so it would likely be unused.

extensions/functions_datetime.yaml
westonpace: +1.  It seems a little odd to have `add`, `subtract`, and then `multiply_interval` instead of `multiply`.

arkanovicz: Ok, operator renamed.

extensions/functions_datetime.yaml
arkanovicz: It is apparently the only engine around supporting this. Is it a sufficient reason to add it?

It does so by approximating a month to 30 days whenever necessary...

EpsilonPrime: Hmm, agreed, that's not an ideal behavior to have lying around.

EpsilonPrime (commented):
Looks good to me.  @westonpace Would you like a second pass at this one?

EpsilonPrime (approved):

extensions/functions_datetime.yaml
westonpace: Yikes, thanks for investigating :)

Issue #353, opened 2022-10-05
Modulus function in functions_arithmetic.yaml is missing division-by-zero behavior


jvanstraten:
IIRC there is no PR open for this at this time, so I guess we glossed over it during the consistency pass. It's also missing a specification of behavior for negative numbers.

jordanvieler:
I have been looking into this issue, and it looks like modulo is defined differently by different languages. My impression is that would result in different results for zero and negative inputs. Off my head, handling this would require a convention across all languages for how to perform modulo or an enumerated type of modulo where all languages perform the operation according to the mathematical rules specified in the message.

EpsilonPrime:
We have been adding options to describe the behavior for corner cases (such as division by zero).  We can add those here.

EpsilonPrime:
Addressed by #583.  Thanks @jordanvieler !

PR #583, opened 2023-12-13
fix: renamed modulus to modulo; updated modulo operator defintion

jordanvieler:
BREAKING CHANGE: Renamed modulus to modulo. 

Added options and documentation for the modulo operator as defined in math and comp sci.

Refs: #353

### DIFF
--- a/extensions/functions_arithmetic.yaml
+++ b/extensions/functions_arithmetic.yaml
@@ -323,7 +323,7 @@ scalar_functions:
           overflow:
             values: [ SILENT, SATURATE, ERROR ]
           on_domain_error:
-            values: [ NULL, ERROR ]
+            values: [ "NULL", ERROR ]
         return: i8
       - args:
           - name: x
--- a/extensions/functions_arithmetic.yaml
+++ b/extensions/functions_arithmetic.yaml
@@ -346,7 +346,7 @@ scalar_functions:
           overflow:
             values: [ SILENT, SATURATE, ERROR ]
           on_domain_error:
-            values: [ NAN, ERROR ]
+            values: [ NULL, ERROR ]
         return: i32
       - args:
           - name: x
--- a/extensions/functions_arithmetic.yaml
+++ b/extensions/functions_arithmetic.yaml
@@ -347,7 +347,7 @@ scalar_functions:
           overflow:
             values: [ SILENT, SATURATE, ERROR ]
           on_domain_error:
-            values: [ NULL, ERROR ]
+            values: [ "NULL", ERROR ]
         return: i32
       - args:
           - name: x
--- a/extensions/functions_arithmetic.yaml
+++ b/extensions/functions_arithmetic.yaml
@@ -322,7 +322,7 @@ scalar_functions:
             values: [ TRUNCATE, FLOOR ]
           overflow:
             values: [ SILENT, SATURATE, ERROR ]
-          on_domain_error: values: [ NAN, ERROR ]
+          on_domain_error: values: [ NULL, ERROR ]
         return: i8
       - args:
           - name: x
--- a/extensions/functions_arithmetic.yaml
+++ b/extensions/functions_arithmetic.yaml
@@ -322,7 +322,8 @@ scalar_functions:
             values: [ TRUNCATE, FLOOR ]
           overflow:
             values: [ SILENT, SATURATE, ERROR ]
-          on_domain_error: values: [ NULL, ERROR ]
+          on_domain_error:
+            values: [ NULL, ERROR ]
         return: i8
       - args:
           - name: x
--- a/extensions/functions_arithmetic.yaml
+++ b/extensions/functions_arithmetic.yaml
@@ -334,8 +334,6 @@ scalar_functions:
             values: [ TRUNCATE, FLOOR ]
           overflow:
             values: [ SILENT, SATURATE, ERROR ]
-          on_domain_error:
-            values: [ NAN, ERROR ]
         return: i16
       - args:
           - name: x
--- a/extensions/functions_arithmetic.yaml
+++ b/extensions/functions_arithmetic.yaml
@@ -288,7 +288,7 @@ scalar_functions:
     description: >
       Calculate the remainder (r) when dividing dividend (x) by divisor (y).
 
-      In mathematics, many conventions for the modulo operation exists. In computing the result of a modulo operation
+      In mathematics, many conventions for the modulo operation exists. The result of a modulo operation
       depends on the software implementation and underlying hardware. Substrait is a format for describing compute
       operations on structured data and designed for interoperability. Therefore the user is responsible for determining
       a definition of division as defined by the quotient (q).
--- a/extensions/functions_arithmetic.yaml
+++ b/extensions/functions_arithmetic.yaml
@@ -299,17 +299,13 @@ scalar_functions:
       (3) abs(r) < abs(y)
       where q is the quotient.
 
-      The `quotient` option determines the mathematical definition of quotient to use in the above definition of
+      The `division_type` option determines the mathematical definition of quotient to use in the above definition of
       division.
 
-      When `quotient`=TRUNCATE, q = trunc(x/y).
-      When `quotient`=FLOOR, q = floor(x/y).
-      When `quotient`=CEILING, q = ceil(x/y).
-      When `quotient`=ROUND, q = round(x/y), where the round function is rounding half to even.
-      When `quotient`=EUCLIDIAN, q = sign(y) * floor(x/abs(y))
+      When `division_type`=TRUNCATE, q = trunc(x/y).
+      When `division_type`=FLOOR, q = floor(x/y).
 
-      In the case of TRUNCATE, FLOOR, AND CEILING division: remainder r = x - round_func(x/y)
-      In the case of EUCLIDIAN division: remainder r = x - abs(y)*floor(x/abs(y))
+      In the cases of TRUNCATE and FLOOR division: remainder r = x - round_func(x/y)
 
       The `on_domain_error` option governs behavior in cases where y is 0, y is +/-inf, or x is +/-inf. In these cases
       modulo is undefined.
@@ -322,12 +318,11 @@ scalar_functions:
           - name: y
             value: i8
         options:
-          quotient:
-            values: [ TRUNCATE, FLOOR, CEILING, ROUND, EUCLIDEAN ]
+          division_type:
+            values: [ TRUNCATE, FLOOR ]
           overflow:
             values: [ SILENT, SATURATE, ERROR ]
-          on_domain_error:
-            values: [ NAN, ERROR ]
+          on_domain_error: values: [ NAN, ERROR ]
         return: i8
       - args:
           - name: x
@@ -335,8 +330,8 @@ scalar_functions:
           - name: y
             value: i16
         options:
-          quotient:
-            values: [ TRUNCATE, FLOOR, CEILING, ROUND, EUCLIDEAN ]
+          division_type:
+            values: [ TRUNCATE, FLOOR ]
           overflow:
             values: [ SILENT, SATURATE, ERROR ]
           on_domain_error:
@@ -348,8 +343,8 @@ scalar_functions:
           - name: y
             value: i32
         options:
-          quotient:
-            values: [ TRUNCATE, FLOOR, CEILING, ROUND, EUCLIDEAN ]
+          division_type:
+            values: [ TRUNCATE, FLOOR ]
           overflow:
             values: [ SILENT, SATURATE, ERROR ]
           on_domain_error:
@@ -361,8 +356,8 @@ scalar_functions:
           - name: y
             value: i64
         options:
-          quotient:
-            values: [ TRUNCATE, FLOOR, CEILING, ROUND, EUCLIDEAN ]
+          division_type:
+            values: [ TRUNCATE, FLOOR ]
           overflow:
             values: [ SILENT, SATURATE, ERROR ]
           on_domain_error:
--- a/extensions/functions_arithmetic.yaml
+++ b/extensions/functions_arithmetic.yaml
@@ -288,9 +288,10 @@ scalar_functions:
     description: >
       Calculate the remainder (r) when dividing dividend (x) by divisor (y).
 
-      In mathematics, many conventions for the modulo operation exists. In computing the result of a modulo operation depends on the software implementation
-      and underlying hardware. Substrait is a format for describing compute operations on structured data and designed for
-      interoperability. Therefore the user is responsible for determining a definition of division as defined by the quotient (y).
+      In mathematics, many conventions for the modulo operation exists. In computing the result of a modulo operation
+      depends on the software implementation and underlying hardware. Substrait is a format for describing compute
+      operations on structured data and designed for interoperability. Therefore the user is responsible for determining
+      a definition of division as defined by the quotient (y).
 
       The following basic conditions of division are satisfied:
       (1) q ∈ ℤ (the quotient is an integer)
@@ -298,7 +299,9 @@ scalar_functions:
       (3) abs(r) < abs(y)
       where q is the quotient.
 
-      The `quotient` option determines the mathematical definition of quotient to use in the above definition of division.
+      The `quotient` option determines the mathematical definition of quotient to use in the above definition of
+      division.
+
       When `quotient`=TRUNCATE, q = trunc(x/y).
       When `quotient`=FLOOR, q = floor(x/y).
       When `quotient`=CEILING, q = ceil(x/y).
@@ -308,8 +311,8 @@ scalar_functions:
       In the case of TRUNCATE, FLOOR, AND CEILING division: remainder r = x - round_func(x/y)
       In the case of EUCLIDIAN division: remainder r = x - abs(y)*floor(x/abs(y))
 
-      The `on_domain_error` option governs behavior in cases where y is 0, y is +/-inf, or x is +/-inf.
-      Therefore, modulus is undefined.
+      The `on_domain_error` option governs behavior in cases where y is 0, y is +/-inf, or x is +/-inf. In these cases
+      modulo is undefined.
       The `overflow` option governs behavior when integer overflow occurs.
       If x and y are both 0 or both +/-infinity, behavior will be governed by `on_domain_error`.
     impls:
--- a/extensions/functions_arithmetic.yaml
+++ b/extensions/functions_arithmetic.yaml
@@ -284,32 +284,86 @@ scalar_functions:
             value: fp64
         return: fp64
   -
-    name: "modulus"
-    description: "Get the remainder when dividing one value by another."
+    name: "modulo"
+    description: >
+      Calculate the remainder (r) when dividing dividend (x) by divisor (y).
+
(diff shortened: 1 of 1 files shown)


westonpace:
This is very helpful information, thank you for adding it.  I think this is all valid for modulo.  I have a question though.  Do you know of any query engines today that support more than just `truncate` behavior?

richtia:
> This is very helpful information, thank you for adding it. I think this is all valid for modulo. I have a question though. Do you know of any query engines today that support more than just `truncate` behavior?

I was wondering the same.  Not exactly query engines, but this shows what different programming languages do: https://en.wikipedia.org/wiki/Modulo#In_programming_languages

richtia (approved):

richtia:
Since this is a breaking change, we'll need 2 SMC approvals

jordanvieler:
Per the Wikipedia link, I believe that ANSI SQL uses truncate division. So, presumably, any query engine compliant with the specification would use Truncate. However, the Python % operator utilizes Floored division, while the Rust % Operator utilizes Truncated division. Therefore, care would need to be taken to ensure that a Substrait producer written in Python and a Substrait consumer written in Rust consistently perform the Modulo operation.
This table from [Division and Modulus for Computer Scientists by Daan Leijen](https://www.microsoft.com/en-us/research/wp-content/uploads/2016/02/divmodnote-letter.pdf) illustrates the differences in results from Truncate (T), Floor(F), and Euclidean (E) division. q is the quotient, and r is the remainder.

modulo_table

westonpace:
Thanks for the example cases.

Rust, datafusion, sqlite, postgres, sql server, mysql, spark, and velox all use truncate.
Python, pandas, and cudf are the only examples I can find that use floor, but those are probably enough to justify it.

Let's update the PR to limit the choices to truncate and floor.  Given there are no systems that use ceil/round/euclidian I think the extra information would be confusing to users.

extensions/functions_arithmetic.yaml
```diff
@@ -284,32 +284,89 @@ scalar_functions:
...
       - args:
           - name: x
             value: i8
           - name: y
             value: i8
+        options:
+          quotient:
+            values: [ TRUNCATE, FLOOR, CEILING, ROUND, EUCLIDEAN ]
+          overflow:
+            values: [ SILENT, SATURATE, ERROR ]
```
westonpace, line 328: Can we get an overflow performing modulo on integers if we are only using truncate and floor?

extensions/functions_arithmetic.yaml
```diff
@@ -284,32 +284,89 @@ scalar_functions:
...
+          on_domain_error:
+            values: [ NAN, ERROR ]
```
westonpace, line 330: This implementation is for `i8` and `on_domain_error` does not make sense here.
```suggestion
```

extensions/functions_arithmetic.yaml
```diff
@@ -284,32 +284,89 @@ scalar_functions:
...
             value: i16
           - name: y
             value: i16
+        options:
+          quotient:
+            values: [ TRUNCATE, FLOOR, CEILING, ROUND, EUCLIDEAN ]
+          overflow:
+            values: [ SILENT, SATURATE, ERROR ]
+          on_domain_error:
+            values: [ NAN, ERROR ]
```
westonpace, line 343:
```suggestion
```

extensions/functions_arithmetic.yaml
```diff
@@ -284,32 +284,89 @@ scalar_functions:
...
             value: i32
           - name: y
             value: i32
+        options:
+          quotient:
+            values: [ TRUNCATE, FLOOR, CEILING, ROUND, EUCLIDEAN ]
+          overflow:
+            values: [ SILENT, SATURATE, ERROR ]
+          on_domain_error:
+            values: [ NAN, ERROR ]
```
westonpace, line 356:
```suggestion
```

extensions/functions_arithmetic.yaml
```diff
@@ -284,32 +284,89 @@ scalar_functions:
...
             value: i64
           - name: y
             value: i64
+        options:
+          quotient:
+            values: [ TRUNCATE, FLOOR, CEILING, ROUND, EUCLIDEAN ]
+          overflow:
+            values: [ SILENT, SATURATE, ERROR ]
+          on_domain_error:
+            values: [ NAN, ERROR ]
```
westonpace, line 369:
```suggestion
```

westonpace (changes requested):
If we want `on_domain_error` then we will need to add floating point implementations.  Also see my change from my non-review comment about removing `ceiling`, `round`, and `euclidian`.

jordanvieler:
Thanks for the feedback. I just have a couple questions.

1) How is on_domain_error meant to be defined in substrait? The Modulo function is undefined (return ∅) when the divisor is 0 or arguments approach +/- inf. It is my understanding this is "out of domain." Therefore, it is out of domain. How relevant is float vs integer division in this case?

2) While python uses floor for its % operator in the example, most languages define a set of functions in their standard library. For example, Rust defines rem_euclid() for the Euclidean definition of division. Swift defines remainder(dividingBy:) for a modulo with a rounded definition.

Because of how drastic the results can be when performing these different types of modulo, I think the distinctions are important for Data Science, Statistics, and Machine Learning applications. For example a Data Scientist might specifically call for a Ceil Modulo to be performed from their python data frame library.

extensions/functions_arithmetic.yaml
jordanvieler: Correct me if I am wrong, but I think all modulo can overflow if the dividend is equal to the minimum negative value for a signed int and the divisor is equal to -1.

extensions/functions_arithmetic.yaml
jordanvieler: I think modulo is undefined for a divisor of 0 and args of +/- inf.

westonpace:
> How is on_domain_error meant to be defined in substrait? The Modulo function is undefined (return ∅) when the divisor is 0 or arguments approach +/- inf. Therefore, I believe it is "out of domain" by definition. How relevant is float vs integer division in this case?

Float vs integer is relevant because different implementations (e.g f32 vs i32) have different options.  Historically we've used `overflow=SILENT, SATURATE, ERROR` for integer operations and `on_domain_error=NAN, ERROR` for floating point operations.  We could have done this other ways too.  For example, integer division could have both `overflow=SILENT,SATURATE,ERROR` (to handle `min_int/-1`) and `on_domain_error=SILENT,ERROR` (for division by zero).  We went with the simplest approach to match what was currently being offered by available engines.

The current PR's proposed `on_domain_error` will not work for integer implementations however because `NAN` is not possible.

This does answer my "is overflow possible for modulo?" question though (if we consider division by zero to be overflow).

> While python uses floor for its % operator in the example, many languages also define a set of other modulo functions in their libraries. For example, Rust defines rem_euclid() for the Euclidean definition of division. Swift defines remainder(dividingBy:) for a modulo with a rounded definition. Are these important cases?

Our current charter for standard functions is to only capture functions that a sufficient number of query engines currently provide.  Rust may offer `rem_euclid` but datafusion does not.  Even if one engine were to provide such a method that would be a better case for an engine-specific extension function than a standard Substrait function.

jordanvieler:
> Float vs integer is relevant because different implementations (e.g f32 vs i32) have different options. Historically we've used `overflow=SILENT, SATURATE, ERROR` for integer operations and `on_domain_error=NAN, ERROR` for floating point operations. We could have done this other ways too. For example, integer division could have both `overflow=SILENT,SATURATE,ERROR` (to handle `min_int/-1`) and `on_domain_error=SILENT,ERROR` (for division by zero). We went with the simplest approach to match what was currently being offered by available engines.
> 
> The current PR's proposed `on_domain_error` will not work for integer implementations however because `NAN` is not possible.
> 
> This does answer my "is overflow possible for modulo?" question though (if we consider division by zero to be overflow).
> 
Ok, I see what you are saying about NaN not being relevant for int types. Thank you for clarifying that. However, I think there is a difference between overflow and out-of-domain for modulo. Do we assume that a Substrait consumer will return an Error for out-of-domain values of the modulus function, namely 0, as a divisor? So, no option is necessary, right? I could also imagine a historical argument for MAX_INT being a return value in these cases.

> 
> Our current charter for standard functions is to only capture functions that a sufficient number of query engines currently provide. Rust may offer `rem_euclid` but datafusion does not. Even if one engine were to provide such a method that would be a better case for an engine-specific extension function than a standard Substrait function.

I understand the rationale, but I am approaching this slightly differently. Shouldn't standard Substrait functions be toward creating a platform, language, and system agnostic specification and network protocol for data computing operations? I think only supporting the current behavior of existing query engines could introduce a historical bias and limit use cases. SQL itself is only loosely based on relational algebra.

westonpace:
> Do we assume that a Substrait consumer will return an Error for out-of-domain values of the modulus function, namely 0, as a divisor? So, no option is necessary, right?

Sadly, it seems we are not so lucky.

Postgres, Datafusion, SQL server, velox raise an error
Sqlite, spark, and MySQL return null
Pandas and cudf silently upcast the entire operation to floating point and return NaN (this would not be valid substrait because the return type is not allowed to depend on input and so we can probably ignore this case)

Even ignoring the invalid case we still have to consider error and null options.

> I understand the rationale, but I am approaching this slightly differently. Shouldn't standard Substrait functions be toward creating a platform, language, and system agnostic specification and network protocol for data compute operations? I think only supporting the current behavior of existing query engines could introduce a historical bias and limit use cases. SQL itself is only loosely based on relational algebra.

This is a fair viewpoint but the flipside is that anyone using Substrait to create plans cannot reasonably assume the plan will run.  Either way, this issue is probably not the place to discuss the charter for functions.  I'm afraid I'm rather inflexible.  I've been reviewing based on the last time we discussed this topic which was at https://github.com/substrait-io/substrait/issues/307.

If we want a new charter then we should have a separate discussion where everyone can weigh in.  Either as a new issue, in a community meeting, or on the mailing list.  Discussing the scope of Substrait functions on a per-function / per-PR basis doesn't really work.

jordanvieler:
> Sadly, it seems we are not so lucky.
> 
> Postgres, Datafusion, SQL server, velox raise an error Sqlite, spark, and MySQL return null Pandas and cudf silently upcast the entire operation to floating point and return NaN (this would not be valid substrait because the return type is not allowed to depend on input and so we can probably ignore this case)
> 
> Even ignoring the invalid case we still have to consider error and null options.
> 

Does this mean that a field to specify domain errors is warranted separately from the overflow error we have discussed?

> This is a fair viewpoint but the flipside is that anyone using Substrait to create plans cannot reasonably assume the plan will run. Either way, this issue is probably not the place to discuss the charter for functions. I'm afraid I'm rather inflexible. I've been reviewing based on the last time we discussed this topic which was at #307.

Per the referenced topic it seems like modulo is exactly the pedantic case that @jvanstraten mentions.

> If we want a new charter then we should have a separate discussion where everyone can weigh in. Either as a new issue, in a community meeting, or on the mailing list. Discussing the scope of Substrait functions on a per-function / per-PR basis doesn't really work.

I completely understand. I am relatively new to contributing to open-source. My attempt to resolve the divide-by-zero behavior highlighted these nuances modulo. Would it then make sense to bring this aspect of the discussion somewhere else?

jordanvieler:
For now how does this look?

extensions/functions_arithmetic.yaml
```diff
@@ -284,32 +284,84 @@ scalar_functions:
...
-    name: "modulus"
-    description: "Get the remainder when dividing one value by another."
+    name: "modulo"
+    description: >
+      Calculate the remainder (r) when dividing dividend (x) by divisor (y).
+
+      In mathematics, many conventions for the modulo operation exists. In computing the result of a modulo operation
+      depends on the software implementation and underlying hardware. Substrait is a format for describing compute
+      operations on structured data and designed for interoperability. Therefore the user is responsible for determining
+      a definition of division as defined by the quotient (y).
```
westonpace, line 294:
```suggestion
      a definition of division as defined by the quotient (q).
```

extensions/functions_arithmetic.yaml
```diff
@@ -284,32 +284,84 @@ scalar_functions:
             value: fp64
         return: fp64
   -
-    name: "modulus"
-    description: "Get the remainder when dividing one value by another."
+    name: "modulo"
+    description: >
+      Calculate the remainder (r) when dividing dividend (x) by divisor (y).
+
+      In mathematics, many conventions for the modulo operation exists. In computing the result of a modulo operation
```
westonpace, line 291:
```suggestion
      In mathematics, many conventions for the modulo operation exists. The result of a modulo operation
```
This read a little odd to me.  I wasn't sure if that is "in the field of computing, the result..." or "while computing the result...".  I think simpler might be better here.

extensions/functions_arithmetic.yaml
```diff
@@ -284,32 +284,84 @@ scalar_functions:
...
+
+      When `division_type`=TRUNCATE, q = trunc(x/y).
+      When `division_type`=FLOOR, q = floor(x/y).
+
+      In the cases of TRUNCATE and FLOOR division: remainder r = x - round_func(x/y)
+
+      The `on_domain_error` option governs behavior in cases where y is 0, y is +/-inf, or x is +/-inf. In these cases
+      modulo is undefined.
+      The `overflow` option governs behavior when integer overflow occurs.
+      If x and y are both 0 or both +/-infinity, behavior will be governed by `on_domain_error`.
```
westonpace, line 313: Integers can't represent infinity.  We can either drop all discussion of infinity for now or add `impls` for `fp32` and `fp64`.

extensions/functions_arithmetic.yaml
```diff
@@ -284,32 +284,84 @@ scalar_functions:
...
           - name: x
             value: i8
           - name: y
             value: i8
+        options:
+          division_type:
+            values: [ TRUNCATE, FLOOR ]
+          overflow:
+            values: [ SILENT, SATURATE, ERROR ]
+          on_domain_error: values: [ NAN, ERROR ]
```
westonpace, line 325:
```suggestion
          on_domain_error: values: [ NULL, ERROR ]
```

extensions/functions_arithmetic.yaml
```diff
@@ -284,32 +284,84 @@ scalar_functions:
...
             value: i16
           - name: y
             value: i16
+        options:
+          division_type:
+            values: [ TRUNCATE, FLOOR ]
+          overflow:
+            values: [ SILENT, SATURATE, ERROR ]
+          on_domain_error:
+            values: [ NAN, ERROR ]
```
westonpace, line 338:
```suggestion
            values: [ NULL, ERROR ]
```

extensions/functions_arithmetic.yaml
```diff
@@ -284,32 +284,84 @@ scalar_functions:
...
             value: i32
           - name: y
             value: i32
+        options:
+          division_type:
+            values: [ TRUNCATE, FLOOR ]
+          overflow:
+            values: [ SILENT, SATURATE, ERROR ]
+          on_domain_error:
+            values: [ NAN, ERROR ]
```
westonpace, line 351:
```suggestion
            values: [ NULL, ERROR ]
```

extensions/functions_arithmetic.yaml
```diff
@@ -284,32 +284,84 @@ scalar_functions:
...
             value: i64
           - name: y
             value: i64
+        options:
+          division_type:
+            values: [ TRUNCATE, FLOOR ]
+          overflow:
+            values: [ SILENT, SATURATE, ERROR ]
+          on_domain_error:
+            values: [ NAN, ERROR ]
```
westonpace, line 364:
```suggestion
            values: [ NULL, ERROR ]
```

westonpace (changes requested):
Let's keep both `overflow` and `on_domain_error` but change `NAN` to `NULL` since we are dealing with integer kernels which cannot represent `NAN`.

extensions/functions_arithmetic.yaml
jordanvieler: Ah yes. Thanks for catching that

extensions/functions_arithmetic.yaml
```diff
@@ -284,32 +284,82 @@ scalar_functions:
...
           - name: x
             value: i8
           - name: y
             value: i8
+        options:
+          division_type:
+            values: [ TRUNCATE, FLOOR ]
+          overflow:
+            values: [ SILENT, SATURATE, ERROR ]
+          on_domain_error: values: [ NULL, ERROR ]
```
EpsilonPrime, line 325: Looks like this should be on two lines.

jordanvieler: My Bad Thanks.

jordanvieler:
It seems like NULL is not valid for the YAML Linter.

extensions/functions_arithmetic.yaml
```diff
@@ -284,32 +284,83 @@ scalar_functions:
...
             value: i8
           - name: y
             value: i8
+        options:
+          division_type:
+            values: [ TRUNCATE, FLOOR ]
+          overflow:
+            values: [ SILENT, SATURATE, ERROR ]
+          on_domain_error:
+            values: [ NULL, ERROR ]
```
westonpace, line 326:
```suggestion
            values: [ "NULL", ERROR ]
```

extensions/functions_arithmetic.yaml
```diff
@@ -284,32 +284,83 @@ scalar_functions:
...
             value: i32
           - name: y
             value: i32
+        options:
+          division_type:
+            values: [ TRUNCATE, FLOOR ]
+          overflow:
+            values: [ SILENT, SATURATE, ERROR ]
+          on_domain_error:
+            values: [ NULL, ERROR ]
```
westonpace, line 350:
```suggestion
            values: [ "NULL", ERROR ]
```

extensions/functions_arithmetic.yaml
```diff
@@ -284,32 +284,83 @@ scalar_functions:
...
             value: i64
           - name: y
             value: i64
+        options:
+          division_type:
+            values: [ TRUNCATE, FLOOR ]
+          overflow:
+            values: [ SILENT, SATURATE, ERROR ]
+          on_domain_error:
+            values: [ NULL, ERROR ]
```
westonpace, line 363:
```suggestion
            values: [ "NULL", ERROR ]
```

westonpace (changes requested):
Seems to be an annoying yaml / yaml-validator gotcha.  `NULL` is being interpreted as the data type "null" and not as a string.

westonpace:
The one thing we haven't talked about is whether the name change justifies breaking backwards compatibility.  We have a community meeting coming up on Wednesday and I think we should bring this up there.

jordanvieler:
Is it necessary? I added that change after making a very deep dive into mod and thought I read modulus referred to something else. But now I think it's just preference.

westonpace:
> Is it necessary? I added that change after making a very deep dive into mod and thought I read modulus referred to something else. But now I think it's just preference.

If you're comfortable with the old name then let's keep it.  There are systems out there which are relying on the old name and so the breaking change would be a bit inconvenient.

westonpace (approved):
This looks good to me now.  Thanks for all your work!  @EpsilonPrime did you want to take another look?

EpsilonPrime (approved):

PR #584, opened 2023-12-22
feat: add min/max for datetime types

LeonidChistov:
Add min/max aggregations for datetime types.


westonpace (approved):
The only potential concern I have is that we haven't defined the ordering for intervals very clearly.  However, this is already a concern given we allow comparisons for intervals so I don't see any reason to hold this up over that issue.

richtia (approved):

EpsilonPrime (approved):

PR #585, opened 2024-01-03
feat: add custom equality behavior to the hash/merge join

westonpace:
This PR (hopefully) concludes various discussions around flags such as `null_equals_null`
(Datafusion) and `null_aware` (Velox).  The goal of these flags is to slightly tweak the
definition of "equality" in an equijoin relation.

This PR introduces a new EquiJoinKey message that can be used by physical join relations
to define how keys should be compared.

These custom equality functions are needed in a variety of scenarios:

## Optimizing set operations

Set operations (e.g. set difference) can sometimes be satisfied by an equi-join.  When this
happens the user typically wants the equality comparison to be "is not distinct from"

## Flattening correlated subqueries

Some kinds of correlated subqueries can be removed during optimization and replaced with
an anti-join.  Depending on the original query ("not in" vs "where not exists") there may be
slightly different behaviors with respect to null an we may want to use "might equals" as
the comparison.

## String collations

Collations define the ordering and equality of a column.  Different columns can have different
collations.  The equi-join must use the comparison function defined by the collation.

### DIFF
--- a/proto/substrait/algebra.proto
+++ b/proto/substrait/algebra.proto
@@ -563,12 +563,16 @@ message WriteRel {
 
 // The hash equijoin join operator will build a hash table out of the right input based on a set of join keys.
 // It will then probe that hash table for incoming inputs, finding matches.
+//
+// Two rows are a match if the equality function returns true for all keys
 message HashJoinRel {
   RelCommon common = 1;
   Rel left = 2;
   Rel right = 3;
-  repeated Expression.FieldReference left_keys = 4;
-  repeated Expression.FieldReference right_keys = 5;
+  repeated Expression.FieldReference left_keys = 4 [deprecated = true];
+  repeated Expression.FieldReference right_keys = 5 [deprecated = true];
+  // One or more keys to join on.  The relation is invalid if this is empty.
+  repeated HashJoinKey keys = 8;
   Expression post_join_filter = 6;
 
   JoinType type = 7;
@@ -583,6 +587,44 @@ message HashJoinRel {
     JOIN_TYPE_RIGHT_SEMI = 6;
     JOIN_TYPE_LEFT_ANTI = 7;
     JOIN_TYPE_RIGHT_ANTI = 8;
+  }
+
+  // Most hash-joins will use one of the following equality behaviors.  To avoid the complexity
+  // of a function lookup we define the common behaviors here.
+  enum SimpleEqualityType {
+    // Returns true only if both values are equal and not null
+    EQ = 0;
+    // Returns true if both values are equal and not null
+    // Returns true if both values are null
+    // Returns false if one value is null and the other value is not null
+    //
+    // This can be expressed as a = b OR (isnull(a) AND isnull(b))
+    IS_NOT_DISTINCT_FROM = 1;
+    // Returns true if both values are equal and not null
+    // Returns true if either value is null
+    //
+    // This can be expressed as a = b OR isnull(a = b)
+    MIGHT_EQUAL = 2;
+  }
+
+  // Describes how the relation should consider if two rows are a match
+  message EqualityType {
+    oneof inner_type {
+      // One of the simple equality behaviors is used
+      SimpleEqualityType simple = 0;
+      // A custom equality behavior is used.  This can happen, for example, when using
+      // collations, where we might want to do something like a case-insensitive comparison
+      uint32 custom_function_reference = 1;
+    }
+  }
+
+  message HashJoinKey {
+    // The key to compare from the left table
+    Expression.FieldReference left = 0;
+    // The key to compare from the right table
+    Expression.FieldReference right = 1;
+    // Describes how to compare the two keys, defaults to the equal function
+    EqualityType equality = 2;
   }
 
   substrait.extensions.AdvancedExtension advanced_extension = 10;
--- a/proto/substrait/algebra.proto
+++ b/proto/substrait/algebra.proto
@@ -576,18 +576,18 @@ message EquiJoinKey {
   // of a function lookup we define the common behaviors here.
   enum SimpleEqualityType {
     // Returns true only if both values are equal and not null
-    EQ = 0;
+    SIMPLE_EQUALITY_TYPE_EQ = 0;
     // Returns true if both values are equal and not null
     // Returns true if both values are null
     // Returns false if one value is null and the other value is not null
     //
     // This can be expressed as a = b OR (isnull(a) AND isnull(b))
-    IS_NOT_DISTINCT_FROM = 1;
+    SIMPLE_EQUALITY_TYPE_IS_NOT_DISTINCT_FROM = 1;
     // Returns true if both values are equal and not null
     // Returns true if either value is null
     //
     // This can be expressed as a = b OR isnull(a = b)
-    MIGHT_EQUAL = 2;
+    SIMPLE_EQUALITY_TYPE_MIGHT_EQUAL = 2;
   }
 
   // Describes how the relation should consider if two rows are a match
@@ -601,7 +601,6 @@ message EquiJoinKey {
     }
   }
 }
-
 
 // The hash equijoin join operator will build a hash table out of the right input based on a set of join keys.
 // It will then probe that hash table for incoming inputs, finding matches.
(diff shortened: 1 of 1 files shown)


vibhatha:
@westonpace this would be a breaking change for existing APIs?

westonpace:
> @westonpace this would be a breaking change for existing APIs? 

It would not be.  If a plan contains the old left_keys / right_keys field then the "equal" equality method should be used.  I believe this is documented.

proto/substrait/algebra.proto
```diff
@@ -561,14 +561,63 @@ message WriteRel {
...
+    // This can be expressed as a = b OR (isnull(a) AND isnull(b))
+    SIMPLE_EQUALITY_TYPE_IS_NOT_DISTINCT_FROM = 2;
+    // Returns true if both values are equal and not null
+    // Returns true if either value is null
+    //
+    // This can be expressed as a = b OR isnull(a = b)
+    SIMPLE_EQUALITY_TYPE_MIGHT_EQUAL = 3;
+  }
+
+  // Describes how the relation should consider if two rows are a match
```
EpsilonPrime, line 594: Half the comments in this PR end in a period, the others don't.  Please make them consistent.

EpsilonPrime (approved):
Thanks for driving this one to the finish line!  Follow up work for this will be to update downstream code to start using this (such as the text format).

proto/substrait/algebra.proto
```diff
@@ -561,14 +561,63 @@ message WriteRel {
...
+  // The key to compare from the left table
+  Expression.FieldReference left = 1;
+  // The key to compare from the right table
+  Expression.FieldReference right = 2;
+  // Describes how to compare the two keys, defaults to the equal function
+  EqualityType equality = 3;
+
+  // Most equi-joins will use one of the following equality behaviors.  To avoid the complexity
+  // of a function lookup we define the common behaviors here.
+  enum SimpleEqualityType {
```
EpsilonPrime, line 577: Should we add a bit more language here reminding folks about collations?

jacques-n:
I'm happy with this so +1 for me. @EpsilonPrime items feel like minor changes so I'm fine with or without them before merge.

proto/substrait/algebra.proto
```diff
@@ -561,14 +561,63 @@ message WriteRel {
...
 
+// Equality joins (equi-joins) are a specialization of the general join where the join expression
+// is an series of equality comparisons between fields that are ANDed together.  The exact definition
+// of "equality" is flexible.
+message EquiJoinKey {
+  // The key to compare from the left table
+  Expression.FieldReference left = 1;
+  // The key to compare from the right table
+  Expression.FieldReference right = 2;
+  // Describes how to compare the two keys, defaults to the equal function
```
vbarua, line 572: When you say equals function, I assume you mean `SIMPLE_EQUALITY_TYPE_EQ`. We should specify this explicitly if so.

proto/substrait/algebra.proto
```diff
@@ -594,8 +643,13 @@ message MergeJoinRel {
...
   Rel right = 3;
-  repeated Expression.FieldReference left_keys = 4;
-  repeated Expression.FieldReference right_keys = 5;
+  // These fields are deprecated in favor of `keys`.  If they are set then
+  // the two lists (left_keys and right_keys) must have the same length and
+  // the comparion function is considered to be SimpleEqualityType::EQ
+  repeated Expression.FieldReference left_keys = 4 [deprecated = true];
+  repeated Expression.FieldReference right_keys = 5 [deprecated = true];
+  // One or more keys to join on.  The relation is invalid if this is empty.
+  repeated EquiJoinKey keys = 8;
```
vbarua, line 652: The fact that we're using EquiJoinKey here makes me want to generalize it further to something like `ComparisonJoinKeys`.

My understanding is that hash joins are restricted to equality-like comparisons for lookup purposes, but for merge joins conditions like `a > b` are perfectly valid. This is representable with what you have now with the `custom_function_reference`, but it would mean putting something that is very much _not_ an equality predicate into a bunch of messages prefixed with `Equality*`.

vbarua (commented):
One documentation comment, and one thought about taking this ever so slightly further with more generic language around the comparisons. See my MergeRel comment.

proto/substrait/algebra.proto
westonpace: I think it's ok because users will start at `EqualityType` (now `ComparisonType`) which mentions collations when referring them to this message.

proto/substrait/algebra.proto
westonpace: I removed this line about default entirely.  We very rarely use defaults anywhere else (`buf`'s influence I think) and it's probably best not to imply that we are doing so here.

proto/substrait/algebra.proto
westonpace: This is correct.  I tossed and turned on this because it's not implemented very consistently at the moment.

In spark's merge join this is an expression, and so it can be anything, even things that don't make sense
In duckdb's merge join this is an expression type, which is overly flexible, similar to spark
In both velox and datafusion the merge join can only do equality conditions

I toyed with the idea of making `EquiJoinKey` not generic (make it part of `HashJoinRel` and revert `MergeJoinRel`.  We could then create a `ComparisonJoinKey` for merge join that, in its list of simple comparison types, includes the basic comparison functions.  But I think this is too much redundancy.  Probably best, if we want to include the simple comparison types (`<`, `<=`, `>`, `>=`) then we go ahead and do that in some future PR and let there be the potential to create an invalid plan (a hash join with `<` for example).  You can already create an invalid plan by choosing a custom comparison operator that either isn't binary and boolean or that doesn't agree with the hash function you use.

This is a very long winded way of saying I agree with your proposal and have switched to it.

EpsilonPrime (approved):

proto/substrait/algebra.proto
```diff
@@ -594,8 +647,15 @@ message MergeJoinRel {
...
-  repeated Expression.FieldReference left_keys = 4;
-  repeated Expression.FieldReference right_keys = 5;
+  // These fields are deprecated in favor of `keys`.  If they are set then
+  // the two lists (left_keys and right_keys) must have the same length and
+  // the comparion function is considered to be SimpleEqualityType::EQ
+  repeated Expression.FieldReference left_keys = 4 [deprecated = true];
+  repeated Expression.FieldReference right_keys = 5 [deprecated = true];
+  // One or more keys to join on.  The relation is invalid if this is empty.
+  // If a custom comparison function is used then it must be consistent with
+  // the ordering of the input data.
```
vbarua, line 657: What does it mean to be consistent with the ordering of the input data?

proto/substrait/algebra.proto
```diff
@@ -561,14 +561,67 @@ message WriteRel {
...
-  repeated Expression.FieldReference left_keys = 4;
-  repeated Expression.FieldReference right_keys = 5;
+  // These fields are deprecated in favor of `keys`.  If they are set then
+  // the two lists (left_keys and right_keys) must have the same length and
+  // the comparion function is considered to be SimpleEqualityType::EQ
+  repeated Expression.FieldReference left_keys = 4 [deprecated = true];
+  repeated Expression.FieldReference right_keys = 5 [deprecated = true];
+  // One or more keys to join on.  The relation is invalid if this is empty.
+  // If a custom comparison function is used then it must be consistent with
+  // the hash function used for the keys.
```
vbarua, line 623: What does it mean to be consistent with the hash function used for the keys?

proto/substrait/algebra.proto
westonpace: E.g. if your comparison function is < then your data needs to be ordered in ascending order.  If your comparison function is > then your data needs to be ordered in descending order.

In other words, for a given value on the left, as we are walking the right side, the comparison function should only return false once there are no more records that will return true.

proto/substrait/algebra.proto
westonpace: A hash function must return the same value for two keys that are "equal" or else your hash table will not work correctly.  Therefore, if your equality function is something like "equals ignoring case" then your hash function should also ignore case.

vbarua (approved):
This seems good to me as well. For the questions I had about the meaning of consistency, it could be worth adding the PR as well as it wasn't immediately clear what consistent meant.

westonpace:
> For the questions I had about the meaning of consistency, it could be worth adding the PR as well as it wasn't immediately clear what consistent meant.

I'll open a follow-up PR.

PR #587, opened 2024-01-17
chore: bump node version from 18 to 20 in CI scripts

westonpace:
It seems node 18 is now in maintenance mode and
some tools are dropping support for it

### DIFF
--- a/.github/workflows/pr.yml
+++ b/.github/workflows/pr.yml
@@ -56,7 +56,7 @@ jobs:
       - uses: actions/checkout@v4
       - uses: actions/setup-node@v4
         with:
-          node-version: "18"
+          node-version: "20"
       - run: npm install -g ajv-cli
       - run: |
           set -euo pipefail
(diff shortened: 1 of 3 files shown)


westonpace:
Example failure: 
https://github.com/substrait-io/substrait/actions/runs/7562755823/job/20593855026?pr=586

.github/workflows/pr.yml
```diff
@@ -75,19 +75,19 @@ jobs:
...
     runs-on: ubuntu-latest
     steps:
-    - uses: actions/checkout@v4
-    - name: Install dependencies
-      run: python3 -m pip install --upgrade pip black==22.3.0 flake8==4.0.1
-    - name: Black
-      run: python3 -m black --diff --check .
-    - name: Flake8
-      run: python3 -m flake8 .
+      - uses: actions/checkout@v4
```
EpsilonPrime, line 84: Does this extra indent do anything?

EpsilonPrime (approved):

.github/workflows/pr.yml
westonpace: :grimacing: break everything?  It was not intended.  I'll remove.

westonpace: Actually, it appears it was inserted by my yaml formatter.  Indenting lists is consistent with the rest of the file.  Do you mind if we leave it in?

EpsilonPrime: Yep, if it's consistent that's good.

PR #591, opened 2024-01-23
fix: add missing RelCommon field to WriteRel and DdlRel

joellubi:
Adding the RelCommon field to WriteRel and DdlRel message types so that they can be handled with the same common functionality as other operators.

Addition of a new field is not a breaking change.


EpsilonPrime (approved):

jacques-n (approved):
Looks good. Thanks @joellubi !

westonpace (approved):
Thanks!

PR #594, opened 2024-01-25
[DRAFT] feat: include precision parameter in timestamp types

richtia:
This PR adds a parameter to the `Timestamp` types in order to specify the fractional second precision up
to 9 digits (nanoseconds).  

BREAKING CHANGE: `Timestamp` now has a precision parameter.  Extension functions using timestamp need to be modified to use the new type as an argument.


richtia:
Keeping this as a draft and only making the changes for `Timestamp` as of right now.  Once we agree on what changes need to be made, I'll make the updates for `Timestamp_tz` as well.

extensions/functions_datetime.yaml
```diff
@@ -32,7 +32,8 @@ scalar_functions:
       * SECOND Return the second (0-59).
       * MILLISECOND Return number of milliseconds since the last full second.
       * MICROSECOND Return number of microseconds since the last full millisecond.
-      * SUBSECOND Return number of microseconds since the last full second of the given timestamp.
+      * NANOSECOND Return number of nanoseconds since the last full microsecond.
+      * SUBSECOND Return number of nanoseconds since the last full second of the given timestamp.
```
EpsilonPrime, line 36: Are we changing the behavior of SUBSECOND here?  Would it be safer to add a new option and deprecate SUBSECOND?

proto/substrait/algebra.proto
```diff
@@ -796,7 +796,8 @@ message Expression {
       string string = 12;
       bytes binary = 13;
       // Timestamp in units of microseconds since the UNIX epoch.
-      int64 timestamp = 14;
+      // Deprecated in favor of `Timestamp timestamp`
+      // int64 timestamp = 14 [deprecated = true];
```
EpsilonPrime, line 800: Technically the number would become reserved to prevent reuse.  But usually one goes through a deprecation period to allow for interoperability.  That means we need to have both around for a while.  We could say that consumers always should use matched sets but I don't think that's always possible.

proto/substrait/algebra.proto
```diff
@@ -845,6 +848,11 @@ message Expression {
       int32 scale = 3;
     }
 
+    message Timestamp {
+      // the maximum precision is 9.
+      int32 precision = 1;
```
EpsilonPrime, line 853: Do we also have a value?

EpsilonPrime, line 853: Default value is likely 6 for backwards compatibility (although this is a completely different message type so I'm not sure that matters).

proto/substrait/algebra.proto
```diff
@@ -845,6 +848,11 @@ message Expression {
       int32 scale = 3;
     }
 
+    message Timestamp {
+      // the maximum precision is 9.
```
EpsilonPrime, line 852: Do we only support 6 and 9 (maybe even 3)?  Zero seems right out.

proto/substrait/type.proto
```diff
@@ -99,8 +99,9 @@ message Type {
   }
 
   message Timestamp {
-    uint32 type_variation_reference = 1;
-    Nullability nullability = 2;
+    optional int32 precision = 1;
```
EpsilonPrime, line 102: The numbers here should not be reordered as that affects backwards compatibility.

site/docs/types/type_classes.md
```diff
@@ -31,16 +30,17 @@ Simple type classes are those that don't support any form of configuration. For
...
+| MAP&lt;K, V&gt;              | An unordered list of type K keys with type V values. Keys may be repeated. While the key type could be nullable, keys may not be null.                                                                                                                                                                                                                                | `repeated KeyValue` (in turn two `Literal`s), all key types matching K and all value types matching V
+| TIMESTAMP&lt;P&gt;           | A naive timestamp within [1000-01-01 00:00:00.000000000..9999-12-31 23:59:59.999999999], with precision (P, number of digits) <= 9. Does not include timezone information and can thus not be unambiguously mapped to a moment on the timeline without context. Similar to naive datetime in Python.                                                                  | `int64` microseconds since 1970-01-01 00:00:00.000000 (in an unspecified timezone)
```
EpsilonPrime, line 43: Still refers to microseconds here.  See notes about precision limits elsewhere.

proto/substrait/algebra.proto
vbarua: > That means we need to have both around for a while. We could say that consumers always should use matched sets but I don't think that's always possible.

For migration purposes we need a period of time where both of these timestamps types coexist. To migrate this properly we need to be able to do something like:
1. Update consumers to handle the new timestamp type.
2. Update producers to generate plans with the new timestamp type and remove the old type.
3. Remove code for handling the old timestamp type in consumers.

If we remove this entirely we have to do 1 and 2 simultaneously, which doesn't work super well if they're entirely different systems 😢

richtia: Makes sense.  So we'll need a different name for the new timestamp.  Maybe something like `ParametrizedTimestamp`?

proto/substrait/type.proto
```diff
@@ -159,6 +161,12 @@ message Type {
     Nullability nullability = 4;
   }
 
+  message ParametrizedTimestamp {
```
EpsilonPrime, line 164: Is this a duplicate of the one in parameterized_types.proto?

proto/substrait/type_expressions.proto
```diff
@@ -33,6 +34,7 @@ message DerivationExpression {
     ExpressionVarChar varchar = 22;
     ExpressionFixedBinary fixed_binary = 23;
     ExpressionDecimal decimal = 24;
+    ExpressionParametrizedTimestamp parametrized_timestamp = 40;
```
EpsilonPrime, line 37: Parameterized

site/docs/types/type_classes.md
```diff
@@ -31,16 +31,17 @@ Simple type classes are those that don't support any form of configuration. For
...
+| MAP&lt;K, V&gt;              | An unordered list of type K keys with type V values. Keys may be repeated. While the key type could be nullable, keys may not be null.                                                                                                                                                                                                                                | `repeated KeyValue` (in turn two `Literal`s), all key types matching K and all value types matching V
+| TIMESTAMP&lt;P&gt;           | A naive timestamp within [1000-01-01 00:00:00.000000000..9999-12-31 23:59:59.999999999], with precision (P, number of digits) <= 9. Does not include timezone information and can thus not be unambiguously mapped to a moment on the timeline without context. Similar to naive datetime in Python.                                                                  | `int64` nanoseconds since 1970-01-01 00:00:00.000000000 (in an unspecified timezone)
```
EpsilonPrime, line 44: Probably should have both the old and new types listed here.

proto/substrait/type.proto
richtia: Since parameterized_types.proto already uses Parameterized as a prefix in all the names, it'll be less confusing to use something like PrecisionTimestamp instead

extensions/functions_datetime.yaml
richtia: Yeah, deprecating might be a better.  Should we have a process to be followed for deprecating an option value?

proto/substrait/algebra.proto
richtia: Since parameterized_types.proto already uses `Parameterized` as a prefix in all the names, it'll be less confusing to use something like `PrecisionTimestamp` instead

proto/substrait/algebra.proto
richtia: Not exactly sure why value would be necessary?

EpsilonPrime: If this is the type, there's no need, but it looked like this was replacing the value version of int timestamp.

extensions/functions_datetime.yaml
EpsilonPrime: If the standard definition uses millisecond for SUBSECOND then we might need to leave this version permanently.

proto/substrait/algebra.proto
westonpace: This is a literal (the type is named `ParameterizedPrecisionTimestamp`?.  In other words, if my SQL filter is `event_time < '1999-01-08 04:05:06'` and `event_time` is a "precision timestamp" column then I need to make a Substrait plan that looks like...

`less_than(input_field(0), ???)` where `???` is my timestamp literal.  So we do need a value (and not a precision) here.

We could do something like:

```
      // If the precision is 6 or less then this is the microseconds since the UNIX epoch
      // If the precision is more than 6 then this is the nanoseconds since the UNIX epoch
      uint64 precision_timestamp = 34;
```

site/docs/types/type_classes.md
westonpace: From the proto it looks like we are deprecating the old type?  Are you saying we should list the old type and mark it deprecated?

westonpace: This does not work.  If you store int64 nanoseconds since the epoch then the maximum value you can get is `2262-04-12 01:47:16.854775807 +0200` (slightly larger if you store unsigned but still not large enough).  However, the first sentence claims `within [1000-01-01 00:00:00.000000000..9999-12-31 23:59:59.999999999]`.

So either the literal representation needs to be larger than 8 bytes or the allowable range needs to depend on the precision.

Note that making literals larger than 8 bytes is tempting, but it could lead to system incompatibility.  For example, you could create a Substrait filter `my_date < year_4000` but if you tried to evaluate that on the consumer it might try and convert that literal to its internal representation and return an error (e.g. this is what would happen if you tried to use pandas, datafusion, or pyarrow to evaluate that).

So I think we have to state that the allowable range depends on the precision.

westonpace (changes requested):

extensions/functions_datetime.yaml
```diff
@@ -32,6 +32,7 @@ scalar_functions:
       * SECOND Return the second (0-59).
       * MILLISECOND Return number of milliseconds since the last full second.
       * MICROSECOND Return number of microseconds since the last full millisecond.
+      * NANOSECOND Return number of nanoseconds since the last full microsecond.
       * SUBSECOND Return number of microseconds since the last full second of the given timestamp.
```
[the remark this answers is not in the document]
richtia: I think all these fractional second components actually need to be updated.  It looks like other engines extract the entire seconds portion, including the fractional part.  


```
SELECT EXTRACT(MILLISECONDS FROM TIMESTAMP '2016-12-31 13:30:15.100259');
postgres: 15100.259
duckdb: 15100

SELECT EXTRACT(MICROSECONDS FROM TIMESTAMP '2016-12-31 13:30:15.100259');
postgres: 15100259
duckdb: 15100259
```

Interesting that postgres will still keep the fractional parts after the conversion.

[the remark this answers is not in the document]
westonpace: Ok.  Let's save that for a different PR I guess :)

proto/substrait/algebra.proto
```diff
@@ -796,7 +796,8 @@ message Expression {
       string string = 12;
       bytes binary = 13;
       // Timestamp in units of microseconds since the UNIX epoch.
-      int64 timestamp = 14;
+      // Deprecated in favor of `PrecisionTimestamp precision_timestamp`
```
westonpace, line 799:
```suggestion
      // Deprecated in favor of `precision_timestamp`
```

proto/substrait/algebra.proto
```diff
@@ -807,10 +808,15 @@ message Expression {
...
       Decimal decimal = 24;
+      // If the precision is 6 or less then this is the microseconds since the UNIX epoch
+      // If the precision is more than 6 then this is the nanoseconds since the UNIX epoch
+      uint64 precision_timestamp = 34;
+      uint64 precision_timestamp_tz = 35;
       Struct struct = 25;
       Map map = 26;
       // Timestamp in units of microseconds since the UNIX epoch.
-      int64 timestamp_tz = 27;
+      // Deprecated in favor of `PrecisionTimestampTZ precision_timestamp_tz`
```
westonpace, line 818:
```suggestion
      // Deprecated in favor of `precision_timestamp_tz`
```

proto/substrait/algebra.proto
```diff
@@ -845,6 +851,18 @@ message Expression {
...
+      // Supported values are 0, 3, 6, and 9.  The default is 6.
+      int32 precision = 1;
+    }
+
+    message PrecisionTimestampTZ {
+      // The maximum number of digits allowed in the value.
+      // Supported values are 0, 3, 6, and 9.  The default is 6.
+      int32 precision = 1;
+    }
+
```
westonpace, line 865:
```suggestion
```
I don't believe these messages are needed any longer

proto/substrait/type.proto
```diff
@@ -159,6 +163,18 @@ message Type {
     Nullability nullability = 4;
   }
 
+  message PrecisionTimestamp {
+    optional int32 precision = 1;
```
westonpace, line 167: I don't know that we've used `optional` much elsewhere.  I seem to recall it being largely a no-op as far as protobuf is concerned.  However, I don't see much harm in it.  Maybe document the default (6 I assume) again?

site/docs/types/type_classes.md
```diff
@@ -31,16 +31,18 @@ Simple type classes are those that don't support any form of configuration. For
...
+| PRECISIONTIMESTAMP&lt;P&gt;   | A timestamp with fractional second precision (P, number of digits) <= 9. Does not include timezone information and can thus not be unambiguously mapped to a moment on the timeline without context. Similar to naive datetime in Python.                                                                                                                             | `uint64` nanoseconds since 1970-01-01 00:00:00.000000000 (in an unspecified timezone)
+| PRECISIONTIMESTAMPTZ&lt;P&gt; | A timezone-aware timestamp, with fractional second precision (P, number of digits) <= 9. Similar to aware datetime in Python.                                                                                                                                                                                                                                                                   | `int64` microseconds since 1970-01-01 00:00:00.000000 UTC
```
westonpace, line 45: It might go without saying but should we mention that `P >= 0` as well?

site/docs/types/type_classes.md
```diff
@@ -31,16 +31,18 @@ Simple type classes are those that don't support any form of configuration. For
...
+| MAP&lt;K, V&gt;               | An unordered list of type K keys with type V values. Keys may be repeated. While the key type could be nullable, keys may not be null.                                                                                                                                                                                                                                | `repeated KeyValue` (in turn two `Literal`s), all key types matching K and all value types matching V
+| PRECISIONTIMESTAMP&lt;P&gt;   | A timestamp with fractional second precision (P, number of digits) <= 9. Does not include timezone information and can thus not be unambiguously mapped to a moment on the timeline without context. Similar to naive datetime in Python.                                                                                                                             | `uint64` nanoseconds since 1970-01-01 00:00:00.000000000 (in an unspecified timezone)
```
westonpace, line 44: This looks like `PRECISIONTIMESTAMP` has a literal representation of `uint64 nanoseconds` and `PRECISIONTIMESTAMPTZ` has a literal representation of `uint64 microseconds`.

However, if I'm reading the proto correctly, they both have a literal representation of "either uint64 microseconds or nanoseconds since ..."

westonpace (changes requested):
Thanks again!  Some suggestions and cleanup.

vbarua (commented):
As part of this, do we need add signature names for the new types as well?

https://github.com/substrait-io/substrait/blob/d9b9672fd3c24285afdee9344fc2f4f7fcd70afb/site/docs/extensions/index.md#L73-L74

richtia:
> As part of this, do we need add signature names for the new types as well?
> 
> https://github.com/substrait-io/substrait/blob/d9b9672fd3c24285afdee9344fc2f4f7fcd70afb/site/docs/extensions/index.md#L73-L74

Thanks for the catch!  Added new signature names

site/docs/types/type_classes.md
```diff
@@ -31,16 +31,18 @@ Simple type classes are those that don't support any form of configuration. For
...
+| PRECISIONTIMESTAMP&lt;P&gt;   | A timestamp with fractional second precision (P, number of digits) 0 >= P <= 9. Does not include timezone information and can thus not be unambiguously mapped to a moment on the timeline without context. Similar to naive datetime in Python.                                                                                                                      | `uint64` microseconds or nanorseconds since 1970-01-01 00:00:00.000000000 (in an unspecified timezone)
+| PRECISIONTIMESTAMPTZ&lt;P&gt; | A timezone-aware timestamp, with fractional second precision (P, number of digits) 0 >= P <= 9. Similar to aware datetime in Python.                                                                                                                                                                                                                                  | `uint64` microseconds or nanorseconds since 1970-01-01 00:00:00.000000000 UTC
```
westonpace, line 45:
```suggestion
| PRECISIONTIMESTAMP&lt;P&gt;   | A timestamp with fractional second precision (P, number of digits) 0 >= P <= 9. Does not include timezone information and can thus not be unambiguously mapped to a moment on the timeline without context. Similar to naive datetime in Python.                                                                                                                      | `uint64` microseconds or nanoseconds since 1970-01-01 00:00:00.000000000 (in an unspecified timezone)
| PRECISIONTIMESTAMPTZ&lt;P&gt; | A timezone-aware timestamp, with fractional second precision (P, number of digits) 0 >= P <= 9. Similar to aware datetime in Python.                                                                                                                                                                                                                                  | `uint64` microseconds or nanoseconds since 1970-01-01 00:00:00.000000000 UTC
```

westonpace (dismissed):
One small typo

proto/substrait/type.proto
EpsilonPrime: Optional is from the version 2 of the protobuf format.  These days most of the outside world is using version 3 (as is this file) where optional is assumed.

EpsilonPrime (approved):

EpsilonPrime (approved):

proto/substrait/algebra.proto
```diff
@@ -807,10 +808,15 @@ message Expression {
       VarChar var_char = 22;
       bytes fixed_binary = 23;
       Decimal decimal = 24;
+      // If the precision is 6 or less then this is the microseconds since the UNIX epoch
+      // If the precision is more than 6 then this is the nanoseconds since the UNIX epoch
+      uint64 precision_timestamp = 34;
+      uint64 precision_timestamp_tz = 35;
```
Blizzara, line 814: Should these have been custom types that'd include the precision? Like Decimal/VarChar below. Or how is the consumer mean to infer the precision? The `uint64` presumably means the timestamp value itself

PR #599, opened 2024-02-08
fix: remove function definitions w/ invalid return types

vbarua:
T? and T|? are not valid return types

T|? is equivalent to setting the nullability to MIRROR
T&? is not currently expressible via nullability

BREAKING CHANGE: least_skip_null and greatest_skip_null functions have been removed

### DIFF
--- a/extensions/functions_comparison.yaml
+++ b/extensions/functions_comparison.yaml
@@ -225,18 +225,8 @@ scalar_functions:
           - value: T
         variadic:
           min: 2
-        return: T|?
-  -
-    name: "least_skip_null"
-    description: >-
-      Evaluates each argument and returns the smallest one. The function will return
-      null only if all arguments evaluate to null.
-    impls:
-      - args:
-          - value: T
-        variadic:
-          min: 2
-        return: T&?
+        return: T
+        nullability: MIRROR
   -
     name: "greatest"
     description: >-
@@ -247,15 +237,5 @@ scalar_functions:
           - value: T
         variadic:
           min: 2
-        return: T|?
-  -
-    name: "greatest_skip_null"
-    description: >-
-      Evaluates each argument and returns the largest one. The function will return
-      null only if all arguments evaluate to null.
-    impls:
-      - args:
-          - value: T
-        variadic:
-          min: 2
-        return: T&?
+        return: T
+        nullability: MIRROR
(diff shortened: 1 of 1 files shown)


richtia:
Here's the PR that introduced those for reference: https://github.com/substrait-io/substrait/pull/247

vbarua:
@richtia thanks for linking that 🙇 

Additional context from @westonpace in Slack

> No.  Let's back those out for now and create a todo.  We had some discussion on this topic in the community meeting.  The conclusion was that we do not want to add new syntax to the standard return types.
Instead, those should use the same "return type is a small block of code that you have to evaluate" approach that is used for decimal arithemtic, e.g.
```
        return: |-
          init_scale = max(S1,S2)
          delta = init_prec - 38
          prec = min(init_prec, 38)
          scale_after_borrow = max(init_scale - delta, min_scale)
          DECIMAL<prec, scale>
```

Part of the discussion in Slack was also around adding a new nullability type like `INTERSECTION` for `returns non-null if any of the input arguments is non-null` as comes up in a number of places (i.e. coalesce).

EpsilonPrime (dismissed):

westonpace (dismissed):
I approve if we want to proceed with this.  However, would it be better to keep the methods around and just change the return type to `T`?  Is "slightly inaccurate but usable" better than "accurate but not usable"?

vbarua:
> However, would it be better to keep the methods around and just change the return type to T? Is "slightly inaccurate but usable" better than "accurate but not usable"?

@westonpace I think it makes sense to keep them available so that they can be used. In this case, using nullability MIRROR is weaker than the actual nullability desired so it seems safe.

Made a change to bring them back in, update the return types and include a note about it.

I've created https://github.com/substrait-io/substrait/issues/601 as an issue to track this new desired nullability as well.

westonpace (approved):

westonpace:
If @EpsilonPrime is still good with the updated version I think we are good to go

bvolpato (approved):
LGTM! Thanks.

Note, this part of the description is not true anymore:
> BREAKING CHANGE: least_skip_null and greatest_skip_null functions have been removed

EpsilonPrime (approved):

vbarua:
Will merge this as this has 3 SMC votes (including myself)

PR #602, opened 2024-02-10
docs: clarify nullability information source for empty literals

vbarua:
Looking at the `oneof literal_type` block I noticed that the `empty_list` variant and `empty_map` variant reuse a `Type` message, much like the `null` variant.
```
      Type null = 29; // a typed null literal
      ...
      Type.List empty_list = 31;
      Type.Map empty_map = 32;
```

Like `Type`, `Type.List` and `Type.Map` _also_ declare their nullability directly. To avoid ambiguity, I updated the docs to specify that the nullability of these literals should come from the inner message here, and not from the `nullable` field of the `Literal` message.

### DIFF
--- a/proto/substrait/algebra.proto
+++ b/proto/substrait/algebra.proto
@@ -825,9 +825,11 @@ message Expression {
       UserDefined user_defined = 33;
     }
 
-    // whether the literal type should be treated as a nullable type. Applies to
-    // all members of union other than the Typed null (which should directly
-    // declare nullability).
+    // Whether the literal_type above should be treated as a nullable type.
+    // Applies to all members of the literal_type oneof EXCEPT:
+    //  * Type null             (must be nullable by definition)
+    //  * Type.List empty_list  (use Type.List::nullability)
+    //  * Type.Map empty_map    (use Type.Map::nullability)
     bool nullable = 50;
 
     // optionally points to a type_variation_anchor defined in this plan.


vbarua:
Looking at the `oneof literal_type` block I noticed that the `empty_list` variant and `empty_map` variant reuse a `Type` message, much like the `null` variant.
```
      Type null = 29; // a typed null literal
      ...
      Type.List empty_list = 31;
      Type.Map empty_map = 32;
```

Like `Type`, `Type.List` and `Type.Map` _also_ declare their nullability directly. To avoid ambiguity, I updated the docs to specify that the nullability of these literals should come from the inner message here, and not from the `nullable` field of the `Literal` message.

proto/substrait/algebra.proto
```diff
@@ -819,9 +819,12 @@ message Expression {
...
 
-    // whether the literal type should be treated as a nullable type. Applies to
-    // all members of union other than the Typed null (which should directly
-    // declare nullability).
+    // Whether the literal_type above should be treated as a nullable type.
+    // Applies to all members of the literal_type oneof EXCEPT:
+    //  * Type null
+    //  * Type.List empty_list
+    //  * Type.Map empty_map
+    // which should declare their nullability directly instead.
```
EpsilonPrime, line 827: I need to verify but the text format appears to set both for empty lists and just the type for empty maps (based on its test cases taken from Isthmus).

westonpace (approved):
Thanks for removing the ambiguity

proto/substrait/algebra.proto
vbarua: @EpsilonPrime where you able to verify this?

EpsilonPrime: It shouldn't matter going forward, I will make sure that the text format follows this guideline by adding some tests.

EpsilonPrime (dismissed):

proto/substrait/algebra.proto
EpsilonPrime: For empty lists the following behavior is present for the text format:

```
{}_list<string> -> nothing nullable
{}_list<string?> -> literal is nullable, list is not nullable
{}_list?<string> -> list is nullable, literal is not
{}_list?<string?> -> both are nullable
```

It seems like this nullability covers the nullability of the list even if it's empty.  Even if it's empty is it legal to have a nullable version of it?

EpsilonPrime: Added the empty_map tests in https://github.com/substrait-io/substrait-cpp/pull/97.

EpsilonPrime: The text format is definitely using this nullability over the one in the type.  It's the same behavior it uses for non-empty lists and maps.

proto/substrait/algebra.proto
```diff
@@ -819,9 +819,12 @@ message Expression {
       UserDefined user_defined = 33;
     }
 
-    // whether the literal type should be treated as a nullable type. Applies to
-    // all members of union other than the Typed null (which should directly
-    // declare nullability).
+    // Whether the literal_type above should be treated as a nullable type.
+    // Applies to all members of the literal_type oneof EXCEPT:
+    //  * Type null
```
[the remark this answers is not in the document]
westonpace: Do you want to clarify this further then?  Maybe...

```
Applies to all members of the literal_type oneof EXCEPT:
 * Type null (must be nullable by definition)
 * Type.List empty_list (use Type.List::nullability)
 * Type.Map empty_map (use Type.List::nullability)

**Review by westonpace — DISMISSED** [review]
This wording is ok.  I have a suggestion for something slightly different if we want.

**Comment by vbarua** [comment]
I want to double check something before merging this, will look at it this week.

File: proto/substrait/algebra.proto (hunk 2 above)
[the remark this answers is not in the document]
Comment (jacques-n) [reply proto/substrait/algebra.proto:824]: Agree with @westonpace that we should update the null literal to be defined differently from list/map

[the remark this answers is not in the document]
Comment (vbarua) [reply proto/substrait/algebra.proto]: Pulled in @westonpace suggestions for this.

**Comment by vbarua** [comment]
>I want to double check something before merging this, will look at it this week.

Weston's suggestion clarified this for me. I think this should be good as is with his changes incorporated.

**Comment by EpsilonPrime** [comment]
Will check this on Thursday.

File: proto/substrait/algebra.proto [hunk 3]
```diff
@@ -825,9 +825,11 @@ message Expression {
       UserDefined user_defined = 33;
     }
 
-    // whether the literal type should be treated as a nullable type. Applies to
-    // all members of union other than the Typed null (which should directly
-    // declare nullability).
+    // Whether the literal_type above should be treated as a nullable type.
+    // Applies to all members of the literal_type oneof EXCEPT:
+    //  * Type null             (must be nullable by definition)
+    //  * Type.List empty_list  (use Type.List::nullability)
```
EpsilonPrime, line 831: If the Type.List::nullability is the nullability of the type inside the list then I don't think this applies.

EpsilonPrime: That matches what I believe.  But for empty lists and empty maps why is that any different than populated lists and maps?

westonpace: Empty lists and empty maps have two spots for list nullability (so we need to pick one to be authoritative):

```
// List nullability (authoritative)
Literal::empty_list::list::nullability
// Also list nullability (redundant)
Literal::nullability
```

Populated lists and maps have one spot for list nullability and one spot for item nullability:

```
// Applies to the item
Literal::list::values...nullability
// Applies to the list (only option, thus authoritative)
Literal::nullability
```

EpsilonPrime: Okay, I see it now.  An alternative would be to change the type of of empty_list and empty_map to Type but that doesn't work simply for maps (which would need two nullability values for its key and type).

EpsilonPrime (approved):

westonpace (approved):

PR #603, opened 2024-02-20
docs: specify Substrait library merge policy

vbarua:


vbarua:
Making the current library merge policy more explicit.

EpsilonPrime (approved):

westonpace (commented):
This effectively means that each Substrait library needs two committers to make progress right?  That seems like it could be a barrier for new / less popular libraries.

westonpace (approved):
We can always loosen later if it becomes a problem I suppose.

jacques-n (approved):

PR #610, opened 2024-03-19
feat: add decimal type support for sum0 function

liuyuhanalex:
This PR adds support for decimal types in the sum0 function.
This is currently supported in Calcite.

### DIFF
--- a/extensions/functions_arithmetic_decimal.yaml
+++ b/extensions/functions_arithmetic_decimal.yaml
@@ -149,3 +149,19 @@ aggregate_functions:
         decomposable: MANY
         intermediate: "DECIMAL?<P, S>"
         return: "DECIMAL?<P, S>"
+  - name: "sum0"
+    description: >
+      Sum a set of values. The sum of zero elements yields zero.
+
+      Null values are ignored.
+    impls:
+      - args:
+          - name: x
+            value: "DECIMAL<P, S>"
+        options:
+          overflow:
+            values: [ SILENT, SATURATE, ERROR ]
+        nullability: DECLARED_OUTPUT
+        decomposable: MANY
+        intermediate: "DECIMAL<38,S>"
+        return: "DECIMAL<38,S>"


extensions/functions_arithmetic_decimal.yaml
```diff
@@ -149,3 +149,19 @@ aggregate_functions:
...
+    impls:
+      - args:
+          - name: x
+            value: "DECIMAL<P, S>"
+        options:
+          overflow:
+            values: [ SILENT, SATURATE, ERROR ]
+        nullability: DECLARED_OUTPUT
+        decomposable: MANY
+        intermediate: "DECIMAL?<38,S>"
```
EpsilonPrime, line 166: Based in the discussion in #465 the intermediate and return values should just be DECIMAL<38,S> since this function should not return null.

EpsilonPrime (commented):
Thank you for the contribution!  This version of sum0 does indeed appear to be missing.

extensions/functions_arithmetic_decimal.yaml
liuyuhanalex: Thanks for your review, I have submitted a new commit.

liuyuhanalex: > Based in the discussion in #465 the intermediate and return values should just be DECIMAL<38,S> since this function should not return null.

Thanks for your review, I have submitted a new commit according to the comment. @EpsilonPrime

extensions/functions_arithmetic_decimal.yaml
```diff
@@ -149,3 +149,19 @@ aggregate_functions:
...
+      - args:
+          - name: x
+            value: "DECIMAL<P, S>"
+        options:
+          overflow:
+            values: [ SILENT, SATURATE, ERROR ]
+        nullability: DECLARED_OUTPUT
+        decomposable: MANY
+        intermediate: "DECIMAL<38,S>"
+        return: "DECIMAL<38,S>"
```
vbarua, line 167: Should this return type be `DECIMAL<P,S>`? Does `sum0` change the precision somehow?

xiedeyantu: > Should this return type be `DECIMAL<P,S>`? Does `sum0` change the precision somehow?

I think this behavior should be the same as sum, always plan will do cast later.

EpsilonPrime: It appears that any operation where decimals aren't manipulated by each other return the same precision.  Any time it does, we are returning the max precision (as we do here).  It doesn't seem likely that we'd increase precision by doing a lot of add operations though.  However it is consistent with sum (and the other similar arithmetic operations).  We can address the correctness for all of them together I suppose.

EpsilonPrime (approved):

vbarua (approved):
The changes look reasonable and are consistent with the behaviour of `sum`

PR #618, opened 2024-04-03
docs: clarify difference between any and any1, any2, etc

vbarua:


vbarua:
Following up on a discussion item from a Substrait sync earlier this year around clarifying the differences between using `any` as a type and `any1` as a type.

site/docs/extensions/index.md
```diff
@@ -98,6 +98,34 @@ Rather than using a full data type representation, the input argument types (`sh
...
+- name: bar
+  impls:
+  - args:
+    - name: a
+      value: any1
+    - name: b
+      value: any1
+    return: int64
+```
+The `any[\d]` types (i.e. `any1`, `any2`, etc) impose an additional restriction. Within a single function invocation, all any types with same numeric suffix _must_ be of the same type. In the `bar` function above, arguments `a` and `b` can be bound to any values within a specific function invocation as long as both values have the same type.
```
westonpace, line 128:
```suggestion
The `any[\d+]` types (i.e. `any1`, `any2`, etc) impose an additional restriction. Within a single function invocation, all any types with same numeric suffix _must_ be of the same type. In the `bar` function above, arguments `a` and `b` can be bound to any values within a specific function invocation as long as both values have the same type.
```
Though I'm not sure we'd ever actually accept a function with more than 10 `any` parameters :laughing: so I'm fine if you want to ignore this.

westonpace, line 128: The "bound to any values within a specific function invocation" wording seems a bit confusing.  Maybe...

"arguments `a` and `b` can have any type as long as both types are the same"?

westonpace (dismissed):
Feel free to go with this as-is but I have a few minor suggestions/nits

site/docs/extensions/index.md
vbarua: Changed this to be 
```
The `any[\d+]` types (i.e. `any1`, `any2`, ..., `any9`) ...
```
> Though I'm not sure we'd ever actually accept a function with more than 10 any parameters

If someone ever wants to do to this we can revisit this. I want to avoid making changes to type grammar.

vbarua: Done

westonpace (approved):

EpsilonPrime (approved):

PR #620, opened 2024-04-04
fix: remove implicit casts in trig extension functions

richtia:
This PR removes some implicit casts that exist in the trig functions.

More discussion here:
https://github.com/substrait-io/substrait/issues/251

Most of the trig functions already follow this behavior. This should handle the remaining ones.


EpsilonPrime (approved):

westonpace (approved):
I checked a few systems and they all seem to mirror this behavior.  They pick one type to be their canonical input and the return type is the same.

PR #623, opened 2024-04-11
feat:  add single match join

EpsilonPrime:
This establishes the physical single join relation.  Similar to other joins it uses an expression to compare the left and right inputs.  What sets it apart from other joins is that it requires that each row be involved in at most one match.  This kind of join is commonly used in [unnesting arbitrary subqueries](https://btw-2015.informatik.uni-hamburg.de/res/proceedings/Hauptband/Wiss/Neumann-Unnesting_Arbitrary_Querie.pdf).

### DIFF
--- a/proto/substrait/algebra.proto
+++ b/proto/substrait/algebra.proto
@@ -444,6 +444,7 @@ message Rel {
     HashJoinRel hash_join = 13;
     MergeJoinRel merge_join = 14;
     NestedLoopJoinRel nested_loop_join = 18;
+    SingleJoinRel single_join = 22;
     ConsistentPartitionWindowRel window = 17;
     ExchangeRel exchange = 15;
     ExpandRel expand = 16;
@@ -721,6 +722,18 @@ message NestedLoopJoinRel {
     JOIN_TYPE_LEFT_ANTI = 7;
     JOIN_TYPE_RIGHT_ANTI = 8;
   }
+
+  substrait.extensions.AdvancedExtension advanced_extension = 10;
+}
+
+// A single join enforces that each row is matched at once most counterpart.
+// If this constraint is violated, an error is raised.
+message SingleJoinRel {
+  RelCommon common = 1;
+  Rel left = 2;
+  Rel right = 3;
+  // optional, defaults to true (a cartesian join)
+  Expression expression = 4;
 
   substrait.extensions.AdvancedExtension advanced_extension = 10;
 }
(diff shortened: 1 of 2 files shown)


site/docs/relations/physical_relations.md
```diff
@@ -67,10 +67,35 @@ The merge equijoin does a join by taking advantage of two sets that are sorted o
...
+| Right Keys          | References to the fields to join on in the right input.                                                                                                                                                                 | Required                    |
 | Post Join Predicate | An additional expression that can be used to reduce the output of the join operation post the equality condition. Minimizes the overhead of secondary join conditions that cannot be evaluated using the equijoin keys. | Optional, defaults true.    |
 | Join Type           | One of the join types defined in the Join operator.                                                                                                                                                                     | Required                    |
 
+
+
+## Single Join Operator
+
+The single match join works by enforcing that each row is involved in only a single match.  If this violated, an error
```
westonpace, line 78: Here you say "only a single match"  In the protobuf you say "at most one match".  Which is it?  At most one?  Or exactly one?

If it "at most one" then what happens if there is no match?  Is this an inner join (no row is emitted) or a left join (a row is emitted with the values from the left side and nulls on the right side)?

site/docs/relations/physical_relations.md
```diff
@@ -67,10 +67,35 @@ The merge equijoin does a join by taking advantage of two sets that are sorted o
...
+is raised.  Typical use would be to have one side's duplicates eliminated and then be pushed into a column scan on the
+other side. This operator is useful when converting subqueries into joins.
```
westonpace, line 80: I'm probably dense but I don't understand how this is helpful.  This doesn't seem like it would be useful in eliminating duplicates (an error would be raised).  Are you saying this is somehow used on the deduplicated values?  How so?  What does "pushed into a column scan on the other side" mean?

proto/substrait/algebra.proto
```diff
@@ -444,6 +444,7 @@ message Rel {
     HashJoinRel hash_join = 13;
     MergeJoinRel merge_join = 14;
     NestedLoopJoinRel nested_loop_join = 18;
+    SingleJoinRel single_join = 22;
```
westonpace, line 447: Note that this was previously attempted as a separate join type: https://github.com/substrait-io/substrait/blob/7b794379c51436620515b5ebad7720c502aab991/proto/substrait/algebra.proto#L175

We should either deprecate that join type or collapse this PR into better documenting that join type.  I'm uncertain which approach I'd prefer and think it hinges on the "at most one" or "exactly one" question

westonpace (changes requested):
Thanks for taking this task on.  I've got a first round of questions up.

EpsilonPrime:
As noted above we already have a single join, closing this one as unneeded.

PR #625, opened 2024-04-12
feat: implement mark join

EpsilonPrime:
This establishes the physical mark join relation.  It utilizes a mark (a nullable boolean field reference) to reduce the amount of data that needs to be processed.

### DIFF
--- a/proto/substrait/algebra.proto
+++ b/proto/substrait/algebra.proto
@@ -445,6 +445,7 @@ message Rel {
     HashJoinRel hash_join = 13;
     MergeJoinRel merge_join = 14;
     NestedLoopJoinRel nested_loop_join = 18;
+    MarkJoinRel mark_join = 23;
     ConsistentPartitionWindowRel window = 17;
     ExchangeRel exchange = 15;
     ExpandRel expand = 16;
@@ -722,6 +723,21 @@ message NestedLoopJoinRel {
     JOIN_TYPE_LEFT_ANTI = 7;
     JOIN_TYPE_RIGHT_ANTI = 8;
   }
+
+  substrait.extensions.AdvancedExtension advanced_extension = 10;
+}
+
+// A mark join utilizes a previously applied mark to greatly reduce the
+// input to be processed.  A mark is a nullable boolean field.
+message MarkJoinRel {
+  RelCommon common = 1;
+  Rel left = 2;
+  Rel right = 3;
+  // optional, defaults to true (a cartesian join)
+  Expression expression = 4;
+
+  // A reference to the mark field.
+  Expression.FieldReference mark_field = 6;
 
   substrait.extensions.AdvancedExtension advanced_extension = 10;
 }
--- a/proto/substrait/algebra.proto
+++ b/proto/substrait/algebra.proto
@@ -727,17 +727,16 @@ message NestedLoopJoinRel {
   substrait.extensions.AdvancedExtension advanced_extension = 10;
 }
 
-// A mark join utilizes a previously applied mark to greatly reduce the
-// input to be processed.  A mark is a nullable boolean field.
+// A mark join internally scans the left side, constructing a hash table that
+// is used to mark the right side as having a join partner on the left side. A
+// mark is a nullable boolean field.  The mark join operator is used to
+// implement semi-joins, anti-joins, and other join types that are not equijoins.
 message MarkJoinRel {
   RelCommon common = 1;
   Rel left = 2;
   Rel right = 3;
   // optional, defaults to true (a cartesian join)
   Expression expression = 4;
-
-  // A reference to the mark field.
-  Expression.FieldReference mark_field = 6;
 
   substrait.extensions.AdvancedExtension advanced_extension = 10;
 }
--- a/site/docs/relations/physical_relations.md
+++ b/site/docs/relations/physical_relations.md
@@ -71,6 +71,30 @@ The merge equijoin does a join by taking advantage of two sets that are sorted o
 | Post Join Predicate | An additional expression that can be used to reduce the output of the join operation post the equality condition. Minimizes the overhead of secondary join conditions that cannot be evaluated using the equijoin keys. | Optional, defaults true.    |
 | Join Type           | One of the join types defined in the Join operator.                                                                                                                                                                     | Required                    |
 
+
+
+## Mark Join Operator
+
+A mark join utilizes a previously applied mark to greatly reduce the input to be processed.
+
+| Signature            | Value                                                        |
+| -------------------- | ------------------------------------------------------------ |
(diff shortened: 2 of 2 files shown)


site/docs/relations/physical_relations.md
```diff
@@ -71,6 +71,30 @@ The merge equijoin does a join by taking advantage of two sets that are sorted o
...
+|-----------------|-------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|-------------------------------------------------------------|
+| Left Input      | A relational input.                                                                                                                                                                 | Required                                                    |
+| Right Input     | A relational input.                                                                                                                                                                 | Required                                                    |
+| Mark Reference  | A nullable boolean field reference that is used to filter the right input.  If the mark is null, the row is not included in the join.                                               | Required.                                                   |
```
westonpace, line 93: That's it?  Why not just use a `FilterRel`?

westonpace (commented):
I think I'm probably missing something.

site/docs/relations/physical_relations.md
Mytherin: This description is not correct - a `mark` join does not change the cardinality of the left table (much like a `single` join). The marker is either `true`, `false` or `NULL` depending on if a match is found and/or if any `NULL` values were encountered. Often the marker is used in a filter later on - but this is not required. If the marker is exclusively used in a filter the join can also be converted into a semi join and the marker is not necessary. It's described in [this paper](http://btw2017.informatik.uni-stuttgart.de/slidesandpapers/F1-10-37/paper_web.pdf).

Mytherin (commented):
Thanks - I'm not that well versed in Substrait but the implementation looks good to me. There just seems to be a mistake in the description of what the join does.

site/docs/relations/physical_relations.md
EpsilonPrime: I've updated the documentation.  It hopefully is more clear what the mark join is trying to accomplish.

I've also removed the mark reference as that appears to be an internal detail for the mark join (an internal hash table is constructed instead of the right join input having a new column introduced).

proto/substrait/algebra.proto
```diff
@@ -726,6 +727,20 @@ message NestedLoopJoinRel {
   substrait.extensions.AdvancedExtension advanced_extension = 10;
 }
 
+// A mark join internally scans the left side, constructing a hash table that
```
westonpace, line 730: What does internally here mean?

westonpace, line 730: This claims that the hash table is built from the left side and then the right side is probed.  This is the reverse of the `HashJoinRel`.  If this intentional, then we should point this out (e.g. "Note that the table is built from the left side which is the opposite of the approach taken in the HashJoinRel)

site/docs/relations/physical_relations.md
```diff
@@ -71,6 +71,29 @@ The merge equijoin does a join by taking advantage of two sets that are sorted o
...
+## Mark Join Operator
+
+A mark join internally scans the left side, constructing a hash table that is used to mark the right side as having a join partner on the left side.  This mark can end up being True, False, or NULL.  The NULL mark is used to indicate that the right side does not have a join partner on the left side.  The mark join operator is used to implement semi-joins, anti-joins, and other join types that are not equijoins.
+
+| Signature            | Value                                                        |
+| -------------------- | ------------------------------------------------------------ |
+| Inputs               | 2                                                            |
+| Outputs              | 1                                                            |
+| Property Maintenance | Distribution is maintained. Orderedness is eliminated.       |
+| Direct Output Order  | Same as the [Join](logical_relations.md#join-operator) operator. |
```
westonpace, line 85: Does the mark join actually join two tables or does it just add the mark column?  In other words, if I apply mark join to a left table `l_key, l_payload` and a right table `r_key, r_payload` then is the output `l_key, l_payload, r_key, r_payload, mark` or is it `l_key, l_payload, mark` (which would be very strange if the left table is the build side, see previous comment)

From my read of the mark join algorithm in the paper it appears that a mark join does not actually join two tables, but only adds the mark (the algorithm scans `R` and `S`.  Only rows from `R` are inserted into the hash table `H`.  The emit at the end is a scan of `H` which only contains rows from `R`.  This is reinforced by the notation of the emit (`emit r, marker of r`)  Compare this to the full outer hash join algorithm which contains lines like `emit r,s` and `emit r,NULL`)

site/docs/relations/physical_relations.md
```diff
@@ -71,6 +71,29 @@ The merge equijoin does a join by taking advantage of two sets that are sorted o
 | Post Join Predicate | An additional expression that can be used to reduce the output of the join operation post the equality condition. Minimizes the overhead of secondary join conditions that cannot be evaluated using the equijoin keys. | Optional, defaults true.    |
 | Join Type           | One of the join types defined in the Join operator.                                                                                                                                                                     | Required                    |
 
+
+
+## Mark Join Operator
+
+A mark join internally scans the left side, constructing a hash table that is used to mark the right side as having a join partner on the left side.  This mark can end up being True, False, or NULL.  The NULL mark is used to indicate that the right side does not have a join partner on the left side.  The mark join operator is used to implement semi-joins, anti-joins, and other join types that are not equijoins.
```
westonpace, line 78: Given the mark might be null then it sounds like this is a right join.  In other words, there is one output row for each row on the right side.  Is that true?  Can we describe that more clearly?

westonpace, line 78: > The NULL mark is used to indicate that the right side does not have a join partner on the left side.

This is not correct.  From the paper:

>  the marker may not only be TRUE (had a join partner) or FALSE (had no
join partner), but also NULL (had a join partner where the comparison result is NULL, but
none where the comparison is TRUE)

westonpace (changes requested):
I made a scan of the paper and I'm not sure it agrees with what is proposed.  Note, the paper also has some mention / notion of "compare type" and suggests the join might actually do a cast of sorts on the key columns.  If this is true, do we need to include this compare type concept in order for these relations to be usable by duckdb?

site/docs/relations/physical_relations.md
Mytherin: You are correct - the mark join only adds the mark column, the output is `l_key, l_payload, mark` (where left is the probe side). The cardinality of the left side (probe side) is preserved - i.e. the mark join neither removes nor adds additional rows.

site/docs/relations/physical_relations.md
Mytherin: A regular right/outer join can still emit multiple output rows per input row, a mark join preserves the cardinality of the left hand side exactly.

EpsilonPrime:

PR #628, opened 2024-04-16
feat: add i64 variant for exp, ln, log10, log2 and logb functions

JazabC:
This PR adds `i64` as an argument to `exp`, `ln`, `log10`, `log2` and `logb` functions

### DIFF
--- a/extensions/functions_logarithmic.yaml
+++ b/extensions/functions_logarithmic.yaml
@@ -7,6 +7,17 @@ scalar_functions:
     impls:
       - args:
           - name: x
+            value: i64
+        options:
+          rounding:
+            values: [ TIE_TO_EVEN, TIE_AWAY_FROM_ZERO, TRUNCATE, CEILING, FLOOR ]
+          on_domain_error:
+            values: [ NAN, "NULL", ERROR ]
+          on_log_zero:
+            values: [NAN, ERROR, MINUS_INFINITY]
+        return: fp64
+      - args:
+          - name: x
             value: fp32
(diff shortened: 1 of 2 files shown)


vbarua (approved):
Thanks for adding these!

The definitions follow a similar pattern as `sqrt`, which only defines an `i64` variant (alongside `fp32` and `fp64`). 

It might make sense to add variants for `i8`, `i16` and `i32`, but it's an open-question (IMO) about whether these should return `fp32` or `fp64`. I think the `i64` to `fp64` make sense generally.

EpsilonPrime (approved):

extensions/functions_logarithmic.yaml
```diff
@@ -5,6 +5,17 @@ scalar_functions:
...
     description: "Natural logarithm of the value"
     impls:
+      - args:
+          - name: x
+            value: i64
+        options:
+          rounding:
+            values: [ TIE_TO_EVEN, TIE_AWAY_FROM_ZERO, TRUNCATE, CEILING, FLOOR ]
+          on_domain_error:
+            values: [ NAN, "NULL", ERROR ]
```
westonpace, line 15: Domain error here is negative integers?

westonpace (approved):
One question but I think the answer is just "yes".

extensions/functions_logarithmic.yaml
JazabC: yes, the domain error will be negative integers/floats

PR #629, opened 2024-04-17
fix: use any1 instead of T in function extensions

vbarua:
The functions:
* least
* least_skip_null
* greatest
* greatest_skip_null
index_in

were declared as consuming a value of type T, which is not a valid type


EpsilonPrime (approved):

richtia (approved):

westonpace (approved):
Thanks for cleaning this up!

PR #631, opened 2024-04-18
fix: duplicate declaration of min:timestamp & max:timestamp

amol-:
Addresses a duplication of `min` and `max` function overloads
for timestamp types.

The functions are declared in `arithmetic` extensions: 

* `min` -> https://github.com/amol-/substrait/blob/main/extensions/functions_arithmetic.yaml#L1217-L1230
* `max` -> https://github.com/amol-/substrait/blob/main/extensions/functions_arithmetic.yaml#L1217-L1230

but are also declared in `datetime` extensions:

* `min` -> https://github.com/amol-/substrait/blob/main/extensions/functions_datetime.yaml#L807-L820
* `max` -> https://github.com/amol-/substrait/blob/main/extensions/functions_datetime.yaml#L852-L865

This seems to be a source of confusion for a system loading those extensions definition, which one of the two should be considered valid?

The PR addresses this by preserving only the definitions in `datetime` for those argument types.

### DIFF
--- a/extensions/functions_arithmetic.yaml
+++ b/extensions/functions_arithmetic.yaml
@@ -1214,20 +1214,6 @@ aggregate_functions:
         decomposable: MANY
         intermediate: fp64?
         return: fp64?
-      - args:
-          - name: x
-            value: timestamp
-        nullability: DECLARED_OUTPUT
-        decomposable: MANY
-        intermediate: timestamp?
-        return: timestamp?
-      - args:
-          - name: x
-            value: timestamp_tz
-        nullability: DECLARED_OUTPUT
-        decomposable: MANY
-        intermediate: timestamp_tz?
-        return: timestamp_tz?
   - name: "max"
     description: Max a set of values.
     impls:
@@ -1273,20 +1259,6 @@ aggregate_functions:
         decomposable: MANY
         intermediate: fp64?
         return: fp64?
-      - args:
-          - name: x
-            value: timestamp
-        nullability: DECLARED_OUTPUT
-        decomposable: MANY
-        intermediate: timestamp?
-        return: timestamp?
-      - args:
-          - name: x
-            value: timestamp_tz
-        nullability: DECLARED_OUTPUT
-        decomposable: MANY
-        intermediate: timestamp_tz?
-        return: timestamp_tz?
   - name: "product"
     description: Product of a set of values. Returns 1 for empty input.
     impls:


richtia (approved):

vbarua:
The duplicated names prevent substrait-java from being updated.

However, there is a question around whether names need to unique across ALL extensions, or just within a single extension file. While these functions were added in error (I think) you could make the case that:
* `functions_arithmetic/min:timestamp,max:timestamp`
* `functions_datetime/min:timestamp,max:timestamp`

_should_ be treated as different functions.

vbarua:
The case I'm think of is more like:
1. `functions_arithmetic/add:int32_int32`
2. `custom_functions_for_fancy_engine/add:int32_int32`

Which is a name collision with names _outside_ of the core spec, because users can choose to provide their own functions. My engine might provide 2 and disallow 1 because it differs from the Substrait semantics for add somehow.

> But how are you supposed to know which one of the two namespaces is the one to use?

_Within_ a Substrait plan, these two functions are distinguishable because we have access to the extension information. 

Outside of Substrait, less so. You're right that for something like `1 + 2`, whatever is parsing and generating the Substrait plan needs to decide on which `add` version to use. For Ishmus, we choose to use the standard functions_arithmetic version and provide a (relatively*) [explicit mapping](https://github.com/substrait-io/substrait-java/blob/ce0f1eea6b89b9bb733edab80c1bedea71345048/isthmus/src/main/java/io/substrait/isthmus/expression/FunctionMappings.java#L20-L23) from Calcite to Substrait. The choice of which functions to use use is generally up to the plan producer, and will depend on the consumer that they are targeting and what they support.

>I think that introducing namespaces is not as simple as allowing duplicated declarations and requires a dedicated discussion.

I agree with this, figuring out duplicating declarations is out of the scope of this PR. 

\* I say relatively because the mapping doesn't include the name of the extension, yet.

extensions/functions_arithmetic.yaml
```diff
@@ -1214,20 +1214,6 @@ aggregate_functions:
...
-        decomposable: MANY
-        intermediate: timestamp?
-        return: timestamp?
-      - args:
-          - name: x
-            value: timestamp_tz
-        nullability: DECLARED_OUTPUT
-        decomposable: MANY
-        intermediate: timestamp_tz?
-        return: timestamp_tz?
```
vbarua, line 1230: This is technically a breaking change. Plans may exist that use this version of the function, especially because it predates the version in `functions_datetime.yaml`.

API wise, it's probably better to remove these versions because the leaves all the time related `min` and `max functions in `functions_datetime.yaml`.

In terms of minimising breakage, it's probably better to remove the definition of these functions in `functions_datetime.yaml` as those have been around for less time.

I have a preference for removing them from `functions_datetime`, but it would be good to have other folks weigh in on this.

westonpace:
I've added my thoughts around function names and uniqueness in #634

vbarua:
From substrait sync, we've decided to remove the old ones or keep the new ones.

vbarua (approved):

EpsilonPrime (approved):

westonpace (approved):

PR #632, opened 2024-04-18
fix: index_in has wrong return type

amol-:
`index_in` is currently declared as returning an `int64`,
the return type should be `i64`, as according to
https://substrait.io/types/type_classes/#simple-types 
there is no int64 type.


richtia (approved):

vbarua (approved):
Heh, I just fixed the argument types being declared incorrectly, but totally missed the return type 😅 

Thanks for fixing this 🙇

PR #636, opened 2024-04-30
feat: sketch of MATCH_RECOGNIZE support

vbarua:
Very early sketch of a MatchRecognizeRel


vbarua:
Halting this work for now due to changes in priorities.

PR #642, opened 2024-05-21
fix: specify a minimum length for the options of enum args

mbrobbel:
This came up in a [discussion on a substrait-rs PR](https://github.com/substrait-io/substrait-rs/pull/185#discussion_r1603513149): it doesn't make sense for the options field of enum args to be empty. This PR modifies the schema to state that the `options` array of enum args should be non-empty.

I recommend reviewers to take a look at the first commit to see what this PR changes, the second commit is applying auto-formatting and moving the variants of `oneOf` to reference other definitions, and adding a def for `enum_options`.

~~This is a breaking change, because extensions with empty options array are no longer considered valid.~~

### DIFF
--- a/text/simple_extensions_schema.yaml
+++ b/text/simple_extensions_schema.yaml
@@ -26,9 +26,9 @@ properties:
           type: string
         structure:
           $ref: "#/$defs/type"
-        parameters:  # parameter list for compound types
+        parameters: # parameter list for compound types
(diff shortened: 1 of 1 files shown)


westonpace (approved):
Makes sense to me

westonpace:
I also don't know if you need to call it a breaking change.  I would say any extension function with an empty options array was already invalid, we just weren't checking it for it.

PR #646, opened 2024-05-30
feat: add CSV (text) file support

EpsilonPrime:
[Gluten's version](https://github.com/apache/incubator-gluten/blob/73eb21db45a83c3180518bfcef3c5d343595f452/gluten-core/src/main/resources/substrait/proto/substrait/algebra.proto#L138) of text file support relies on a named schema which shouldn't be necessary given that by its given nature a text file is comprised of strings.  This version doesn't attempt to define a schema or ways to mangle text into that schema.

### DIFF
--- a/proto/substrait/algebra.proto
+++ b/proto/substrait/algebra.proto
@@ -136,13 +136,16 @@ message ReadRel {
       message ArrowReadOptions {}
       message OrcReadOptions {}
       message DwrfReadOptions {}
-      message TextReadOptions {
+      message DelimeterSeparatedTextReadOptions {
+        // Delimiter separated files may be compressed.  The reader should
+        // autodetect this and decompress as needed.
+
         // The character(s) used to separate fields.  Common values are comma,
-        // tab, and pipe.
+        // tab, and pipe.  Multiple characters are allowed.
         string field_delimiter = 1;
-        // The maximum number of bytes to read from a single row.  If a row
+        // The maximum number of bytes to read from a single line.  If a line
         // exceeds this limit the resulting behavior is undefined.
-        uint64 max_block_size = 2;
+        uint64 max_line_size = 2;
         // The character(s) used to quote strings.  Common values are single
         // and double quotation marks.
         string quote = 3;
@@ -153,10 +156,11 @@ message ReadRel {
         // the most common value.
         string escape = 5;
         // If this value is encountered (including empty string), the resulting
-        // value is null instead.  Leave unset to disable.  If this value is
-        // provided, the effective schema of this file is comprised entirely of
-        // nullable strings.  If not provided, the effective schema is
-        // instead made up of non-nullable strings.
+        // value is null instead.  This value is checked before quotes are
+        // removed.  Leave unset to disable.  If this value is provided, the
+        // effective schema of this file is comprised entirely of nullable
+        // strings.  If not provided, the effective schema is instead made up of
+        // non-nullable strings.
         optional string null_value = 6;
       }
 
@@ -167,7 +171,7 @@ message ReadRel {
         OrcReadOptions orc = 11;
         google.protobuf.Any extension = 12;
         DwrfReadOptions dwrf = 13;
-        TextReadOptions text = 14;
+        DelimeterSeparatedTextReadOptions text = 14;
       }
     }
   }
--- a/proto/substrait/algebra.proto
+++ b/proto/substrait/algebra.proto
@@ -136,7 +136,7 @@ message ReadRel {
       message ArrowReadOptions {}
       message OrcReadOptions {}
       message DwrfReadOptions {}
-      message DelimeterSeparatedTextReadOptions {
+      message DelimiterSeparatedTextReadOptions {
         // Delimiter separated files may be compressed.  The reader should
         // autodetect this and decompress as needed.
 
@@ -151,9 +151,10 @@ message ReadRel {
         string quote = 3;
         // If true, consume the first row as names of the columns.  These names
         // are not used elsewhere in the plan.
-        uint64 header = 4;
-        // The character(s) used to escape characters in strings.  Backslash is
-        // the most common value.
+        uint64 treat_first_row_as_header = 4;
+        // The character used to escape characters in strings.  Backslash is
+        // a common value.  Note that a double quote mark can also be used as an
+        // escape character but the external quotes should be removed first.
         string escape = 5;
         // If this value is encountered (including empty string), the resulting
         // value is null instead.  This value is checked before quotes are
@@ -161,7 +162,7 @@ message ReadRel {
         // effective schema of this file is comprised entirely of nullable
         // strings.  If not provided, the effective schema is instead made up of
         // non-nullable strings.
-        optional string null_value = 6;
+        optional string value_treated_as_null = 6;
       }
 
       // The format of the files along with options for reading those files.
@@ -171,7 +172,7 @@ message ReadRel {
         OrcReadOptions orc = 11;
         google.protobuf.Any extension = 12;
         DwrfReadOptions dwrf = 13;
-        DelimeterSeparatedTextReadOptions text = 14;
+        DelimiterSeparatedTextReadOptions text = 14;
       }
     }
   }
--- a/proto/substrait/algebra.proto
+++ b/proto/substrait/algebra.proto
@@ -153,7 +153,7 @@ message ReadRel {
         uint64 skip_lines = 4;
         // If true, consume the first unskipped line as names of the columns.
         // These names are not used elsewhere in the plan.
-        uint64 treat_first_row_as_header = 5;
+        bool treat_first_row_as_header = 5;
         // The character used to escape characters in strings.  Backslash is
         // a common value.  Note that a double quote mark can also be used as an
         // escape character but the external quotes should be removed first.
--- a/proto/substrait/algebra.proto
+++ b/proto/substrait/algebra.proto
@@ -149,20 +149,22 @@ message ReadRel {
         // The character(s) used to quote strings.  Common values are single
         // and double quotation marks.
         string quote = 3;
-        // If true, consume the first row as names of the columns.  These names
-        // are not used elsewhere in the plan.
-        uint64 treat_first_row_as_header = 4;
+        // The number of lines to skip at the beginning of the file.
+        uint64 skip_lines = 4;
+        // If true, consume the first unskipped line as names of the columns.
+        // These names are not used elsewhere in the plan.
+        uint64 treat_first_row_as_header = 5;
         // The character used to escape characters in strings.  Backslash is
         // a common value.  Note that a double quote mark can also be used as an
         // escape character but the external quotes should be removed first.
-        string escape = 5;
+        string escape = 6;
         // If this value is encountered (including empty string), the resulting
         // value is null instead.  This value is checked before quotes are
         // removed.  Leave unset to disable.  If this value is provided, the
         // effective schema of this file is comprised entirely of nullable
         // strings.  If not provided, the effective schema is instead made up of
         // non-nullable strings.
-        optional string value_treated_as_null = 6;
+        optional string value_treated_as_null = 7;
       }
 
       // The format of the files along with options for reading those files.
--- a/proto/substrait/algebra.proto
+++ b/proto/substrait/algebra.proto
@@ -156,11 +156,10 @@ message ReadRel {
         // escape character but the external quotes should be removed first.
         string escape = 5;
         // If this value is encountered (including empty string), the resulting
-        // value is null instead.  This value is checked before quotes are
-        // removed.  Leave unset to disable.  If this value is provided, the
-        // effective schema of this file is comprised entirely of nullable
-        // strings.  If not provided, the effective schema is instead made up of
-        // non-nullable strings.
+        // value is null instead.  Leave unset to disable.  If this value is
+        // provided, the effective schema of this file is comprised entirely of
+        // nullable strings.  If not provided, the effective schema is instead
+        // made up of non-nullable strings.
         optional string value_treated_as_null = 6;
       }
 
--- a/proto/substrait/algebra.proto
+++ b/proto/substrait/algebra.proto
@@ -150,21 +150,18 @@ message ReadRel {
         // and double quotation marks.
         string quote = 3;
         // The number of lines to skip at the beginning of the file.
-        uint64 skip_lines = 4;
-        // If true, consume the first unskipped line as names of the columns.
-        // These names are not used elsewhere in the plan.
-        bool treat_first_row_as_header = 5;
+        uint64 header_lines_to_skip = 4;
         // The character used to escape characters in strings.  Backslash is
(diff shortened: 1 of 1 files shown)


proto/substrait/algebra.proto
```diff
@@ -131,14 +131,38 @@ message ReadRel {
       message ArrowReadOptions {}
       message OrcReadOptions {}
       message DwrfReadOptions {}
+      message TextReadOptions {
```
westonpace, line 134: I think I prefer CSV / DSV / TSV over "Text".  Text is pretty commonly used in the file format domain to mean "CSV or JSON or YAML or XML".

proto/substrait/algebra.proto
```diff
@@ -131,14 +131,38 @@ message ReadRel {
...
+        // The character(s) used to separate fields.  Common values are comma,
+        // tab, and pipe.
+        string field_delimiter = 1;
```
westonpace, line 137: Can this be multiple characters or is it restricted to one?

proto/substrait/algebra.proto
```diff
@@ -131,14 +131,38 @@ message ReadRel {
...
+        // The maximum number of bytes to read from a single row.  If a row
+        // exceeds this limit the resulting behavior is undefined.
+        uint64 max_block_size = 2;
```
westonpace, line 140: At first I thought this was spark specific (it is not in pandas) but I see that duckdb's csv reader does have an equivalent property so I think it's fine.

Minor nit: Why not `max_line_size`?

proto/substrait/algebra.proto
```diff
@@ -131,14 +131,38 @@ message ReadRel {
...
+        uint64 header = 4;
+        // The character(s) used to escape characters in strings.  Backslash is
+        // the most common value.
+        string escape = 5;
+        // If this value is encountered (including empty string), the resulting
+        // value is null instead.  Leave unset to disable.  If this value is
+        // provided, the effective schema of this file is comprised entirely of
+        // nullable strings.  If not provided, the effective schema is
+        // instead made up of non-nullable strings.
+        optional string null_value = 6;
```
westonpace, line 155: Can protobuf impls differentiate between an "unset string" and a "null string"?

proto/substrait/algebra.proto
```diff
@@ -131,14 +131,38 @@ message ReadRel {
...
+      message TextReadOptions {
+        // The character(s) used to separate fields.  Common values are comma,
+        // tab, and pipe.
+        string field_delimiter = 1;
+        // The maximum number of bytes to read from a single row.  If a row
+        // exceeds this limit the resulting behavior is undefined.
+        uint64 max_block_size = 2;
+        // The character(s) used to quote strings.  Common values are single
+        // and double quotation marks.
+        string quote = 3;
```
westonpace, line 143: Can `null_value` be quoted?  E.g. if `null_value` is `NULL` then is `"NULL"` a null or the string `NULL`?

westonpace (commented):
This seems like a reasonably small set of the most common options.  I think any arguing about CSV options quickly becomes subjective but I'll pitch the following:

**Common options not included here:**

 * Some ability to set the column types (are you proposing that casts be used for this purpose?)
 * Compression (many CSV files are compressed with some kind of compression scheme)
 * Newline character (\r, \n, \r\n, though I think this can often be inferred pretty accurately)
 * Skip (some number of lines at the top of the file to skip)

**Options included here that don't feel very common:**

 * `max_block_size` strikes me as something only needed in more exotic files

proto/substrait/algebra.proto
EpsilonPrime: Changed to DelimiterSeperatedText (instead of just text).

proto/substrait/algebra.proto
EpsilonPrime: Added clarifying text that multiple characters are allowed.

proto/substrait/algebra.proto
EpsilonPrime: Changed.

proto/substrait/algebra.proto
EpsilonPrime: Optional is now supported which allows the distinction to be determined.

proto/substrait/algebra.proto
EpsilonPrime: Clarified that null_value happens is checked first and then quote removal happens.

EpsilonPrime:
> This seems like a reasonably small set of the most common options. I think any arguing about CSV options quickly becomes subjective but I'll pitch the following:
> 
> **Common options not included here:**
> 
> * Some ability to set the column types (are you proposing that casts be used for this purpose?)
> * Compression (many CSV files are compressed with some kind of compression scheme)
> * Newline character (\r, \n, \r\n, though I think this can often be inferred pretty accurately)
> * Skip (some number of lines at the top of the file to skip)
> 
> **Options included here that don't feel very common:**
> 
> * `max_block_size` strikes me as something only needed in more exotic files

Setting the schema here doesn't seem appropriate.  All of the other formats have a schema defined internally.  In absence of anything internal definition, the schema for CSVs is actually text.  That said there is a base schema in the ReadRel relation itself that one could go try to cast the data to if one wanted to force the schema of the relation for ease of use.  I do prefer casting, however.

I've added a comment about compression.  All of the compression formats are easily determined by the MAGIC file header so specifying the compression externally seems extraneous.

The header option removes the first row to get the names (which we don't use).  I could replace that with a number of lines to skip instead.

I agree that newlines seem to be fairly standard these days and probably don't need to be included as an option.

jacques-n:
I agree with @westonpace that schema/parsing desires should be included in configuration. As a concrete example, pushing down filters requires data type to be done with CSVs and is pretty common to avoid things like building up large arrow datasets and than deleting most of the records.

I also agree on skip lines. Super common and frequently useful.

Compression also feels like it should be an option but I'm torn on what that should be and would prefer to solve once we have more real use cases.

proto/substrait/algebra.proto
```diff
@@ -131,14 +131,42 @@ message ReadRel {
...
+        // If true, consume the first row as names of the columns.  These names
+        // are not used elsewhere in the plan.
+        uint64 header = 4;
+        // The character(s) used to escape characters in strings.  Backslash is
+        // the most common value.
+        string escape = 5;
+        // If this value is encountered (including empty string), the resulting
+        // value is null instead.  This value is checked before quotes are
```
jacques-n, line 154: Why would this value be before quotes removed? That's super weird. The quotes are part of the container, not the value.

proto/substrait/algebra.proto
```diff
@@ -131,14 +131,42 @@ message ReadRel {
...
+        string field_delimiter = 1;
+        // The maximum number of bytes to read from a single line.  If a line
+        // exceeds this limit the resulting behavior is undefined.
+        uint64 max_line_size = 2;
+        // The character(s) used to quote strings.  Common values are single
+        // and double quotation marks.
+        string quote = 3;
+        // If true, consume the first row as names of the columns.  These names
+        // are not used elsewhere in the plan.
+        uint64 header = 4;
```
jacques-n, line 149: can we give this a more meaningful name e.g. treat_as_header

proto/substrait/algebra.proto
```diff
@@ -131,14 +131,42 @@ message ReadRel {
...
+        // The character(s) used to escape characters in strings.  Backslash is
+        // the most common value.
+        string escape = 5;
```
jacques-n, line 152: Saying this can be more than character is interesting. Where have you seen multiple characters used? High performance readers are often hand tuned for this kind of thing so I'm wondering if we really need multiple characters.

proto/substrait/algebra.proto
```diff
@@ -131,14 +131,42 @@ message ReadRel {
...
+        // If this value is encountered (including empty string), the resulting
+        // value is null instead.  This value is checked before quotes are
+        // removed.  Leave unset to disable.  If this value is provided, the
+        // effective schema of this file is comprised entirely of nullable
+        // strings.  If not provided, the effective schema is instead made up of
+        // non-nullable strings.
+        optional string null_value = 6;
```
jacques-n, line 159: I would give this a clearer name. For example

value_treated_as_null

proto/substrait/algebra.proto
```diff
@@ -131,14 +131,42 @@ message ReadRel {
       message ArrowReadOptions {}
       message OrcReadOptions {}
       message DwrfReadOptions {}
+      message DelimeterSeparatedTextReadOptions {
```
jacques-n, line 134: I think we should add boolean  treat_empty_string_as_null

proto/substrait/algebra.proto
EpsilonPrime: If this is done after, you wouldn't be able to express both "NULL" and NULL.  By checking before you allow yourself the capability to find magic strings that wouldn't be present afterwards.

jacques-n: Can you show me a set of systems that enable this behavior?

As an aside, I think that assumes a strange behavior for CSVs. My experience is that in CSV parsers, quotes are optional so "NULL" and NULL are fundamentally the same.

proto/substrait/algebra.proto
```diff
@@ -136,14 +136,42 @@ message ReadRel {
...
+        // Delimiter separated files may be compressed.  The reader should
+        // autodetect this and decompress as needed.
+
```
jacques-n, line 142: one other question here. Should "trim" be an option. For example, will these two records return different values. 

```
A, B, C
A,B,C
```
If so, it that typically desired? My recollection is many readers will trim whitespace but I could be misremembering.

westonpace: Spark, DuckDb, and Pandas do not have such an option.  I think we should avoid anything that isn't already implemented somewhere.  We can always add more options later without introducing a backwards incompatible change.

proto/substrait/algebra.proto
```diff
@@ -136,14 +136,42 @@ message ReadRel {
       message ArrowReadOptions {}
       message OrcReadOptions {}
       message DwrfReadOptions {}
+      message DelimeterSeparatedTextReadOptions {
```
CurtHagenlocher, line 139: typo in the message name here and in file_format

proto/substrait/algebra.proto
CurtHagenlocher: In at least some flavors of CSV, quotes are not optional when fields contain special values; e.g. if the text contains a delimiter, the value must be quoted to prevent the character from understood as a delimiter. If your mental model of quoting is "treat this value literally" then I think it makes sense for NULL to mean NULL and "NULL" to mean "NULL". That said, while I think this is potentially useful behavior I also don't know of any systems which work this way. By contrast, I can immediately think of one that does have specialized NULL-handling for which "NULL" and NULL are treated identically -- R's read.csv.

proto/substrait/algebra.proto
```diff
@@ -136,14 +136,42 @@ message ReadRel {
...
+        // The maximum number of bytes to read from a single line.  If a line
+        // exceeds this limit the resulting behavior is undefined.
+        uint64 max_line_size = 2;
+        // The character(s) used to quote strings.  Common values are single
+        // and double quotation marks.
+        string quote = 3;
+        // If true, consume the first row as names of the columns.  These names
+        // are not used elsewhere in the plan.
+        uint64 header = 4;
+        // The character(s) used to escape characters in strings.  Backslash is
```
CurtHagenlocher, line 155: While the pernicious influence of C has caused its escape style to spread into new areas, in CSV the most common style is still that the quote character is doubled inside quoted strings in order to escape it.

proto/substrait/algebra.proto
EpsilonPrime: thanks, fixed

proto/substrait/algebra.proto
EpsilonPrime: If null_value is empty would that achieve enough of the same result as treating empty string as NULL?  (Note that if we have empty string and quoted empty string it's likely only one would match depending on null_value behavior.)

proto/substrait/algebra.proto
EpsilonPrime: treat_first_row_as_header works for me

proto/substrait/algebra.proto
EpsilonPrime: Using a quote to escape a quote should work if provided here.  I'll add it for clarity.

proto/substrait/algebra.proto
EpsilonPrime: Done.

EpsilonPrime:
> I also agree on skip lines. Super common and frequently useful.

Added skip_lines which will skip lines.  The treat_first_row_as_header option is defined to work after that but we could conceivably just not have a treat_first_row_as_header option since we don't use the names in Substrait.

proto/substrait/algebra.proto
jacques-n: I'm okay without the option. What is the specified behavior? Trim or not trim?

proto/substrait/algebra.proto
jacques-n: I think we're saying the same thing. 

I'm saying I'm not aware of any systems that will treat "NULL" and NULL differently in CSV reading. This is because many csv files only quote things when the values include delimiters. 

If systems don't treat them differently, I'm not clear why we're giving an option that forces us to treat them differently for null handling.

proto/substrait/algebra.proto
jacques-n: I'm fine using that configuration to represent this assuming we remove the "with quotes" part of it.

proto/substrait/algebra.proto
EpsilonPrime: The standard behavior is no trimming.

proto/substrait/algebra.proto
EpsilonPrime: Fine.  Leaving treat_empty_string_as_null as part of value_treated_as_null.

proto/substrait/algebra.proto
EpsilonPrime: Updated the documentation so the value is no longer evaluate unquoting.

jacques-n (approved):
Thanks for the patience. LGTM +1

EpsilonPrime:
The only topic left that hasn't reached consensus is the type handling.  Do we want to assume the return type is all strings (nullable or not) as currently written or should we expect that the reader uses the ReadRel's schema?

proto/substrait/algebra.proto
jacques-n: ok, captured here. ideally would make itself into docs at some point.

jacques-n:
> The only topic left that hasn't reached consensus is the type handling. Do we want to assume the return type is all strings (nullable or not) as currently written or should we expect that the reader uses the ReadRel's schema?

My suggestion is we merge this and that can be an enhancement. All strings are nullable if the null value is defined. Otherwise non-nullable I suppose. I'm also okay with all just being nullable.

EpsilonPrime:
Anyone else interested in weighing in?  I believe this PR is ready to merge.

Issue #634, opened 2024-04-18
Function names & URIs


westonpace:
This is in relation to some discussion that came up in #631

> The duplicated names prevent substrait-java from being updated.
> 
> However, there is a question around whether names need to unique across ALL extensions, or just within a single extension file. While > these functions were added in error (I think) you could make the case that:
> 
>     functions_arithmetic/min:timestamp,max:timestamp
>     functions_datetime/min:timestamp,max:timestamp
> 
> should be treated as different functions.

My understanding is that a function name is the triple:

```
<function uri>,<function name>,<function arguments>
```

Therefore, yes, these are different functions.  However, there are three problems that have never really been fully resolved:

1. No one uses the function URI

The major producers that I am aware of today (isthmus, duckdb, ibis) either set the function URI to undefined, the empty string, or `/` (I think we actually have all three behaviors across the three producers :face_exhaling: )

Correspondingly, consumers tend to ignore this field.  The one exception I'm aware of is Acero which will tolerate `/`, empty string, and undefined (Acero goes into a "fallback" mode where it does name-only matching and will match any registered function with the same name regardless of the URI) but which will accept URLs of the form `https://github.com/substrait-io/substrait/blob/main/extensions/functions_arithmetic.yaml` and also has a special URI `urn:arrow:substrait_simple_extension_function` which means "use the arrow compute function with the given name" (this is how we support UDFs).

2. What is the canonical URL?

There are several choices.  For example:

```
# Stable URI but potentially unstable definition
https://github.com/substrait-io/substrait/blob/main/extensions/functions_arithmetic.yaml
# Stable definition but unstable URI
https://github.com/substrait-io/substrait/blob/v0.47.0/extensions/functions_arithmetic.yaml
```

My preference is the former, for practical reasons.

3. How is versioning handled?

This is discussed in more detail in https://github.com/substrait-io/substrait/issues/274

But the basic question is "what if we have tons of users and we decide to make a change to some function?"

joellubi:
I'm currently facing some design choices related to this and this is roughly where my thinking is at this point:

_TLDR at bottom..._

Generally, uniquely qualifying function names is not a problem that consumers have to deal with because the extension references are already resolved in the plan they are receiving.

For plan producers, it seems some specific preferences do need to be adopted in order to produce plans under practical constraints. Specifically, the input most producers are serializing into substrait will typically contain the function name and input types, but not the URI. Generally speaking this is coming from SQL but dataframes have the same issue.

[...]

PR #652, opened 2024-06-26
feat: add null input handling options for `any_value`

Blizzara:
This adds a "ignore_nulls" option for `any_value` that can be used when converting e.g. Spark's first()/first_value()/any_value()

### DIFF
--- a/extensions/functions_aggregate_generic.yaml
+++ b/extensions/functions_aggregate_generic.yaml
@@ -32,6 +32,11 @@ aggregate_functions:
     impls:
       - args:
           - name: x
-            value: any
+            value: any1
+        options:
+          ignore_nulls:
+            values: [ "TRUE", "FALSE" ]
         nullability: DECLARED_OUTPUT
-        return: any?
+        decomposable: MANY
+        intermediate: any1?
+        return: any1?


extensions/functions_arithmetic.yaml
```diff
@@ -1563,6 +1563,43 @@ aggregate_functions:
...
+      If `null_handling` is set to `IGNORE_NULLS`, the function will look for the first non-null value.
+      If set to `ACCEPT_NULLS`, the function will return the first value it sees, even if null.
+
+      If all values are null or there are no values, returns null.
+    impls:
+      - args:
+          - name: x
+            value: any
+        options:
+          null_handling:
```
Blizzara, line 1580: I copied this from [concat](https://github.com/substrait-io/substrait/blob/fe7eba4454d404705cd124151856260c4a2b13bf/extensions/functions_string.yaml#L20), does it make sense here or is better to add e.g. a boolean arg?

extensions/functions_arithmetic.yaml
```diff
@@ -1563,6 +1563,43 @@ aggregate_functions:
             values: [ TIE_TO_EVEN, TIE_AWAY_FROM_ZERO, TRUNCATE, CEILING, FLOOR ]
         nullability: DECLARED_OUTPUT
         return: fp64?
+  - name: "first"
```
Blizzara, line 1566: the window functions are called `first/last_value`, should we stick to that naming? 

Spark seems to have both `first` and `first_value`, though first_value is only supported in SQL while first is supported as a method. 
DataFusion has `first_value`

Blizzara:
Onne thing I'm not sure - is it better to have these as both window and aggregate functions, or could we remove the window function version and just replace with this?

extensions/functions_arithmetic.yaml
```diff
@@ -1563,6 +1563,43 @@ aggregate_functions:
...
+
+      If all values are null or there are no values, returns null.
+    impls:
+      - args:
+          - name: x
+            value: any
+        nullability: DECLARED_OUTPUT
+        decomposable: MANY
+        intermediate: any?
+        return: any?
```
vbarua, line 1602: I think both of these functions should be in `functions_aggregate_generic.yaml` instead of `functions_arithmetic.yaml`

extensions/functions_arithmetic.yaml
```diff
@@ -1563,6 +1563,43 @@ aggregate_functions:
...
+
+      The `null_handling` option determines whether or not null values will be recognized by the function.
+      If `null_handling` is set to `IGNORE_NULLS`, the function will look for the first non-null value.
+      If set to `ACCEPT_NULLS`, the function will return the first value it sees, even if null.
+
+      If all values are null or there are no values, returns null.
+    impls:
+      - args:
+          - name: x
+            value: any
```
vbarua, line 1578: minor: if we replace `any` with `any1` in the definition, it enforces that they are all the same type.

See https://substrait.io/extensions/#any-types

extensions/functions_arithmetic.yaml
vbarua: I think options make sense for this. The names `ACCEPT_NULLS` and `IGNORE_NULLS` sound a little weird to me, but I can't think of better ones currently.

extensions/functions_arithmetic.yaml
vbarua: Looking at other engines, Postgres and Trino only have support for `first_value` and `last_value` as window functions, and don't have `first` and `last` aggregate functions.

I think it make sense to keep first/last and first_value/last_value as seperate.

vbarua (commented):
Overall this seems reasonable to me, but I'm curious to see what others think.

extensions/functions_arithmetic.yaml
richtia: I added that original option to concat, but I agree it does sound a little weird.  How about changing the option name to `ignore_nulls` and have the options be `["True", "False"]`.  I think True/False may need to be quoted.  If i recall correctly I didn't do this originally because no other options were quoted, but that's changed since then.

extensions/functions_aggregate_generic.yaml
```diff
@@ -35,3 +35,41 @@ aggregate_functions:
             value: any
         nullability: DECLARED_OUTPUT
         return: any?
+  - name: "first"
+    description: >-
+      First value from a group of values.
```
EpsilonPrime, line 40: It's probably worth noting that order matters for these two functions.  Acero will reject plans that don't have a defined ordering on the input which might be a reasonable practice.

extensions/functions_arithmetic.yaml
Blizzara: Yeah, looks like this is a mess overall - postgres has only `first_value` as window, Spark has `first` and `first_value` being the same, duckdb has `first` for aggregate and both for window, DataFusion has `first_value` as aggregate...

The purpose of the functions is the same, though. I actually think it'd make sense for Substrait to only have one set of these (as aggregate), and maybe it should be the `_value` option to match already existing `any_value`. But dunno if we can remove the window versions, I guess that'd be a breaking change?

extensions/functions_arithmetic.yaml
Blizzara: renamed in https://github.com/substrait-io/substrait/pull/652/commits/b18cecd00aee3aeab891f4e095c9ee25942be255 - and quoted after, as that was needed to keep them as strings. Though it hurts my soul a bit to have a string "TRUE". Maybe should make them YES/NO instead to be less confusing 😅

extensions/functions_arithmetic.yaml
Blizzara: ah right, I knew that but forgot, thanks! 3dc5d18bf37ce08a4aa2faec2f07877261290a4b

extensions/functions_aggregate_generic.yaml
```diff
@@ -35,3 +35,41 @@ aggregate_functions:
             value: any
         nullability: DECLARED_OUTPUT
         return: any?
+  - name: "first"
```
EpsilonPrime: Should we call out to first_value with the difference to make it easier for folks to choose one or the other?

extensions/functions_arithmetic.yaml
Blizzara: I renamed them to have the "_value" postfix now - small annoyance there is that if someone includes now both these and window functions they'll get duplicate signatures...

EpsilonPrime: Currently the base name has to be unique across all of the files.  So having two first_value functions with the same signature will likely mess things up.  (And yes, we need a check.)

extensions/functions_aggregate_generic.yaml
Blizzara: Added a note c93c326bfd5d41c94825a0dffd2cde2089e6fce0!

extensions/functions_aggregate_generic.yaml
Blizzara: Just to make sure I understand, what do you see as the difference? My hope would be to replace the window versions with these completely, given that aggregate functions are also valid window functions, and the engines I looked at don't seem to make a difference between these - but maybe I missed something

EpsilonPrime: The only difference I see is what context they can be called in.

Blizzara:
Okays, I revamped the PR a bit - now it moves the first_value and last_value from window funcs into aggregate funcs, and adds the null handling options. 

This makes most sense to me, but lmk what you think!

EpsilonPrime (approved):

extensions/functions_arithmetic.yaml
```diff
@@ -1704,28 +1704,6 @@ window_functions:
...
-    description: >
-      Returns the last value in the window.
-    impls:
-      - args:
-          - value: any1
-            name: expression
-        nullability: DECLARED_OUTPUT
-        decomposable: NONE
-        return: any1
-        window_type: PARTITION
```
vbarua, line 1728: Heads up, moving _existing_ functions is a breaking change, and a relatively painful one to workaround. I would prefer if we could avoid breakage here.

See: https://github.com/substrait-io/substrait/issues/634#issuecomment-2091386661

Blizzara: I guess the options are:
1. move existing functions (breaking change
2. duplicate existing functions (bad)
3. add new functions with new name (leads to confusing end state as there's now two functions with same functionality but different name, unless we can deprecate the existing functions somehow)

For my need, I think both 1 and 3 work fine, so I don't have strong opinions - I guess it's a question of breaking now vs keeping a worse state for ever/until breaking later?

vbarua: 4. We could also keep the functions in the same file, but move them into the aggregate functions block.

> I guess it's a question of breaking now vs keeping a worse state for ever/until breaking later?

I think it would good to move these eventually, but it would be nice to do _after_ substrait-java can handle duplicate functions in different files, because then we could make the change by duplicating the functions into the new file in one release, and then removing the old functions in the next release.


I've filed https://github.com/substrait-io/substrait-java/issues/275 to track this work in substrait-java.

EpsilonPrime: Keeping the same file for now, fixing substrait-java, and then moving files works for me.

Blizzara: Ah, I thought "moving" meant moving also from window -> aggregation. Keeping them in the old file is fine for me -like fc3fa78a7a1e3c3cf9ba43abdb2cc7fbfc635aae ?

EpsilonPrime (dismissed):

westonpace:
I don't love it but I won't necessarily vote against it.  It isn't supported by some significant engines (e.g. Postgres, SQL server, Snowflake, ). However, it does appear to be supported by `DataFusion` and `DuckDB` so there is some representation.  Databricks has `first`/`last` but it marks them as "order is non-deterministics" and it's not clear that Databricks supports "ORDER BY" in an aggregate expression.

My main problem is that `first(x order by x)` is the same as `min(x)` and `first(x order by y)` can be obtained by `arg_min(x, y)`.  I think `min` / `arg_min` are better since they don't require an `order by` statement and so they are more easily implemented by engines (yes, any engine can optimize `first` into `min` / `arg_min` but that's just introducing extra steps for no real gain).  I think the more common result would be an engine falling back to its `order by` implementation and introducing an expensive sort into the query (`arg_min` and `arg_max` can be calculated without a sort).  For example, this appears to be what DataFusion does today:

```
❯ EXPLAIN SELECT first_value(val ORDER BY val) FROM foo;
+---------------+------------------------------------------------------------------------------------------+
| plan_type     | plan                                                                                     |
+---------------+------------------------------------------------------------------------------------------+
| logical_plan  | Aggregate: groupBy=[[]], aggr=[[FIRST_VALUE(foo.val) ORDER BY [foo.val ASC NULLS LAST]]] |
|               |   TableScan: foo projection=[val]                                                        |
| physical_plan | AggregateExec: mode=Single, gby=[], aggr=[FIRST_VALUE(foo.val)]                          |
|               |   SortExec: expr=[val@0 ASC NULLS LAST]                                                  |
|               |     MemoryExec: partitions=1, partition_sizes=[1]                                        |
|               |                                                                                          |
+---------------+------------------------------------------------------------------------------------------+
```

That plan is more expensive than `SELECT min(val) FROM foo` even though they are identical.  That being said, I do think we should finish up the `arg_min` / `arg_max` PR (https://github.com/substrait-io/substrait/pull/326)

Blizzara:
> I don't love it but I won't necessarily vote against it. It isn't supported by some significant engines (e.g. Postgres, SQL server, Snowflake, ). However, it does appear to be supported by `DataFusion` and `DuckDB` so there is some representation. Databricks has `first`/`last` but it marks them as "order is non-deterministics" and it's not clear that Databricks supports "ORDER BY" in an aggregate expression.

By Databricks, do you mean Spark? Turns out Spark has also `any_value`, and they are "interchanged" ie the implementation for `any_value` is just `first`. Substrait already has `any_value`, so one option is to just use that.

> My main problem is that `first(x order by x)` is the same as `min(x)` and `first(x order by y)` can be obtained by `arg_min(x, y)`. 

FWIW, I don't think those are the main uses for first/last, rather I think the need is more for the "any_value" concept. Now that I think of it, I'm not sure why Spark even has a `last`, but maybe there's a reason.

The main reason I started this PR was to support Spark's `distinct`/`dropDuplicates`. Those get rewritten by the optimizer into a `first` aggregate: https://github.com/apache/spark/blob/df13ca05c475e98bf5c218a4503513065611a47f/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/optimizer/Optimizer.scala#L2262

We do have some uses of `last` as well, but it's possible those would be within windows, I don't have an easy way to check.

In the end, if it's preferable, I think I can actually turn Spark's `first()` aggregate into Substrait's `any_value()` and then that again into DataFusion's `first_value()`. Then if I do end up needing the `last()` as aggregate, I can add it as a Spark specific mapping or something.

Would that be better? Then I could change this PR to instead add the null-handling options into `any_value`, since those would be nice to have still.

EpsilonPrime:
An alternative to first is to use a fetch relation but it becomes a lot more complicated to modify a complicated subquery to introduce it.  Last does weird me out and it is a available in less places.  Probably implemented for completeness and not actual use.

westonpace:
> Would that be better? Then I could change this PR to instead add the null-handling options into any_value, since those would be nice to have still.

Yes, I'd prefer that.  Sorry for the churn.  Agree the null handling is good.

Blizzara:
> Yes, I'd prefer that. Sorry for the churn. Agree the null handling is good.

Done! All good, makes sense to be careful when adding stuff into the standard (or standard extensions)!

westonpace (approved):
+1, but I will point out that "ignore nulls" / "skip nulls" is a property that can apply to many (potentially all) aggregate functions.

EpsilonPrime (approved):

extensions/functions_arithmetic.yaml
jacques-n: > Currently the base name has to be unique across all of the files

I just saw this comment. I believe this is inconsistent with both the spirit and spec of extensions. Extensions should allow people to create new extensions that completely conflict with other extensions. That's why the spec specifies that functions are identified by a combination of their name and URI. Two different URIs can define the same names as entirely different things since they are in different namespaces.

westonpace: Recent discussion on the topic:

https://github.com/substrait-io/substrait/pull/631#issuecomment-2065325055
https://github.com/substrait-io/substrait/issues/634

IIRC, the general consensus is "yes, different filenames should be able to have duplicate functio names but we're not there yet, there isn't much motivation to tackle the issue, and keeping the core Substrait functions unique makes life easier in the short term"

PR #653, opened 2024-06-26
feat: make optimization a repeated field

vbarua:
BREAKING CHANGE: consumers must now check for multiple optimization messages within AdvancedExtensions

### DIFF
--- a/proto/substrait/extensions/extensions.proto
+++ b/proto/substrait/extensions/extensions.proto
@@ -72,7 +72,7 @@ message SimpleExtensionDeclaration {
 message AdvancedExtension {
   // An optimization is helpful information that don't influence semantics. May
   // be ignored by a consumer.
-  google.protobuf.Any optimization = 1;
+  repeated google.protobuf.Any optimization = 1;
 
   // An enhancement alter semantics. Cannot be ignored by a consumer.
   google.protobuf.Any enhancement = 2;


vbarua:
Following up on https://github.com/substrait-io/substrait/pull/649#issuecomment-2179423468

proto/substrait/extensions/extensions.proto
```diff
@@ -72,7 +72,7 @@ message SimpleExtensionDeclaration {
 message AdvancedExtension {
   // An optimization is helpful information that don't influence semantics. May
   // be ignored by a consumer.
-  google.protobuf.Any optimization = 1;
+  repeated google.protobuf.Any optimization = 1;
 
   // An enhancement alter semantics. Cannot be ignored by a consumer.
   google.protobuf.Any enhancement = 2;
```
vbarua, line 78: The discussion in https://github.com/substrait-io/substrait/pull/649 indicated that `optimization` should be made repeated, but called out that `enhancement` _should not_ be repeated. Is there a specific reason why we wouldn't want that?

@westonpace @jacques-n

EpsilonPrime (approved):

proto/substrait/extensions/extensions.proto
westonpace: Yes.  An enhancement feels more like an "is a" relationship and less like a "has a" relationship.  In other words, an enhancement is likely to change the semantic intent of the relation in some way.  Combining two enhancements is unlikely to be successful.

For example, let's assume that we had a hash join relation enhancement that applies a pre-filter (I know that such a property already exists but we'll assume it was forgotten).  At the same time let's assume we have some strange enhancement that specifies which collation to use for the equality comparison.  This enhancements would not be compatible (the pre-filter likely would perhaps not be expecting the collation to be used).

westonpace (approved):

proto/substrait/extensions/extensions.proto
westonpace: That being said, I don't think it was ruled out.  Partly we just don't really have any idea what valid use cases for `enhancement` look like yet.  Until someone is using it though it probably doesn't have to be decided one way or the other.

vbarua: Makes sense to me. We can limit this to `optimization` for now.

proto/substrait/extensions/extensions.proto
```diff
@@ -72,7 +72,7 @@ message SimpleExtensionDeclaration {
 message AdvancedExtension {
   // An optimization is helpful information that don't influence semantics. May
   // be ignored by a consumer.
-  google.protobuf.Any optimization = 1;
+  repeated google.protobuf.Any optimization = 1;
```
vbarua, line 75: meta: I would have liked to pluralize this (i.e. `optimizations`) this, but that is a breaking change due to json encoding.

PR #644, opened 2024-05-22
fix(ci): pin `conventional-changelog-conventionalcommits` to `7.0.2`

mbrobbel:
This fixes the release scripts.

PR #655, opened 2024-06-26
ci: stop pinning conventional commits

vbarua:


vbarua:
This was pinned in https://github.com/substrait-io/substrait/pull/644 to fix the release, but now it's broken again because it's pinned :sad:

In substrait-java unpinning fixed a similar issue: https://github.com/substrait-io/substrait-java/pull/267/files

There's probably a proper way to do this to have this stop happening generally, but my motivation actually figure out Javascript dependency weirdness is basically zero at this point.

ci/release/run.sh
```diff
@@ -12,5 +12,5 @@ npx --yes \
   -p "@semantic-release/github" \
   -p "@semantic-release/exec" \
   -p "@semantic-release/git" \
-  -p "conventional-changelog-conventionalcommits@7.0.2" \
+  -p "conventional-changelog-conventionalcommits" \
```
gforsyth, line 15:
```suggestion
  -p "conventional-changelog-conventionalcommits@8.0.0" \
```

If we do this and also pin semantic-release itself to `"semantic-release@24.0.0"` that should prevent this breaking in the future (we just hit this in Ibis, which is where we copied this release process from)

vbarua:
Closing in favour of https://github.com/substrait-io/substrait/pull/657

PR #657, opened 2024-06-26
chore(ci): pin semantic release

gforsyth:
Resolves #655

If we pin both `semantic-release` and  `conventionalcommits` it should stop the random breakages when one of them pushes a new release.

Updated in both the `dry_run.sh` and `run.sh` scripts.


vbarua (approved):
This makes me happier than my attempt at this in https://github.com/substrait-io/substrait/pull/655

PR #660, opened 2024-07-09
feat: add arithmetic function "power" with decimal type

anshuldata:

 - `feat: add feature X`
 - `fix: X in case of Y`

### DIFF
--- a/extensions/functions_arithmetic_decimal.yaml
+++ b/extensions/functions_arithmetic_decimal.yaml
@@ -166,6 +166,23 @@ scalar_functions:
           - name: "n"
             value: "DECIMAL<P,0>"
         return: "DECIMAL<38,0>"
+  -
+    name: "power"
+    description: "Take the power with x as the base and y as exponent.
+    Behavior for complex number result is indicated by option complex_number_result"
+    impls:
+      - args:
+          - name: x
+            value: "DECIMAL<P1,S1>"
+          - name: y
+            value: "DECIMAL<P2,S2>"
+        options:
+          overflow:
+            values: [ SILENT, SATURATE, ERROR ]
+          complex_number_result:
+            values: [ NAN, ERROR ]
+        return: fp64
+
 aggregate_functions:
   - name: "sum"
     description: Sum a set of values.


extensions/functions_arithmetic_decimal.yaml
```diff
@@ -174,3 +174,95 @@ aggregate_functions:
...
+          - name: x
+            value: "DECIMAL<P,S>"
+          - name: y
+            value: "DECIMAL<P,S>"
+        return: "DECIMAL<P,S>"
+
+  -
+    name: "bitwise_xor"
+    description: >
+      Return the bitwise XOR result for two integer inputs.
```
EpsilonPrime, line 261: integer -> decimal (and elsewhere)

extensions/functions_arithmetic_decimal.yaml
```diff
@@ -174,3 +174,95 @@ aggregate_functions:
         decomposable: MANY
         intermediate: "DECIMAL<38,S>"
         return: "DECIMAL<38,S>"
+
```
EpsilonPrime, line 177: no need for a blank line between function definitions

extensions/functions_arithmetic_decimal.yaml
```diff
@@ -174,3 +174,95 @@ aggregate_functions:
...
+      The factorial of 0! is 1 by convention.
+      Negative inputs will raise an error.
+    impls:
+      - args:
+          - value: "DECIMAL<P,S>"
+            name: "n"
+        options:
+          overflow:
+            values: [ SILENT, SATURATE, ERROR ]
+        return: "DECIMAL<P,S>"
```
EpsilonPrime, line 219: Will the precision and scale stay the same as a result of this operation or do we need to extend it?  I suppose overflow is quickly likely so it's not a big deal if we don't rescale.

extensions/functions_arithmetic_decimal.yaml
```diff
@@ -174,3 +174,95 @@ aggregate_functions:
         decomposable: MANY
         intermediate: "DECIMAL<38,S>"
         return: "DECIMAL<38,S>"
+
+  -
+    name: "power"
```
EpsilonPrime, line 179: This the aggregate functions section.  These definitions should be placed in the scalar functions section.

extensions/functions_arithmetic_decimal.yaml
anshuldata: fixed it, Thanks

extensions/functions_arithmetic_decimal.yaml
anshuldata: fixed it

extensions/functions_arithmetic_decimal.yaml
anshuldata: fixed it

extensions/functions_arithmetic_decimal.yaml
anshuldata: max it will be Decimal(37, S), which  supports factorial of max 33 (verified in duckdb and Snowflake). In both these engines type of return datatype was max size and didn't depend on input size. So I made it Decimal(38, S)

EpsilonPrime (dismissed):

anshuldata:
@EpsilonPrime I don't have write access to branch. Can you kindly merge ?
Thanks

EpsilonPrime:
As this PR is somewhere between a new feature and a change to an existing one I'd like to get another approval for this PR before we merge it.  I'll check with the other folks at the Substrait community meeting on Wednesday and will likely merge it then.

extensions/functions_arithmetic_decimal.yaml
```diff
@@ -102,14 +102,96 @@ scalar_functions:
...
           - name: x
             value: decimal<P,S>
         return: decimal<P,S>
+  -
+    name: "power"
+    description: "Take the power with x as the base and y as exponent."
+    impls:
+      - args:
+          - name: x
+            value: "DECIMAL<P,S>"
```
jacques-n, line 119: I don't think these data types are right. This says that both arguments must have the same precision and scale. I think this and all the ones below are not right.

extensions/functions_arithmetic_decimal.yaml
```diff
@@ -102,14 +102,96 @@ scalar_functions:
...
+     name: "bitwise_and"
+     description: >
+        Return the bitwise AND result for two decimal inputs.
+     impls:
+      - args:
+          - name: x
+            value: "DECIMAL<P,S>"
+          - name: y
+            value: "DECIMAL<P,S>"
+        return: "DECIMAL<P,S>"
```
jacques-n, line 172: Isn't this some pattern of max of others?

jacques-n (changes requested):
I don't think the precision/scale handling is correct here.

extensions/functions_arithmetic_decimal.yaml
anshuldata: Fixed power function. I have removed all other functions, I will add them in separate PR

extensions/functions_arithmetic_decimal.yaml
anshuldata: I have add this function in separate PR. I will take care of this comment in that future PR

jacques-n (approved):
This lgtm. @EpsilonPrime ?

jacques-n:
@anshuldata can you rebase?

anshuldata:
> @anshuldata can you rebase?

rebased it

extensions/functions_arithmetic_decimal.yaml
anshuldata: resolving since added this functions to #675

jacques-n (approved):

PR #661, opened 2024-07-10
feat: add output_schema to ExpandRel message

andrew-coleman:
Background: I’m currently working on improving the test pass rate of the TPC-DS suite for the `spark` module in `substrait-java`.  One of the relations that is currently not supported by the spark translator is the `Expand` relation.  It’s not implemented in the `core` module either.  The Spark catalyst query optimiser injects `Expand` into the logical plan when it encounters [distinct aggregations](https://github.com/apache/spark/blob/master/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/optimizer/RewriteDistinctAggregates.scala#L171), so I’m implementing `Expand` in `substrait-java` to support this scenario (and fix a number of test cases).

However, the Spark Expand object requires an [extra parameter](https://github.com/apache/spark/blob/master/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/plans/logical/basicLogicalOperators.scala#L1394) that is currently not available in the Substrait Expand protobuf message.  This extra parameter defines the schema of the output that gets generated by applying each of the projections.

This PR proposes an addition to the proto message that would support this.

### DIFF
--- a/proto/substrait/algebra.proto
+++ b/proto/substrait/algebra.proto
@@ -387,6 +387,8 @@ message ExchangeRel {
 message ExpandRel {
   RelCommon common = 1;
   Rel input = 2;
+  // The columns emitted as a result of applying the Expand fields
+  NamedStruct output_schema = 3;
   // There should be one definition here for each input field.  Any fields beyond the provided
   // definitions will be emitted as is (as if a consistent_field record with an identity
   // expression was provided).
--- a/site/docs/relations/physical_relations.md
+++ b/site/docs/relations/physical_relations.md
@@ -236,10 +236,11 @@ The expand operation creates duplicates of input records based on the Expand Fie
 
 ### Expand Properties
 
-| Property  | Description                          | Required |
-| --------- |--------------------------------------| -------- |
-| Input     | The relational input.                | Required |
-| Direct Fields | Expressions describing the output fields.  These refer to the schema of the input.  Each Direct Field must be an expression or a Switching Field  | Required |
+| Property      | Description                                                                                                                                      | Required |
+|---------------|--------------------------------------------------------------------------------------------------------------------------------------------------| -------- |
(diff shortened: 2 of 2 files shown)


andrew-coleman:
> Isn't it possible to calculate the output schema from the message itself?

Yes, we can derive the type information from the field expressions, but we can't determine the names that spark gives the new columns.  That's the reason for adding a `NamedStruct` rather than a `Struct` (which, as you say, would have been redundant).  It needs to contain the information for building a spark `AttributeReference`, similar to a `ReadRel` message.

andrew-coleman:
Just wondering if there has been any discussion on this?

EpsilonPrime:
The only names that truly matter on the ones that are emitted by the plan.  The root's names allow you to specify these.

For all of the intermediate names one shouldn't need to keep track of them.  For the text version of the Substrait plan I ended up automatically generating names and using those generated names as references later on in the plan.  Since they don't matter the round trip from binary to text and back was just fine.  That said, the intermediate names would change if I went from text to binary and back which is the same problem with Spark.  But as they're intermediate I'd argue that they don't really matter what they are.

If we do add the names they shouldn't be required for the execution to succeed so having something like root names to provide intermediate names as a metadata item seems like the right approach.  There are also some similarities to the emit logic that we may be able to leverage.

jacques-n:
I'm going to close this ticket as the specific approach seems to be no inline what how we should solve the underlying problem in Substrait. Suggest @andrew-coleman open a new PR that introduces optional metadata for this per @westonpace comments.

andrew-coleman:
Many thanks for the suggestion and apologies for the delay in responding (holiday season!).

I have opened a new PR https://github.com/substrait-io/substrait/pull/696, which I hope is consistent with your suggestion.

PR #662, opened 2024-07-11
feat: normalize the join types

EpsilonPrime:
There are two groups of join types in the definition with differing enums.

This PR leaves JoinRel's SEMI and ANTI as the canonical names for
LEFT_SEMI and LEFT_ANTI.  Aliases are not allowed due to JSON (and
text) serialization behavior.

This PR also adds RIGHT_SEMI and RIGHT_ANTI to JoinRel's JoinType.

RIGHT_SINGLE is added to all types.  The PR correspondingly renames
SINGLE to LEFT_SINGLE, ANTI TO LEFT_ANTI, and SEMI to LEFT_SEMI.
Finally this PR adds LEFT_SINGLE to all of the other join types.

BREAKING CHANGE:    JoinRel's type enum now has LEFT_SINGLE
instead of SINGLE.  Similarly there is now LEFT_ANTI and LEFT_SEMI.
Other values are available in all join type enums.  This affects JSON and
text formats only (binary plans -- the interoperable part of Substrait --
will still be compatible before and after this change).


site/docs/relations/logical_relations.md
```diff
@@ -231,6 +231,8 @@ The join operation will combine two separate inputs into a single output, based
 | Semi | Returns records from the left input. These are returned only if the records have a join partner on the right side. |
 | Anti  | Return records from the left input. These are returned only if the records do not have a join partner on the right side. |
 | Single | Returns one join partner per entry on the left input. If more than one join partner exists, there are two valid semantics. 1) Only the first match is returned. 2) The system throws an error. If there is no match between the left and right inputs, NULL is returned. |
+| Right Semi | Returns records from the right input. These are returned only if the records have a join partner on the left side. |
```
westonpace, line 234: The description (above) for single join seems incorrect according to https://15721.courses.cs.cmu.edu/spring2018/papers/16-optimizer2/hyperjoins-btw2017.pdf

Also, the description for single join is inconsistent with the other join descriptions.  From the paper we have:

> To express this in relational algebra we introduce a new operator, the single join. It behaves
nearly identical to an outer join, but reports an error if more than one join partner is found

I am going to choose to interpret "outer join" here as "left outer join" based on surrounding context.

I recommend:

```
Return all records from the left input.  For each cross input match, return a record including the data from both sides.  For any remaining non-matching records from the left input, return the left record along with nulls for the right input.  If a record from the left input has more than one match in the right input then an error must be raised.
```

Do we have any reference that the "only the first match is returned" semantic is needed by DuckDb?  I agree it is a thing that could be done but I'd rather not introduce options / variations unless there is a concrete use for them.

proto/substrait/algebra.proto
```diff
@@ -173,11 +173,13 @@ message JoinRel {
...
     JOIN_TYPE_RIGHT = 4;
-    JOIN_TYPE_SEMI = 5;
-    JOIN_TYPE_ANTI = 6;
+    JOIN_TYPE_SEMI = 5; // This is the left semi join.
+    JOIN_TYPE_ANTI = 6; // This is the left anti join.
     // This join is useful for nested sub-queries where we need exactly one record in output (or throw exception)
     // See Section 3.2 of https://15721.courses.cs.cmu.edu/spring2018/papers/16-optimizer2/hyperjoins-btw2017.pdf
     JOIN_TYPE_SINGLE = 7;
+    JOIN_TYPE_RIGHT_SEMI = 8;
+    JOIN_TYPE_RIGHT_ANTI = 9;
```
westonpace, line 182: Are there cases where a right semi join or a right anti join is fundamentally different than swapping the inputs to a left semi join or a left anti join?  Or are we just adding them for completeness' sake?  Do we need a right single join as well?

proto/substrait/algebra.proto
```diff
@@ -654,6 +656,9 @@ message HashJoinRel {
     JOIN_TYPE_RIGHT_SEMI = 6;
     JOIN_TYPE_LEFT_ANTI = 7;
     JOIN_TYPE_RIGHT_ANTI = 8;
+    // This join is useful for nested sub-queries where we need exactly one record in output (or throw exception)
+    // See Section 3.2 of https://15721.courses.cs.cmu.edu/spring2018/papers/16-optimizer2/hyperjoins-btw2017.pdf
```
westonpace, line 660: Can we just remove these comments from all the proto files?  If we want to further justify the single join (which might make sense as it is not ubiquitous yet) then perhaps we put the link to section 3.2 in the site docs where the single join operator is described?

westonpace (commented):
I'm generally ok with this but we could clean the language up a bit.

proto/substrait/algebra.proto
EpsilonPrime: Here we're adding it because it's present in all of the other join type enums.

Overall, the argument for having left and right is ease of roundtrip.  One can obviously switch the inputs but it doesn't appear to be a common practice despite its simplicity.

proto/substrait/algebra.proto
EpsilonPrime: Done.

site/docs/relations/logical_relations.md
EpsilonPrime: @pdet Does DuckDB return only the first match?  If so we might need to make the behavior an option.

proto/substrait/algebra.proto
EpsilonPrime: Based on the description of the join (returning NULLs for one side) it seems like it should be added.

site/docs/relations/logical_relations.md
pdet: In DuckDB the `SINGLE join` is like the `LEFT OUTER JOIN`. However it returns at most one join partner per entry on the LEFT side, and NULL if no partner is found.

There is no guarantee it will be the first one it encounters, though.

site/docs/relations/logical_relations.md
```diff
@@ -229,8 +229,11 @@ The join operation will combine two separate inputs into a single output, based
...
 | Anti  | Return records from the left input. These are returned only if the records do not have a join partner on the right side. |
-| Single | Returns one join partner per entry on the left input. If more than one join partner exists, there are two valid semantics. 1) Only the first match is returned. 2) The system throws an error. If there is no match between the left and right inputs, NULL is returned. |
+| Right Anti  | Return records from the right input. These are returned only if the records do not have a join partner on the left side. |
+| Single | Return all records from the left input.  For each cross input match, return a record including the data from both sides.  For any remaining non-matching records from the left input, return the left record along with nulls for the right input.  If a record from the left input has more than one match in the right input then an error must be raised.  Useful for nested sub-queries where we need exactly one record in output (or throw exception).  See Section 3.2 of https://15721.courses.cs.cmu.edu/spring2018/papers/16-optimizer2/hyperjoins-btw2017.pdf for more information. |
```
jacques-n, line 235:
```suggestion
| Single | Return all records from the primary (left) input with no join expansion. If at least one record from the secondary (right) input matches the primary, return one arbitrary matching record from the secondary (right) input. For any primary (left) records without matching secondary (right) records, return the primary (left) record along with nulls for the secondary (right) input. Similar to a left outer join but only returns one right match at most. Useful for nested sub-queries where we need exactly one record in output (or throw exception).  See Section 3.2 of https://15721.courses.cs.cmu.edu/spring2018/papers/16-optimizer2/hyperjoins-btw2017.pdf for more information. |
```

jacques-n (commented):
Think that we just need to update description of single input. I proposed a description that will hopefully be correct.

site/docs/relations/logical_relations.md
westonpace: I'm ok with this.  Not a huge fan of introducing new terminology (primary, secondary) and its confusing to me how this actually solves the nested subquery problem (since it doesn't raise an error) but I agree that we should document how engines are actually working when they are in disagreement with the paper.

jacques-n: I introduced it since the other one referred to primary but it wasn't defined above.

The more I think about this I think we should call this Left Single and Right Single. That would be more consistent with other names. Having Single and Right Single is just weird.

EpsilonPrime: I agree that it should be left single and right single.  But for the purposes of this PR I just want to normalize the types available in the two join types (essentially a non-breaking change).  A rename could break code so I would rather accomplish that in a later step.

jacques-n: > A rename could break code so I would rather accomplish that in a later step.

I'm inclined to fix it now while we're already fixing the other stuff instead of punting it. If we're changing the definition of what SINGLE means (since it was vague before), that's basically a breaking spec change even if the protobuf change itself isn't breaking. 

Who wants to spend the time to come back and fix that? If we need to fix something, I'd prefer to just fix it while we're in that exact code already.

EpsilonPrime:
I've updated the PR to rename SINGLE to LEFT_SINGLE and have removed primary/secondary from the text.

There's still a looming breaking change in the future which is replacing all of the enums with a single version with the same values but that's best handled in a major release with a number of other major breaking changes.

site/docs/relations/logical_relations.md
```diff
@@ -229,8 +229,11 @@ The join operation will combine two separate inputs into a single output, based
...
 | Anti  | Return records from the left input. These are returned only if the records do not have a join partner on the right side. |
-| Single | Returns one join partner per entry on the left input. If more than one join partner exists, there are two valid semantics. 1) Only the first match is returned. 2) The system throws an error. If there is no match between the left and right inputs, NULL is returned. |
+| Right Anti  | Return records from the right input. These are returned only if the records do not have a join partner on the left side. |
+| Left Single | Return all records from the left input with no join expansion. If at least one record from the right input matches the primary, return one arbitrary matching record from the right input. For any left records without matching right records, return the primary (left) record along with nulls for the right input. Similar to a left outer join but only returns one right match at most. Useful for nested sub-queries where we need exactly one record in output (or throw exception).  See Section 3.2 of https://15721.courses.cs.cmu.edu/spring2018/papers/16-optimizer2/hyperjoins-btw2017.pdf for more information. |
```
jacques-n, line 235: missed changing some of the primary references.

jacques-n (commented):
minor change around primary wording. 

also, would add to the breaking change that the change is only breaking at the name of the field/json plans. Protobuf plans will continue to be cross compatible.

site/docs/relations/logical_relations.md
EpsilonPrime: Removed those.  Didn't touch the original use of primary in the set portion of the documentation.

site/docs/relations/logical_relations.md
```diff
@@ -222,15 +222,18 @@ The join operation will combine two separate inputs into a single output, based
...
+| Inner | Return records from the left side only if they match the right side. Return records from the right side only when they match the left side. For each cross input match, return a record including the data from both sides. Non-matching records are ignored.                                                                                                                                                                                                                                                                                                                                                     |
```
jacques-n, line 227: can you avoid the reformatting. makes it really hard to see what is changed in diff.

EpsilonPrime: Done, also renamed ANTI and SEMI to LEFT_ANTI and LEFT_SEMI.

westonpace (approved):

jacques-n (approved):
LGTM. Thanks for pushing this along @EpsilonPrime !

PR #665, opened 2024-07-16
feat: add precision to IntervalDay and new IntervalCompound type

### LINKED ISSUE #664: Support for an interval type covering both months and (sub)seconds
Substrait currently defines two interval types, [IntervalYearToMonth](https://github.com/substrait-io/substrait/blob/f9e5f9c515d4b8be079bc7d9dfcd89a6fa5e6c7e/proto/substrait/algebra.proto#L876) which contains a number of years and months, and [IntervalDayToSecond](https://github.com/substrait-io/substrait/blob/f9e5f9c515d4b8be079bc7d9dfcd89a6fa5e6c7e/proto/substrait/algebra.proto#L881).

Neither of these types is capable of expressing full intervals that combine years, months, days, minutes, seconds, etc. (For YearToMonth that's probably clear; for DayToSecond one could in theory use large numbers for days but given the number of days in a month/year is not unique, there is a semantic difference between an interval of "40 DAYS 1 MINUTE" and "1 MONTH 10 DAYS 1 MINUTE"

- Spark has an [CalendarIntervalType](https://spark.apache.org/docs/latest/api/scala/org/apache/spark/sql/types/CalendarIntervalType.html) that it uses for this purpose, 
- Arrow (and by proxy DataFusion) has [IntervalMonthDayNano](https://docs.rs/arrow/latest/arrow/datatypes/struct.IntervalMonthDayNano.html)
- [Postgres](https://www.postgresqltutorial.com/postgresql-tutorial/postgresql-interval/) says `Internally, PostgreSQL stores interval values as months, days, and seconds. The months and days values are integers while the seconds field can have fractions`
- [Duckdb](https://duckdb.org/docs/sql/functions/interval.html) doesn't go into details of implementation but the docs seem like their interval supports a combination of all magnitudes, and they have the same [issue when converting substrait](https://github.com/duckdb/substrait/blob/ccf848f6ab4a4673dc93413080fc640796c09b86/src/to_substrait.cpp#L186) 

Would you be open to adding an third interval type that combines three numbers: months (i32), days (i32), (micro)seconds (i32 for micro or decimal for seconds)?

Blizzara:
* Update IntervalDay to support multiple subsecond precisions (0-9). This is done in a way that should be effectively compatible with old systems. Old plans should be able to be continue to be consumed treating the old values as microsecond precision.
* Add new IntervalCompound type which is a combination of IntervalMonth and IntervalDay

BREAKING CHANGE: The encoding of IntervalDay literals has changed in a strictly backwards incompatible way. However, the logical meaning across encoding is maintained using a oneof. Moving a field into a oneof makes unset/set to zero unclear with older messages but the fields are defined such that the logical meaning of the two is indistinct. If neither microseconds nor precision is set, the value can be considered a precision 6 value. If you aren't using IntervalDay type, you will not need to make any changes.

BREAKING CHANGE: TypeExpression and Parametertized type protobufs (used to serialize output derivation) are updated to match the now compound nature of IntervalDay. If you use protobuf to serialize output derivation that refer to IntervalDay type, you will need to rework that logic.

fixes #664


proto/substrait/parameterized_types.proto
```diff
@@ -27,6 +27,7 @@ message ParameterizedType {
     Type.Time time = 17;
     Type.IntervalYear interval_year = 19;
     Type.IntervalDay interval_day = 20;
+    Type.IntervalMonth interval_month = 36;
```
Blizzara, line 30: I aligned with the existing names, but personally I think I'd prefer `IntervalMonthDayNano` or similar to be more descriptive - thoughts?

proto/substrait/algebra.proto
```diff
@@ -810,6 +810,7 @@ message Expression {
       int64 time = 17;
       IntervalYearToMonth interval_year_to_month = 19;
       IntervalDayToSecond interval_day_to_second = 20;
+      IntervalMonthToNano interval_month_to_nano = 36;
```
Blizzara, line 813: I aligned with existing names, but the `To` feels weird to me - I guess it indicates the literal contains data for precisions from X to Y, but I feel these would be clearer without it. Can't rename the existing ones w/o breaking things, but should this be `IntervalMonthDayNano` ?

proto/substrait/algebra.proto
```diff
@@ -884,6 +885,12 @@ message Expression {
       int32 microseconds = 3;
     }
 
+    message IntervalMonthToNano {
+      int32 months = 1;
+      int32 days = 2;
+      int64 nanoseconds = 3;
```
Blizzara, line 891: I opted for nanoseconds here for simplicity, but alternative is to use a variable precision instead like PrecisionTimestamp(Tz) - preferences?

proto/substrait/parameterized_types.proto
vbarua: `IntervalMonthDayNano` or some variant seems reasonable (and better) here.

I don't think it makes sense to align with the existing names. If anything I want
* IntervalYear -> IntervalYearToMonth
* IntervalDay -> IntervalDayToSecond

to better align with the other names in the file, but that's not a backwards compatible change.

jacques-n (changes requested):
I've posted my main comments on #664 . One more cosmetic comment here.

site/docs/types/type_classes.md
```diff
@@ -8,24 +8,25 @@ Implementations of a Substrait type must support *at least* this set of values,
...
-| uuid            | A universally-unique identifier composed of 128 bits. Typically presented to users in the following hexadecimal format: `c48ffa9e-64f4-44cb-ae47-152b4e60e77b`. Any 128-bit value is allowed, without specific adherence to RFC4122. | 16-byte `binary`
+| Type Name      | Description                                                                                                                                                                                                                                                                                                                                      | Protobuf representation for literals
```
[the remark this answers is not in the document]
Blizzara: Done! (otherwise I'll wait for consensus on #664 before changing the impl here)

jacques-n:
Per other comment, I think the discussion is complete in #664. Would be great to get PR update.

jacques-n (approved):

jacques-n:
I've updated the PR here. @westonpace @EpsilonPrime @vbarua , can you all take a look? 

This is technically a breaking change although the way I did it, it isn't logically one. Nonetheless, I think we should have sufficient +1s before merging.

proto/substrait/algebra.proto
```diff
@@ -881,7 +882,21 @@ message Expression {
     message IntervalDayToSecond {
       int32 days = 1;
       int32 seconds = 2;
-      int32 microseconds = 3;
+
+      // oneof ensures that a user can only use precision (and subseconds) or microseconds
```
EpsilonPrime, line 886: I don't think we need the comment (a reader of protobufs should know how oneof works).

proto/substrait/algebra.proto
```diff
@@ -881,7 +882,21 @@ message Expression {
...
+      oneof precision_mode {
+        int32 microseconds = 3 [deprecated = true]; // use precision and subseconds below, they cover and replace microseconds.
+        // Sub-second precision, 0 means the value given is in seconds, 3 is milliseconds, 6 microseconds, 9 is nanoseconds.
+        int32 precision = 4;
+      }
+
+      // the number of fractional seconds using  1e(-precision) units. Should be ignored if microseconds is set above.
```
EpsilonPrime, line 893: Squash the double space into a single one here?

proto/substrait/type.proto
```diff
@@ -122,14 +123,28 @@ message Type {
...
     Nullability nullability = 2;
   }
 
+  // An interval consisting of days, seconds, and microseconds
   message IntervalDay {
     uint32 type_variation_reference = 1;
     Nullability nullability = 2;
+
+    // Sub-second precision, 0 means the value given is in seconds, 3 is milliseconds, 6 microseconds, 9 is nanoseconds, etc.
+    // if unset, treat as 6. this allows non-backwards breaking change
```
EpsilonPrime, line 138: The part about being non-backwards belongs in the PR and probably not the future implementation.  I suggest removing it.

proto/substrait/type.proto
```diff
@@ -122,14 +123,28 @@ message Type {
...
+    optional int32 precision = 3;
+  }
+
+  // An interval consisting of months, days, and nanoseconds
```
EpsilonPrime, line 142: Does this also contain years?

proto/substrait/algebra.proto
```diff
@@ -881,7 +882,21 @@ message Expression {
...
+      int64 subseconds = 5;
+    }
+
+    message IntervalCompound {
```
EpsilonPrime, line 897: Essentially this is struct<iyear, iday>.  To specify it in text one would need to use the nested struct syntax.  Would this be better flattened with the fields from both types included at the top level?

proto/substrait/algebra.proto
jacques-n: moved things around a little. the pairing of precision with subseconds is the part that is a little unusual with oneof.

proto/substrait/algebra.proto
jacques-n: Are you talking about the JSON representation or something else? I wasn't really worried about the json representation. Felt like this better clearly communicated the nature as opposed to field copying. It's a weak preference though.

EpsilonPrime: Both JSON and the Substrait Text format (which was what I was thinking about since it'll require implementation) would be affected.  I prefer struct<iyear, iday> to the compound approach.  I would also classify my preference as weak.

EpsilonPrime (dismissed):

proto/substrait/algebra.proto
jacques-n: I would expect the substrait text format might have the fields be flattened. I don't think it is critical for it to follow the same structure as the protobufs as long as translation from one to the other is supported, right?

jacques-n (approved):

westonpace (approved):
We could potentially make this a non-breaking change if we documented something like:

```
If the microseconds value is non-zero then ignore the other two fields and assume microsecond precision.
```

However, that burdens consumers with (a slight amount) of technical debt.

From a quick glance it appears that:

 * Arrow is incapable of producing `IntervalDayToSecond`
 * Arrow is consuming `IntervalDayToSecond` incorrectly already (it is completely ignoring the microseconds field)
 * Datafusion can produce `IntervalDayToSecond` with a microseconds value (this change will break anyone using this behavior)
 * Datafusion consumes `IntervalDayToSecond` but, since DF only has millisecond precision, it silently discards some precision. (this change will break users relying on this behavior but at least we have some positive benefit here of turning that silent refusal into a noisy refusal).

Overall, I think I'm on the fence but will lean towards progress.

jacques-n:
Given @westonpace seeing a lack of clarity about the fact that the literal is backwards compatible, I added an additional comment to the IntervalDay literal to better clarify that consumers need to understand that two different variations are supported. @westonpace and @EpsilonPrime, if you're both good, I'll merge this. Thanks

westonpace (approved):

jacques-n (approved):

jacques-n:
Just need an approval from @vbarua or @EpsilonPrime and/or @cpcloud and then we can merge.

EpsilonPrime (approved):
I thought about it and the text format should be fine reaching down into the other two types as an implementation detail.  This approach is good enough.

vbarua:
Taking a look at this today. I want to verify that backwards compatibility of this internally.

site/docs/types/type_classes.md
```diff
@@ -43,6 +42,8 @@ Compound type classes are type classes that need to be configured by means of a
...
+| INTERVAL_DAY&lt;P&gt;         | Interval day to second. Supports a range of [-3,650,000..3,650,000] days with fractional second precision (P, number of digits) 0 <= P <= 9. Usually stored as separate integers for various components, but only the total number of fractional seconds is significant, i.e. `1d 0s` is considered equal to `0d 86400s`. | `int32` days, `int32` seconds, and `int64` fractional seconds, with the added constraint that each component can never independently specify more than 10,000 years, even if the components have opposite signs (e.g. `3650001d -86400s 0us` is **not** allowed)
+| INTERVAL_COMPOUND&lt;P&gt;    | A compound interval type that is composed of elements of the underlying elements and rules of both interval_month and interval_day to express arbitrary durations across multiple grains. Substrait gives no definition for the conversion of values between independent grains (e.g. months to days).
```
vbarua, line 46: > across multiple grains

What is a grain in this context?

site/docs/types/type_classes.md
```diff
@@ -43,6 +42,8 @@ Compound type classes are type classes that need to be configured by means of a
...
 | PRECISIONTIMESTAMPTZ&lt;P&gt; | A timezone-aware timestamp, with fractional second precision (P, number of digits) 0 <= P <= 9. Similar to aware datetime in Python.                                                                                                                                                                                                                                  | `int64` seconds, milliseconds, microseconds or nanoseconds since 1970-01-01 00:00:00.000000000 UTC
+| INTERVAL_DAY&lt;P&gt;         | Interval day to second. Supports a range of [-3,650,000..3,650,000] days with fractional second precision (P, number of digits) 0 <= P <= 9. Usually stored as separate integers for various components, but only the total number of fractional seconds is significant, i.e. `1d 0s` is considered equal to `0d 86400s`. | `int32` days, `int32` seconds, and `int64` fractional seconds, with the added constraint that each component can never independently specify more than 10,000 years, even if the components have opposite signs (e.g. `3650001d -86400s 0us` is **not** allowed)
```
vbarua, line 45: Now that this is a parameterised type, do we need to update all the function extensions that refer to it as `interval_day` to use `interval_day<P>` instead?

jacques-n: Good point. I'm inclined to do in a follow-along once this is merged.

vbarua (approved):
Overall looks good to me!

site/docs/types/type_classes.md
jacques-n: year/month vs day/second. there are no consistent reliable translations between each.

PR #666, opened 2024-07-19
feat: add precision timestamp datetime fn variants

vbarua:


vbarua:
This PR adds functions variants for PrecisionTimestamp and PrecisionTimestampTZ to aid in migrating off of the the deprecated Timestamp and TimestampTZ types.

Not all functions can be given variants easily. For example for
```
  -
    name: "add"
      - args:
          - name: x
            value: date
          - name: y
            value: interval_day
        return: timestamp
```

We cannot add a `precision_timestamp` variant because it would only differ _in the output type_. Choosing not to handle these for now.

vbarua:
I've loaded these into substrait-java successfully.

EpsilonPrime (approved):

jacques-n (approved):

PR #668, opened 2024-07-23
fix: use int64 instead of uint64 for PrecisionTimestamp(Tz) literal value

Blizzara:
This allows timestamps to refer to time before epoch, and aligns with other systems (Spark, DataFusion/Arrow, DuckDB, Postgres, Parquet at least)

BREAKING CHANGE: PrecisionTimestamp(Tz) literal's value is now int64 instead of uint64

In #659 I created this `PrecisionTimestamp` message to include the precision, but unfortunately copied the value's type as-is, not realizing the unsignedness is a problem.


westonpace (approved):
Yikes, good catch :sweat_smile:

jacques-n:
I think we should probably update content here:
https://substrait.io/types/type_classes/

So that it is very clear the valid range of timestamps. Right now it is much more vague than other types.

Blizzara:
> I think we should probably update content here: https://substrait.io/types/type_classes/
> 
> So that it is very clear the valid range of timestamps. Right now it is much more vague than other types.

Would you have a proposal? Happy to include in this PR if so, otherwise I’d suggest doing it as followup to get this break in as soon as possible.

(Though I see I need to update the page you linked for uint->int, thanks!)

Blizzara:
@westonpace @jacques-n et al, any further comments or would this be good to go? (I'd like to get it in next release if possible..)

westonpace (approved):

westonpace:
> So that it is very clear the valid range of timestamps. Right now it is much more vague than other types.

I seem to recall this being intentional.  The PR to introduce precision timestamps originally stated `A naive timestamp within [1000-01-01 00:00:00.000000000..9999-12-31 23:59:59.999999999]` but this is not possible with nanosecond resolution on systems that store timestamps as 8 byte integers (some store them as a pair of 8 byte integers for second / subsecond).  Rather than make the range based on the precision we decided to just be vague about the range.

site/docs/types/type_classes.md
```diff
@@ -41,8 +41,8 @@ Compound type classes are type classes that need to be configured by means of a
...
-| PRECISIONTIMESTAMPTZ&lt;P&gt; | A timezone-aware timestamp, with fractional second precision (P, number of digits) 0 <= P <= 9. Similar to aware datetime in Python.                                                                                                                                                                                                                                  | `uint64` microseconds or nanoseconds since 1970-01-01 00:00:00.000000000 UTC
+| PRECISIONTIMESTAMP&lt;P&gt;   | A timestamp with fractional second precision (P, number of digits) 0 <= P <= 9. Does not include timezone information and can thus not be unambiguously mapped to a moment on the timeline without context. Similar to naive datetime in Python.                                                                                                                      | `int64` seconds, milliseconds, microseconds or nanoseconds since 1970-01-01 00:00:00.000000000 (in an unspecified timezone)
```
jacques-n, line 44: Given the discussion on ranges, I'm and P, I'm struggling with this description. How do I do a literal if P = 1,2,4,5,7 or 8? I'm going to approve this PR but we should open a follow-up to address the inconsistency here. Either:

P can only be 0,3,6 or 9
OR
P can be 0..9 and we need additional specification around literal data.

jacques-n (approved):

PR #669, opened 2024-07-24
feat: add logarithmic functions with decimal type args

scgkiran:


jacques-n (approved):
This seems fine to me. I can imagine some systems returning a decimal but suspect most return floats.

PR #670, opened 2024-07-24
feat: add aggregate count functions with decimal return type

scgkiran:


extensions/functions_aggregate_decimal.yaml
```diff
@@ -0,0 +1,41 @@
+%YAML 1.2
+---
+aggregate_functions:
+  - name: "count"
```
Blizzara, line 4: overloading functions based only on return type causes problems, see e.g. https://github.com/substrait-io/substrait/pull/606#issuecomment-1964054841

scgkiran: I have them in a different yaml file. can this still cause problems?

jacques-n: different files should be ok.

Blizzara: I believe it'll still break substrait-java, see https://github.com/substrait-io/substrait-java/issues/275

jacques-n: I think we should fix that in substrait-java then as opposed to holding back the specification. Extensions being independent is key to extensions being extensions :)

extensions/functions_aggregate_decimal.yaml
```diff
@@ -0,0 +1,41 @@
+%YAML 1.2
+---
+aggregate_functions:
```
jacques-n, line 3: the name of this file is weird. in other cases type names in filenames are about input types and this one is about output types. I suggest we rename file to something that is more descriptive. Maybe also add a little text to the functions clarifying how they are different than classic count functions.

jacques-n (changes requested):

extensions/functions_aggregate_decimal.yaml
vbarua: It's possible for us to work around duplicates in substrait-java by excluding them from the default extensions for now. See https://github.com/substrait-io/substrait-java/issues/275#issuecomment-2257176686

vbarua (changes requested):
Overall these changes look reasonable to me. Just one suggestion about the name `count_star`.

extensions/functions_aggregate_decimal_output.yaml
```diff
@@ -0,0 +1,41 @@
...
+          - name: x
+            value: any
+        options:
+          overflow:
+            values: [SILENT, SATURATE, ERROR]
+        nullability: DECLARED_OUTPUT
+        decomposable: MANY
+        intermediate: decimal<38,0>
+        return: decimal<38,0>
+  - name: "count_star"
```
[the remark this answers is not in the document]
scgkiran: changed it count

vbarua (approved):
Changes look good to me.

jacques-n (approved):
lgtm

PR #671, opened 2024-07-24
feat: add rounding functions with decimal type args

scgkiran:


extensions/functions_rounding.yaml
```diff
@@ -265,6 +273,86 @@ scalar_functions:
...
+                - AWAY_FROM_ZERO: round negative values with FLOOR rule, round positive values with CEILING rule
+                - TIE_DOWN: round ties with FLOOR rule
+                - TIE_UP: round ties with CEILING rule
+                - TIE_TOWARDS_ZERO: round ties with TRUNCATE rule
+                - TIE_TO_ODD: round to nearest value; if exactly halfway, tie
+                  to the odd option.
+            values: [ TIE_TO_EVEN, TIE_AWAY_FROM_ZERO, TRUNCATE, CEILING, FLOOR,
+              AWAY_FROM_ZERO, TIE_DOWN, TIE_UP, TIE_TOWARDS_ZERO, TIE_TO_ODD ]
+        nullability: DECLARED_OUTPUT
+        return: decimal?<P1,S2>
```
jacques-n, line 358: where does s2 come from?

jacques-n (changes requested):

extensions/functions_rounding.yaml
```diff
@@ -265,6 +273,88 @@ scalar_functions:
...
+                - TIE_DOWN: round ties with FLOOR rule
+                - TIE_UP: round ties with CEILING rule
+                - TIE_TOWARDS_ZERO: round ties with TRUNCATE rule
+                - TIE_TO_ODD: round to nearest value; if exactly halfway, tie
+                  to the odd option.
+            values: [ TIE_TO_EVEN, TIE_AWAY_FROM_ZERO, TRUNCATE, CEILING, FLOOR,
+              AWAY_FROM_ZERO, TIE_DOWN, TIE_UP, TIE_TOWARDS_ZERO, TIE_TO_ODD ]
+        nullability: DECLARED_OUTPUT
+        return: |-
+          S2 = iff(s > 0, s, 0)
```
jacques-n, line 359: output type derivation doesn't support referring to argument values. we've talked about supporting constant arguments but I don't think we've added to spec, have we?

PR #676, opened 2024-08-01
add arithmetic function for bitwise not with decimal arguments

anshuldata:

 - `feat: add feature X`
 - `fix: X in case of Y`


extensions/functions_arithmetic_decimal.yaml
```diff
@@ -166,6 +166,17 @@ scalar_functions:
...
+    description: >
+      Return the bitwise NOT result for one decimal input.
+      In inputs scale must be 0 (i.e. only integer types are allowed).
+      Result precision should be one more than input precision.
+      Result exceeding precision limit will raise an error.
+    impls:
+      - args:
+          - name: x
+            value: "DECIMAL<P,0>"
+        return: "DECIMAL<P+1,0>"
```
jacques-n, line 179: I think you need to do this through a SSA expression.

Why does the precision increase by one when you not? That seems strange. If I were to not a int32, I would expect the same output precision as input precision. If I do `bitwise_not(1::decimal(38,0)` I wouldn't expect to receive decimal(39,0).

anshuldata: Fixed function to return same behavior. I was testing decimal Snowflake. For input precision 1 Sf returns precision 2. I have captured this behavior in a separate function in snowflake specific file

jacques-n (approved):
additional test cases would be good but this is a good start.

PR #679, opened 2024-08-11
feat: update interval_day function extensions to include precision param

jacques-n:


EpsilonPrime (approved):

extensions/functions_datetime.yaml
```diff
@@ -281,31 +281,31 @@ scalar_functions:
           - name: x
             value: timestamp
           - name: y
-            value: interval_day
+            value: interval_day<P>
         return: timestamp
```
vbarua, line 285: Possible not super relevant as we're slowly deprecating `timestamp`, but given that `timestamp` has microsecond precision should this function only apply to `interval_day<6>` values, or this implicit in the output having a microsecond precision?

vbarua (approved):
Overall LGTM. Left one comment.

extensions/functions_datetime.yaml
jacques-n: i would assume we'd allow any type but the output would always be reduced to precision 6. 

mostly inclined not to worry about since this type is going away anyway.

PR #680, opened 2024-08-14
feat: add test file format

scgkiran:


tests/cases/arithmetic_decimal/power.test
```diff
@@ -0,0 +1,21 @@
...
+### SUBSTRAIT_INCLUDE: extensions/functions_arithmetic_decimal.yaml
+
+# basic: Basic examples without any special cases
+power(8::decimal, 2::decimal<38, 0>) = 64::fp64
+power(1.0::decimal, -1.0::decimal<38, 0>) = 1.0::fp64
+power(2.0::decimal<38, 0>, -2.0::decimal<38, 0>) = 0.25::fp64
+power(13::decimal<38, 0>, 10::decimal<38, 0>) = 137858491849::fp64
+
+# result_more_than_input_precison: Examples demonstrating result with more precision than input
+power(16::decimal<2, 0>, 4::decimal<38, 0>) = 65536::fp64
```
EpsilonPrime, line 11: Perhaps we could reuse the text format's definition of literals?

EpsilonPrime:
I'm guessing we're going to need to make sure that we have something that tests the validity of the test files.  Would a parse test be sufficient enough for that purpose?  Alternatively we could step it up and validate the tests against the function definitions ensuring that they refer to existing functions.

tests/cases/arithmetic_decimal/power.test
```diff
@@ -0,0 +1,21 @@
+### SUBSTRAIT_SCALAR_TEST: 1.0
+### SUBSTRAIT_INCLUDE: extensions/functions_arithmetic_decimal.yaml
+
+# basic: Basic examples without any special cases
+power(8::decimal, 2::decimal<38, 0>) = 64::fp64
```
EpsilonPrime, line 5: BFT will need some way to say "this test is disabled".  Can we add a way to reference the tests?

jacques-n: I don't really understand why BFT would disable a test. A function makes sense but not a test from my pov.

If important (albeit strange), I'd suggest that BFT just list known test failures. Test cases are only a single line.

tests/cases/datetime/lt_datetime.test
```diff
@@ -0,0 +1,25 @@
...
+# timestamp_tz: examples using the timestamp_tz type
+lt(1999-01-08 01:05:05-08:00::timestamp_tz, 1999-01-08 04:05:06-05:00::timestamp_tz) = True::boolean
+lt(1999-01-08 01:05:06-08:00::timestamp_tz, 1999-01-08 04:05:06-05:00::timestamp_tz) = False::boolean
+
+# date: examples using the date type
+lt(2020-12-30::date, 2020-12-31::date) = True::boolean
+lt(2020-12-31::date, 2020-12-30::date) = False::boolean
+
+# interval: examples using the interval type
+lt(INTERVAL 'P7D'::interval, INTERVAL 'P6D'::interval) = False::boolean
```
vbarua, line 17: We don't actually have an `interval` type, we have `iyear`, `idate` and `icompound`. Instead of using
```
INTERVAL 'P7D'::interval
```
why not just something like
```
'P7D'::iday
```
which is less verbose and more precise?

tests/cases/datetime/lt_datetime.test
```diff
@@ -0,0 +1,25 @@
...
+lt(INTERVAL 'P5D'::interval, INTERVAL '6D'::interval) = True::boolean
+lt(INTERVAL 'P5Y'::interval, INTERVAL 'P6Y'::interval) = True::boolean
+lt(INTERVAL 'P7Y'::interval, INTERVAL 'P6Y'::interval) = False::boolean
+
+# null_input: examples with null args or return
+lt(None::interval, INTERVAL 'P5D'::interval) = Null::boolean
+lt(None::date, 2020-12-30::date) = Null::boolean
+lt(None::timestamp, 2018-12-31 13:30:15::timestamp) = Null::boolean
```
vbarua, line 25: Should we use `ts` instead of `timestamp` in the tests themselves?

tests/cases/datetime/lt_datetime.test
```diff
@@ -0,0 +1,25 @@
+### SUBSTRAIT_SCALAR_TEST: 1.0
+### SUBSTRAIT_INCLUDE: /extensions/functions_datetime.yaml
+
+# timestamps: examples using the timestamp type
+lt(2016-12-31 13:30:15::timestamp, 2017-12-31 13:30:15::timestamp) = True::boolean
+lt(2018-12-31 13:30:15::timestamp, 2017-12-31 13:30:15::timestamp) = False::boolean
+
+# timestamp_tz: examples using the timestamp_tz type
+lt(1999-01-08 01:05:05-08:00::timestamp_tz, 1999-01-08 04:05:06-05:00::timestamp_tz) = True::boolean
+lt(1999-01-08 01:05:06-08:00::timestamp_tz, 1999-01-08 04:05:06-05:00::timestamp_tz) = False::boolean
```
vbarua, line 10: Should we use `ts_tz` instead of `timestamp_tz` in the tests?

tests/README.md
```diff
@@ -0,0 +1,138 @@
...
+
+#### Date and Time
+All date and time literals use ISO 8601 format:
+
+- **date**: `YYYY-MM-DD`, example: `2021-01-01`
+- **time**: `HH:MM:SS[.fraction]`, example: `12:00:00.000`
+- **timestamp**: `YYYY-MM-DD HH:MM:SS[.fraction]`, example: `2021-01-01 12:00:00`
+- **timestamp_tz**: `YYYY-MM-DD HH:MM:SS[.fraction]±HH:MM`, example: `2021-01-01 12:00:00+05:30`
+- **interval year**: `INTERVAL 'P[n]Y[n]M'`, example: `INTERVAL 'P2Y3M'` (2 years, 3 months)
+- **interval days**: `INTERVAL 'P[n]DT[n]H[n]M[n]S'`, example: `INTERVAL 'P2DT3H2M9S'` (2 days, 3 hours, 2 minutes, 9 seconds)
```
vbarua, line 100: This a somewhat recent change, but also need to encode precision for interval_day now.

See https://substrait.io/types/type_classes/#compound-types

tests/cases/arithmetic_decimal/power.test
jacques-n: My preference would be to stick with :: format generally. It is a cast format consistent with many systems the people looking at these tests will be familiar with and I believe much more intuitive understood than underscore.

I'm similarly inclined to try to find more standard ways to represent other things as opposed to what the text format has done. There are several examples of alternative ways to implement things to what is commonly done in SQL (which has the most consistent way to approach these things across systems). Examples beyond the use of cast syntax versus underscores include. (1) String literals are double quoted and backticked while sql string literals are single quoted and (2) brackets being overloaded to mean any of intervals, lists, maps and structs.

tests/cases/arithmetic_decimal/power.test
richtia: We'd want to disable/skip a specific test if the function has multiple options.  Where different engines support different options.

jacques-n: I don't understand your comment Rich. The dialect should declare what option values are allowed for which functions. The test should skip test cases whose combination is not supported by a dialect. Why would we need to refer to test cases?

EpsilonPrime:
It's a different use case.  You're interested in testing whether a subset
is compliant or not.  BFT, until now, has been focused on showing all of
the support (positive and negative) to provide a world view (the website).
It can also be used as a way of showing total compliance which can be used
to drive adoption.

tests/cases/arithmetic_decimal/power.test
vbarua: I think the issue is that if we don't declare an option, engines are free to choose _any_ option they support. For testing purposes this is tricky because we can't assume that all engines have the same defaults, and in this case we can get different answers for the same test.

tests/cases/datetime/lt_datetime.test
scgkiran: fixed this.

tests/cases/datetime/lt_datetime.test
scgkiran: changed it

tests/cases/arithmetic_decimal/power.test
scgkiran: When options specified as part of the testcase, those tests are run only with the dialects where the option is supported. AFAIK currently in BFT, all tests without any options they are run on all dialects.

richtia: @scgkiran is correct.  If a test case doesn't specify an option in the BFT, all dialects will run it.  So in this new test file format.  If the dialect doesn't support the option specified, that test case would be skipped.  BFT also skips tests if a type is unsupported (i.e. some dialects don't have decimal support)

tests/cases/arithmetic/add.test
```diff
@@ -0,0 +1,31 @@
...
+
+# overflow: Examples demonstrating overflow behavior tests: overflow with SILENT
+add(120::i8, 10::i8) [overflow:SILENT] = undefined
+
+# floating_exception: Examples demonstrating exceptional floating point cases
+add(1.5e+308::fp64, 1.5e+308::fp64) = inf::fp64
+add(-1.5e+308::fp64, -1.5e+308::fp64) = -inf::fp64
+
+# rounding: Examples demonstrating floating point rounding behavior
+add(4.5::fp32, 2.500001::fp32) [rounding:TIE_TO_EVEN] = 7.000001::fp32
```
westonpace, line 28: This doesn't really demonstrate rounding error.  If we're sticking with the 6/15 digits of precision thing then it is impossible to demonstrate rounding error.  Should we just remove these tests?

tests/cases/arithmetic_decimal/power.test
westonpace: So I think we are ok if I read this correctly?

If no options are specified then the test should not depend on the options (should give the same answer regardless of what options are present)

If a test case's answer depends on one or more options then those options **must** be listed as part of the test.

tests/README.md
```diff
@@ -0,0 +1,138 @@
...
+### Version Declaration
+The version declaration must be the first line of the file. It specifies the version of the test file format. The version declaration must be in the following format:
+```
+### SUBSTRAIT_SCALAR_TEST: V1
+```
+
+### Include Statements
+Include statements should have at least one include statement. The include statement specifies the path to substrait extension functions. The include statement must be in the following format:
+```
+### SUBSTRAIT_INCLUDE: /extensions/functions_aggregate_approx.yaml
```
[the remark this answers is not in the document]
scgkiran: 1. Functions in the same extension yaml file will differ in the signature (function name or argument types)
2. In case of a conflict with function from two different extensions is to avoid the conflict by having test cases for such functions in separate test files.

jacques-n:
> It's a different use case...

I believe the original request was some way to refer to test cases individually. My main thought is that the case itself should probably be the "pointer" (or a hash of the case) as opposed to some kind of hand-defined naming system. From a substrait core perspective, I think there should be no nuance to describe "I support `funcA` with `optionA` and `optionB` but sometimes I return different results than what is defined in the test cases for `funcA`". To me, that's not `funcA`--it's an entirely different function. That system should say "I support a `funcAprime` which might be similar to `funcA` but has a different set of test cases".  

My inclination is we should avoid trying to create some system that says "these portion of test cases for `funcA` also apply to `funcAprime`" Feels like an over-engineer.

EpsilonPrime:
The usage is more like "this system doesn't support this function" sufficiently.  A compatibility matrix lists all the systems that support a functionality and which ones that don't.  While we could say that only these tests pass, the set of passes and failures provide a more complete story of compatibility.

tests/cases/arithmetic_decimal/power.test
EpsilonPrime: There should be a clear way of referring to a test so that an external system can skip a particular function's test if necessary.  That may be just <group name> + <test number within that group>.

tests/README.md
```diff
@@ -0,0 +1,138 @@
...
+```
+doc         := <version>
+               (<include>)+
+               ((<test_group>)?(<test_case>)+\n)+
+version     := ### SUBSTRAIT_SCALAR_TEST: <test_library_version>
+include     := ### SUBSTRAIT_INCLUDE: <uri>
+test_group  := # <description>
+test_case   := <function>(<arguments>) ([<options>])? = <result> (#<description>)?
+description := string
+function    := string
```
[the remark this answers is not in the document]
EpsilonPrime: I suggest using the definition of string already implemented by the textplan Antlr grammar.

tests/cases/arithmetic_decimal/power.test
EpsilonPrime: The :: format is a type subclass operator.  As such it requires a mental context switch every time I look at it.  The underscore was chosen in the text format to create a visual separation between the literal value and its type thus aiding in readability (and it's also requires less keystrokes).

Braces are used for literal values and angle brackets are used for type composition and parameterization.  This helps to reinforce the distinction between value and type.  I have yet to be able to determine what the value of any interval literal is in the proposed format.  However the text format's use of labels makes it ultra-clear what each part of the interval is.

The text format's string implementation does draw from other sources than SQL but that's mostly because the SQL version has its own warts which make it harder to read.  SQL's implicit concatenation is the main SQL string feature that is difficult for humans to parse (it requires the counting of active single quotes).


The text format was reviewed in the community meeting twice (February 1st, 2023 and April 12th, 2023) so presumably there is some community agreement about its design validity.

westonpace (dismissed):
Thanks for tackling the changes.

> I support funcA with optionA and optionB but sometimes I return different results than what is defined in the test cases for funcA

@jacques-n 

There are some functions (maybe very few) where avoiding this will be challenging.  I'm not saying you're wrong, I really like the idea of "if engine supports function X then behavior is exactly defined" (and I don't think we need to figure it out in this PR).  It's more food for thought.  My primary example is the `divide` function:

| LHS | RHS | Postgres | DuckDb | Datafusion |
| --- | --- | --- | --- | --- |
| 0 | 0 | ERR | NAN | NAN |
| 0 | NAN | NULL | NAN | NAN |
| 0 | INF | 0 | 0 | 0 |
| NAN | 0 | NULL | NULL | NAN |
| NAN | NAN | NULL | NAN | NAN |
| NAN | INF | NULL | NAN | NAN |
| INF | 0 | ERR | NULL | INF |
| INF | NAN | NULL | NAN | NAN |
| INF | INF | NULL | NAN | NAN |

The division options we have today (`on_domain_error` and `on_division_by_zero`) are insufficient to describe any one of these three (fairly significant) engines.  There is no engine that matches any combination of these values.  In fact, we can't even introduce more values to the options because engines are inconsistent (e.g. postgres' divide by zero behavior changes if the numerator is NAN).

Some options:

 * Just state that there is no engine that provides the `divide` function and hope we can drive compliance.
 * Come up with lot's of options in the YAMLs (we almost need a `divide_by_zero_but_numerator_is_inf` option or at the extreme, one option per combination of inputs)
 * State that the engines do implement the `divide` function "except at the boundaries"
 * Add an `undefined` to the two options we have today and just accept that all engines will use this value

tests/cases/arithmetic_decimal/power.test
jacques-n: It's not that the text format defined in substrait-cpp isn't valid. It's definitely valid. 

I think we come from different perspectives. I'm coming from the perspective that we should try to leverage existing patterns as much as possible. My expectation is that the vast majority of systems that will be tested are SQL systems so leveraging SQL patterns is desirable as much as reasonably possible and comes with the added benefit of being obvious in meaning to a large group of people reviewing the test case. If we look at the set of BFT dialects, for example, we have the following:

| Dialect    | API       | Single Quotes for Literals | Supports :: for Cast |
|------------|-----------|----------------------------|----------------------|
| cudf       | cudf api  | no                         | no                   |
| datafusion | SQL       | yes                        | no                   |
| duckdb     | SQL       | yes                        | yes                  |
| postgres   | SQL       | yes                        | yes                  |
| snowflake  | SQL       | yes                        | yes                  |
| sqllite    | SQL       | yes                        | no                   |
| velox      | velox api | no                         | no                   |


So a decent portion are SQL based. Single-quoted literals are the most common way to represent strings in those systems. 3/7 have support for the double colon as meaning cast. Given that, I prefer that over something invented.

jacques-n:
> There are some functions (maybe very few) where avoiding this will be challenging. [snip]... example is the divide function...

These are just different functions from my pov. We shouldn't be fearful of clearing documenting different divide functions (through options and/or variants). In fact, I see this as one of the key values of Substrait for people--saying there are 3 distinct divide functions in each of these systems and how each differs. (FWIW, in this particular example I see both using variant and/or options as reasonable.)

My point in general here is that whether a particular engine implements a particular function is a property of the dialect and not the function. Functions should be well defined outside of a specific instance of use.

Now, many people may be fine to do lossy translation between these divide functions and we should make that easy with substrait tooling. But I don't think we should confound that with the functions themselves. I'm generally of the perspective that options and different functions should solve these problems without having to introduce additional primitives.

For example, we may need to introduce the following kind of psuedo-mappings in substrait:

```
mappings:
  from: postgres.yaml:div(float,float)
  to:  duckdb.yaml:div(float,float) 
  differences:
    - divide by NAN => NAN in duckdb, NULL in postgres
    - ...
    - ...
 ```

Then someone using the two systems could get a list of lossy translations and evaluate whether that list is compatible with their use case. (Maybe okay for general analytics, not for financial analytics).

In many cases, I would suspect that we could also define expression trees that provide lossless conversions in these cases. Odds are those scenarios might be more compute expensive.

File: tests/README.md [hunk 10]
```diff
@@ -0,0 +1,157 @@
...
+Use short names listed in https://substrait.io/extensions/#function-signature-compound-names
+
+- **bool**: Boolean
+- **i8**: 8-bit signed integer
+- **i16**: 16-bit signed integer
+- **i32**: 32-bit signed integer
+- **i64**: 64-bit signed integer
+- **f32**: 32-bit floating point number
+- **f64**: 64-bit floating point number
+- **dec**: Fixed-point `decimal<P,S>`
```
EpsilonPrime, line 125: Do short forms still need to be parameterized?  Can we refer to the list elsewhere?  Can long names be interchangeably used?

tests/README.md
```diff
@@ -0,0 +1,157 @@
...
+
+```code
+### SUBSTRAIT_SCALAR_TEST:V1
+### SUBSTRAIT_INCLUDE: /extensions/functions_arithmetic.yaml
+
+# Common Maths
+add(126::i8, 1::i8) = 127::i8
+
+# Arithmetic Overflow Tests
+add(127::i8, 1::i8) [overflow:ERROR] = ERROR  #check overflow
```
EpsilonPrime, line 155: Can comments sometimes be recognized as group names?  Do we care?

tests/cases/arithmetic/add.test
```diff
@@ -0,0 +1,31 @@
+### SUBSTRAIT_SCALAR_TEST: 1.0
+### SUBSTRAIT_INCLUDE: /extensions/functions_arithmetic.yaml
+
+# basic: Basic examples without any special cases
```
EpsilonPrime, line 4: Is the name the whole line or only up to the semicolon?

tests/cases/arithmetic_decimal/power.test
```diff
@@ -0,0 +1,21 @@
...
+power(1.0::decimal, -1.0::decimal<38, 0>) = 1.0::fp64
+power(2.0::decimal<38, 0>, -2.0::decimal<38, 0>) = 0.25::fp64
+power(13::decimal<38, 0>, 10::decimal<38, 0>) = 137858491849::fp64
+
+# result_more_than_input_precison: Examples demonstrating result with more precision than input
+power(16::decimal<2, 0>, 4::decimal<38, 0>) = 65536::fp64
+
+# floating_exception: Examples demonstrating exceptional floating point cases
+power(1.5e+10::decimal<38, 0>, 1.5e+20::decimal<38, 0>) = inf::fp64
```
EpsilonPrime, line 14: Should this be the short form names as noted above or are both legal?

tests/cases/datetime/lt_datetime.test
```diff
@@ -0,0 +1,25 @@
...
+lt(1999-01-08 01:05:05-08:00::tstz, 1999-01-08 04:05:06-05:00::tstz) = true::bool
+lt(1999-01-08 01:05:06-08:00::tstz, 1999-01-08 04:05:06-05:00::tstz) = false::bool
+
+# date: examples using the date type
+lt(2020-12-30::date, 2020-12-31::date) = true::bool
+lt(2020-12-31::date, 2020-12-30::date) = false::bool
+
+# interval: examples using the interval type
+lt('P7D'::iday, 'P6D'::iday) = false::bool
+lt('P5D'::iday, '6D'::iday) = true::bool
```
EpsilonPrime, line 18: Is the P here optional?

tests/README.md
```diff
@@ -0,0 +1,157 @@
...
+- Valid values: true, false
+
+#### Date and Time
+All date and time literals use ISO 8601 format:
+
+- **date**: `YYYY-MM-DD`, example: `2021-01-01`
+- **time**: `HH:MM:SS[.fraction]`, example: `12:00:00.000`
+- **timestamp**: `YYYY-MM-DD HH:MM:SS[.fraction]`, example: `2021-01-01 12:00:00`
+- **timestamp_tz**: `YYYY-MM-DD HH:MM:SS[.fraction]±HH:MM`, example: `2021-01-01 12:00:00+05:30`
+- **interval year**: `'P[n]Y[n]M'`, example: `'P2Y3M'` (2 years, 3 months)
```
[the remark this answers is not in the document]
scgkiran: Period

scgkiran:
Closing this PR, as changes are merged using separate PRs for test file format grammar and testcases

PR #682, opened 2024-08-14
feat: define mark join

EpsilonPrime:
Defines the Mark Join used by a number of engines in query rewrites.


site/docs/relations/logical_relations.md
```diff
@@ -234,6 +234,7 @@ The join operation will combine two separate inputs into a single output, based
 | Right Anti  | Return records from the right input. These are returned only if the records do not have a join partner on the left side. |
 | Left Single | Return all records from the left input with no join expansion. If at least one record from the right input matches the left, return one arbitrary matching record from the right input. For any left records without matching right records, return the left record along with nulls for the right input. Similar to a left outer join but only returns one right match at most. Useful for nested sub-queries where we need exactly one record in output (or throw exception).  See Section 3.2 of https://15721.courses.cs.cmu.edu/spring2018/papers/16-optimizer2/hyperjoins-btw2017.pdf for more information. |
 | Right Single | Same as left single except that the right and left inputs are switched. |
+| Mark | Returns at most one join partner for each of the primary input records. Uses a hash to first determine if there are matches in the secondary input. |
```
westonpace, line 237:
```suggestion
| Left Mark | Returns one record for each record from the left input.  Adds one "mark" column to the output.  If there is at least one join partner in the right input where the join condition evaluates to true then the mark column will be set to true.  Otherwise, if there is at least one join partner in the left input where the join condition evaluates to NULL then the mark column will be set to NULL.  Otherwise the mark column will be set to false. |
| Right Mark | Returns records from the right input.  Adds one "mark" column to the output.  If there is at least one join partner in the left input where the join condition evaluates to true then the mark column will be set to true.  Otherwise, if there is at least one join partner in the left input where the join condition evaluates to NULL then the mark column will be set to NULL.  Otherwise the mark column will be set to false. |
```

Updated to my understanding of mark joins.

> Returns at most one join partner for each of the primary input records

I believe a mark join returns _exactly_ one join partner for each of the primary input records.

Also, let's please be consistent and use left/right instead of primary/secondary.

> Uses a hash to first determine if there are matches in the secondary input

This is a logical relation.  Consumers are free to implement in any manner they would like.

westonpace (changes requested):
I don't find the explanation sufficient and have proposed some alternate wording.

westonpace (approved):

site/docs/relations/logical_relations.md
```diff
@@ -234,6 +234,8 @@ The join operation will combine two separate inputs into a single output, based
...
 | Left Single | Return all records from the left input with no join expansion. If at least one record from the right input matches the left, return one arbitrary matching record from the right input. For any left records without matching right records, return the left record along with nulls for the right input. Similar to a left outer join but only returns one right match at most. Useful for nested sub-queries where we need exactly one record in output (or throw exception).  See Section 3.2 of https://15721.courses.cs.cmu.edu/spring2018/papers/16-optimizer2/hyperjoins-btw2017.pdf for more information. |
 | Right Single | Same as left single except that the right and left inputs are switched. |
+| Left Mark | Returns one record for each record from the left input.  Adds one "mark" column to the output.  If there is at least one join partner in the right input where the join condition evaluates to true then the mark column will be set to true.  Otherwise, if there is at least one join partner in the right input where the join condition evaluates to NULL then the mark column will be set to NULL.  Otherwise the mark column will be set to false. |
```
jacques-n, line 237:
```suggestion
| Left Mark | Returns one record for each record from the left input.  Appends one additional "mark" column to the output of the join. The new column will be listed after all columns from both sides and will be of type nullable boolean.  If there is at least one join partner in the right input where the join condition evaluates to true then the mark column will be set to true.  Otherwise, if there is at least one join partner in the right input where the join condition evaluates to NULL then the mark column will be set to NULL.  Otherwise the mark column will be set to false. |
```

site/docs/relations/logical_relations.md
```diff
@@ -234,6 +234,8 @@ The join operation will combine two separate inputs into a single output, based
...
+| Right Mark | Returns records from the right input.  Adds one "mark" column to the output.  If there is at least one join partner in the left input where the join condition evaluates to true then the mark column will be set to true.  Otherwise, if there is at least one join partner in the left input where the join condition evaluates to NULL then the mark column will be set to NULL.  Otherwise the mark column will be set to false. |
```
jacques-n, line 238:
```suggestion
| Right Mark | Returns records from the right input.  Appends one additional "mark" column to the output of the join. The new column will be listed after all columns from both sides and will be of type nullable boolean.  If there is at least one join partner in the left input where the join condition evaluates to true then the mark column will be set to true.  Otherwise, if there is at least one join partner in the left input where the join condition evaluates to NULL then the mark column will be set to NULL.  Otherwise the mark column will be set to false. |
```

jacques-n (commented):
A minor request to be super clear on output column types and ordering.

It doesn't necessarily need to be resolved here but we should really update the text of join operations to better clarify their output column ordering and types. The current text reads:

> Direct Output Order: The emit order of the left input followed by the emit order of the right input.

That is no longer true with many of these join types.

jacques-n (approved):

PR #688, opened 2024-08-21
docs: naming rules for Named Structs

vbarua:


site/docs/types/named_structs.md
```diff
@@ -0,0 +1,79 @@
...
+
+For example, the following Struct
+```
+        a     b
+struct<i64, fp64>
+
+```
+has 2 names, one for each of its inner fields.
+
+### Nested Structs
```
EpsilonPrime, line 25: Basically this applies to any struct in any complex type (maps, lists, and structs).  Not sure if introducing that term makes the document simpler.

site/docs/types/named_structs.md
```diff
@@ -0,0 +1,79 @@
...
+* 2 names [b, c] for the struct fields used as a key
+* 3 names [d, e, f] for the struct fields used as a value
+
+#### Structs within List
+If a List contains Structs, the Struct fields must be named. For example the following List
+```
+              b     c
+list<struct<i64, fp64>>
+```
+has 2 named fields [a, b] for the struct fields.
```
EpsilonPrime, line 44: doesn't match the above

site/docs/types/named_structs.md
```diff
@@ -0,0 +1,79 @@
...
+
+#### Named Struct w/ Nested Structs
+```
+NamedStruct {
+    names: [a, b, c, d, e, f, g, h]
+    //               a     b          c    d      e                f     g       h
+    struct: struct<i64, list<struct<i64, i64>>, map<i64, struct<fp64, fp64>>, fp64>
+}
+```
+
```
EpsilonPrime, line 79: Do you need an example with a struct multiple levels deep?

site/docs/types/named_structs.md
vbarua: Switched to Compound Types which is the language we use in the docs: https://substrait.io/types/type_classes/#compound-types

site/docs/types/named_structs.md
vbarua: Probably reasonable. Added an example with structs in structs.

site/docs/types/named_structs.md
```diff
@@ -0,0 +1,88 @@
...
+A Named Struct is a special type construct that combines:
+* A Struct type
+* A list of names for the fields in the Struct, in depth-first search order
+
+The depth-first search order for names arises from the the ability to nest Structs within other types. All struct fields must be named, even nested fields.
+
+Named Structs are most commonly used to model the schema of Read relations.
+
+## Determining Names
+When producing/consuming names for a NamedStruct, some types requires special handling:
```
acruise, line 12: nit: s/requires/require ;)

site/docs/types/named_structs.md
```diff
@@ -0,0 +1,88 @@
...
+
+### Struct
+A struct has names for each of its inner fields.
+
+For example, the following Struct
+```
+        a     b
+struct<i64, fp64>
+
```
acruise, line 21: Is this the common syntax here to put the names above the types?  Or are `a` and `b` supposed to be metasyntactic variables that aren't referenced here?

acruise: or, looking at the later examples, is the user responsible for flattening the tree of names?

site/docs/types/named_structs.md
```diff
@@ -0,0 +1,88 @@
...
+```
+has 2 names, one for each of its inner fields.
+
+### Structs within Compound Types
+Struct types nested in compound types must also be be named.
+
+#### Structs within Maps
+If a Map contains Structs, the Struct fields must be named. For example the following Map
```
Blizzara, line 29:
```suggestion
If a Map contains Structs, either as keys or values or both, the Struct fields must be named. Keys are named before values. For example the following Map
```

site/docs/types/named_structs.md
Blizzara: +1 for the `struct<a: i64, b: fp64>` form, maybe with bolded names

vbarua: The [NamedStruct](https://github.com/substrait-io/substrait/blob/8ae1084761c0c384c9c54bd4d7af62c4a58ea1cc/proto/substrait/type.proto#L258-L262) messages consists of a [Type.Struct](https://github.com/substrait-io/substrait/blob/8ae1084761c0c384c9c54bd4d7af62c4a58ea1cc/proto/substrait/type.proto#L195-L199) message, which only contains type information, no names, and a `names` field which is the depth-first encoded list of names associated with the struct.

> Is it this project's style to put the names above the types?

The Struct message itself does not contain any name information. The format I'm using the type names as defined in https://substrait.io/types/type_classes/#compound-types to represent the types.

I'm wary of `struct<a: i64, b: fp64>` because it makes it look like the Structs may have names. There actually is a `NSTRUCT` pseudo-construct for this, but it's not really used anywhere (that I know of).

vbarua: And changed my mind after more feedback. Added a note to the top of the page and inlined the names.

site/docs/types/named_structs.md
```diff
@@ -0,0 +1,83 @@
...
+A Named Struct is a special type construct that combines:
+* A Struct type
+* A list of names for the fields in the Struct, in depth-first search order
+
+The depth-first search order for names arises from the the ability to nest Structs within other types. All struct fields must be named, even nested fields.
+
+Named Structs are most commonly used to model the schema of Read relations.
+
+> Note: In this page Structs are annotated with names for readabilty.
+> In practice Structs to not contain name information.
```
Blizzara, line 12: super nit:
```suggestion
> In practice Structs do not contain name information.
```

site/docs/types/named_structs.md
```diff
@@ -0,0 +1,83 @@
...
+* 3 names [e, f, g] for the fields within the 2nd nested struct
+
+### Putting It All Together
+
+#### Simple Named Struct
+```
+NamedStruct {
+    names: [a, b, c, d]
+    struct: struct<a: i64, b: list<i64>, c: map<i64, i64>, d: i64>
+}
```
Blizzara, line 65: You have a fair point about the <a: i64> being potentially confusing since that's not how substrait actually works. How about making that explicit in these examples, something like:

```suggestion
The struct `struct<a: i64, b: list<i64>, c: map<i64, i64>, d: i64>` would be represented as a NamedStruct as follows:
```
NamedStruct {
    names: [a, b, c, d]
    struct: struct<i64, list<i64>, map<i64, i64>, i64>
}
```

File: site/docs/types/named_structs.md (hunk 7 above)
Comment (vbarua) [reply site/docs/types/named_structs.md]: Not a nit, cannot unsee 🙈

File: site/docs/types/named_structs.md (hunk 8 above)
Comment (vbarua) [reply site/docs/types/named_structs.md:65]: I think/hope the note about names in Structs at the top of the page should be sufficient. Users attempting to build these in code should notice that Structs don't have names. 

I'm inclined to leave this as is for now given that it's more readable, and change it in the future if we get feedback that it is confusing.

Comment (jacques-n) [reply site/docs/types/named_structs.md:65]: I think it is confusing. How about something like this:

```
names:  [a,b,c,d]
struct: 
        struct<i64, list<i64>, map<i64, i64>, i64>
               ↑    ↑          ↑              ↑
               a    b          c              d
```

Note that I chose to put the value of struct on a separate line to avoid the key "struct" looking like it is part of the format. I think it is also really important to use the first character of the type declarations as the place to annotate so no confusion exists about inner field behaviors.

site/docs/types/named_structs.md
acruise: To be clear, I wasn't arguing for a change in the internal representation... It's just that I found this particular part of Substrait quite confusing, so other passersby are likely to as well. :)

acruise: I'm certain that it would be duplicative, but it would probably help to have a paragraph or two--or at least a _very_ specific hyperlink--about Substrait's approach here. With that context, the separate listings of types and names makes more sense...

I might even recommend going so far as to add vertical tree visualizations of the nested structures, to make the depth-first part even clearer... but pretty soon you're talking about real effort ;)

site/docs/types/named_structs.md
vbarua: The arrows due make it look more intentional 🙂 

Updated

EpsilonPrime (approved):

jacques-n (approved):
nicely clarified. Thanks @vbarua !

PR #690, opened 2024-08-23
feat: add JSON file support

EpsilonPrime:


proto/substrait/algebra.proto
```diff
@@ -162,6 +162,14 @@ message ReadRel {
         // made up of non-nullable strings.
         optional string value_treated_as_null = 6;
       }
+      message JsonReadOptions {
+        // JSON encoded files are considered to be in UTF-8.  If any errors are
```
jacques-n, line 166: It would be good to declare some additional expectations here. 

For example, two common file formats for json are:
- Files that are one json object per line with a line break between each object. (Each line is a valid json object but the file isn't)
- Files that are a giant array where each item in the array is a record.

Is this for both? If only one, should we name this more specifically. (I'm guessing it's for the first type given the max_line_size parameter.)

There are several other common patterns. Some are about strictness to the json spec. A lot of json is not exactly strict. (E.g. using single quotes instead of double, not quoting field names, etc.) I don't know if we need to cover them here but elsewhere I've had to state more lenient behavior.

PR #691, opened 2024-08-23
feat: split unbounded bound into lower and upper

EpsilonPrime:
BREAKING CHANGE: splits the WindowFunction Unbounded bound into preceding and following types.

The intention of the unbounded bound type was to work two ways depending on whether the bound type was preceding or following.  However since the preceding and following types are part of the same one of structure this was not possible.


EpsilonPrime:
After checking the substrait-java code for how they handle unbounded_preceding and unbounded_following, it's clear that the existing definition works just fine.  I will clean up the corresponding change here from the Gluten code.

PR #692, opened 2024-08-23
docs: remove stray words in description of Aggregate Operation

ingomueller-net:
One sentence was broken, probably due to left-over words from a previous formulation.


mbrobbel (approved):

vbarua (approved):
Thanks for the cleanup!

PR #697, opened 2024-08-28
feat: add 'first' function

andrew-coleman:
Add `first` to the set of generic aggregate functions.

This is required to support queries containing `distinct` aggregations that get rewritten by the Spark catalyst query optimiser to contain `Expand` relations and `first` function calls.

This PR, together with https://github.com/substrait-io/substrait/pull/696 will allow the implementation of the `Expand` relation in `substrait-java`, improving the test pass rate of the TPC-DS suite in the `spark` module.

### DIFF
--- a/extensions/functions_aggregate_generic.yaml
+++ b/extensions/functions_aggregate_generic.yaml
@@ -40,3 +40,13 @@ aggregate_functions:
         decomposable: MANY
         intermediate: any1?
         return: any1?
+  - name: "first"
+    description: First of a set of values.
+    impls:
+      - args:
+          - name: x
+            value: any
+        nullability: DECLARED_OUTPUT
+        decomposable: MANY
+        intermediate: any
+        return: any


extensions/functions_aggregate_generic.yaml
```diff
@@ -40,3 +40,13 @@ aggregate_functions:
         decomposable: MANY
         intermediate: any1?
         return: any1?
+  - name: "first"
+    description: First of a set of values.
```
EpsilonPrime, line 44: Probably returns null if given an empty set.

EpsilonPrime, line 44: Additional caveat:  Assumes the input is ordered.  If the input is not ordered then the result is undefined.

EpsilonPrime (commented):
Thanks for the submission.  The most similar function is any_value which does not have ordering requirements.  This still represents a different enough concept to me for inclusion purposes.

Blizzara:
FWIW, we had some [discussion](https://github.com/substrait-io/substrait/pull/652#issuecomment-2195527513) around this earlier, and then I concluded to just map first into any_value since in Spark they're the same anyways. That said I'm def fine with having it 🤷

Issue #700, opened 2024-09-05
Precise definition of equality of grouping columns in the aggregate relation


ingomueller-net:
This issue summarizes [two](https://substrait.slack.com/archives/C02D7CTQXHD/p1725290660378019) [threads](https://substrait.slack.com/archives/C02D7CTQXHD/p1725467943844479) in the Slack channel, which will be behind a paywall at some point.

The aggregate relation uses a definition of expression equality that affects its semantics, including its output schema. The current definition is ambiguous and based on protobuf message equality. The first is a problem because it makes the precise semantics of the relation ambiguous; the second is problematic because the protobuf definition should be derived from the spec and not the other way around.

This is the relevant part of the [spec](https://substrait.io/relations/logical_relations/#aggregate-operation) (emphasis mine):

> It’s possible to specify multiple grouping sets in a single aggregate operation. The grouping sets behave more or less independently, with each returned record belonging to one of the grouping sets. The values for the grouping expression columns that are not part of the grouping set for a particular record will be set to null. Two grouping expressions will be returned using the same column if **the protobuf messages describing the expressions are equal**.

To elaborate the problem, let's consider an example. Let's say we have the following three grouping sets, where `a` and `b` are two arbitrary but different expressions:

```
(a, b)
(a)
(b)
```

Then the set of grouping expressions will be `(a, b)` and the output schema will have a column for each each of `a` and `b`.

The challenge arises during the deserialization of the protobuf messages. At that point, we do not have `a` and `b` yet, but four different messages:

```
(m1, m2)
(m3)
(m4)
```

In order to get to the representation above, we need to establish that `m1 == m3`, `m2 == m4`, and !(`m1 == m2)`. For different definitions of `==`, we may either come to the interpretation above, i.e., that the grouping set expressions is `(m1, m2)`, or any other subset of `(m1, m2, m3, m4)`, most of which have different semantics and, thus, break the common understanding between systems that don't agree on `==`.

The current formulation of the spec allows for at least the following interpretations:

* The protobuf messages are considered equal if their underlying buffers are byte-wise equal.
  * This doesn't make much sense since the fields of a protobuf can be serialized in any order.
* They are considered equal if their field values are recursively equal. This is what [`MessageDifferencer`](https://github.com/protocolbuffers/protobuf/blob/main/src/google/protobuf/util/message_differencer.h#L58) implements.
   There are still multiple questions with this approach:
   * Are the advanced_extensions fields of relations relevant for equality?
[...]

PR #706, opened 2024-09-13
feat: change grouping expressions in AggregateRel to references

ingomueller-net:
BREAKING CHANGE: This PR changes the definition of grouping sets in `AggregateRel` to consist of references into a list of grouping expressions instead of consisting of expressions directly.

With the previous definition, consumers had to deduplicate the expressions in the grouping sets in order to execute the query or even derive the output schema (which is problematic, as explained below). With this change, the responsibility of deduplicating expressions is now on the producer. Concretely, consumers are now expected to be simpler: The list of grouping expressions immediately provides the information needed to derive the output schema and the list of grouping sets explicitly and unambiguously provides the equality of grouping expressions. Producers now have to specify the grouping sets explicitly. If their internal representation of grouping sets consists of full grouping expressions (rather than references), then they must deduplicate these expressions according to their internal notion of expression equality in order to produce grouping sets consisting of references to these deduplicated expressions.

If the previous format is desired, it can be obtained from the new format by (1) deduplicating the grouping expressions (according to the previously applicable definition of expression equality), (2) re-establishing the duplicates using the emit clause, and (3) "dereferencing" the references in the grouping sets, i.e., by replacing each reference in the grouping sets with the expression it refers to.

The previous version was problematic because it required the *consumers* to deduplicate the expressions from the grouping sets. This, in turn, requires to parse and understand 100% of these expression even in cases where that understanding is otherwise optional, which is in opposition to the general philosophy of allowing for simple-minded consumers. The new version avoids that problem and, thus, allows consumers to be simpler. The issues are discussed in even more detail in #700.


ingomueller-net:
This is still work in progress. The text part of the specification still needs to be adapted. But I guess the protobuf definition allows to agree on what our desired end goal is?

proto/substrait/algebra.proto
```diff
@@ -244,18 +244,30 @@ message AggregateRel {
...
   substrait.extensions.AdvancedExtension advanced_extension = 10;
 
   message Grouping {
-    repeated Expression grouping_expressions = 1;
+    // Deprecated in favor of `expression_reference` below.
+    repeated Expression grouping_expressions = 1 [deprecated = true];
+
+    // A list of zero or more references to grouping expressions, i.e., indices
+    // into the `grouping_expression` list.
+    repeated uint32 expression_reference = 2;
```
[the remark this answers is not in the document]
ingomueller-net: Good point! Fixed in new commit.

jacques-n:
I'm supportive of this change although it makes my stomach a little sick. Thoughts @westonpace @EpsilonPrime @vbarua ?

EpsilonPrime (approved):
We probably need a description of why and how to use the grouping expressions on the website but the approach is soundly backwards compatible --fix newer consumers will do the right thing for older plans.  Producers may duplicate the old grouping expressions with the references to work with older consumers too.

jacques-n:
I'm +1 on the change but I think the release note/commit message needs to better clarify what needs to separate what was done versus the breaking change part. In the breaking change part we should be describing what specifically needs to be done for migration and what is actually breaking.

ingomueller-net:
Thanks, all, for being supporting and providing feedback! I have added a new commit that propagates the changes to the textual specification and changed/extended the commit message. Please let me know if those changes reflect what you had in mind.

site/docs/relations/logical_relations.md
```diff
@@ -345,13 +345,13 @@ The aggregate operation groups input data on one or more sets of grouping keys,
 | Inputs               | 1                                                            |
 | Outputs              | 1                                                            |
 | Property Maintenance | Maintains distribution if all distribution fields are contained in every grouping set. No orderedness guaranteed. |
-| Direct Output Order  | The list of distinct columns from each grouping set (ordered by their first appearance) followed by the list of measures in declaration order, followed by an `i32` describing the associated particular grouping set the value is derived from (if applicable). |
+| Direct Output Order  | The list of grouping expressions in declaration order followed by the list of measures in declaration order, followed by an `i32` describing the associated particular grouping set the value is derived from (if applicable). |
```
vbarua, line 348: An added benefit of this is that it makes it much easier to compute the output order. Declaration order is very clear ✨

vbarua (approved):
Thanks for working on this. While this does add a little complexity, I think it removes a lot of ambiguity and avoids the need to specify expression equality which IMO is much harder.

site/docs/relations/logical_relations.md
```diff
@@ -345,13 +345,13 @@ The aggregate operation groups input data on one or more sets of grouping keys,
...
+Grouping sets can be used for finer-grained control over which records are folded. A grouping set consists of zero or more references to the list of grouping expressions. Within a grouping set, two records will be folded together if and only if they have the same values for each of the expressions in the grouping set. The values returned by the grouping expressions will be returned as columns to the left of the columns for the aggregate expressions. Each of the grouping expressions must occur in at least one of the grouping sets. If a grouping set contains no grouping expressions, all rows will be folded for that grouping set. (Having a single grouping set with no grouping expressions is thus equivalent to not haveing any grouping sets.)
```
westonpace, line 352:
```suggestion
Grouping sets can be used for finer-grained control over which records are folded. A grouping set consists of zero or more references to the list of grouping expressions. Within a grouping set, two records will be folded together if and only if they have the same values for each of the expressions in the grouping set. The values returned by the grouping expressions will be returned as columns to the left of the columns for the aggregate expressions. Each of the grouping expressions must occur in at least one of the grouping sets. If a grouping set contains no grouping expressions, all rows will be folded for that grouping set. (Having a single grouping set with no grouping expressions is thus equivalent to not having any grouping sets.)
```

westonpace (approved):
I think this might actually be easier to understand than the original formulation (though maybe it's just because I know more now)

ingomueller-net:
(I accepted @westonpace's suggestion to fix a typo -- now I need a new approval to be able to merge 😉)

jacques-n (approved):

jacques-n:
Thanks for your patience here @ingomueller-net , really appreciate the contribution and updates.

ingomueller-net:
Awesome, great to see this being wrapped up. Thank you all for the good discussions!

PR #711, opened 2024-09-26
feat: enhance VirtualTable to have expression as value

anshuldata:
* This enables expressing Virtual table having values as expression
* Example query: `select * from (values (1+2, 'Hello'||'World'))`
* Mark existing literal field "values" as deprecated

### DIFF
--- a/proto/substrait/algebra.proto
+++ b/proto/substrait/algebra.proto
@@ -73,11 +73,10 @@ message ReadRel {
 
   // Definition of which type of scan operation is to be performed
   oneof read_type {
-    VirtualTable virtual_table = 5 [deprecated = true];
+    VirtualTable virtual_table = 5;
     LocalFiles local_files = 6;
     NamedTable named_table = 7;
     ExtensionTable extension_table = 8;
-    VirtualExpressionTable virtual_expression_table = 9;
   }
 
   // A base table. The list of string is used to represent namespacing (e.g., mydb.mytable).
@@ -89,12 +88,8 @@ message ReadRel {
 
   // A table composed of literals.
   message VirtualTable {
-    repeated Expression.Literal.Struct values = 1;
-  }
-
-  // A table composed of expressions.
-  message VirtualExpressionTable {
-    repeated Expression values = 1;
+    repeated Expression.Literal.Struct values = 1 [deprecated = true];
+    repeated Expression expression_values = 2;
   }
 
   // A stub type that can be used to extend/introduce new table types outside
--- a/proto/substrait/algebra.proto
+++ b/proto/substrait/algebra.proto
@@ -86,10 +86,10 @@ message ReadRel {
     substrait.extensions.AdvancedExtension advanced_extension = 10;
   }
 
-  // A table composed of literals.
+  // A table composed of expressions.
   message VirtualTable {
     repeated Expression.Literal.Struct values = 1 [deprecated = true];
-    repeated Expression expression_values = 2;
+    repeated Expression expressions = 2;
   }
 
   // A stub type that can be used to extend/introduce new table types outside
--- a/proto/substrait/algebra.proto
+++ b/proto/substrait/algebra.proto
@@ -73,10 +73,11 @@ message ReadRel {
 
   // Definition of which type of scan operation is to be performed
   oneof read_type {
-    VirtualTable virtual_table = 5;
+    VirtualTable virtual_table = 5 [deprecated = true];
     LocalFiles local_files = 6;
     NamedTable named_table = 7;
     ExtensionTable extension_table = 8;
+    VirtualExpressionTable virtual_expression_table = 9;
   }
 
   // A base table. The list of string is used to represent namespacing (e.g., mydb.mytable).
@@ -89,6 +90,11 @@ message ReadRel {
   // A table composed of literals.
   message VirtualTable {
     repeated Expression.Literal.Struct values = 1;
+  }
+
+  // A table composed of expressions.
+  message VirtualExpressionTable {
+    repeated Expression values = 1;
   }
 
   // A stub type that can be used to extend/introduce new table types outside
--- a/site/docs/relations/logical_relations.md
+++ b/site/docs/relations/logical_relations.md
@@ -51,16 +51,7 @@ Read definition types (like the rest of the features in Substrait) are built by 
 #### Virtual Table
 
 A virtual table is a table whose contents are embedded in the plan itself.  The table data
-is encoded as records consisting of literal values.
-
-| Property | Description | Required |
-| -------- | ----------- | -------- |
-| Data     | Required    | Required |
-
-#### Virtual Expression Table
-
-A virtual expression table is a table whose contents are embedded in the plan itself.
-Each column is an expression that can be resolved without referencing any input data.
+is encoded as records consisting of literal values or an expression that can be resolved without referencing any input data.
 For example, a literal, a function call involving literals, or any other expression that does
 not require input.
 
--- a/site/docs/relations/logical_relations.md
+++ b/site/docs/relations/logical_relations.md
@@ -51,7 +51,7 @@ Read definition types (like the rest of the features in Substrait) are built by 
 #### Virtual Table
 
 A virtual table is a table whose contents are embedded in the plan itself.  The table data
-is encoded as records consisting of literal values or an expression that can be resolved without referencing any input data.
+is encoded as records consisting of literal values or expressions that can be resolved without referencing any input data.
 For example, a literal, a function call involving literals, or any other expression that does
 not require input.
 
--- a/site/docs/relations/logical_relations.md
+++ b/site/docs/relations/logical_relations.md
@@ -59,8 +59,10 @@ is encoded as records consisting of literal values.
 
 #### Virtual Expression Table
 
-A virtual expression table is a table whose contents are embedded in the plan itself.  The table data
-is encoded as records consisting of expression values.
+A virtual expression table is a table whose contents are embedded in the plan itself.
+Each column is an expression that can be resolved without referencing any input data.
+For example, a literal, a function call involving literals, or any other expression that does
+not require input.
 
 | Property | Description | Required |
 | -------- | ----------- | -------- |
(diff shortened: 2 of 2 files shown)


jacques-n:
Patch makes sense. We should update the description message in this PR to also explain the deprecation.

proto/substrait/algebra.proto
```diff
@@ -91,6 +92,11 @@ message ReadRel {
     repeated Expression.Literal.Struct values = 1;
```
acruise, line 92: Rather than deprecate `VirtualTable` and create `VirtualExpressionTable`, would we consider deprecating the `values` _field_ and adding a new `expressions` field?

site/docs/relations/logical_relations.md
```diff
@@ -57,6 +57,16 @@ is encoded as records consisting of literal values.
...
 | Data     | Required    | Required |
 
+#### Virtual Expression Table
+
+A virtual expression table is a table whose contents are embedded in the plan itself.  The table data
+is encoded as records consisting of expression values.
+
+| Property | Description | Required |
+| -------- | ----------- | -------- |
+| Data     | Required    | Required |
```
westonpace, line 67: I think an important property of a virtual (expression) table is that it doesn't contain any field references.  This may not be obvious to new readers to Substrait.  In other words, they may look at a virtual table and think "this is weird, it looks like a projection?  Maybe this is some kind of select relation?"

Maybe something like...

```
A virtual expression table is a table whose contents are embedded in the plan itself.  Each column is an expression that can be resolved without referencing any input data.  These expressions will not contain field references but may contain function calls, literals, and other expression nodes.
```

westonpace (commented):
I don't have a strong opinion on "VirtualTable with expressions field" vs. "VirtualExpressionTable as new relation" but if I had to pick I would go with @acruise 's suggestion (just add a new field, not a new relation).

site/docs/relations/logical_relations.md
vbarua: Adding to this, I assume that these expression shouldn't contain other relations (i.e. subqueries)?

proto/substrait/algebra.proto
vbarua: Expanding on this, I'm almost inclined to have both? Like
```
 message VirtualTable {
    repeated Expression.Literal.Struct values = 1;
    repeated Expression expression = 2;
}
```
which emits first the `values` and then `expressions`. Makes the common case really easy to produce (literals only).

jacques-n: I'm not a fan of this. It creates this confusing situation where you have something like 
```
literal1, literal2, expression1, literal3
```
and it gets stored as 

```
values: [literal1, literal2]
expression: [expresion1, literal3]


> Rather than deprecate VirtualTable and create VirtualExpressionTable, would we consider deprecating the values field and adding a new expressions field?

I don't see how that is better (or clearer for someone new to the format).

File: site/docs/relations/logical_relations.md (hunk 2 above)
Comment (jacques-n) [reply site/docs/relations/logical_relations.md:67]: I don't think we should have prohibitions against either (field references and subqueries).

That being said, a valid field reference in this context is impossible because there are zero fields to reference. So having field[47] here is just as illegal as having a readrel that has two columns and trying to do field[47].

Comment (jacques-n) [reply site/docs/relations/logical_relations.md:67]: As a fun aside, in Snowflake the following is legal.
```
select * from (values ((select 1 where (select true)), 'Hello'||'World'))
```

**Comment by jacques-n** [comment]
>I'm not a fan of Protobuf quirks becoming load-bearing, but [this](https://stackoverflow.com/a/60079101) seems to indicate a field can be replaced with a oneof while maintaining binary compatibility. 😅

I believe that only works for optional/required fields. I believe repeated are disallowed.

File: site/docs/relations/logical_relations.md (hunk 2 above)
Comment (EpsilonPrime) [reply site/docs/relations/logical_relations.md:67]: The problem with nested queries having a field reference is that the resulting structure is no longer scalar.  For systems without virtual table support I have had to write tables to disk and then register named tables.  For those systems I would have to reject field references.

File: proto/substrait/algebra.proto (hunk 1 above)
Comment (EpsilonPrime) [reply proto/substrait/algebra.proto:92]: It looks like we have a protobuf check set up so if we can't use the oneof it will let us know.  I do know that it is less permissive than I'm used to:

https://github.com/substrait-io/substrait/blob/main/.github/workflows/pr.yml#L35

Comment (anshuldata) [reply proto/substrait/algebra.proto:92]: protobuf oneof can't have repeated fields [link](https://protobuf.dev/programming-guides/proto3/)
`A oneof cannot be repeated.`

Comment (anshuldata) [reply proto/substrait/algebra.proto:92]: We can't add oneof as mentioned above so both fields can contain values 
I too share concern with @jacques-n having both fields where initial literals are in Expression_Literal and rest in Expression (assuming there is one expression) doesn't seem intuitive .

File: site/docs/relations/logical_relations.md (hunk 2 above)
Comment (westonpace) [reply site/docs/relations/logical_relations.md:67]: > I don't think we should have prohibitions against either (field references and subqueries).

That's fine.  I don't know that we need to go so far as prohibit things.  My goal is just to make it as obvious as possible what the message is used for.  I'd be find with wording like...

```
A virtual expression table is a table whose contents are embedded in the plan itself.
Each column is an expression that can be resolved without referencing any input data.
For example, a literal, a function call involving literals, or any other expression that does
not require input.
```

Comment (anshuldata) [reply site/docs/relations/logical_relations.md]: > I'd be find with wording like
Fixed doc to indicate virtual expression table are not dependent on any input

**Comment by anshuldata** [comment]
> We should update the description message in this PR to also explain the deprecation.
Updated PR description to indicate VirutalTable is deprecated

**Comment by jacques-n** [comment]
I feel like the main outstanding question is whether we create new table type versus adding field to existing and deprecating old fields. @westonpace, @vbarua and @EpsilonPrime, thoughts? It's subjective but I'm definitely a fan of keeping the two table types separate so nobody can make a mistake and set both fields.

**Comment by westonpace** [comment]
My opinion hasn't changed.  I'll approve either approach.  If I need to take a side I still prefer one relation.  We have, several times now, deprecated fields and not worried too much about users setting both values.

I think confusion about "why are there two relations that seem to do the same thing" is going to be more confusing than "why are there two fields, one which is marked deprecated, and one that isn't".

**Review by westonpace — APPROVED** [review]

**Comment by jacques-n** [comment]
I'm fine with @westonpace's preferred change. @anshuldata can you update to single message with two sets of repeated fields?

**Comment by anshuldata** [comment]
> If I need to take a side I still prefer one relation.
Fixed as suggested. Basically to add another field in VirtualTable instead of another Relation type

**Review by vbarua — DISMISSED** [review]
Given the constraints of proto compatibility, this seems like as good as it's going to get. Thanks for working on this @anshuldata.

File: proto/substrait/algebra.proto [hunk 3]
```diff
@@ -88,7 +88,8 @@ message ReadRel {
 
   // A table composed of literals.
```
vbarua, line 89: minor/nit: For clarity it would be good to update this to
```
A table composed of expressions.
```

proto/substrait/algebra.proto
```diff
@@ -88,7 +88,8 @@ message ReadRel {
...
   message VirtualTable {
-    repeated Expression.Literal.Struct values = 1;
+    repeated Expression.Literal.Struct values = 1 [deprecated = true];
+    repeated Expression expression_values = 2;
```
vbarua, line 92: minor/nit: I'm partial to `expressions` instead of `expression_values` as it's less verbose an tacking `_values` on the end feels redundant. I'm not wedded to this.

site/docs/relations/logical_relations.md
```diff
@@ -51,12 +51,15 @@ Read definition types (like the rest of the features in Substrait) are built by
 #### Virtual Table
 
 A virtual table is a table whose contents are embedded in the plan itself.  The table data
-is encoded as records consisting of literal values.
+is encoded as records consisting of literal values or an expression that can be resolved without referencing any input data.
```
vbarua, line 54: minor/nit
```
literal values or an expression that can
  to
literal values or expressions that can
```
it's a little dramatically wonky that expression is not pluralized.

vbarua (commented):
Added some minor comments as well which I don't consider blocking.

proto/substrait/algebra.proto
anshuldata: fixed it

proto/substrait/algebra.proto
anshuldata: fixed it

site/docs/relations/logical_relations.md
anshuldata: fixed it

westonpace (approved):

vbarua (approved):

jacques-n (approved):

akoshchiy:
@anshuldata Hi! Trying to adapt the changes in [datafusion](https://github.com/apache/datafusion/issues/12984). It's not clear how to handle multiple rows in values, eq `SELECT * FROM VALUES (1+1, 2), (2+2, 3)`.
Should it be wrapped into `Expression.Nested.List` or `Expression.Nested.Struct`, or, maybe, there is another preferred way to handle it?

tokoko:
@akoshchiy `expressions` is a repeated field. You can set it to a list of `Expression` messages with `Expression.Literal.Struct` set within. It will be basically the same as the current behavior with deprecated `values` field with the only difference being that you need to wrap `Expression.Literal.Struct` messages in `Expression` messages now.

tokoko:
I think this is a really good question actually. with the `values` field, I'm assuming a single `Literal.Struct` value would be a single row. If that approach stays the same with the new `expressions` field, I think it only ever makes sense for the Expression to hold either `Literal.Struct` or `Nested.Struct`, it should never be a scalar as it's supposed to represent a row, not an individual value.

Wouldn't it make have made more sense in the first place for the new field to have been `repeated Expression.Nested.Struct` type instead of `repeated Expression`? Nested.Struct can contain all literals as well, after all.

anshuldata:
> Wouldn't it make have made more sense in the first place for the new field to have been repeated Expression.Nested.Struct type instead of repeated Expression
Yes, agree. I missed this in my original change. I will raise a PR to fix it

PR #713, opened 2024-09-26
feat: add decimal argument support to round function

andrew-coleman:
The round function has a number of variants to support different numeric types.  This commit adds support for rounding decimals. This is required for the spark module.

### DIFF
--- a/extensions/functions_rounding_decimal.yaml
+++ b/extensions/functions_rounding_decimal.yaml
@@ -1,0 +1,82 @@
+%YAML 1.2
+---
+scalar_functions:
+  -
+    name: "ceil"
+    description: >
+      Rounding to the ceiling of the value `x`.
+    impls:
+      - args:
+          - value: decimal<P,S>
+            name: x
+        return: |-
+          integral_least_num_digits = P - S + 1
+          precision = min(integral_least_num_digits, 38)
+          decimal?<precision, 0>
+  -
+    name: "floor"
+    description: >
+      Rounding to the floor of the value `x`.
+    impls:
+      - args:
+          - value: decimal<P,S>
+            name: x
+        return: |-
+          integral_least_num_digits = P - S + 1
+          precision = min(integral_least_num_digits, 38)
+          decimal?<precision, 0>
+  -
+    name: "round"
+    description: >
+      Rounding the value `x` to `s` decimal places.
+    impls:
+      - args:
+          - value: decimal<P,S>
+            name: x
+            description: >
+              Numerical expression to be rounded.
+          - value: i32
+            name: s
+            description: >
+              Number of decimal places to be rounded to.
+
+              When `s` is a positive number, the rounding
+              is performed to a `s` number of decimal places.
+
+              When `s` is a negative number, the rounding is
+              performed to the left side of the decimal point
+              as specified by `s`.
+
+              The precision of the resultant decimal type is one
+              more than the precision of the input decimal type to
+              allow for numbers that round up or down to the next
+              decimal magnitude.
+              E.g. `round(9.9, 0)` -> `10.0`.
+              The scale of the resultant decimal type cannot be
+              larger than the scale of the input decimal type.
+        options:
+          rounding:
+            description: >
+              When a boundary is computed to lie somewhere between two values,
+              and this value cannot be exactly represented, this specifies how
+              to round it.
+
+                - TIE_TO_EVEN: round to nearest value; if exactly halfway, tie
+                  to the even option.
+                - TIE_AWAY_FROM_ZERO: round to nearest value; if exactly
+                  halfway, tie away from zero.
+                - TRUNCATE: always round toward zero.
+                - CEILING: always round toward positive infinity.
+                - FLOOR: always round toward negative infinity.
+                - AWAY_FROM_ZERO: round negative values with FLOOR rule, round positive values with CEILING rule
+                - TIE_DOWN: round ties with FLOOR rule
+                - TIE_UP: round ties with CEILING rule
+                - TIE_TOWARDS_ZERO: round ties with TRUNCATE rule
+                - TIE_TO_ODD: round to nearest value; if exactly halfway, tie
+                  to the odd option.
+            values: [ TIE_TO_EVEN, TIE_AWAY_FROM_ZERO, TRUNCATE, CEILING, FLOOR,
+              AWAY_FROM_ZERO, TIE_DOWN, TIE_UP, TIE_TOWARDS_ZERO, TIE_TO_ODD ]
+        nullability: DECLARED_OUTPUT
+        return: |-
+          precision = min(P + 1, 38)
+          decimal?<precision, S>
--- a/tests/cases/rounding_decimal/ceil.test
+++ b/tests/cases/rounding_decimal/ceil.test
@@ -1,0 +1,7 @@
+### SUBSTRAIT_SCALAR_TEST: v1.0
+### SUBSTRAIT_INCLUDE: '/extensions/functions_rounding_decimal.yaml'
+
+# basic: Basic examples without any special cases
+ceil(2.25::dec<3,2>) = 3::dec<2,0>
+ceil(-65.5::dec<3,1>) = -65::dec<3,0>
+ceil(9.9::dec<2,1>) = 10::dec<2,0>
--- a/tests/cases/rounding_decimal/floor.test
+++ b/tests/cases/rounding_decimal/floor.test
@@ -1,0 +1,6 @@
+### SUBSTRAIT_SCALAR_TEST: v1.0
+### SUBSTRAIT_INCLUDE: '/extensions/functions_rounding_decimal.yaml'
+
+# basic: Basic examples without any special cases
+floor(2.25::dec<3,2>) = 2::dec<2,0>
+floor(-65.5::dec<3,1>) = -66::dec<3,0>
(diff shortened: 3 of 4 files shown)


extensions/functions_rounding.yaml
```diff
@@ -268,3 +268,43 @@ scalar_functions:
...
+              performed to the left side of the decimal point
+              as specified by `s`.
+        options:
+          rounding:
+            description: >
+              When a boundary is computed to lie somewhere between two values,
+              and this value cannot be exactly represented, this specifies how
+              to round it.
+
+                - TIE_TO_EVEN: round to nearest value; if exactly halfway, tie
```
EpsilonPrime, line 294: Can this happen with decimal representations?  I'd argue all of the floating point handling stuff here does not apply.

extensions/functions_rounding.yaml
```diff
@@ -268,3 +268,43 @@ scalar_functions:
               AWAY_FROM_ZERO, TIE_DOWN, TIE_UP, TIE_TOWARDS_ZERO, TIE_TO_ODD ]
         nullability: DECLARED_OUTPUT
         return: fp64?
+      - args:
```
EpsilonPrime, line 271: For all other decimal functionality we have placed them in `_decimal.yaml` files.  Not sure if we want to have just this one function in a file by itself though.

extensions/functions_rounding.yaml
```diff
@@ -268,3 +268,43 @@ scalar_functions:
...
+            name: "s"
+            description: >
+              Number of decimal places to be rounded to.
+
+              When `s` is a positive number, the rounding
+              is performed to a `s` number of decimal places.
+
+              When `s` is a negative number, the rounding is
+              performed to the left side of the decimal point
+              as specified by `s`.
```
EpsilonPrime, line 286: Does this operation affect the scale?  We should probably clarify that here.

andrew-coleman: I guess this function _could_ return a different decimal type (i.e. reduce the precision and the scale parameters), but I was working on the assumption that it would just return a different _value_.  I'm not sure if that is what you are asking.

extensions/functions_rounding.yaml
andrew-coleman: Why would these not apply?  For example, the input value `2.5` could be represented exactly in a decimal type, but rounding it to the nearest integer would result in `2` if the rounding mode is `TIE_TO_EVEN` or `3` if the mode is `TIE_AWAY_FROM_ZERO`.

extensions/functions_rounding.yaml
jacques-n: Let's not assume. Let's add some expected behaviors here. Once we get tests inside core, we can transplant those into test cases.

andrew-coleman:
Just revisiting this.  I've updated the parameters of the decimal return type to match the logic used by the Spark round functions:
https://github.com/apache/spark/blob/master/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/mathExpressions.scala#L1492

I note the comment here: https://github.com/substrait-io/substrait/pull/671#discussion_r1696305841
Does this still apply?  If so, how would you suggest I work around this since the output decimal scale depends on that second argument?

Thanks!

tests/cases/rounding_decimal/round_decimal.test
```diff
@@ -0,0 +1,11 @@
+### SUBSTRAIT_SCALAR_TEST: v1.0
+### SUBSTRAIT_INCLUDE: '/extensions/functions_rounding_decimal.yaml'
+
+# basic: Basic examples without any special cases
+round(2::dec<2,0>, 2::i32) = 2::dec<3,0>
+round(2.75::dec<8,2>, 1::i32) = 2.8::dec<8,1>
+
+# negative_rounding: Examples with negative rounding
+round(2::dec<2,0>, -2::i32) = 0::dec<2,0>
+round(123::dec<2,0>, -2::i32) = 100::dec<2,0>
```
EpsilonPrime, line 10: there are three digits here so the precision needs to be 3

EpsilonPrime (commented):
I don't think we can use any constants in the calculation.  The s parameter in round could be not a constant although I suspect some backends only will allow constants here.  If it is not a constant then we don't know a lot.  If the number turns out to be negative the scale could even increase.  Because that second value could change the value in any manner we likely have to return maximum scale and precision here.


I will take it as an action to try running these tests against a backend (as the consumer testing does) to see if the tests are functional/correct.

tests/cases/rounding_decimal/round_decimal.test
```diff
@@ -0,0 +1,11 @@
+### SUBSTRAIT_SCALAR_TEST: v1.0
+### SUBSTRAIT_INCLUDE: '/extensions/functions_rounding_decimal.yaml'
+
+# basic: Basic examples without any special cases
+round(2::dec<2,0>, 2::i32) = 2::dec<3,0>
```
EpsilonPrime, line 5: DuckDB returns:

2,0
8,1
2,0
3,0
4,0

andrew-coleman:
Thanks @EpsilonPrime, that's helpful.  It looks like the scale of the decimal returned by DuckDB is also a function of the second parameter of `round()` as is the case with Spark.

Given the following query:
```sql
select num, floor(num), ceil(num),
     round(num, -2), round(num, -1), round(num, 0),
     round(num, 1), round(num, 2), round(num, 3)
from (values (0.5), (-0.5), (999.9), (-999.9), (2.75)) as table(num)
```
Spark produces the following output and type schema:
```
+-------+----------+---------+--------------+--------------+-------------+-------------+-------------+-------------+
|    num|FLOOR(num)|CEIL(num)|round(num, -2)|round(num, -1)|round(num, 0)|round(num, 1)|round(num, 2)|round(num, 3)|
+-------+----------+---------+--------------+--------------+-------------+-------------+-------------+-------------+
|   0.50|         0|        1|             0|             0|            1|          0.5|         0.50|         0.50|
|  -0.50|        -1|        0|             0|             0|           -1|         -0.5|        -0.50|        -0.50|
| 999.90|       999|     1000|          1000|          1000|         1000|        999.9|       999.90|       999.90|
|-999.90|     -1000|     -999|         -1000|         -1000|        -1000|       -999.9|      -999.90|      -999.90|
|   2.75|         2|        3|             0|             0|            3|          2.8|         2.75|         2.75|
+-------+----------+---------+--------------+--------------+-------------+-------------+-------------+-------------+

root
 |-- num: decimal(5,2) (nullable = false)
 |-- FLOOR(num): decimal(4,0) (nullable = true)
 |-- CEIL(num): decimal(4,0) (nullable = true)
 |-- round(num, -2): decimal(4,0) (nullable = true)
 |-- round(num, -1): decimal(4,0) (nullable = true)
 |-- round(num, 0): decimal(4,0) (nullable = true)
 |-- round(num, 1): decimal(5,1) (nullable = true)
 |-- round(num, 2): decimal(6,2) (nullable = true)
 |-- round(num, 3): decimal(6,2) (nullable = true)
```

I propose, then, that the given an input of type `decimal<P, S>`, the return type expression should be:
```yaml
        return: |-
          precision = min(P + 1, 38)
          decimal?<precision, S>
```

Which is not necessarily what it actually returns, but is the maximum precision/scale of what it could return (taking into account your earlier [comment](https://github.com/substrait-io/substrait/pull/713#pullrequestreview-2616911376)).

How does that sound?

EpsilonPrime (approved):
Works for me.

andrew-coleman:
Thanks for the approval @EpsilonPrime.
Would it be possible to merge this?  I can then follow up with the Java implementation. 
cc @vbarua @jacques-n @westonpace @cpcloud 

Thanks!

andrew-coleman:
@jacques-n, would you be happy with this? ^^

westonpace (approved):
If I understand the two points correctly they are:

1. If we had a notion of constant arguments we could potentially express a more accurate return type.
2. Some engines choose to return data-dependent return types that are more narrow than the worst-case scenario.

Regarding (1) I don't think this corner case warrants the additional complexity of constant arguments.  I'm not aware of any significant planner improvements that could be made by more clearly knowing the decimal scale/precision.

Regarding (2) I don't think this is a Substrait concern.  Engines are always permitted to return more efficient encodings of data as long as the values fit the range defined by the return type.  It's not possible for Substrait (which doesn't have access to data) to make a better decision with the data it does have.

So I approve.

The only thing that might be nice to clarify is what should happen if a 38 digit number consisting of all 9's is rounded up.  Does it saturate (return the number unchanged) or emit an error?

andrew-coleman:
> The only thing that might be nice to clarify is what should happen if a 38 digit number consisting of all 9's is rounded up. Does it saturate (return the number unchanged) or emit an error?

I guess that would be implementation specific?

In the case of Spark, it throws an error:

`spark.sql("select round(99999999999999999999999999999999999999, -1)").show()`

`org.apache.spark.SparkArithmeticException: [DECIMAL_PRECISION_EXCEEDS_MAX_PRECISION] Decimal precision 39 exceeds max precision 38.`

westonpace:
Ok.  Let's worry about it in a follow-up.  We can either document the error behavior or add an overflow option as we learn more.

jacques-n:
> If we had a notion of constant arguments

We do. We just haven't figured out yet how to expose to type derivation.

> Some engines choose to return data-dependent return types that are more narrow than the worst-case scenario.

Then those engines don't implement these functions.

PR #716, opened 2024-09-29
fix: add missing udt identifier in functions_geometry

tokoko:
argument type for is_closed function is missing `u!` before geometry udt causing a parser to treat it as a non-existent builtin data type.

### DIFF
--- a/extensions/functions_geometry.yaml
+++ b/extensions/functions_geometry.yaml
@@ -77,7 +77,7 @@ scalar_functions:
     impls:
       - args:
           - name: geom
-            value: geometry
+            value: u!geometry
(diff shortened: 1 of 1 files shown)


EpsilonPrime (approved):

EpsilonPrime:
Thanks for the fix!  Getting a second look to verify.  Since it's broken I'm not sure we need to mark it as a breaking change.

jacques-n (approved):
yeah, agree that this isn't breaking. this file was introduced before the spec was updated to require u! so that change is actually the breaking change and this fix was missed then.

vbarua (approved):

Issue #725, opened 2024-10-11
FR: Add plan_anchor field to PlanRel


tokoko:
I'm working on a [library](https://github.com/tokoko/subframe) that lets you build up substrait plans using dataframe API. Over the course of dataframe transformations various unrelated plans need to be merged to construct the final substrait plan. For the most part it's pretty straightforward, but you need to deal with extension functions and references during the merges. 

Functions in substrait fortunately let you define "pointers" in a relaxed fashion with anchors. This means that as long as there is some global way of assigning a unique identifier to each function beforehand, merging plans become straightforward. Unlike functions, relationship between PlanRels and RefereceRels is modeled differently, RefereceRels need to be aware of the ordinal position of the target rel in the relations array. This makes merging plans very tricky. One has to either traverse the whole plan to adjust ordinal positions on each merge or use some sort of placeholders and do a single-pass traversal at the end.

To make this simpler, substrait could use the same anchor mechanism with PlanRels. An additional `plan_anchor` field can be introduced in PlanRel message, which will be used by ReferenceRel instead of relying on ordinal positions.

drin:
This sounds cool and potentially related to something I'm working on.

> Functions in substrait fortunately let you define "pointers" in a relaxed fashion with anchors. This means that as long as there is some global way of assigning a unique identifier to each function beforehand, merging plans become straightforward

Do you mean merging plans is straightforward with respect to not having to compare function semantics?

> To make this simpler, substrait could use the same anchor mechanism with PlanRels

How would you define the same type of anchor for a PlanRel? It seems easy to think that a function extension URI maps 1-1 with a particular semantic definition of a function. I'm not sure how to map that to PlanRels

drin:
In my own project, I use `PlanAnchor` to refer to an operator where I will split a plan for distributed processing [code-plananchor]. The operator becomes a leaf of the original plan and a parent of the subplan that was originally rooted at the operator. The subplan becomes a new, independent query plan that carries the `PlanAnchor` as a reference for later merging.

Discussion of splitting and merging aside, the `PlanAnchor` marks a spot in a plan that says "someone else will execute this sub-graph, and when they're done the results should stream back into this spot;" so, I think my intention is the same but I have the benefit of knowing that I can create a `PlanAnchor` at planning time in a way that there is an origin to reconnect it back to.

[...]

PR #726, opened 2024-10-25
feat: add anchor for non-ordinal plan reference

drin:
This adds an "anchor" to `PlanRel` that can be referenced from a `ReferenceRel`. This anchor and reference relationship provides a non-ordinal method for identifying and accessing a "subtree" or sub-graph. To maintain consistency with `type_anchor` and `function_anchor`, the new `subtree_anchor` attribute is `uint32`.

This PR leaves in the original `subtree_ordinal` attribute since it seems a (mildly) more performant method for referencing a subtree, but also since it is still relevant in the typical case. The new anchor attribute is an improvement for cases where multiple plans are merged and at least one already contains a `ReferenceRel`.

It is expected that only one of `subtree_ordinal` or `subtree_reference` will be used, however I don't see a good reason to enforce the use of only one, so I did not group the attributes in a `oneof` constraint.

If either of the above points ((1) keeping `subtree_ordinal` and (2) not enforcing a `oneof` constraint) are changed, then this PR would likely become a breaking change.

Also, this PR addresses #725 , but uses the name `subtree_reference` and `subtree_anchor` (instead of `plan_anchor`) to be consistent with the pre-existing `subtree_ordinal` attribute. I don't know that relations in a substrait plan are strictly trees, so I think this could be a good time to change "subtree" to "planrel" or "subgraph" or something similar.

### DIFF
--- a/proto/substrait/algebra.proto
+++ b/proto/substrait/algebra.proto
@@ -1633,8 +1633,12 @@ message AggregateFunction {
   }
 }
 
-// This rel is used  to create references,
+// This rel is used to create references,
 // in case we refer to a RelRoot field names will be ignored
 message ReferenceRel {
+  // points to a PlanRel in Plan.relations (ordinal reference)
   int32 subtree_ordinal = 1;
+
+  // points to a subtree_anchor in PlanRel (for non-ordinal reference)
+  uint32 subtree_reference = 2;
 }
--- a/proto/substrait/plan.proto
+++ b/proto/substrait/plan.proto
@@ -19,6 +19,9 @@ message PlanRel {
     // The root of a relation tree
     RelRoot root = 2;
   }
+
+  // provides a non-ordinal anchor for ReferenceRel
+  uint32 subtree_anchor = 3;
 }
 
 // Describe a set of operations to complete.


drin:
I should have time to finish this up in the next few days

proto/substrait/algebra.proto
```diff
@@ -1633,8 +1633,12 @@ message AggregateFunction {
   }
 }
 
-// This rel is used  to create references,
+// This rel is used to create references,
 // in case we refer to a RelRoot field names will be ignored
 message ReferenceRel {
+  // points to a PlanRel in Plan.relations (ordinal reference)
```
jacques-n, line 1639: let's put these in on a one of (subtree_reference and subtree_ordinal) and deprecate subtree ordinal.

drin:
rebased on main

proto/substrait/algebra.proto
drin: I named the oneof `ref_type` since that seemed descriptive enough without being too verbose.

drin:
hm, according to [a stackoverflow post][web-so-protobuf-breakingchange], moving an optional field into a `oneof` is not breaking. This PR essentially makes `subtree_ordinal` an optional field instead of an implicit field by moving it into a `oneof`, but that seems backwards-compatible. However, the check for breaking changes still flags it so I decided to add a `BREAKING CHANGES:` footer to the PR description.


[web-so-protobuf-breakingchange]: https://stackoverflow.com/a/60079101

EpsilonPrime:
By making the anchor on the plan object does that mean you can no longer use a ReferenceRel to point to a subtree of the existing plan?  For instance, TPC-H 15 processes the same subtree twice and often ends up with different results (when using float).  By using a reference you could process the subtree once and point to it avoiding differences.

drin:
> By making the anchor on the plan object...

I think you may be misunderstanding where the anchor is? I don't think of the PlanRel as the Plan object

> ...does that mean you can no longer use a ReferenceRel to point to a subtree of the existing plan?

The change is essentially to change a relative ID (ordinal) to an absolute ID (anchor). I think in either case you need to know what you're pointing to and you can point to a subtree of the existing plan.

proto/substrait/algebra.proto
```diff
@@ -1734,8 +1734,14 @@ message AggregateFunction {
...
+// This rel is used to create references,
 // in case we refer to a RelRoot field names will be ignored
 message ReferenceRel {
-  int32 subtree_ordinal = 1;
+  oneof ref_type {
+    // points to a PlanRel in Plan.relations (ordinal reference)
+    int32 subtree_ordinal = 1 [deprecated = true];
+
+    // points to a subtree_anchor in PlanRel (for non-ordinal reference)
+    uint32 subtree_reference = 2;
```
vbarua: What do we think about making this even more generic and just calling it a `rel_reference`?

proto/substrait/algebra.proto
```diff
@@ -1734,8 +1734,14 @@ message AggregateFunction {
   }
 }
 
-// This rel is used  to create references,
+// This rel is used to create references,
 // in case we refer to a RelRoot field names will be ignored
 message ReferenceRel {
-  int32 subtree_ordinal = 1;
```
vbarua, line 1770: If we're going to do this, we should deprecate the old field, with the eventual plan of removing `subtree_ordinal` entirely.

And if we're going to do that, I don't know how much sense it makes to put them both in a oneof. With the way you've worded the anchor on the PlanRel, 0 isn't a valid rel anchor, so while these field coexist, if you have a ReferenceRel with `subtree_reference` set to 0 you know you have to use the `subtree_ordinal`.

proto/substrait/plan.proto
```diff
@@ -19,6 +19,10 @@ message PlanRel {
     // The root of a relation tree
     RelRoot root = 2;
   }
+
+  // provides a non-ordinal anchor for ReferenceRel.
+  // A default value (0) should be considered unset.
+  uint32 subtree_anchor = 3;
```
vbarua: Why don't just we call this `rel_anchor`, because its the anchor for a relation. The word `subtree` isn't used anywhere in the PlanRel message so it feels more consistent to just call this what it is, and I feel like it will save us having to explain that when we say subtree we mean Rel, and that it's not a separate concept from a Rel.

vbarua (commented):
Let me see if I understand the intent of this change.

### Function / Type References

For things like functions, we define Plan-level anchors https://github.com/substrait-io/substrait/blob/d430e521f203aec6a4e06731d4bfd68cdf61f443/proto/substrait/plan.proto#L34-L35

https://github.com/substrait-io/substrait/blob/d430e521f203aec6a4e06731d4bfd68cdf61f443/proto/substrait/extensions/extensions.proto#L57-L59

which are then used within the plan
https://github.com/substrait-io/substrait/blob/d430e521f203aec6a4e06731d4bfd68cdf61f443/proto/substrait/algebra.proto#L1178-L1183

This mechanism is also used for types.

### Rel References
For RelReference we don't currently use an anchor, but instead an ordinal reference
https://github.com/substrait-io/substrait/blob/d430e521f203aec6a4e06731d4bfd68cdf61f443/proto/substrait/algebra.proto#L1756-L1760
into an list of relations in the Plan
https://github.com/substrait-io/substrait/blob/d430e521f203aec6a4e06731d4bfd68cdf61f443/proto/substrait/plan.proto#L37-L38

### Your Proposal
Bring references to relations into line with functions and types by attaching an anchor to the PlanRel message.

### Notes
Bringing this inline with how we do references generally make sense to me. That being said:

> The new anchor attribute is an improvement for cases where multiple plans are merged and at least one already contains a ReferenceRel.

How does this make merging plans easier. I assume that as with ordinal-based references you would still have to rewrite anchors when they collide between plans?

> It is expected that only one of subtree_ordinal or subtree_reference will be used, however I don't see a good reason to enforce the use of only one

If we're going to do this, I think it would be better to have only 1 way to do it and we should plan to get rid of the ordinal-based resolution.

proto/substrait/algebra.proto
vbarua: @jacques-n is there a reason to both deprecate the existing field AND put both of the deprecaetd and new field in a oneof. It feels noisy to create a oneof if the eventual plan is to remove one of the fields, and hence the need for the oneof.

proto/substrait/algebra.proto
vbarua: @drin you hada better suggestion in Slack to call this`planrel_reference` because it points to a PlanRel specifically, and not Rel.

proto/substrait/algebra.proto
vbarua: Talking through this in the Substrait sync with @jacques-n,  we agreed that keeping this _out_ of a `oneof` make sense.

For migration purposes, its actually useful to allow both to be set for a while. Effectively, we can add support for this in producers and consumers in a decoupled fashion, and once everything supports both, remove the old field.

As part of this, we would need to document that the new relation anchor should be preferred when both are set.

vbarua (commented):
I brought this PR up in the Substrait sync this week. I've updated some comments with what we talked about.

tokoko:
@vbarua I'm not opposed to `planrel_reference`, but one slight caveat regarding naming: this field will most likely act as a reference to the list of relations in `ExtendedExpression` message as well (see #794). In my PR I decided to introduce new `ExtendedExpressionRel` message to hold Rels. If we go that route, calling the field `planrel_reference` doesn't make as much sense anymore. Alternatively we can reuse `PlanRel` message for `ExtendedExpression` relations, which is doable but slightly awkward, because relations in `PlanRel` for extended expressions will only ever be of type `Rel`, never `RelRoot`.

drin:
I think more context about the interaction with `ExtendedExpressionRel` would be helpful.

I commented on #794 (moved my comment from here).

vbarua:
@tokoko left a comment on your proposal in https://github.com/substrait-io/substrait/pull/794 with regards to including relations  in ExtendedExpressions.

proto/substrait/plan.proto
drin: I'll change this to `planrel_anchor` to be more descriptive. `rel_anchor` is a misleading since it doesn't reference a `Rel`.

drin: Actually, I just realized it may be better to call it a `reference_anchor` to signify that it is an anchor, but I don't think it has to specify that it is a `PlanRel` anchor (since it's in `PlanRel`). The comment will be updated to still clarify it as an anchor for a `ReferenceRel` to point to.

drin:
I accommodated feedback from @vbarua and rebased on main

drin:
@tokoko @vbarua as long as `planrel_reference` is a new field in `ReferenceRel` anyways, what if we actually nest it in a `oneof`, that way we can be explicit about other things the reference may point to (e.g. `extendedexpr_reference`) in a way that is a bit future proof (although a bit more verbose). Or, we can change it to be `anchor_reference` and have an enum that specifies what the expected type of message being pointed to is?

proto/substrait/algebra.proto
```diff
@@ -1764,8 +1764,14 @@ message AggregateFunction {
...
+  // points to a PlanRel in Plan.relations by position (ordinal reference)
+  // this is deprecated in favor of `planrel_reference`
+  int32 subtree_ordinal = 1 [deprecated = true];
+
+  // points to a PlanRel in Plan.relations by use of an anchor (non-ordinal reference),
+  // PlanRel.reference_anchor. A default value (0) is considered unset.
```
benbellick, line 1775: Just curious, is it required to have a default value of `0` be considered unset? It seems a bit strange considering that all of the other places where we have reference / anchors, we allow `0` (e.g. [here](https://github.com/substrait-io/substrait/blob/793c64ba26e337c22f5e91b658be58b1eea7efd3/proto/substrait/algebra.proto#L1220)).

benbellick (commented):
Can you also update the docs in `logical_relations.md`? There is still only the old field mentioned: 
```
### Reference Properties

| Property                    | Description                                                                    | Required                    |
|-----------------------------|--------------------------------------------------------------------------------| --------------------------- |
| Referred Rel                | A zero-indexed positional reference to a `Rel` defined within the same `Plan`. | Required                    |
```

proto/substrait/algebra.proto
drin: defining 0 as unset prevents reference rel from pointing to arbitrary plans accidentally. although we haven't agreed on cross plan semantics, this helps leave that possibility open  (ergonomically, at least)

tokoko: I agree with Ben. During transition phase, both of these values need to be set and pointing to identical plans for the plan to be considered valid so there's not much chance of accidentally pointing to an incorrect plan. Afterwards (when `subtree_ordinal` will be removed) there's essentially no difference between this and other anchors, so I don't see a reason for disallowing 0 in this instance alone.

drin: okay, but I think that it makes sense that if you intend for it to be used, then you set it to a value. The only "extra work" you need to do is set the value for the first PlanRel and first ReferenceRel, but you need to set it for every subsequent one regardless.

I can yield on this, but I am using it for distributed query planning, so I would just be needing to extend these semantics or add another field later if so.

drin: IMO, a PlanRel anchor is semantically independent from a ReferenceRel. It can be referenced by many separate ones. But, in an aggregation, the anchor can only be referenced by fields in the same operator, those anchors are not semantically independent from their references.

This is why the default value is fine there.

benbellick: To me it makes sense to say that that `reference_anchor` must be unique. For usage of substrait at my job, the majority of plans only have one relation in them, so the default value of `0` is not a problem. But if your plan has two relations in it, then both using the default value should be considered invalid substrait. 

I realize that this puts an additional constraint on plan production / validation, but I think that it's worth it to be consistent on our usage of the 0 anchor.

drin: then should it be optional? otherwise there is no way to know if it was set unless you go through the work of checking they're all unique. does the anchor matter at all if there are no reference rels? or can you only have multiple planrels if you have at least 1 reference rel?

drin: also, during the deprecation period, if 0 is valid, then do you always have to check the second/last planrel to see if the anchor should be used or if the subtree ordinal should be used?

tokoko: I think we should do the same as in uri/urn migration, consumer should not have to check which one to use, both should yield the exact same references.

drin: hmm, for my own clarification, if 0 is a valid value and during the deprecation period they can/should be set to the same values (meaning the anchor only has a valid value within the plan itself and not across plans), then when are anchor semantics beneficial over ordinal semantics?

tokoko: good point, we'll start to see the value when we drop the deprecated field (you're free to violate the rule I suggested in your project if you can get away with it of course.

I also think if we can get this merged, we can speedrun (relatively speaking) the rest. From what I have seen, not many consumers/producers support ReferenceRel in the first place anyway.

benbellick:
@drin I just got the email that this autoclosed but considering that @vbarua had already approved it, I think it's appropriate to open it back up and nudge another PMC about it. 

I'll try and open it back up today. Thanks!

PR #727, opened 2024-10-28
fix: virtualTable expression should represent a row of expression

anshuldata:
BREAKING CHANGE: changes the message type for Expressions field in VirtualTable

Fixed issue mentioned in [this](https://github.com/substrait-io/substrait/pull/711#issuecomment-2439930751)
I have fixed it by fixing recently introduced field. Though this is a breaking change but since this field is introduced very recently so I assume this won't have many consumers. So, instead of having another field I decided to fix existing field

### DIFF
--- a/proto/substrait/algebra.proto
+++ b/proto/substrait/algebra.proto
@@ -126,7 +126,7 @@ message ReadRel {
   // A table composed of expressions.
   message VirtualTable {
     repeated Expression.Literal.Struct values = 1 [deprecated = true];
-    repeated Expression expressions = 2;
+    repeated Expression.Nested.Struct expressions = 2;
   }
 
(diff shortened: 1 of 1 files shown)


jacques-n (approved):
Oops!

Makes sense.

jacques-n:
@vbarua @EpsilonPrime  @westonpace would be good to get your eyes on this so we can fix quickly

westonpace (approved):

jacques-n:
@anshuldata , this shouldn't have been merged until we had at least 2 SMC approvals since it is a spec change. Please refer to https://substrait.io/governance/#substrait-voting-process in future to make sure we're following process. Thanks!

PR #728, opened 2024-10-28
feat: add antlr grammar for test file format

scgkiran:
Add parser changes for testcase files
Add function test coverage changes


vbarua:
Would it make sense to split this into 2 PRs, one for the type grammar which is effectively just an import of the existing grammar from substrait-java, and then one for the new test grammar which requires more careful review?

grammar/FuncTestCaseLexer.g4
```diff
@@ -0,0 +1,103 @@
...
+    : 'v' DIGIT+ ('.' DIGIT+)?
+    ;
+
+DescriptionLine
+    : '# ' ~[\r\n]* '\r'? '\n'
+    ;
+
+ErrorResult: '<!ERROR>';
+UndefineResult: '<!UNDEFINED>';
+Overflow: 'overlfow';
```
EpsilonPrime, line 24: overflow, probably should use consistent case as other literals nearby (all upper or lower)

grammar/FuncTestCaseLexer.g4
```diff
@@ -0,0 +1,103 @@
...
+import SubstraitLexer;
+
+options {
+    caseInsensitive = true;
+}
+
+Whitespace    : [ \t\n\r]+ -> channel(HIDDEN) ;
+
+SubstraitScalarTest: '### SUBSTRAIT_SCALAR_TEST:';
+SubstraitInclude: '### SUBSTRAIT_INCLUDE:';
```
EpsilonPrime, line 12: This feels more like a pragma than a token.  Perhaps change this to a grammar rule like '###' Identifier ':'?

grammar/FuncTestCaseParser.g4
```diff
@@ -0,0 +1,227 @@
...
+    : DecimalLiteral | IntegerLiteral | floatLiteral
+    ;
+
+floatLiteral
+    : FloatLiteral | NaN
+    ;
+
+nullArg: NullLiteral DoubleColon datatype;
+
+i8Arg: IntegerLiteral DoubleColon I8;
```
EpsilonPrime, line 71: Is there any value to having i8Arg separate from any other integer -- no matter what happens you will need to check that the value fits the chosen size.

For that matter, the same applies to other numbers.  Another benefit is that instead of someone being told that they need to put an i8/i16/i132/i64/fp32/fp64 when they have a syntax error is that it would tell them they need a number.

tests/coverage/antlr_parser/FuncTestCaseLexer.tokens
EpsilonPrime, line 1: I don't believe this needs to be checked in.

tests/coverage/antlr_parser/FuncTestCaseParser.interp
EpsilonPrime, line 1: This shouldn't need to be checked in.

tests/coverage/antlr_parser/FuncTestCaseLexer.tokens
scgkiran: removed it

tests/coverage/antlr_parser/FuncTestCaseParser.interp
scgkiran: removed it

grammar/FuncTestCaseLexer.g4
scgkiran: Changed it to a parser rule, but retained the `SUBSTRAIT_*` keywords as tokens.

grammar/FuncTestCaseLexer.g4
scgkiran: fixed it.

grammar/FuncTestCaseLexer.g4
```diff
@@ -0,0 +1,104 @@
...
+    : 'v' DIGIT+ ('.' DIGIT+)?
+    ;
+
+DescriptionLine
+    : '# ' ~[\r\n]* '\r'? '\n'
+    ;
+
+ErrorResult: '<!ERROR>';
+UndefineResult: '<!UNDEFINED>';
+Overflow: 'OVERLFOW';
```
EpsilonPrime, line 25: overflow

grammar/FuncTestCaseParser.g4
EpsilonPrime: It may help to add tests for syntax errors so that behavior can be seen.

grammar/FuncTestCaseLexer.g4
scgkiran: fixed the typo

grammar/FuncTestCaseParser.g4
scgkiran: Added tests. Combined few rules.

EpsilonPrime (approved):

jacques-n (approved):
Looks good. Can we add a follow-up task to get this incorporated into CI? As it stands, it looks like nothing guarantees that this runs before merges and in main.

scgkiran:
> Looks good. Can we add a follow-up task to get this incorporated into CI? As it stands, it looks like nothing guarantees that this runs before merges and in main.

I will do it in next PR.

PR #730, opened 2024-10-31
feat: add antlr grammar for types

scgkiran:


grammar/SubstraitLexer.g4
```diff
@@ -0,0 +1,157 @@
...
+
+// Whitespace and comment handling
+LineComment   : '//' ~[\r\n]* -> channel(HIDDEN) ;
+BlockComment  : ( '/*' ( ~'*' | '*'+ ~[*/] ) '*'* '*/' ) -> channel(HIDDEN) ;
+Whitespace    : [ \t\r]+ -> channel(HIDDEN) ;
+
+// Substrait is case-insensitive, ANTLR is not. So, in order to define our
+// keywords in a somewhat readable way, we have to define these shortcuts.
+
+fragment A : [aA];
```
tokoko, line 11: @EpsilonPrime suggested swapping this with `caseInsensitive = true;` option over in substrait-io/substrait-python#63

grammar/SubstraitLexer.g4
```diff
@@ -0,0 +1,157 @@
...
+fragment U : [uU];
+fragment V : [vV];
+fragment W : [wW];
+fragment X : [xX];
+fragment Y : [yY];
+fragment Z : [zZ];
+
+fragment DIGIT: [0-9];
+
+fragment INTEGER
```
EpsilonPrime, line 40: This fragment doesn't appear to be used elsewhere.

grammar/SubstraitType.g4
```diff
@@ -0,0 +1,71 @@
...
+  : scalarType isnull=QMark?
+  | parameterizedType
+  | anyType
+  ;
+
+expr
+  : OParen expr CParen                                                                      #ParenExpression
+  | Identifier Eq expr Newline+ (Identifier Eq expr Newline+)* finalType=typeDef Newline*   #MultilineDefinition
+  | typeDef                                                                                 #TypeLiteral
+  | number=Number                                                                           #LiteralNumber
```
EpsilonPrime, line 63: Is "number=" here an ad-hoc lexer symbol?  If so, should be explicitly part of the lexer?  Does it prevent the word number from being used as an identifier (i.e. did it become a reserved word)?

grammar/SubstraitType.g4
```diff
@@ -0,0 +1,71 @@
...
+  ;
+
+numericParameter
+  : Number              #numericLiteral
+  | Identifier          #numericParameterName
+  | expr                #numericExpression
+  ;
+
+anyType
+    : Any isnull=QMark?
```
EpsilonPrime, line 49: Use the same indentation as above.

grammar/SubstraitType.g4
```diff
@@ -0,0 +1,71 @@
...
+  | parameterizedType
+  | anyType
+  ;
+
+expr
+  : OParen expr CParen                                                                      #ParenExpression
+  | Identifier Eq expr Newline+ (Identifier Eq expr Newline+)* finalType=typeDef Newline*   #MultilineDefinition
+  | typeDef                                                                                 #TypeLiteral
+  | number=Number                                                                           #LiteralNumber
+  | identifier=Identifier isnull=QMark?                                                     #TypeParam
```
tokoko, line 64: Is calling this a TypeParam correct? The value might be either a type or numeric, no?

grammar/SubstraitLexer.g4
scgkiran: fixed it.

grammar/SubstraitLexer.g4
scgkiran: Removed it. I was using this for TestCase file format.

grammar/SubstraitType.g4
scgkiran: fixed it.

grammar/SubstraitType.g4
tokoko: Are you sure that's accurate? I think `varchar<L1>` will fall under TypeLiteral and as a result L1 will be covered by `numericParameter` (typeDef -> parameterizedType -> varchar) which makes sense because L1 can only ever be a numeric, never a type.

#TypeParam should cover any time parameters are used outside of the parameterized type context. for example in the "add" function:
```
init_scale = max(S1,S2)
delta = init_prec - 38
prec = min(init_prec, 38)
scale_after_borrow = max(init_scale - delta, min_scale)
DECIMAL<prec, scale>
```

Every time a parameter is used (except for the last expression) params will be resolved to `TypeParam`, while `prec` and `scale` at the end will be resolved to `NumericParameterName` instead.

All I'm saying is that when we have two expressions named `NumericParameterName` and `TypeParam`, it makes you think that one can only hold numerics and the other can only hold types, when in fact `TypeParam` can hold both.

I'd suggest using more generic `ParameterName` instead of `TypeParam` to make that distinction less confusing, plus make the name more consistent with `NumericParameterName`.

scgkiran: You are Right, for varchar `numericParameter` covers `L1` and `10`. I was confusing it with parameters of UserDefinedType.

I will change the name to `ParameterName`

tokoko: thanks. let me take the discussion a bit further though. I was thinking, is there a point in having a separate `numericParameter` token? If the goal is to somehow limit what is or is not applicable in that context, I don't think that can be achieved anyway. `numericParameter` can contain an `expr` which may or may not be numeric. On the flip side, both number (literal) and identifier (parameter) can be expressed by expr anyway. Can't we have exprs instead of numericParameters inside parameterizedType?:

```
FixedChar isnull=QMark? Lt length=expr Gt                                   #fixedChar
...
```

grammar/SubstraitLexer.g4
```diff
@@ -0,0 +1,129 @@
...
+options {
+    caseInsensitive = true;
+}
+
+// Whitespace and comment handling
+LineComment   : '//' ~[\r\n]* -> channel(HIDDEN) ;
+BlockComment  : ( '/*' ( ~'*' | '*'+ ~[*/] ) '*'* '*/' ) -> channel(HIDDEN) ;
+Whitespace    : [ \t\r]+ -> channel(HIDDEN) ;
+
+// Substrait is case-insensitive, ANTLR is not. So, in order to define our
```
EpsilonPrime, line 12: Comment no longer needed.

grammar/SubstraitType.g4
```diff
@@ -0,0 +1,75 @@
...
+  | parameterizedType
+  | anyType
+  ;
+
+expr
+  : OParen expr CParen                                                                      #ParenExpression
+  | Identifier Eq expr Newline+ (Identifier Eq expr Newline+)* finalType=typeDef Newline*   #MultilineDefinition
+  | typeDef                                                                                 #TypeLiteral
+  | number=Number                                                                           #LiteralNumber
+  | identifier=Identifier isnull=QMark?                                                     #ParameterName
```
EpsilonPrime, line 68: Not sure renaming identifier here helps (nor number above).

grammar/SubstraitLexer.g4
scgkiran: removed it

grammar/SubstraitType.g4
scgkiran: Removed it. I wasn't sure if it is helpful in any language, if needed we can add it back.

EpsilonPrime (approved):

grammar/SubstraitType.g4
scgkiran: I see the duplication of rules, but I also see current grammar is helpful in parser code to create different type of objects based on the context. 

I haven't yet worked on the type derivation, but this is the grammar that is used in other substrait repos, so there must be a good reason.

jacques-n (approved):

grammar/SubstraitType.g4
tokoko: I just found it redundant when I was doing an implementation in python. I'm fine with leaving it as is as well, though.

PR #734, opened 2024-11-07
feat: update operator to update a table

scgkiran:


proto/substrait/algebra.proto
```diff
@@ -619,7 +620,7 @@ message WriteRel {
   // The type of operation to perform
   WriteOp op = 4;
 
-  // The relation that determines the records to add/remove/modify
```
EpsilonPrime, line 622: Modification using this particular method is still reasonable -- one would use this if you had ids to merge against.

site/docs/relations/logical_relations.md
```diff
@@ -474,6 +474,32 @@ Write definition types are built by the community and added to the specification
 | Format   | Enumeration of available formats. Only current option is PARQUET. | Required |
 
 
+## Update Operator
+
+The update operator is an operator that consumes one input and writes the column updates to a storage. 
+
+| Signature            | Value                                 |
+| -------------------- |---------------------------------------|
+| Inputs               | 1                                     |
```
EpsilonPrime, line 483: Technically this new operator doesn't have an input (besides the table name).

proto/substrait/algebra.proto
scgkiran: In current state it is not possible. Right?
We can modify the comment if we modify this proto.

site/docs/relations/logical_relations.md
scgkiran: We can consider the table name on which the operator applied as input?
0 inputs for a update operator sounds strange.

proto/substrait/algebra.proto
scgkiran: ok got it, will revert the changes related to WriteRel.

site/docs/relations/logical_relations.md
EpsilonPrime: Inputs are traditionally other relations.  You could say zero and add a comment explaining that it does get data from somewhere.  The closest analog we have for existing relations is ReadRel.

scgkiran: ok fixed it. Please take a look.

EpsilonPrime (approved):

jacques-n (approved):

PR #736, opened 2024-11-08
feat: update test file format to support aggregate functions

scgkiran:


tests/coverage/antlr_parser/FuncTestCaseParser.py
EpsilonPrime, line 1: Reminder:  Eventually we want to stop checking these in if possible.

scgkiran: Once the grammar changes are stable, the checkins to auto-generated files will stop (I think it will be sooner than later, we have to add existing BFT tests in new format here).
We have a pre-commit check on test coverage so we need this file in this repo.

grammar/FuncTestCaseParser.g4
```diff
@@ -126,53 +184,56 @@ timeInterval
...
+  : (Bool | Boolean)        #boolean
+  | I8                      #i8
+  | I16                     #i16
+  | I32                     #i32
+  | I64                     #i64
+  | FP32                    #fp32
+  | FP64                    #fp64
+  | (Str | String)          #string
+  | (Binary | VBin)         #binary
+  | (Ts | Timestamp)        #timestamp
```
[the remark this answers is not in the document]
scgkiran: fixed it.

EpsilonPrime (approved):

jacques-n (approved):
Agree with @EpsilonPrime . We can have the coverage test include:
* generate the parser
* run the tests
So we should ultimately land with no need to check in generated files. (Their only purpose is to complete the tests.)

One other ticket we should think about: add CI checks that new functions have test coverage and test coverage shouldn't go down.

scgkiran:
>  We can have the coverage test include:
> 
> * generate the parser
> * run the tests
>   So we should ultimately land with no need to check in generated files. (Their only purpose is to complete the tests.)

I will address this in another PR.

> One other ticket we should think about: add CI checks that new functions have test coverage and test coverage shouldn't go down.

I have added a check [here](https://github.com/substrait-io/substrait/blob/a6ba11dc68f8f7bff732d89740f0263147f8f58e/tests/test_extensions.py#L32). This should catch if a new function is added without testcases. Once we have more tests checked in, we can improve the check to report functions with no coverage or less coverage.

PR #739, opened 2024-11-13
feat: update add_intervals functions with option for days in a year

scgkiran:


scgkiran:
This is used by BFT [tests](https://github.com/substrait-io/bft/blob/65351e5b173180a85ed8241febb6fb2a60dae616/cases/datetime/add_intervals.yaml#L31) and in [duckdb](https://github.com/substrait-io/bft/blob/65351e5b173180a85ed8241febb6fb2a60dae616/dialects/duckdb.yaml#L392) & [postgres](https://github.com/substrait-io/bft/blob/65351e5b173180a85ed8241febb6fb2a60dae616/dialects/postgres.yaml#L428) dialects

extensions/functions_datetime.yaml
```diff
@@ -374,6 +374,9 @@ scalar_functions:
             value: interval_year
           - name: y
             value: interval_year
+        options:
```
EpsilonPrime, line 377: Would these options also apply to any place an interval was being used?

extensions/functions_datetime.yaml
```diff
@@ -374,6 +374,9 @@ scalar_functions:
...
+          days_in_year:
+            values: [ YEAR_366DAYS, YEAR_365DAYS, YEAR_360DAYS ]
```
jacques-n, line 379: I think we should hold off on introducing this option without thinking more broadly about this.

Additionally, this concept (days in year) is irrelevant to interval_day and interval_year. Neither of those mix days and years. It is really only relevant when you bring in interval_compound.

PR #740, opened 2024-11-13
feat: add substrait test files to go embedded fs

scgkiran:


EpsilonPrime (commented):
So the purpose of this PR is to make the function compliance tests visible to the Go module for use by systems wanting to run the compliance checks elsewhere?

That's fine with me but I wonder if it should be a separate Go module so that folks only wanting the definitions can get to them separately.  It would be additionally nice if we could keep the non-definition files out of the root directory for the repository.

scgkiran:
> So the purpose of this PR is to make the function compliance tests visible to the Go module for use by systems wanting to run the compliance checks elsewhere?
> 

Yes

> That's fine with me but I wonder if it should be a separate Go module so that folks only wanting the definitions can get to them separately. It would be additionally nice if we could keep the non-definition files out of the root directory for the repository.

I felt its too small to make it a separate module. I moved tests to a new embedFS.

EpsilonPrime (approved):

jacques-n (approved):

PR #742, opened 2024-11-14
feat: misc fixes and cleanup to func testcase grammar

scgkiran:
- to support missing
 - to support zero argument aggregate function


grammar/Makefile
```diff
@@ -13,3 +13,4 @@ generate_type_parser:
 clean:
 	rm -rf $(TYPE_OUTPUT_DIR)/*.py  $(TYPE_OUTPUT_DIR)/*.tokens $(TYPE_OUTPUT_DIR)/*.interp
 	rm -rf $(TESTCASE_OUTPUT_DIR)/*.py $(TESTCASE_OUTPUT_DIR)/*.tokens $(TESTCASE_OUTPUT_DIR)/*.interp
+	rm -rf ./*.tokens
```
EpsilonPrime, line 16: Also consider deleting the interp files which are typically used by interpreters.

tests/coverage/coverage.py
```diff
@@ -62,12 +62,14 @@ def to_dict(self):
...
     file_coverage: dict[str, FileTestCoverage]
     test_count: int
+    num_bad_tests: int
     num_covered_function_variants: int
     total_function_variants: int
 
     def __init__(self, ext_uris):
         self.file_coverage = dict()
         self.test_count = 0
+        self.num_bad_tests = 0
```
EpsilonPrime, line 72: Is a bad test something that had a catastrophic failure or does it merely mean that the result is not what it should be?  Consider using another name.

grammar/Makefile
scgkiran: It is done in line 15. In the current folder it creates token file, not sure why it creates in two places.

tests/coverage/coverage.py
scgkiran: Fixed it

EpsilonPrime (approved):

jacques-n (approved):

PR #748, opened 2024-11-22
feat: add expression support for count and offset in the fetch operator

jcamachor:
* Update FetchRel in Substrait to support expressions for both offset and count. 
This is done in a way that should be effectively compatible with old systems. Old plans should be able to continue to be consumed treating the old values as int64.

BREAKING CHANGE: The encoding of FetchRel has changed in a strictly backwards incompatible way. The change involves transitioning offset and count from a standalone int64 field to a oneof structure, where the original int64 field is marked as deprecated, and a new field of Expression type is introduced. Using a oneof may cause ambiguity between unset and set-to-zero states in older messages. However, the fields are defined such that their logical meaning remains indistinguishable, ensuring consistency across encodings.

### DIFF
--- a/proto/substrait/algebra.proto
+++ b/proto/substrait/algebra.proto
@@ -273,6 +273,7 @@ message FetchRel {
     int64 offset = 3 [deprecated = true];
     // Expression evaluated into an integer specifying the number of records to
     // skip.
+    // Recommended type for offset is int64.
     Expression offset_expr = 5;
   }
   oneof count_mode {
@@ -282,6 +283,7 @@ message FetchRel {
     int64 count = 4 [deprecated = true];
     // Expression evaluated into an integer specifying the number of records to
     // return. -1 signals that all records should be returned.
+    // Recommended type for count is int64.
     Expression count_expr = 6;
   }
   substrait.extensions.AdvancedExtension advanced_extension = 10;
--- a/proto/substrait/algebra.proto
+++ b/proto/substrait/algebra.proto
@@ -269,26 +269,27 @@ message FetchRel {
   Rel input = 2;
   // Note: A oneof field is inherently optional, whereas individual fields
   // within a oneof cannot be marked as optional. The unset state of offset
-  // should therefore be checked at the oneof level.
+  // should therefore be checked at the oneof level. Unset is treated as 0.
   oneof offset_mode {
     // the offset expressed in number of records
     // Deprecated: use `offset_expr` instead
     int64 offset = 3 [deprecated = true];
     // Expression evaluated into an integer specifying the number of records to
-    // skip.
+    // skip. An expression evaluating to null is treated as 0.
     // Recommended type for offset is int64.
     Expression offset_expr = 5;
   }
   // Note: A oneof field is inherently optional, whereas individual fields
   // within a oneof cannot be marked as optional. The unset state of count
-  // should therefore be checked at the oneof level.
+  // should therefore be checked at the oneof level. Unset is treated as ALL.
   oneof count_mode {
     // the amount of records to return
     // use -1 to signal that ALL records should be returned
     // Deprecated: use `count_expr` instead
     int64 count = 4 [deprecated = true];
     // Expression evaluated into an integer specifying the number of records to
-    // return. -1 signals that all records should be returned.
+    // return. An expression evaluating to null signals that ALL records should
+    // be returned.
     // Recommended type for count is int64.
     Expression count_expr = 6;
   }
--- a/proto/substrait/algebra.proto
+++ b/proto/substrait/algebra.proto
@@ -267,6 +267,9 @@ message CrossRel {
 message FetchRel {
   RelCommon common = 1;
   Rel input = 2;
+  // Note: A oneof field is inherently optional, whereas individual fields
+  // within a oneof cannot be marked as optional. The unset state of offset
+  // should therefore be checked at the oneof level.
   oneof offset_mode {
     // the offset expressed in number of records
     // Deprecated: use `offset_expr` instead
@@ -276,6 +279,9 @@ message FetchRel {
     // Recommended type for offset is int64.
     Expression offset_expr = 5;
   }
+  // Note: A oneof field is inherently optional, whereas individual fields
+  // within a oneof cannot be marked as optional. The unset state of count
+  // should therefore be checked at the oneof level.
   oneof count_mode {
     // the amount of records to return
     // use -1 to signal that ALL records should be returned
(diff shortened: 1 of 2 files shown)


site/docs/relations/logical_relations.md
```diff
@@ -338,11 +338,11 @@ The fetch operation eliminates records outside a desired window. Typically corre
...
+| Property          | Description                                                                                                                                                            | Required                 |
+|-------------------|------------------------------------------------------------------------------------------------------------------------------------------------------------------------| ------------------------ |
+| Input             | A relational input, typically with a desired orderedness property.                                                                                                     | Required                 |
+| Offset Expression | An expression which evaluates to a non-negative integer. Declares the offset for retrieval of records.                                                                 | Optional, defaults to 0. |
+| Count Expression  | An expression which evaluates to a non-negative integer or -1. Declares the number of records that should be returned. -1 signals that ALL records should be returned. | Required |
```
ingomueller-net, line 345: Nit: pad last column with spaces?

proto/substrait/algebra.proto
```diff
@@ -267,11 +267,23 @@ message CrossRel {
...
-  // use -1 to signal that ALL records should be returned
-  int64 count = 4;
+  oneof offset_mode {
+    // the offset expressed in number of records
+    // Deprecated: use `offset_expr` instead
+    int64 offset = 3 [deprecated = true];
+    // Expression evaluated into an integer specifying the number of records to
+    // skip.
+    Expression offset_expr = 5;
+  }
```
ingomueller-net, line 277: I may not know protobuf well enough: is this a recommended way to evolve a message type? It seems like `FetchRel` does not have an `offset` field anymore -- isn't that more breakage than necessary (compared with *only* marking that field `decprecated`)?

site/docs/relations/logical_relations.md
jcamachor: Fixed (though it appears padding is not applied consistently throughout the documentation).

proto/substrait/algebra.proto
jcamachor: Through `oneof`, `offset` remains exposed as a top-level field, but it prevents both fields from being set simultaneously. Initially, I used the approach you proposed, but later switched to this method based on @jacques-n's recommendation in the mailing list discussion, which also appears to align with the process followed in similar changes, e.g., #665.

proto/substrait/algebra.proto
```diff
@@ -267,11 +267,23 @@ message CrossRel {
...
-  // the amount of records to return
-  // use -1 to signal that ALL records should be returned
-  int64 count = 4;
+  oneof offset_mode {
+    // the offset expressed in number of records
+    // Deprecated: use `offset_expr` instead
+    int64 offset = 3 [deprecated = true];
+    // Expression evaluated into an integer specifying the number of records to
+    // skip.
+    Expression offset_expr = 5;
```
EpsilonPrime, line 276: Based on the discussion at the meeting this week we should likely have some sort of recommendation for the expected type here.  Obviously if a consumer can't handle the type (it comes back as a string) it should not accept the plan but we should also recommend a type.

jcamachor: Thanks, @EpsilonPrime . Just a bit of guidance since I didn't join the meeting: Should the recommendation be added here as a comment? Should it also go into the spec documentation? And should we proceed with `i64` as the recommendation, given that it was originally used for `offset` and `count`?

jcamachor: I have made the corresponding changes and pushed a new commit.

EpsilonPrime (dismissed):

proto/substrait/algebra.proto
```diff
@@ -267,11 +267,25 @@ message CrossRel {
...
+    // Expression evaluated into an integer specifying the number of records to
+    // skip.
+    // Recommended type for offset is int64.
+    Expression offset_expr = 5;
+  }
+  oneof count_mode {
+    // the amount of records to return
+    // use -1 to signal that ALL records should be returned
+    // Deprecated: use `count_expr` instead
+    int64 count = 4 [deprecated = true];
```
jacques-n, line 283: As I've thought about this more, I don't like that the "not set" state is now inside an expression (e.g. -1 and 0).

I suggest we add another branch to each of these that is 

`google.protobuf.Empty no_count = 7`

and

`google.protobuf.Empty no_offset = 6`

or similar. My other thought is we could just keep the existing fields as not-deprecated (oneof is sufficient) and clearly state that the -1 behavior isn't available to count_expr.

proto/substrait/algebra.proto
jacques-n: I think the update you made is sufficient for this patch. We can address the broader issue in that other ticket.

proto/substrait/algebra.proto
EpsilonPrime: How about the expression returns a nullable int64?  Presumably the expression needs to have the same expressibility as the non-expression forms.

That wouldn't change the not set behavior for count though.

jcamachor: @jacques-n , I was looking into how `oneof` works in protobuf and I think your concern might already be addressed in the current PR. The unset state is handled at the `oneof` level, i.e., a `oneof` field is always optional so you can just check at the top level whether it's set. So based on that, it seems that `no_count`/`no_offset` would be redundant, since when oneof is unset, it already kind of works like `no_count`/`no_offset`. Regarding the other question, a field within `oneof` cannot be optional (I assume this is what you meant by nullable @EpsilonPrime?, protobuf doesn't allow an optional modifier for `oneof` fields.

jacques-n: > would be redundant

Good point. Maybe add a comment in docs clarifying that this is the expected pattern. It's a bit non-obvious (especially given how non-oneof int fields work)

EpsilonPrime: Currently the guidance for what type count_expr returns is int64.  Also -1 means returns all records.  If we changed the guidance for the count_expr rexpression to return a nullable int64 we could have null mean no filter/count limit.

jcamachor: @EpsilonPrime , I have updated the documentation. Please let me know if this is what you had in mind.

EpsilonPrime (dismissed):

westonpace (dismissed):
Some pedantic nits but looks good otherwise!

jcamachor:
@EpsilonPrime , @westonpace , could you please re-cast your approvals so we can merge? Thanks

westonpace (approved):

EpsilonPrime (approved):

PR #750, opened 2024-11-28
feat: support missing variants for regexp string functions

srikrishnak:


tests/cases/aggregate_approx/approx_count_distinct.test
```diff
@@ -7,8 +7,8 @@ approx_count_distinct((-32767, -20000, 30000, 5, 32767)::i16) = 5::i64
 approx_count_distinct((-2147483648, -10000000, 30000000, 2147483647)::i32) = 4::i64
 approx_count_distinct((-214748364800000, -1000000000, 0, 922337203685477580)::i64) = 4::i64
 approx_count_distinct((1)::i8) = 1::i64
-approx_count_distinct(('abc', 'def', 'ghi')::str) = 3::i64
-approx_count_distinct(('abc', Null, 'ghi')::str) = 2::i64
+approx_count_distinct(('abc', 'def', 'ghi')::str) = 2::i64
```
jacques-n, line 10: these changes look wrong.

extensions/functions_string.yaml
```diff
@@ -197,6 +197,19 @@ scalar_functions:
           dotall:
             values: [ DOTALL_DISABLED, DOTALL_ENABLED ]
         return: "string"
+      - args:
```
jacques-n, line 200: this addition doesn't match the description. do in separate definition with separate description.

extensions/functions_string.yaml
```diff
@@ -778,6 +791,19 @@ scalar_functions:
           dotall:
             values: [ DOTALL_DISABLED, DOTALL_ENABLED ]
         return: i64
+      - args:
```
jacques-n, line 794: as above, different args should be in separate described impl.

extensions/functions_string.yaml
```diff
@@ -1198,6 +1224,24 @@ scalar_functions:
           dotall:
             values: [ DOTALL_DISABLED, DOTALL_ENABLED ]
         return: "varchar<L1>"
+      - args:
```
jacques-n, line 1227: as above.

tests/cases/string/regexp_match_substring.test
```diff
@@ -2,34 +2,34 @@
 ### SUBSTRAIT_INCLUDE: '/extensions/functions_string.yaml'
 
 # basic: Basic examples without any special cases
-regexp_match_substring('foobarboopzoo'::str, 'o{1,}'::str, 1::i64, 1::i64, 0::i64) = 'oo'::str
```
jacques-n, line 5: looks like you're removing some function tests and adding others. Why not keep existing and add new ones?

tests/cases/aggregate_approx/approx_count_distinct.test
srikrishnak: understood! I will remove these failing tests for now instead.

tests/cases/string/regexp_match_substring.test
srikrishnak: understood! will update.

extensions/functions_string.yaml
```diff
@@ -197,6 +197,36 @@ scalar_functions:
           dotall:
             values: [ DOTALL_DISABLED, DOTALL_ENABLED ]
         return: "string"
+  -
+    name: regexp_match_substring
```
EpsilonPrime, line 201: I believe having the same function name twice will cause an issue in the Java client.  Should this alternative substring function have a different name?

srikrishnak: @EpsilonPrime 
There is an existing case for [count](https://github.com/substrait-io/substrait/blob/1890e6a7814c5161f38a31aba3e284dde1bc79d4/extensions/functions_aggregate_generic.yaml#L4-L26) with same function name twice. I was using that as reference. 

Shall I change that as well?

EpsilonPrime: as discussed yesterday it's okay to keep them with duplicate names in the same file (this is like the files used to be)

EpsilonPrime (approved):

srikrishnak:
Hi @jacques-n , Can you please review this PR. It needs second approval. Thanks

jacques-n (approved):

PR #756, opened 2024-12-17
fix: change the type of the bool_and testcase to aggregate

EpsilonPrime:


tests/cases/boolean/bool_and.test
```diff
@@ -1,4 +1,4 @@
-### SUBSTRAIT_SCALAR_TEST: v1.0
+### SUBSTRAIT_AGGREGATE_TEST: v1.0
```
scgkiran, line 1: `bool.and` is a scalar function with variadic args. We should fix the tests.

EpsilonPrime: This is bool.bool_and:  https://github.com/substrait-io/substrait/blob/main/extensions/functions_boolean.yaml#L118

scgkiran: oh I missed the _ part.

scgkiran (approved):
LGTM

jacques-n (approved):

PR #757, opened 2024-12-19
fix: update the doc to clarify that function names are case-sensitive

scgkiran:


EpsilonPrime (approved):

jacques-n (approved):
thanks for update. +1 for me.

@westonpace , @vbarua , @EpsilonPrime can you confirm.  

Note the only actual change here is being explicit that the name of a function in substrait definitions is case sensitive. This was always intended but not explicit. It is to support the goal of keeping things simple for consumers. (They don't need to canonicalize or something to match.)

vbarua (approved):
It's good to clarify this

PR #758, opened 2024-12-19
feat: introduce Iceberg table type using metadata file

jacques-n:
Adds Iceberg table type and first sub-variety, reading manifest files directly.

### DIFF
--- a/proto/substrait/algebra.proto
+++ b/proto/substrait/algebra.proto
@@ -126,7 +126,6 @@ message ReadRel {
 
   // Read an Iceberg Table
   message IcebergTable {
-
     oneof table_type {
       MetadataFileRead direct = 1;
       // future: add catalog table types (e.g. rest api, latest metadata in path, etc)
@@ -141,8 +140,6 @@ message ReadRel {
       optional string snapshot_id = 2;
     }
   }
-
-
 
   // A table composed of expressions.
   message VirtualTable {
--- a/proto/substrait/algebra.proto
+++ b/proto/substrait/algebra.proto
@@ -114,6 +114,7 @@ message ReadRel {
     LocalFiles local_files = 6;
     NamedTable named_table = 7;
     ExtensionTable extension_table = 8;
+    IcebergTable iceberg_table = 9;
   }
 
   // A base table. The list of string is used to represent namespacing (e.g., mydb.mytable).
@@ -122,6 +123,26 @@ message ReadRel {
     repeated string names = 1;
     substrait.extensions.AdvancedExtension advanced_extension = 10;
   }
+
+  // Read an Iceberg Table
+  message IcebergTable {
+
+    oneof table_type {
+      MetadataFileRead direct = 1;
+      // future: add catalog table types (e.g. rest api, latest metadata in path, etc)
+    }
+
+    // Read an Iceberg table using a metadata file
+    message MetadataFileRead {
+      // the specific uri of a metadata file (e.g. s3://mybucket/mytable/<ver>-<uuid>.metadata.json)
+      string metadata_uri = 1;
+
+      // the snapshot id to read. if not set, uses the current snapshot listed in the metadata file
+      optional string snapshot_id = 2;
+    }
+  }
+
+
 
   // A table composed of expressions.
   message VirtualTable {
--- a/proto/substrait/algebra.proto
+++ b/proto/substrait/algebra.proto
@@ -141,8 +141,8 @@ message ReadRel {
         // the snapshot id to read.
         string snapshot_id = 2;
 
-        // the timestamp that should be used to select the snapshot
-        PrecisionTimestamp snapshot_timestamp = 3;
+        // the timestamp that should be used to select the snapshot (Time passed in microseconds since 1970-01-01 00:00:00.000000 in UTC)
+        int64 snapshot_timestamp = 3;
       }
     }
   }
--- a/proto/substrait/algebra.proto
+++ b/proto/substrait/algebra.proto
@@ -138,7 +138,7 @@ message ReadRel {
 
       // snapshot options. if none set, uses the current snapshot listed in the metadata file
       oneof snapshot {
-        // the snapshot id to read. 
+        // the snapshot id to read.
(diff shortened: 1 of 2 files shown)


proto/substrait/algebra.proto
```diff
@@ -123,6 +124,23 @@ message ReadRel {
...
+      // future: add catalog table types (e.g. rest api, latest metadata in path, etc)
+    }
+
+    // Read an Iceberg table using a metadata file
+    message MetadataFileRead {
+      // the specific uri of a metadata file (e.g. s3://mybucket/mytable/<ver>-<uuid>.metadata.json)
+      string metadata_uri = 1;
+
+      // the snapshot id to read. if not set, uses the current snapshot listed in the metadata file
+      optional string snapshot_id = 2;
```
rymurr, line 140: Typically we have the choice of specifying snapshot or timestamp. Is the intention that the engine will know enough to derive snapshot if given a timestamp

site/docs/relations/logical_relations.md
```diff
@@ -95,6 +95,19 @@ possible approach is that a chunk should only be read if the midpoint of the chu
 %%% proto.algebra.ReadRel %%%
     ```
 
+#### Iceberg Table Type
+
+A Iceberg Table is a table built on [Apache Iceberg](https://iceberg.apache.org/). Iceberg tables can be read by either directly reading a [metadata file](https://iceberg.apache.org/spec/#table-metadata) or by consulting a [catalog](https://iceberg.apache.org/concepts/catalog/). 
+
+##### Metadata File Reading
+
+Points to a [Iceberg metadata file](https://iceberg.apache.org/spec/#table-metadata) and uses that as a starting point for reading an Iceberg table. This is the simplest form of Iceberg table access but only works for reading.
```
rymurr, line 104: just a nit: 

> but only works for reading.

There is nothing preventing writing from the manifest file its just not a good idea.

proto/substrait/algebra.proto
```diff
@@ -123,6 +124,23 @@ message ReadRel {
...
+  // Read an Iceberg Table
+  message IcebergTable {
+    oneof table_type {
+      MetadataFileRead direct = 1;
+      // future: add catalog table types (e.g. rest api, latest metadata in path, etc)
+    }
+
+    // Read an Iceberg table using a metadata file
+    message MetadataFileRead {
+      // the specific uri of a metadata file (e.g. s3://mybucket/mytable/<ver>-<uuid>.metadata.json)
```
rymurr, line 136: Is it worth calling out that substrait is explicitly not touching credentials? I don't think it should and want to be explicit about it.

site/docs/relations/logical_relations.md
westonpace, line 104:
```suggestion
Points to an [Iceberg metadata file](https://iceberg.apache.org/spec/#table-metadata) and uses that as a starting point for reading an Iceberg table. This is the simplest form of Iceberg table access but only works for reading.
```

site/docs/relations/logical_relations.md
```diff
@@ -95,6 +95,19 @@ possible approach is that a chunk should only be read if the midpoint of the chu
...
+
+| Property | Description                                                      | Required                |
+| -------- | ---------------------------------------------------------------- | ----------------------- |
+| metadata_uri    | A URI for a Iceberg metadata file. This current snapshot will be read from this file.  | Required |
```
westonpace, line 108:
```suggestion
| metadata_uri    | A URI for an Iceberg metadata file. This current snapshot will be read from this file.  | Required |
```

westonpace (dismissed):
This is fun to see.  Guess I should add a `LanceTable` soon :laughing:

proto/substrait/algebra.proto
jacques-n: added

site/docs/relations/logical_relations.md
jacques-n: updated.

proto/substrait/algebra.proto
jacques-n: updated to add snapshot timestamp as well.

site/docs/relations/logical_relations.md
jacques-n: done

site/docs/relations/logical_relations.md
jacques-n: done

jacques-n:
Ping @westonpace @vbarua @EpsilonPrime for approval. Would love to get this merged and pulled through to duckdb.

westonpace (approved):
@EpsilonPrime and I discussed this a little offline.  One thing that he noted was that Gluten had done similar work on adapting Substrait for Iceberg.  However, it appears their work was on adding relations and details needed to create "post-Iceberg, planned" queries (I doubt this is the correct phrasing).  Primarily this seems to be adding the concept of deletion files to the read relation (which I'm not sure is the way I would prefer those to be handled).

In other words, this relation (the new `read_type`) is very abstract and would represent the input to some kind of Iceberg-aware query engine or planner.  It would then get transformed into a query that any old query engine could evaluate by some kind of Iceberg planner (or just directly evaluated).

I'm not sure any action needs to be taken, I don't think anything in here is necessarily misleading or vague, I'm just summarizing the offline conversation.

EpsilonPrime (approved):
I've reviewed the differences between this implementation and Gluten's.  As @westonpace mentioned above, the Gluten format is more physical, having the capability of specifying files and portions of files to ignore.  Ignoring content could be evolved into a generic feature at some point.  This change doesn't prevent something like Gluten's iceberg solution to be included so this logical iceberg table type works for me.

One thing I didn't like about the Gluten implementation was the iceberg file format referencing other file formats.  When the physical Iceberg implementation is added I'd like to see it done as part of IcebergTable.

jacques-n:
Agree with @westonpace and @EpsilonPrime comments here. Gluten focuses on physical iceberg reading (post scan planning) so a very different thing than what is done here. Agree they are compatible and we should try to keep that in mind as we progress.

PR #760, opened 2025-01-06
docs: fix some usages of the word "dataset"

ingomueller-net:
This PR fixes several usages of the word "dataset". In one occasion, the word "transformation" was used when really the *result* of such a transformation was meant, for which the spec uses the word "dataset". In two other occasions, the word "dataset" was used when really the *transformation* that produces such a dataset was meant, for which the spec uses the words "transformation", "relation", "operation", or "operator."


mbrobbel (approved):

westonpace (approved):
Thanks for the keen eye for detail

vbarua (approved):
Thanks for the cleanup ✨ 🧹

PR #765, opened 2025-01-07
feat: run pytest in pr workflow to check function test coverage

scgkiran:


EpsilonPrime (dismissed):

.github/workflows/pr.yml
```diff
@@ -83,11 +83,14 @@ jobs:
     steps:
```
EpsilonPrime, line 83: Should we update the name of this section to make it clear that tests are being run as well?

scgkiran: Done

EpsilonPrime (approved):

jcamachor:
@scgkiran , any insights on why this failure might be occurring?
https://github.com/substrait-io/substrait/actions/runs/12662036941/job/35286264139?pr=766
It started happening after I rebased to include this PR and I cannot reproduce it locally. Does it have to do with some dependency version mismatch?

scgkiran:
>  Does it have to do with some dependency version mismatch?

@jcamachor can you try [this](https://github.com/substrait-io/substrait/pull/768/files) change?

jcamachor:
> > Does it have to do with some dependency version mismatch?
> 
> @jcamachor can you try [this](https://github.com/substrait-io/substrait/pull/768/files) change?

@scgkiran , thanks, that seems to fix the issue: https://github.com/substrait-io/substrait/actions/runs/12664721192/job/35293275122?pr=766

PR #766, opened 2025-01-07
feat: add advanced extension field to DdlRel, WriteRel, and UpdateRel

jcamachor:


jacques-n:
What are general thoughts on creating field here versus the one already available in RelCommon? RelCommon is kind of like an abstract superclass (if you will) except that protobuf makes you invert the construction.

jcamachor:
Thanks, @jacques-n . I was trying to follow the pattern used for other operators. My assumption was that both fields are included because it allows to have an extension implementation common to all operators (the one under RelCommon) as well as an extension specific to a given operator.

jacques-n:
> Thanks, @jacques-n . I was trying to follow the pattern used for other operators....

Fair. I think we never really figured it out and it shows.

jacques-n (approved):

jacques-n:
The one problem I struggle with is the in relation location means these are by default semantic changing so  a system receiving something at this position must understand or fail the query.

The benefit of using inside hint of relcommon is that an engine can ignore if it likes.

jacques-n:
> > in relation location means these are by default semantic changing
> 
> Isn't it the case that whether these extensions are semantic changing by default depends on whether we use the `optimization` field or the `enhancement` field within the `AdvancedExtension` to define the extension?

lol. yes.

jacques-n (approved):

proto/substrait/algebra.proto
```diff
@@ -649,6 +649,9 @@ message DdlRel {
     // An operation that removes an object from the system (without throwing an exception if the object did not exist)
     DDL_OP_DROP_IF_EXIST = 5;
   }
+
+  substrait.extensions.AdvancedExtension advanced_extension = 10;
```
vbarua, line 653: Does it make sense for this to just be 9 instead of 10? The last used field is 8.

proto/substrait/algebra.proto
```diff
@@ -710,6 +713,8 @@ message WriteRel {
     // subplans in the body of the Rel input) and return those with anounter PlanRel.relations.
     OUTPUT_MODE_MODIFIED_RECORDS = 2;
   }
+
+  substrait.extensions.AdvancedExtension advanced_extension = 10;
```
vbarua, line 717: Should we use 8 instead of 10 here?

proto/substrait/algebra.proto
```diff
@@ -728,6 +733,8 @@ message UpdateRel {
     Expression transformation = 1; // the transformation to apply
     int32 column_target = 2; // index of the column to apply the transformation to
   }
+
+  substrait.extensions.AdvancedExtension advanced_extension = 10;
```
vbarua, line 737: 5 instead of 10 here?

vbarua (commented):
Overall looks reasonable to me. This brings these inline with all the other rels which also have AdvancedExtensions available.

I did leave some comments about the field number we're using. I don't know if it makes sense to reserve fields as opposed to just using the next available one.

jcamachor:
> I don't know if it makes sense to reserve fields as opposed to just using the next available one.

Thanks, @vbarua . For this, I used the same convention that I observed in extensions defined in other ops. Maybe let's reach consensus with other committers before I change it?

jcamachor:
@vbarua , @jacques-n , how do you think we should proceed here? Thanks

proto/substrait/algebra.proto
jcamachor: 8 is used, I used 9

jcamachor:
> @vbarua , @jacques-n , how do you think we should proceed here? Thanks

I have updated the field numbers.

jacques-n:
I agree with @vbarua that we should just use next available.  I didn't notice that on my review. It looks like we're still using not-next?

jcamachor:
In the most recent commit, I changed it to use next available value.

jcamachor:
@jacques-n , @vbarua , all feedback has been addressed, could we merge this change? (I believe current check failure is unrelated to this change but I'm not super familiar with it.) Thanks

EpsilonPrime (approved):

jacques-n (approved):

PR #769, opened 2025-01-09
build: improve automation of ANTLR parser generation

vbarua:

PR #772, opened 2025-01-15
fix:  Make new comparison tests parseable

EpsilonPrime:


tests/cases/comparison/is_false.test
```diff
@@ -1,6 +1,7 @@
 ### SUBSTRAIT_SCALAR_TEST: v1.0
 ### SUBSTRAIT_INCLUDE: '/extensions/functions_comparison.yaml'
 
+# complete
```
jacques-n, line 4: Does our grammar require description? Did we make that deliberately make that decision? (can't remember)

scgkiran: This is test group name/description. should it be optional?

EpsilonPrime: It is not optional, hence the need to fix the test.

scgkiran: I was proposing if we should make it optional?
1. all tests without group description, will assign a default name something like `default_<func>_tests`
2. OR add code to fail the pre-commit/PR workflow if there is no group description

vbarua: I tried making it optional in https://github.com/substrait-io/substrait/pull/769 but according to David there are issues downstream with that.

vbarua: Something to add, as part of those test changes I updated the Python parser to treat a missing description as an empty string.

I don't have strong opinions either way about whether the descriptions should be optional. The current ambiguity stems from the following.

In the text spec, they are optional
https://github.com/substrait-io/substrait/blob/7e3c1394fe6c5a0cce1c1afcc6db9731db2dfc52/tests/README.md?plain=1#L84-L86
based on `(<test_group>)?`

whereast in the ANTLR grammer they are required
https://github.com/substrait-io/substrait/blob/7e3c1394fe6c5a0cce1c1afcc6db9731db2dfc52/grammar/FuncTestCaseParser.g4#L33-L36

Once we decide on the behaviour we can fix whichever one we need.

EpsilonPrime:
fixed elsewhere

PR #777, opened 2025-01-22
feat: support picoseconds in precisionTimestamp and precistionTimestampTZ

abbas-gadhia:
feat: PRECISIONTIMESTAMP<P> and PRECISIONTIMESTAMPTZ<P> now support upto 12 precision ie picoseconds


abbas-gadhia:
I wish to add an additional test for the extract function, but it seems that the existing test case does not conform to the specification 
for ex. `extract('MICROSECONDS'::str, '2016-12-31T13:30:15.220000'::ts) = 15220000::i64`

As per the specification, `MICROSECOND Return number of microseconds since the last full millisecond.` The current test case seems to be more aligned with the behavior of PostgreSQL i think

vbarua:
Trino is an example of a sytem that supports picosecond precision 
https://trino.io/docs/current/language/types.html#time-p

EpsilonPrime:
> I wish to add an additional test for the extract function, but it seems that the existing test case does not conform to the specification for ex. `extract('MICROSECONDS'::str, '2016-12-31T13:30:15.220000'::ts) = 15220000::i64`
> 
> As per the specification, `MICROSECOND Return number of microseconds since the last full millisecond.` The current test case seems to be more aligned with the behavior of PostgreSQL i think

Based on the discussion in the Substrait Community meeting today the tests should be updated.  I'm hoping that the docs reflect the primary use case.  For the secondary behavior we can add another function with the different semantics.

EpsilonPrime (approved):
I'm okay fixing the tests separately (since they aren't directly related to this picosecond change).

vbarua (approved):
LGMT! Thanks for adding this.

abbas-gadhia:
Thanks for approving :)

I'll open another PR for fixing the tests after some investigations into how other engines behave

PR #780, opened 2025-01-24
feat: add dynamic parameter expression

jcamachor:
feat: This PR adds a dynamic parameter expression to Substrait.

A dynamic parameter expression represents a placeholder within an expression whose value is determined at runtime.
This is particularly useful for parameterized queries where certain values are not known until execution.
Additionally, using dynamic parameters can enable other use cases, such as sharing execution plans without embedding sensitive information.


proto/substrait/algebra.proto
```diff
@@ -1604,6 +1605,16 @@ message Expression {
   }
 }
 
+// Expression that represents a dynamic parameter.
+// Dynamic parameters are identified by their 0-based index within a plan.
+message DynamicParameter {
+  // The inferred type of the dynamic parameter.
```
EpsilonPrime, line 1611: Given that Substrait is fully typed I believe we can require that the type be an exact match.  From context it feels like that an untyped string would be coerced to this type but we can assume that is handled elsewhere.  Thus the provided additional input for a dynamic plan would merely be a list of types.

A follow on to having a list of types is where would one specify that additional info.  We could shove it into the plan's root (to fully qualify a dynamic plan) or we could provide a separate datastructure that would be supplied in addition to the plan.  I'm more inclined to keep them together although that would mean the plan would need to be modified for each use which could be a disadvantage somewhere.

jcamachor: Thanks for the feedback, @EpsilonPrime !

Regarding your first point, I agree--typically, the type should act as validation and align with the type of expression replacing the dynamic parameter (perhaps the comment in the proto should state 'type' of the dynamic parameter rather than 'inferred type' to avoid confusion). As for the additional input, it would be a list of expressions rather than types (was that what you were referring to?). For example, this could be a list of Substrait literal expressions to populate the dynamic parameters in the plan.

On your second point, our team has been discussing the same idea. We felt it deserved a longer discussion, so we didn't include any related changes/proposal in this PR. There are trade-offs and maybe providing flexibility is the way to go? We could introduce an _optional_ field that allows it to be supplied within the plan, which would give the possibility to keep it self-contained but also the option to supply it as a separate message alongside the plan.

proto/substrait/algebra.proto
```diff
@@ -1604,6 +1605,16 @@ message Expression {
...
+  Type type = 1;
+
+  // The 0-based index of the dynamic parameter.
+  int32 index = 2;
```
vbarua, line 1615: What do we think about keying this using strings as opposed to integers? If we have a query with lots of dynamic parameters, it's more readable/debuggable (IMO) if we can reference them using case-sensitive names like `customer_id`, `order_id`, `supplier_id` instead of `0`, `1`, `2`

proto/substrait/algebra.proto
EpsilonPrime: Agreed, it should be a list of literals (not just types) at least.  I'm less inclined to include expressions since that'll make it harder for the consumer to just plug the value in.

In the Substrait Community meeting today we felt that the method of resolving the dynamic parameters should be included with this PR.  I do think that the list of literals that would need to be optional -- there is value in having a plan with dynamic parameters without the values filled in yet (and there are plans which won't have any dynamic parameters too).  I'm inclined to stuff them in the plan since that means all of the existing ways of sharing Substrait plans won't need to be updated in order to fill in the values but as you point out we can revisit that later on since the field would be optional (and if we don't stuff it into a submessage the list itself is implemented as repeated in protobuf which is essentially optional).

proto/substrait/algebra.proto
EpsilonPrime: The discussion in the community meeting leaned towards using anchors (no names) as we do elsewhere in the project.

jcamachor: I have updated the code and comments to make it consistent with other anchors, please take a look. Thanks

proto/substrait/algebra.proto
jcamachor: I have updated the code and added documentation according to the feedback, please take a look. Thanks

proto/substrait/plan.proto
```diff
@@ -46,6 +46,10 @@ message Plan {
   // unused. In many cases, a consumer may be able to work with a plan even if
   // one or more message types defined here are unknown.
   repeated string expected_type_urls = 5;
+
+  // An optional list of bindings for dynamic parameters used in this plan.
+  // Each binding maps a parameter_anchor to its corresponding runtime value.
+  repeated ParameterBinding parameter_bindings = 7;
```
EpsilonPrime, line 52: If this was `repeated Expression.Literal parameter_bindings = 7;` then the anchor could just be the parameter's position in the list.

jcamachor: Yes, I had it that way originally, then I changed it to make it consistent with other anchors such as the functions. I'll change it back.

EpsilonPrime (approved):

proto/substrait/plan.proto
```diff
@@ -46,6 +46,11 @@ message Plan {
   // unused. In many cases, a consumer may be able to work with a plan even if
   // one or more message types defined here are unknown.
   repeated string expected_type_urls = 5;
+
+  // An optional list of bindings for dynamic parameters used in this plan.
+  // The index of each entry in the list corresponds to the parameter_anchor
+  // of a DynamicParameter expression, mapping it to its runtime value.
+  repeated Expression.Literal parameter_bindings = 7;
```
jacques-n, line 53: My inclination is to match the SimpleExtensionURI pattern here. In that pattern, the anchors are in a message and this repeated element doesn't necessarily have to be in order. This means numbers are more explicit and we have a message to add other properties in the future. I regret using the implicit pattern in ReferenceRel and think we should probably avoid further impls of that pattern. Thoughts?

EpsilonPrime: I'm okay with that but it is half the work of names without the strings.  You will have to index the entries here so you can refer to them later just like you would need to do with strings.  If we do take that route we could consider using a map here so that at least the indexing would be done by the proto library instead of everyone needing to roll their own.

jcamachor: One option is to use the SimpleExtensionURI pattern since that's what is already in place and existing libraries can leverage the same existing deserialization logic built for it (that's what I did in this PR originally). If we want to handle things differently (i.e., dictionary-based), it could be worth having a broader discussion, including migrating other existing fields? That way, we avoid introducing yet another approach (i.e., explicit integer anchor, implicit list positional anchor, and now a dictionary-based anchor). That said, I'm fine with either approach, let me know and I'll adapt the PR accordingly.

jacques-n: I'm inclined to match SimpleExtensionURI. I don't think we should use map here. Part of my sensitivity is that in many packages we do things like rewrites of these structures. Having the ID outside the structure is frustrating. The same thing happens in rest apis. Some people are inclined to say "hey, the url has the id, we don't need it in the object". The reality is that having it in the object fits most languages much better since odds are that is how they will be manipulated.

jcamachor: OK, I have updated the code accordingly.

proto/substrait/algebra.proto
```diff
@@ -1611,6 +1612,16 @@ message Expression {
...
 }
 
+// Expression that represents a dynamic parameter.
+// Dynamic parameters are identified by a surrogate key within a plan.
+message DynamicParameter {
+  // The type of the dynamic parameter.
+  Type type = 1;
+
+  // The surrogate key used within a plan to reference a specific parameter binding.
+  uint32 parameter_anchor = 2;
```
jacques-n, line 1622: I believe we call these references. They are references to the anchors. (e.g. function_reference refers to function_anchor).

proto/substrait/plan.proto
```diff
@@ -80,3 +84,12 @@ message Version {
   // created the plan.
   string producer = 5;
 }
+
+// Represents a binding for a dynamic parameter.
+message ParameterBinding {
+  // The parameter anchor that identifies the dynamic parameter reference.
+  uint32 parameter_anchor = 1;
+
+  // The literal value assigned to the parameter at runtime.
```
jacques-n, line 93: Maybe add a comment that the type of this is only valid if it matches the type declared in the DynamicParameter

proto/substrait/plan.proto
```diff
@@ -80,3 +84,12 @@ message Version {
   // created the plan.
   string producer = 5;
 }
+
+// Represents a binding for a dynamic parameter.
+message ParameterBinding {
```
jacques-n, line 89: Maybe call DynamicParameterBinding to be explicit about which parameter.

jacques-n (approved):
LGTM. Thanks for the iteration patience @jcamachor !

jacques-n:
@vbarua @EpsilonPrime @westonpace thoughts?

jcamachor:
@vbarua , @EpsilonPrime , @westonpace , I'd appreciate your thoughts! Thanks

EpsilonPrime (approved):

jcamachor:
Thanks, @jacques-n , @EpsilonPrime , @vbarua !

Issue #782, opened 2025-01-31
Support ROLLUP and CUBE grouping expansion


yongchul:
# Motivation

ROLLUP and CUBE are established way to specify multiple groupings. In many databases, ROLLUP and CUBE can appear as a grouping expression so that one does not have to tediously expand all the combinations of the grouping expressions (see the survey of state of RDBMS).
 In Substrait, Aggregate rel supports grouping sets. ROLLUP and CUBE can be spelled out by expanding all possible combinations in the plan. The only problem is that CUBE will expand to exponential number of groupings, and sending the fully expanded form is not compact and not aligned well with other parts of Substrait, where compact representation is preferred (e.g., anchor and referring by indexes). Also, fully expanded groupings are hard to grasp the actual operations, thus not helping readability.

The proposal is to support ROLLUP and CUBE as an expansion method of Grouping when specified. The default is point, thus backward compatible, and rollup and cube will specify how to expand the specified sequence of grouping expression. This not only prevents exponential expansions and is easier to understand the higher-level semantic.

# State of the Art

Following databases (but not limited to) all support ROLLUP and CUBE as valid expansions of grouping.

* [DB2](https://www.ibm.com/docs/en/db2/12.1?topic=subselect-group-by-clause)
* [Oracle](https://docs.oracle.com/cd/E49933_01/server.770/es_eql/src/reql_aggregation_grouping_sets.html)
* [PostgreSQL](https://www.postgresql.org/docs/current/queries-table-expressions.html#QUERIES-GROUPING-SETS)
* [SQL Server](https://learn.microsoft.com/en-us/sql/t-sql/queries/select-group-by-transact-sql?view=sql-server-ver16#syntax)
* [ZetaSQL](https://github.com/google/zetasql/blob/master/docs/query-syntax.md#group_by_grouping_sets)

# Proposed Change

The proposed changes are two fold.

1. Add an optional grouping expansion enum field to `Grouping` to describe the type of expansion. If omitted, compatible to prior semantic.
2. Add repeated field to specify grouping of expressions (e.g., `CUBE( (a, b), c, (d, e))` where `(a, b)` and `(d, e)` are treated as a unit in expansion).

```protobuf
enum GroupingExpansion {
  // Default.
  Unspecified = 0;

  // The N grouped expressions are expanded to N + 1 groupings.
  ROLLUP = 1;
  // The N grouped expressions are expanded to 2^N  groupings.
  CUBE = 2;
}

message Grouping {
[...]

PR #783, opened 2025-02-01
feat: add ROLLUP, CUBE grouping expansion

yongchul:
feat: add ROLLUP, CUBE grouping expansion

Adds ROLLUP and CUBE grouping expansions to `AggregateRel`. See #782 for rationale.

* Adds `AggregateRel.GroupingExpansion` enum to denote expansion type.
* Adds `expansion` field to `AggregateRel.Grouping` to specify expansion type in a grouping set.
* Adds `composite_grounping_expressions` field to `AggregateRel.Grouping` to support composite grouping expression.

Refs: #782

### DIFF
--- a/proto/substrait/algebra.proto
+++ b/proto/substrait/algebra.proto
@@ -353,6 +353,31 @@ message AggregateRel {
     // A list of zero or more references to grouping expressions, i.e., indices
     // into the `grouping_expression` list.
     repeated uint32 expression_references = 2;
+
+    // An optional list of zero or more grouping id of `expression_references`
+    // at corresponding index. The expressions sharing the same non-zero values
+    // in this field formulates a composite grouping expression (i.e., those
+    // columns considered as one grouping expression).
+    // 
+    // For GROUNPING_EXPANSION_UNSPECIFIED, this field is ignored.
+    //
+    // Example:
+    // {
+    //   expression_references: [ 3, 1, 4, 5, 2 ],
+    //   composite_grouping_expressions: [ 1, 1, 0, 2, 2 ],
+    //   expansion: GROUPING_EXPANSION_CUBE
+    // },
+    //
+    // means CUBE((3, 1), 4, (5, 2)), thus expands to following grouping sets:
+    //
+    // (),
+    // (3, 1), (4), (5, 2),
+    // (3, 1, 4), (3, 1, 5, 2), (4, 5, 2),
+    // (3, 1, 4, 5, 2)
+    repeated uint32 composite_grouping_expressions = 3;
+
+    // Optional grouping expansion method. Default: no expansion.
+    GroupingExpansion expansion = 4;
   }
 
   message Measure {
@@ -363,6 +388,17 @@ message AggregateRel {
     // within the measure.
     // Helps to support SUM(<c>) FILTER(WHERE...) syntax without masking opportunities for optimization
     Expression filter = 2;
+  }
+
+  enum GroupingExpansion {
+    // No grouping expansion. Default.
+    GROUPING_EXPANSION_UNSPECIFIED = 0;
+
+    // N grouping expression groups are expanded into N+1 groupings.
+    GROUPING_EXPANSION_ROLLUP = 1;
+
+    // N grouping expression groups are expanded into 2^N grounpings.
+    GROUPING_EXPANSION_CUBE = 2;
   }
 }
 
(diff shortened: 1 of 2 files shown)


proto/substrait/algebra.proto
```diff
@@ -353,6 +353,30 @@ message AggregateRel {
     // A list of zero or more references to grouping expressions, i.e., indices
     // into the `grouping_expression` list.
     repeated uint32 expression_references = 2;
+
+    // An optional list of zero or more grouping id of `expression_references`
+    // at corresponding index. The expressions sharing the same non-zero values
+    // in this field formulates a composite grouping expression (i.e., those
+    // columns considered as one grouping expression).
+    // For GROUNPING_EXPANSION_UNSPECIFIED, this field is ignored.
```
jcamachor, line 361: nit. Typo: GROUNPING_EXPANSION_UNSPECIFIED

site/docs/relations/logical_relations.md
```diff
@@ -381,10 +381,14 @@ Grouping sets can be used for finer-grained control over which records are folde
...
 
 If at least one grouping expression is present, the aggregation is allowed to not have any aggregate expressions. An aggregate relation is invalid if it would yield zero columns.
 
+Optionally, a grouping set can be specified with rollup or cube expansions rather than fully spelling out the grouping sets. When either rollup or cube is specified for a grouping set, permutations of grouping expressions will be generated accordingly. Given `n` grouping expressions for a grouping set, ROLLUP expands to n + 1 grouping sets. CUBE expands to 2^`n` grouping sets (i.e., all subsets of the `n` grouping expressions). Given `i`th grouping set with ROLLUP expansion, the indicator values are assigned from `i` (i.e., an empty grouping) to `i + n + 1` (i.e., the grouping as if no expansion was done). For CUBE, the indicator values are assigned from `i` (i.e., an empty grouping) to `i + 2^n` (i.e., the grouping as if no expanson was done) where the values are assigned as `n`-bit integer where the bit is set when a grouping expression at corresponding position is being grouped.
```
jacques-n, line 388: Can you add some concrete examples in a table or similar here?

proto/substrait/algebra.proto
```diff
@@ -364,6 +388,17 @@ message AggregateRel {
     // Helps to support SUM(<c>) FILTER(WHERE...) syntax without masking opportunities for optimization
     Expression filter = 2;
   }
+
+  enum GroupingExpansion {
+    // No grouping expansion. Default.
+    GROUPING_EXPANSION_UNSPECIFIED = 0;
```
jacques-n, line 394: I think in other locations NONE is specified separately from UNSPECIFICED. Can you confirm? If so, we should match that pattern here.

proto/substrait/algebra.proto
```diff
@@ -353,6 +353,30 @@ message AggregateRel {
...
+    //   expansion: GROUPING_EXPANSION_CUBE
+    // },
+    //
+    // means CUBE((3, 1), 4, (5, 2)), thus expands to following grouping sets:
+    //
+    // (),
+    // (3, 1), (4), (5, 2),
+    // (3, 1, 4), (3, 1, 5, 2), (4, 5, 2),
+    // (3, 1, 4, 5, 2)
+    repeated uint32 composite_grouping_expressions = 3;
```
jacques-n, line 376: I think it might be better to create an alternative representation rather than the double array representation. This feels a bit complicated. 

As part of that, it might be best if it also includes the indiciator value explicitly rather than the more complex pattern you describe below.

jacques-n: Not that this is proposal, not a demand. What are you thoughts?

yongchul: I was never happy with this representation but just put out here to get some feedback. :-) The reason why I chose to this was because of the abundant usage of index/offset in other places. If this feels complex to you, we should make it more explicit, perhaps, use `oneof` and more descriptive message to encode these albeit it will introduce another level of nesting.

I don't know whether we will like it but I will draft that. Let's see how we like it.

yongchul: @jacques-n BTW, could you elaborate `the indicator value` in your comment? You mean the indicator value used to bundle (i'll use different word as "group" is very overloaded here) the grouping columns (i.e., the values in the `composite_grouping_expressions`)? Or something else?

PR #796, opened 2025-03-20
feat: remove ordering guarantees from Hash Equijoin operator

yongchul:
BREAKING CHANGE: Hash Equijoin no longer preserves ordering for inner joins

The original `Property Maintenance` of hash join operator is following.

> Orderedness of the left set is maintained in INNER join cases, otherwise it is eliminated. 

This holds ONLY very specific implementation of a hash join, for instance, when build side input completely fits in memory, and probe side input is streamed in single thread. It is also strange why INNER JOIN is specifically called out because other joins can preserve order of probe (LEFT) when build (RIGHT) fits in memory.

Nonetheless, if you throw some kicks and chops, this order preserving claim quickly falls apart unless implementation does some non-trivial work under following scenarios.

* build does not fit in memory (i.e., spill to storage)
* parallel probe

So in general, we should not say hash join preserves order of probe. It *may* assuming a specific implementation under  particular conditions, which is more of optimization or hint territory.


site/docs/relations/physical_relations.md
```diff
@@ -12,8 +12,9 @@ The hash equijoin join operator will build a hash table out of the right input b
 | -------------------- | ------------------------------------------------------------ |
 | Inputs               | 2                                                            |
 | Outputs              | 1                                                            |
-| Property Maintenance | Distribution is maintained. Orderedness of the left set is maintained in INNER join cases, otherwise it is eliminated. |
-| Direct Output Order  | Same as the [Join](logical_relations.md#join-operation) operator. |
+| Property Maintenance | Distribution is maintained. Orderedness is eliminated.       |
```
vbarua, line 15: Looking at this now, does this actually maintain the distribution of the input? I would assume the output is now distributed based on the join keys, but that doesn't necessarily match the incoming distribution of either side?

yongchul: @vbarua Good question. Let's consider following scenarios.

1. If the build have been broadcasted, the probe distribution is maintained by definition. Note that in this case, probe distribution may not be along with join key as you described.
2. If the join is distributed join (i.e., build and probe are distributed in compatible way with respect to join key), both inputs are distributed in the same way, thus it does preserve probe distribution.
3. broadcasting probe does not make sense so exclude.

So in (1) and (2), probe distribution is preserved. (3) does not happen (or I am not aware of these and curious to learn if there is any and why).

There is a chance that some sort of dynamic or adaptive join implementation may destroy this property (e.g., switching build and probe with broadcast decision at runtime) but I believe those should be captured as a custom rel.

yongchul: @vbarua In fact, I realized a hindsight of distribution with respect to join type because outer joins can destroy distribution property due to NULL padding especially for the RIGHT and FULL OUTER. Thank you for bringing this up! And now I remember my past work as well.

For INNER and LEFT joins, we can retain `Probe` distribution.

For RIGHT (or build), you can't retain `Probe` distribution because the probe side may have NULL padded rows in the output and this will happen for every partition thus the distribution on `Probe` is broken. The preserved distribution is `Build` when it is **NOT** a broadcast join.

FULL OUTER join includes NULL padding on both build and probe thus can't derive any distribution. Well, technically, it is still distributed by `COALESCE(L.jk_i, R.jk_i)` (i.e., choosing the non-null value) but that's beyond the scope of `JoinRel`.

So below is the complete matrix.

| Join Type | Distribution |
| --- | --- |
| INNER       | Probe  |
| LEFT OUTER | Probe |
| LEFT SEMI | Probe |
| LEFT ANTI SEMI | Probe |
| RIGHT OUTER | Build if not broadcast. Otherwise, N/A |
| RIGHT SEMI | Build if not broadcast. Otherwise, N/A |
| RIGHT ANTISEMI | Build if not broadcast. Otherwise, N/A |
| FULL OUTER | N/A |

I'm not sure how best to express this in the document. Perhaps, I can break it into three: INNER, LEFT joins, RIGHT joins, and FULL OUTER joins...?

vbarua: I think we've stumbled into a bit of a can of words with regards to distribution. I also think we can leave that for its own change. I'm asking about this in the Slack and can file a followup issue once I gather some information.

vbarua (approved):
Removing the ordering guarantees from Hash Equijoin makes sense to me.

site/docs/relations/physical_relations.md
yongchul: +1 to handle it in a follow up PR. Yes, the properties should be actually with respect to the actual emit. We could probably define it as a general convention.

bvolpato (approved):
That makes sense to me as well. Thanks!

yongchul:
@jacques-n @cpcloud @westonpace @EpsilonPrime could you take a look? Thank you!

jacques-n (approved):
Agreed on weakening guarantee.

westonpace (approved):

EpsilonPrime (approved):

PR #799, opened 2025-04-01
feat: bitwise shift functions

andrew-coleman:
Proposed addition of the bitwise shift functions to the core arithmetic function yaml as supported by a number of SQL implementations. 
E.g. Spark - https://spark.apache.org/docs/latest/api/sql/index.html#shiftleft

### DIFF
--- a/tests/cases/arithmetic/shift_left.test
+++ b/tests/cases/arithmetic/shift_left.test
@@ -1,0 +1,14 @@
+### SUBSTRAIT_SCALAR_TEST: v1.0
+### SUBSTRAIT_INCLUDE: '/extensions/functions_arithmetic.yaml'
+
+# basic: Basic examples without any special cases
+shift_left(2::i32, 1::i32) = 4::i32
+shift_left(511::i32, 8::i32) = 130816::i32
+shift_left(301989888::i64, 3::i32) = 2415919104::i64
+shift_left(301989888::i64, 8::i32) = 19791209299968::i64
+shift_left(-3::i32, 1::i32) = -6::i32
+shift_left(-3::i32, 2::i32) = -12::i32
+shift_left(-3::i64, 1::i32) = -6::i64
+shift_left(-3::i64, 2::i32) = -12::i64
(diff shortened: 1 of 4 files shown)


tests/cases/arithmetic/shift_right_unsigned.test
```diff
@@ -0,0 +1,10 @@
+### SUBSTRAIT_SCALAR_TEST: v1.0
+### SUBSTRAIT_INCLUDE: '/extensions/functions_arithmetic.yaml'
+
+# basic: Basic examples without any special cases
+shift_right_unsigned(2::i32, 1::i32) = 1::i32
+shift_right_unsigned(1024::i32, 8::i32) = 4::i32
+shift_right_unsigned(301989888::i64, 3::i32) = 37748736::i64
```
EpsilonPrime, line 7: Should there be a test case that is different between shift right and shift right unsigned?

tests/cases/arithmetic/shift_left.test
EpsilonPrime, line 1: Should there be a shift_left_unsigned?

EpsilonPrime: If we had unsigned integer types (which we briefly discussed in yesterday's community meeting) then we'd probably need one.  For now since we are treating integers as storage (containing unsigned or signed integers) then this works.

EpsilonPrime (approved):

tests/cases/arithmetic/shift_right.test
```diff
@@ -0,0 +1,14 @@
+### SUBSTRAIT_SCALAR_TEST: v1.0
+### SUBSTRAIT_INCLUDE: '/extensions/functions_arithmetic.yaml'
+
+# basic: Basic examples without any special cases
```
bvolpato, line 4: nit I was wondering if we should cover shifts with 0 -- but this could apparently be a special case.

bvolpato (approved):
💯

andrew-coleman:
This has two approvals - is it possible to merge this?  Thanks :)

PR #801, opened 2025-04-11
build: upgrade buf to v2

vbarua:


vbarua:
Checking locally, the code that is generated in `gen/` remains unchanged after these changes.

vbarua:
Looking at https://github.com/substrait-io/substrait/blob/main/tools/proto_prefix.py is a real "Here Be Dragons" moment.

Original context for that script in https://github.com/substrait-io/substrait/pull/207

PR #803, opened 2025-04-15
feat: clarify behavior and direct output order of joins

yongchul:
feat: clarify behavior and direct output order of joins

BREAKING CHANGE: direct output order of semi/anti/mark joins no longer includes invalid side (i.e., right of lefty joins, and left of righty joins).
BREAKING CHANGE: single join raises runtime error if there more than one matching rows to adhere to the original proposed usage (unnesting scalar subqueries).

# Problems

1. Semi-joins and anti-joins are one-sided and fields from the other side of joins are not valid. The `Direct Output Order` in the document is okay for other types of joins but not in one-side joins.
2. Single joins are proposed to be used unnesting scalar subqueries where exactly 1 row is expected. The current documentation citing the paper and behavior although relaxed the behavior so that the implementation can silently produce wrong result.

# What this PR do?

1. Clarify the direct output order by explicitly stating the semi, anti, and mark joins. Introducing `Input Order` (the previous `Direct Output Order`) so that all the properties referencing `Input Order` to reduce ambiguity.
2. Single joins expecting at most one row for each join key. Otherwise, runtime error. This behavior can be extended in the future if such generalization is justified with correct use cases.


site/docs/relations/logical_relations.md
```diff
@@ -225,15 +225,16 @@ The join operation will combine two separate inputs into a single output, based
 | Inputs               | 2                                                            |
 | Outputs              | 1                                                            |
 | Property Maintenance | Distribution is maintained. Orderedness is empty post operation. Physical relations may provide better property maintenance. |
-| Direct Output Order  | The emit order of the left input followed by the emit order of the right input. |
+| Direct Input Order   | The direct input order is the left input followed by the emit order of the right input. All field references of [Join Properties](#join-properties) are over this order. |
```
tokoko, line 228: If you think this needs a clarification here, we should probably add the same for `CrossRel`. Also, I think calling it `Input Order` should be sufficient as there're not multiple ways for operators to get inputs.

yongchul: Done. Thank you for suggestions!

site/docs/relations/physical_relations.md
```diff
@@ -13,6 +13,7 @@ The hash equijoin join operator will build a hash table out of the right input b
 | Inputs               | 2                                                            |
 | Outputs              | 1                                                            |
 | Property Maintenance | Distribution is maintained. Orderedness is eliminated.       |
+| Direct Input Order   | Same as the [Join](logical_relations.md#join-operator) operator. |
```
jcamachor, line 16: Should these be changed `Direct Input Order` -> `Input Order` as well?

jcamachor (dismissed):

site/docs/relations/physical_relations.md
yongchul: Indeed! Done.

site/docs/relations/logical_relations.md
```diff
@@ -199,7 +199,8 @@ The cross product operation will combine two separate inputs into a single outpu
 | Inputs               | 2                                                            |
 | Outputs              | 1                                                            |
 | Property Maintenance | Distribution is maintained. Orderedness is empty post operation. |
-| Direct Output Order  | The emit order of the left input followed by the emit order of the right input. |
+| Input Order          | The input order is the left input followed by the emit order of the right input. All field references of [Cross Product Properties](#cross-product-properties) are over this order. |
```
westonpace, line 202:
```suggestion
| Input Order          | The input order is the left input followed by the the right input. All field references of [Cross Product Properties](#cross-product-properties) are over this order. |
```

Nit: minor inconsistency between left & right

Let's avoid `emit order` when defining `Input Order`.  Or we could say `emit order of the left input followed by emit order of the right input` but I think "emit has been applied" is implicit at this point so we can just leave it out.

site/docs/relations/logical_relations.md
```diff
@@ -225,15 +226,16 @@ The join operation will combine two separate inputs into a single output, based
 | Inputs               | 2                                                            |
 | Outputs              | 1                                                            |
 | Property Maintenance | Distribution is maintained. Orderedness is empty post operation. Physical relations may provide better property maintenance. |
-| Direct Output Order  | The emit order of the left input followed by the emit order of the right input. |
+| Input Order          | The input order is the left input followed by the emit order of the right input. All field references of [Join Properties](#join-properties) are over this order. |
```
westonpace, line 229:
```suggestion
| Input Order          | The input order is the left input followed by the right input. All field references of [Join Properties](#join-properties) are over this order. |
```

site/docs/relations/logical_relations.md
```diff
@@ -244,15 +246,15 @@ The join operation will combine two separate inputs into a single output, based
...
+| Right Anti | Same as *Left Anti* except that the right and left inputs are switched. |
+| Left Single | Return all records from the left input with no join expansion.  If there exists exactly one record from the right input matches the left, return the matching record from the right input. For any left records without matching right records, return the left record along with nulls for the right input. Otherwise, runtime error. Similar to a left outer join but only returns one right match at most. Useful for nested sub-queries where we need exactly one record in output (or throw exception).  See Section 3.2 of https://15721.courses.cs.cmu.edu/spring2018/papers/16-optimizer2/hyperjoins-btw2017.pdf for more information. |
```
westonpace, line 254:
```suggestion
| Left Single | Return all records from the left input with no join expansion.  If there exactly one record from the right input matches the left, return the matching record from the right input. For any left records without matching right records, return the left record along with nulls for the right input. If any left records match multiple right records, it is a runtime error. Similar to a left outer join but only returns one right match at most. Useful for nested sub-queries where we need exactly one record in output (or return an error).  See Section 3.2 of https://15721.courses.cs.cmu.edu/spring2018/papers/16-optimizer2/hyperjoins-btw2017.pdf for more information. |
```

Minor nits

westonpace (dismissed):
This looks good, I have some minor suggestions.  Input order confused me for a minute but I think I understand the problem now.

If there are multiple inputs, since we are defining them with terms like "left" and "right", then it is not clear which one is first (e.g. is it "left right" or "right left"?).  The purpose of input order is to describe this.

westonpace:
I'm not entirely sure I agree with the first breaking change.  I think that was always the intention and so this isn't technically breaking.

That being said, I agree the second change is a breaking change (albeit minor).  CC @pdet for second opinion.

For background, I think what happened was:

* The paper says single join returns an error if there are multiple matches
* During implementation, DuckDb originally chose to not return an error and instead return an arbitrary record (match sqlite behavior)
* Later, after this relation was added to Substrait, DuckDb decided to change that default to return an error (match postgres behavior)

So I believe this change is aligning the definition of single join to the default behavior in DuckDb (this is configurable in DuckDb but I don't think we need to make it configurable in Substrait) and thus it would be a welcome breaking change.

yongchul:
> This looks good, I have some minor suggestions. Input order confused me for a minute but I think I understand the problem now.
> 
> If there are multiple inputs, since we are defining them with terms like "left" and "right", then it is not clear which one is first (e.g. is it "left right" or "right left"?). The purpose of input order is to describe this.

Thank you for bringing this up. The main reason I spelled out "Input Order" is that all the other properties are defined over those input order and sometimes (for those one side joins) the output order can drift from the input order due to semantic of joins. By calling out input order (as you suggested left followed right), we can clarify what the properties can reference  explicitly and also clarify the output order per join type.

yongchul:
> I'm not entirely sure I agree with the first breaking change. I think that was always the intention and so this isn't technically breaking.

For the first one, I don't know. This can still break some implementations although we claim that this has been the intent and you were holding it wrong. :) Happy to remove the first one if other members agree.

EpsilonPrime (approved):

westonpace (approved):

jcamachor (approved):

jcamachor:
@EpsilonPrime , @westonpace , could we merge this one? It seems it has the required approvals? Thanks

PR #807, opened 2025-04-23
feat: clarify the expected behavior, and rationale, of the post join filter

westonpace:
The post join filter has very little explanation.  It can also be confusing because, from a purely logical perspective, it is possible to see the post join filter as redundant.  This PR attempts to clarify the description of the post join filter.

- - -
This change is [Reviewable](https://reviewable.io/reviews/substrait-io/substrait/807)

### DIFF
--- a/proto/substrait/algebra.proto
+++ b/proto/substrait/algebra.proto
@@ -255,6 +255,13 @@ message JoinRel {
   Rel left = 2;
   Rel right = 3;
   Expression expression = 4;
+  // The post-join filter is a filter that is applied to the result of the join before an output
+  // record is produced.  If the filter evaluates to false then the record is not considered a
+  // match.
+  //
+  // For purely logical plans, this field is redundant, as the post_join_filter could be included in
+  // `expression`.  However, for backwards compatibility, and potentially to support engines that do
+  // not support highly logical plans, this field is provided.
   Expression post_join_filter = 5;
 
   JoinType type = 6;
@@ -815,6 +822,9 @@ message HashJoinRel {
   // hash function for a given comparsion function or to reject the plan if it cannot
   // do so.
   repeated ComparisonJoinKey keys = 8;
+  // A boolean condition to be applied to each potential match between the left and right
+  // inputs. If it evaluates to false then the potential match is not considered a match.
+  // This allows the HashJoinRel to express more complex join conditions than just equality.
   Expression post_join_filter = 6;
 
   JoinType type = 7;
@@ -865,6 +875,9 @@ message MergeJoinRel {
   // is free to do so as well).  If possible, the consumer should verify the sort
   // order and reject invalid plans.
   repeated ComparisonJoinKey keys = 8;
+  // A boolean condition to be applied to each potential match between the left and right
+  // inputs. If it evaluates to false then the potential match is not considered a match.
+  // This allows the MergeJoinRel to express more complex join conditions than just equality.
   Expression post_join_filter = 6;
 
   JoinType type = 7;
--- a/site/docs/relations/logical_relations.md
+++ b/site/docs/relations/logical_relations.md
@@ -234,7 +234,9 @@ The join operation will combine two separate inputs into a single output, based 
 | Left Input       | A relational input.                                          | Required                           |
 | Right Input      | A relational input.                                          | Required                           |
 | Join Expression  | A boolean condition that describes whether each record from the left set "match" the record from the right set. Field references correspond to the direct output order of the data. | Required. Can be the literal True. |
-| Post-Join Filter | A boolean condition to be applied to each result record after the inputs have been joined, yielding only the records that satisfied the condition. | Optional                           |
+| Post-Join Filter | A boolean condition to be applied to each potential match between the left and right
+inputs.  If it evaluates to false then the potential match is not considered a match.  A join relation with
+Join Expression X and Post-Join Filter Y is equivalent to a join relation with Join Expression X AND Y. | Optional                           |
 | Join Type        | One of the join types defined below.                         | Required                           |
 
 ### Join Types


site/docs/relations/logical_relations.md
```diff
@@ -234,7 +234,9 @@ The join operation will combine two separate inputs into a single output, based
 | Left Input       | A relational input.                                          | Required                           |
 | Right Input      | A relational input.                                          | Required                           |
 | Join Expression  | A boolean condition that describes whether each record from the left set "match" the record from the right set. Field references correspond to the direct output order of the data. | Required. Can be the literal True. |
-| Post-Join Filter | A boolean condition to be applied to each result record after the inputs have been joined, yielding only the records that satisfied the condition. | Optional                           |
+| Post-Join Filter | A boolean condition to be applied to each potential match between the left and right
+inputs.  If it evaluates to false then the potential match is not considered a match.  A join relation with
+Join Expression X and Post-Join Filter Y is equivalent to a join relation with Join Expression X AND Y. | Optional                           |
```
yongchul, line 239: Thank you! Much better than my local draft! :)

Two more things.
* Align Hash/MergeJoin post-join filter description with this. We could refer `JoinRel` there and leave what's different.
* Should this be `Optional, default True` like hash/merge join?

yongchul (approved):

site/docs/relations/logical_relations.md
westonpace: > Align Hash/MergeJoin post-join filter description with this. We could refer JoinRel there and leave what's different.

Can you expand on what you mean here?  The PR does currently update the hash/merge join descriptions.  I don't include the `A join relation with
Join Expression X and Post-Join Filter Y is equivalent to a join relation with Join Expression X AND Y` statement because this is not true for hash/merge join (the join expression for these relations is a series of equality conditions).

yongchul: I meant the language of the description. The way you describe is more explicit that the `post_join_filter` IS part of the join condition, say saying what "matches" and what "does not match". This is not for try to reduce the output.

Also, can we drop `Equi` form Hash`Equi`Join? :)

drin:
> A join relation with Join Expression X and Post-Join Filter Y is equivalent to a join relation with Join Expression X AND Y

Is this strictly true? As in a consumer must resolve both expressions on the same inputs? If so, I think it'd be nice to add a comment in the .proto file to the effect of "`post_join_filter` should be resolved in conjunction (AND) with `expression`."

proto/substrait/algebra.proto
```diff
@@ -255,6 +255,13 @@ message JoinRel {
   Rel left = 2;
   Rel right = 3;
   Expression expression = 4;
+  // The post-join filter is a filter that is applied to the result of the join before an output
+  // record is produced.  If the filter evaluates to false then the record is not considered a
+  // match.
```
drin, line 260: > ...applied to the result of the join

I believe this is semantically ambiguous for functions that aggregate over many tuples. I think you would get different results if such a function is applied over the input records (what is evaluated by `expression`) vs applied on the records passing the filter.

drin, line 260: I think this is ambiguous for functions that aggregate over many tuples. I think a "simple" example is:
1. the `post_join_filter` is a comparison (`lte`) that uses a window function (`count`)
2. the `expression` is a predicate with selectivity between 0 and 100%
3. the `expression` produces many tuples from one input for a single tuple of the other input

(2) and (3) are necessary for ambiguous scenarios to occur and (1) is where the ambiguity is expressed.

yongchul: Agree that it's better to clarify the predicates are evaluated over the inputs. Like @drin 's suggestion.

westonpace: I get very confused easily when talking about when this filter is applied.  Here is my understanding, in naive pseudocode, of how it is applied.  I'm omitting right joins, full outer joins, single joins, and mark joins for simplicity.

```
for left_record in left_records:
  has_match = False
  for right_record in right_records:
    if join_expression(left_record, right_record) and post_join_filter(left_record, right_record):
      has_match = True
      if join_type == Inner or join_type == Left:
        emit(combine(left_record, right_record):
  if has_match and join_type == LeftSemi:
    emit(left_record)
  elif not has_match and join_type == LeftAnti:
    emit(left_record)
```

If someone has an alternate proposal, is it possible to share your own pseudocode representation?

yongchul: @tokoko the predicates in join condition does not follow join type. The join type becomes into play depending on whether there is a matching row (i.e., intersection) or not. outer joins and antisemi joins should ensure that you have no rows that matches according to `JoinRel.expression AND JoinRel.post_join_filter` to correctly behave (i.e., whether to produce null padded rows (outer) or include in the output (antisemi)).

tokoko: makes sense. For JoinRel, can't we write something like "post-join filter is supposed to be evaluated as if it's part of the join expression" or something similar? It would be a lot simpler to understand imho rather than thinking through when during the operation it's supposed to be applied/evaluated.

drin: I agree with that pseudocode weston. My intention is to make the wording clearly reflect that `post_join_filter(left, record, right_record)` is valid and `post_join_filter(combine(left_record, right_record))` is invalid.

Note (for completeness) that my naive reading of "post join" was incorrect and would have been to implement:
```
# above pseudocode here
...
if post_join_filter(emitted_record):
  really_emit(emitted_record)
```

yongchul: Yup 'post_join' really tripped me and reason i started the thread. In the systems i worked, used residual join condition/predicate rather than post.

EpsilonPrime (approved):

jacques-n:
It's difficult to follow the threads in this discussion. 

One can think of a a join with a post join filter as a composite operation that is a join followed by a filter relation. It is entirely valid translation to take a post join filter out of the join and put in a filter relation directly afterwards and vice versa. 

The post join filter does not logically interact with the join type at all. The composite exists because many systems have it and it can be a beneficial physical pattern. The reason it has to be stated separately from the join predicate is to have covering behavior of all possible filter conditions. I always have to remind myself of [which conditions can and cannot be moved into a join evaluation clause](https://chatgpt.com/share/6817d298-5078-8007-8770-3049e4742959).

I'm supportive of clarifying the text if people are unclear as to what post join filter means.

westonpace:
> One can think of a a join with a post join filter as a composite operation that is a join followed by a filter relation. It is entirely valid translation to take a post join filter out of the join and put in a filter relation directly afterwards and vice versa.

This is _not_ the conclusion we came to.  I believe the content of the PR is still accurate with the threads, so you can just review the content and ignore the discussion.

For example:

```
SELECT * FROM a LEFT OUTER JOIN b ON a.id = b.id AND b.other_field IS NOT NULL
```

As it stands this filter will emit one row for each row in `a`.  If the filter is moved into the `WHERE` clause then it will emit fewer rows (less or equal to the number of rows emitted by an inner join).

From your GPT link this matches:

> Conditions on the non-preserved side that would otherwise eliminate rows that should remain when there's no match

yongchul:
Resurrecting this PR as I really can't get past on this `post` join filter. 😆  @westonpace do you mind if I carry this forward?

westonpace:
Oh yes, please carry on.

jacques-n:
> This is _not_ the conclusion we came to...

I think the conclusion and content updates are redefining the original purpose of this field when we should simply add another field. The changes do not align with my intentions when I [added the field](https://github.com/substrait-io/substrait/commit/d602e95732030b69e5985466cbca9e41453526a1#diff-cc469d4c718140101e2469c528afa30d690b56b870250b0040ebe867012ddd77R102)).  

I think the [spec](https://substrait.io/relations/logical_relations/#join-properties) content was clear (albeit a bit terse). The key bit is (my bold added):

> A boolean condition to be applied to each result record **after the inputs have been joined**, yielding only the records that satisfied the condition.

IOW, this filter is done after the join algorithm is completed. It does **not** interplay with join's matching decision.

To clarify things, I think there are two different concepts being discussed:
1. A filter that happens could be run in a separate relation but just happens to be included in the join operator. 
2. A filter that is part of the join matching condition and influences matching.

This patch tries to redefine the behavior as (2). My original intention was (1). 

Both (1) and (2) are useful. While I'm not an originalist, I think it is important to point out that I think the addition of a field for (2) only makes sense for HashJoinRel. it doesn't make sense for JoinRel or MergeJoinRel (since those already carry arbitrary expressions that are applied as part of matching). I would further say that the name isn't quite right if it applies as part of matching as that isn't "post join", it's during join. If we stick with the post_join_filter equating to (1) the field serves a useful purpose in all three of JoinRel, MergeJoinRel and HashJoinRel. 

My recommendation is improve content to clarify that the purpose of this field is (1) and introduce a separate (2) for HashJoinRel if so desired . If we do that, the following statements in this patch are incorrect:

- ~~For purely logical plans, this field is redundant, as the post_join_filter could be included in
expression~~
- ~~A boolean condition to be applied to each potential match between the left and right
  inputs. If it evaluates to false then the potential match is not considered a match~~

yongchul:
@jacques-n  Thanks for much needed clarification! So, the original intent was *really* (join - filter) (i.e., a filter happens to be squeezed into join). This was my interpretation when I just saw the name but unfortunately the [FAQ](https://substrait.io/faq/#what-is-the-purpose-of-the-post-join-filter-field-on-join-relations) was implying this is actually the *residual join condition* (this is my way of describing something that is part of join condition but not natively incorporated by the join algorithms (i.e., non-equality predicates).

Having said that, I much prefer to do following.
1. keep the `post_join_filter` as `post_join_filter` with much clearer documentation.
2. introduce `redisual_expression` field to HashJoinRel and MergeJoinRel to represent the extra predicates not accounted by the algorithms.
3. rewrite or remove the confusing FAQ entry (or refer to the latest documentation)

How does that sound? @jacques-n  @drin @tokoko @westonpace

yongchul:
Created #1044 , making `post_join_filter` as `post_join_filter` and introduce `residual_expression` to incorporate non-equality join predicate to HashJoinRel and MergeJoinRel.

PR #810, opened 2025-04-25
feat: specify build input of hash join operator

yongchul:
# Background

Substrait declares build is right and probe is left today. Unfortunately there are other systems (SQL Server family, Oracle, Snowflake, Google Cloud Spanner, etc) chose opposite side and it is a major headache for those systems to deal with the mirror images.

# Proposal

Adding a `build_left` field to signify which side is the build of the hash join. Default (false) implies the current behavior. Left and right does not change. Spark SQL/Spanner employs this approach and be flexible.

The change is backward/binary compatible.


proto/substrait/algebra.proto
```diff
@@ -788,8 +788,8 @@ message ComparisonJoinKey {
   }
 }
 
-// The hash equijoin join operator will build a hash table out of the right input based on a set of join keys.
-// It will then probe that hash table for incoming inputs, finding matches.
+// The hash equijoin operator will build a hash table out of one input (defualt `right`) based on a set of join keys.
```
jcamachor, line 791: typo. defualt -> default

jcamachor (dismissed):

EpsilonPrime:
If the behavior is different why not just switch left and right when implementing the join?  Is there value in maintaining the source left as left?

yongchul:
> If the behavior is different why not just switch left and right when implementing the join? Is there value in maintaining the source left as left?

There are multiple things. We also need to massage join types and there are issues with plumbing with existing tooling (e.g., anything presenting plans somehow) where left is implicitly build. It's also hard to rewire people's brain as well: a lot of people got easily confused in seconds internally when I tried to push different build/probe order. It appears that Spark is another benefiter of this given that the system can specify the build side.

jcamachor (dismissed):

westonpace (approved):
My initial reaction was similar to @EpsilonPrime 's, which is that this feels redundant.  @yongchul and I spoke on Slack and I will summarize where I'm at now.

I think the problem is that users have some concept of "left" and "right" and they communicate this through the join type.  In other words, if a user says `RIGHT_SEMI` then they are communicating that they want the right half of the query (in SQL or whatever) to be the side that is returned.

Then an optimizer often decides which side is ideal for building.  There is no guarantee this will be the right side.  Let's pretend it is the left side.

Without a `build_left` flag that would mean the optimized query would have the two inputs swapped (and the `JOIN_TYPE_RIGHT` would turn into `JOIN_TYPE_LEFT`).  This is a more significant rewrite than just changing a boolean flag.

I'm still not sure if this alone is enough justification for a new field.  However, this is also a physical relation and maybe not as stable / widely used, so I think I could go either way.  My default, when I'm conflicted, is to side with the people doing the work, and so I'll mark approve.

proto/substrait/algebra.proto
```diff
@@ -835,6 +835,10 @@ message HashJoinRel {
     JOIN_TYPE_RIGHT_MARK = 12;
   }
 
+  // When true, build hash table out of the `left` input. Otherwise, build hash table out the `right` input.
+  // Default is building hash table from the `right` input.
+  bool build_left = 9;
```
jacques-n, line 840: If we're adding a field, I suggest using an enum to make things clearer.

yongchul: Thank you! I changed the bool to enum.

yongchul:
> I'm pretty lukewarm on this change similar to Weston and David. Output field emit allows join sides to be swapped without impacting any other node in the plan so I don't feel like the new field is necessary.

The join type also needs to change when you swap the input and I agree that the output can be unaffected. Completely dropping the notion of left/right from hash join would be another solution but I am not sure about the acceptance -- I'm fine with that approach as well.

westonpace (approved):

vbarua (approved):
These changes make sense to me overall.

While it is possible to deal with this by swapping the left and the right inputs, swapping the comparison key fields and then adding a remap, it's a somewhat annoying transformation.

This sugar isn't free, as it pushes a bit more complexity into the consumer, but I'm generally in favor of making plan production simpler at the cost of consumer complexity, because consumer code is usually written by engine-owners and/or engine adjacent folks with plenty of knowledge, but plans can be produced by folks who may not have as much engine knowledge.

vbarua:
@jacques-n

> I've been saying for a while that we should build for dumb consumers in my intro deck to substrait. Not sure if your comment is in conflict with this or it simply means we need to do more work to give people "canonicalization tools".

Interesting, I think we might be in a (little) bit of conflict. I'm not advocating for producers to be able to generate garbage plans, but generally I assume that there are going to be many more plan producers than there are consumers. I'm definitely a little biased because the primary consumer I work on is an optimizer which has lots of tooling to manipulate plans, and our producers are ideally just translating from query languages to Substrait with as minimal processing as possible.

For this specific sugar, I don't think it introduces a ton of complexity into the consumer side and makes 1-1 translation into Substrait easier.

> it simply means we need to do more work to give people "canonicalization tools"

I think we need this generally, specifically good plan builders.

Issue #611, opened 2024-03-19
If a precision timestamp literal is encountered there is no way to know the precision


westonpace:
Currently precision timestamp literals are defined as:

```
      // If the precision is 6 or less then this is the microseconds since the UNIX epoch
      // If the precision is more than 6 then this is the nanoseconds since the UNIX epoch
      uint64 precision_timestamp = 34;
      uint64 precision_timestamp_tz = 35;
```

However, there is no way to know if the precision is 6 or less because the literal doesn't have to be associated with a type.  For example, consider:

```
SELECT * FROM my_table WHERE my_ts_col < timestamp (3) '2004-10-19 10:23:54.123'
```

This will turn into a plan with a literal value in the "less than" expression:

```
"scalarFunction":{
  "functionReference":3,
  "outputType":{..},
  "arguments":[
    { ... },
   ...
   "literal":{ "precision_timestamp": 1098181434123000 }
```

However, there is no way of knowing what the precision of that timestamp is.  I could infer it by tracing back and realizing that the other value in the expression has precision X but:

 * That's a lot of work, and I don't have to do this for other literals
 * They don't technically have to be the same, depending on what range of functions the producer / consumer supports

EpsilonPrime:
This behavior is identical to what happens with fixed_char, var_char, and decimal literals.  The way this is addressed in Isthmus is to immediately cast the literal to the appropriate precision.   For instance, a fixed character would be represented with the string "fruit" and then later is cast to the actual desired with such as fixed_char[20].

westonpace:
First, I'm not sure that's true with varchar or decimal, the precision is part of the literal:

```
    message VarChar {
      string value = 1;
      uint32 length = 2;
    }
```

```
    message Decimal {
      // little-endian twos-complement integer representation of complete value
      // (ignoring precision) Always 16 bytes in length
      bytes value = 1;
      // The maximum number of digits allowed in the value.
      // the maximum precision is 38.
      int32 precision = 2;
      // declared scale of decimal literal
      int32 scale = 3;
    }
```

With fixed_char / fixed_binary I can at least form some kind of "minimum precision literal".  E.g. given the literal "foo" I can turn it into a fixed_char<3>.

I can't do that with timestamp.  Given the literal 1710989532 how am I supposed to know if this is:

Timestamp(0) - Thursday, March 21, 2024 2:52:12 AM
Timestamp(3) - Tuesday, January 20, 1970 7:16:29.532 PM
[...]

PR #814, opened 2025-05-13
feat: add timezone to precision timestamp tz literals

EpsilonPrime:
Precision timestamp TZ literals now use a different (but backwards compatible datastructure).  Old plans loaded in to newer consumers will sometimes encounter an empty timezone value which may need to be handled during the transition period.

### DIFF
--- a/proto/substrait/algebra.proto
+++ b/proto/substrait/algebra.proto
@@ -1007,7 +1007,7 @@ message Expression {
       // Time in precision units past midnight.
       PrecisionTime precision_time = 37;
       PrecisionTimestamp precision_timestamp = 34;
-      PrecisionTimestamp precision_timestamp_tz = 35;
+      PrecisionTimestampTz precision_timestamp_tz = 35;
       Struct struct = 25;
       Map map = 26;
       // Timestamp in units of microseconds since the UNIX epoch.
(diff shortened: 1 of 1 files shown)


proto/substrait/algebra.proto
```diff
@@ -1059,10 +1059,19 @@ message Expression {
...
       int32 precision = 1;
-      // Time passed since 1970-01-01 00:00:00.000000 in UTC for PrecisionTimestampTZ and unspecified timezone for PrecisionTimestamp
+      // Time passed since 1970-01-01 00:00:00.000000 an unspecified timezone.
       int64 value = 2;
     }
 
+    message PrecisionTimestampTz {
+      // Sub-second precision, 0 means the value given is in seconds, 3 is milliseconds, 6 microseconds, 9 is nanoseconds, 12 is picoseconds
+      int32 precision = 1;
+      // Time passed since 1970-01-01 00:00:00.000000 in the specified timezone.
```
EpsilonPrime, line 1069: I specifically use the provided timezone since conversion to UTC and back can lose clarity due to daylight savings time.

vbarua (approved):
Makes sense to me.

vbarua:
This will fully complete https://github.com/substrait-io/substrait/issues/611 as well.

proto/substrait/algebra.proto
Blizzara: There is no loss of clarity when talking about _timestamps_, those are always well-defined, since they indicate an absolute point on a timeline (with possibly some nuances around how leap seconds are handled..).

What is the reason why a timezone would be needed here?

Blizzara:
> #611

Arguable that had already been fixed by https://github.com/substrait-io/substrait/pull/659. In fact, _if_ this PR is to be merged (which I tentatively disagree with), then I think we should also add the timezone into the PrecisionTimestampTz type. I see no benefit at all to having it only on the literal.

EpsilonPrime:
Closing this for now.   The existing PrecisionTimezoneTZ behavior matches that of DuckDB but not other systems (such as Snowflake).  We likely need an additional PrecisionTimezoneTz which preserves the timezone as part of the literal (allowing different timezones in one column) but that can wait until someone needs it.

PR #815, opened 2025-05-16
feat: add JSON schema for the dialect YAML file used in Substrait Go

EpsilonPrime:


EpsilonPrime:
Closing in favor of #816 which goes to the desired end state instead of pausing at the current Substrait Go state first (and then doing #816).

PR #816, opened 2025-05-18
feat:  add required relations and expressions to the dialect YAML definition

EpsilonPrime:
In addition to adding relations and expressions this also updates the dialect grammar:
 - Removes type which is currently unused.
 - Adds substrait_version to the top level.
 - Converts infix/postfix into an enum (now also including prefix and the previous default of function).

### DIFF
--- a/dialects/tests/test1.yaml
+++ b/dialects/tests/test1.yaml
@@ -1,0 +1,38 @@
+name: testdb
+type: sql
+dependencies:
+  arithmetic: 
+    https://github.com/substrait-io/substrait/blob/main/extensions/functions_arithmetic.yaml
+  comparison: 
+    https://github.com/substrait-io/substrait/blob/main/extensions/functions_comparison.yaml
+supported_types:
+  i32:
+    sql_type_name: TINYINT,
+    supported_as_column: true
+  i64:
+    sql_type_name: TINYINT,
+    supported_as_column: true
+scalar_functions:
+- name: comparison.is_null
+  local_name: IS NULL
+  postfix: true
+  supported_kernels:
+  - any1
+aggregate_functions:
+- name: arithmetic.max
+  aggregate: true
+  supported_kernels:
+  - i32
+  - i64
+- name: arithmetic.min
+  aggregate: true
+  supported_kernels:
+  - i32
+  - i64
+window_functions:
+- name: arithmetic.ntile
+  supported_kernels:
+  - i32
+- name: arithmetic.rank
+  supported_kernels:
+  - ""
--- a/dialects/tests/test1.yaml
+++ b/dialects/tests/test1.yaml
@@ -7,10 +7,10 @@ dependencies:
     https://github.com/substrait-io/substrait/blob/main/extensions/functions_comparison.yaml
 supported_types:
   i32:
-    sql_type_name: TINYINT,
+    sql_type_name: TINYINT
     supported_as_column: true
   i64:
-    sql_type_name: TINYINT,
+    sql_type_name: TINYINT
     supported_as_column: true
 scalar_functions:
 - name: comparison.is_null
--- a/dialects/tests/test1.yaml
+++ b/dialects/tests/test1.yaml
@@ -1,5 +1,4 @@
 name: testdb
-type: sql
 dependencies:
   arithmetic: 
     https://github.com/substrait-io/substrait/blob/main/extensions/functions_arithmetic.yaml
@@ -15,17 +14,15 @@ supported_types:
 scalar_functions:
 - name: comparison.is_null
   local_name: IS NULL
-  postfix: true
+  notation: POSTFIX
   supported_kernels:
   - any1
 aggregate_functions:
 - name: arithmetic.max
-  aggregate: true
   supported_kernels:
   - i32
   - i64
 - name: arithmetic.min
-  aggregate: true
   supported_kernels:
   - i32
   - i64
--- a/dialects/tests/test2.yaml
+++ b/dialects/tests/test2.yaml
@@ -1,0 +1,67 @@
+name: test
+type: sql
+dependencies:
+  arithmetic: 
+    https://github.com/substrait-io/substrait/blob/main/extensions/functions_arithmetic.yaml
+  boolean: 
+    https://github.com/substrait-io/substrait/blob/main/extensions/functions_boolean.yaml
+  comparison: 
+    https://github.com/substrait-io/substrait/blob/main/extensions/functions_comparison.yaml
+  string: 
+    https://github.com/substrait-io/substrait/blob/main/extensions/functions_string.yaml
+supported_types:
+  i32:
+    sql_type_name: INTEGER
+    supported_as_column: true
+  i64:
+    sql_type_name: BIGINT
+    supported_as_column: true
+  fp32:
+    sql_type_name: REAL
+    supported_as_column: true
+  fp64:
+    sql_type_name: DOUBLE
+    supported_as_column: true
+  bool:
+    sql_type_name: BOOLEAN
+    supported_as_column: true
+scalar_functions:
+- name: arithmetic.add
+  local_name: +
+  infix: true
+  required_options:
+    overflow: ERROR
+    rounding: TIE_TO_EVEN
+  supported_kernels:
+  - i32_i32
+  - i64_i64
+  - fp32_fp32
+  - fp64_fp64
+- name: arithmetic.subtract
+  local_name: '-'
+  infix: true
+  required_options:
+    overflow: ERROR
+    rounding: TIE_TO_EVEN
+  supported_kernels:
+  - fp32_fp32
+  - fp64_fp64
+- name: boolean.and
+  local_name: and
+  infix: true
+  supported_kernels:
+  - bool
+- name: comparison.is_null
+  local_name: IS NULL
+  postfix: true
+  supported_kernels:
+  - any1
+- name: string.concat
+  local_name: '||'
+  required_options:
+    null_handling: ACCEPT_NULLS
+  infix: true
+  supported_kernels:
+  - vchar
+  - str
+  variadic: 1
--- a/dialects/tests/test2.yaml
+++ b/dialects/tests/test2.yaml
@@ -1,5 +1,4 @@
 name: test
-type: sql
 dependencies:
   arithmetic: 
     https://github.com/substrait-io/substrait/blob/main/extensions/functions_arithmetic.yaml
@@ -28,7 +27,7 @@ supported_types:
 scalar_functions:
 - name: arithmetic.add
   local_name: +
-  infix: true
+  notation: INFIX
   required_options:
     overflow: ERROR
     rounding: TIE_TO_EVEN
@@ -39,7 +38,7 @@ scalar_functions:
   - fp64_fp64
 - name: arithmetic.subtract
   local_name: '-'
-  infix: true
+  notation: INFIX
   required_options:
     overflow: ERROR
     rounding: TIE_TO_EVEN
@@ -48,19 +47,19 @@ scalar_functions:
   - fp64_fp64
 - name: boolean.and
   local_name: and
-  infix: true
+  notation: INFIX
   supported_kernels:
   - bool
 - name: comparison.is_null
   local_name: IS NULL
-  postfix: true
+  notation: POSTFIX
   supported_kernels:
   - any1
 - name: string.concat
   local_name: '||'
   required_options:
     null_handling: ACCEPT_NULLS
-  infix: true
+  notation: INFIX
   supported_kernels:
   - vchar
   - str
--- a/text/dialect_schema.yaml
+++ b/text/dialect_schema.yaml
@@ -1,0 +1,112 @@
+# Substrait Dialect Schema Definition
+# This schema defines the structure for Substrait dialect files, which specify how
+# functions and types are supported in different execution environments.
+
+$id: http://substrait.io/schemas/dialect
+$schema: https://json-schema.org/draft/2020-12/schema
+title: Dialect
+additionalProperties: false
+type: object
+
+# A valid dialect must have at least one defined function (of any time).
+anyOf:
+  - properties:
+      scalar_functions:
+        type: array
+        minItems: 1
+    required:
+      - scalar_functions
+  - properties:
+      aggregate_functions:
+        type: array
+        minItems: 1
+    required:
+      - aggregate_functions
+  - properties:
+      window_functions:
+        type: array
+        minItems: 1
+    required:
+      - window_functions
+properties:
+  # The name of the dialect (e.g., "MySQL", "PostgreSQL", "Spark")
+  name:
+    type: string
+  # The type or category of the dialect (e.g., "SQL", "Compute Engine")
+  type:
+    type: string
+  # External dependencies required by this dialect
+  # Each key is a dependency name and value is the version string
+  dependencies:
+    type: object
+    additionalProperties:
+      type: string
+    minProperties: 1
+  # Types supported by this dialect
+  # Maps type identifiers to their properties
+  supported_types:
+    type: object
+    additionalProperties:
+      type: object
+      properties:
+        # The SQL name for this type in the dialect
+        sql_type_name:
+          type: string
+        # Whether this type can be used as a column type
+        supported_as_column:
+          type: boolean
+      minProperties: 1
+    minProperties: 1
+  # Regular functions that operate on scalar values
+  scalar_functions:
+    type: array
+    items:
+      $ref: "#/definitions/dialect_function"
+  # Functions that aggregate multiple rows into a single result
+  aggregate_functions:
+    type: array
+    items:
+      $ref: "#/definitions/dialect_function"
+  # Functions that operate over a window of rows
+  window_functions:
+    type: array
+    items:
+      $ref: "#/definitions/dialect_function"
+
+definitions:
+  # Function definition used for all function types
+  dialect_function:
+    type: object
+    additionalProperties: false
+    required: [name, supported_kernels]
+    properties:
+      # The canonical name of the function in Substrait
+      name:
+        type: string
+        # The name used in the specific dialect implementation
+        # (e.g., "+" might be the local_name for "add")
+      local_name:
+        type: string
+      # Whether the function uses infix notation (e.g., "a + b")
+      infix:
+        type: boolean
+      # Whether the function uses postfix notation
+      postfix:
+        type: boolean
+      # Function-specific options that must be specified
+      required_options:
+        type: object
+      # Indicates if this is an aggregation function
+      aggregate:
+        type: boolean
+        # List of kernel identifiers supported by this function
+      supported_kernels:
+        type: array
+        minItems: 1
+        items:
+          type: string
+      # If present, indicates the function accepts a variable number of arguments
+      # The value specifies the minimum number of arguments required (must be ≥ 1)
+      variadic:
+        type: integer
+        minimum: 1
--- a/text/dialect_schema.yaml
+++ b/text/dialect_schema.yaml
@@ -8,7 +8,7 @@ title: Dialect
 additionalProperties: false
 type: object
 
-# A valid dialect must have at least one defined function (of any time).
+# A valid dialect must have at least one defined function (of any type).
 anyOf:
   - properties:
       scalar_functions:
@@ -29,84 +29,88 @@ anyOf:
     required:
       - window_functions
 properties:
-  # The name of the dialect (e.g., "MySQL", "PostgreSQL", "Spark")
+  # Optional name of the dialect (e.g., "MySQL", "PostgreSQL", "Spark").
   name:
     type: string
-  # The type or category of the dialect (e.g., "SQL", "Compute Engine")
+  # Optional type or category of the dialect (e.g., "SQL", "Compute Engine").
   type:
     type: string
-  # External dependencies required by this dialect
-  # Each key is a dependency name and value is the version string
+  # A map of function YAML files depended on by this dialect.
+  # Each key is an alias used by functions with the value being a URI.
   dependencies:
     type: object
     additionalProperties:
       type: string
     minProperties: 1
-  # Types supported by this dialect
-  # Maps type identifiers to their properties
+  # A map of types supported by this dialect.
+  # The key is the Substrait name for the type.
   supported_types:
     type: object
     additionalProperties:
       type: object
       properties:
-        # The SQL name for this type in the dialect
+        # The SQL name for this type in the dialect.
         sql_type_name:
           type: string
-        # Whether this type can be used as a column type
+        # Whether this type can be used as a column type.
         supported_as_column:
           type: boolean
       minProperties: 1
     minProperties: 1
-  # Regular functions that operate on scalar values
+  # Regular functions that operate on scalar values.
   scalar_functions:
     type: array
     items:
       $ref: "#/definitions/dialect_function"
-  # Functions that aggregate multiple rows into a single result
+  # Functions that aggregate multiple rows into a single result.
   aggregate_functions:
     type: array
     items:
       $ref: "#/definitions/dialect_function"
-  # Functions that operate over a window of rows
+  # Functions that operate over a window of rows.
   window_functions:
     type: array
     items:
       $ref: "#/definitions/dialect_function"
 
 definitions:
-  # Function definition used for all function types
+  # Function definition used for all function types.
   dialect_function:
     type: object
     additionalProperties: false
     required: [name, supported_kernels]
     properties:
-      # The canonical name of the function in Substrait
+      # The canonical name of the function in Substrait.  This typically takes the
+      # form of a dependency alias followed by a period and then the function name
+      # from the YAML file.
       name:
         type: string
-        # The name used in the specific dialect implementation
-        # (e.g., "+" might be the local_name for "add")
+      # The name used in the specific dialect implementation
+      # (e.g., "+" might be the local_name for "add").
       local_name:
         type: string
-      # Whether the function uses infix notation (e.g., "a + b")
+      # Whether the function uses infix notation (e.g., "a + b").
       infix:
         type: boolean
-      # Whether the function uses postfix notation
+      # Whether the function uses postfix notation (e.g., "a b + ").
       postfix:
         type: boolean
-      # Function-specific options that must be specified
+      # Function-specific options that must be supported by the dialect.
       required_options:
         type: object
-      # Indicates if this is an aggregation function
+      # Indicates if this is an aggregation function.
       aggregate:
         type: boolean
-        # List of kernel identifiers supported by this function
+      # List of kernels (parameter type lists) supported by this function.
+      # For example, i32 or str_str.  Variadic arguments are not included in
+      # the kernel's list of parameters.
       supported_kernels:
         type: array
         minItems: 1
         items:
           type: string
       # If present, indicates the function accepts a variable number of arguments
-      # The value specifies the minimum number of arguments required (must be ≥ 1)
+      # The value specifies the minimum number of arguments required (must be ≥ 1).
       variadic:
         type: integer
         minimum: 1
--- a/text/dialect_schema.yaml
+++ b/text/dialect_schema.yaml
@@ -32,8 +32,8 @@ properties:
   # Optional name of the dialect (e.g., "MySQL", "PostgreSQL", "Spark").
   name:
     type: string
-  # Optional type or category of the dialect (e.g., "SQL", "Compute Engine").
-  type:
+  # The version of Substrait that this dialect applies for.  Optional.
+  substrait_version:
     type: string
   # A map of function YAML files depended on by this dialect.
   # Each key is an alias used by functions with the value being a URI.
@@ -89,18 +89,20 @@ definitions:
       # (e.g., "+" might be the local_name for "add").
       local_name:
         type: string
-      # Whether the function uses infix notation (e.g., "a + b").
-      infix:
-        type: boolean
-      # Whether the function uses postfix notation (e.g., "a b + ").
-      postfix:
-        type: boolean
+      notation:
+        type: string
+        enum:
+          - INFIX
+          - POSTFIX
+          - PREFIX
+          - FUNCTION
+        default: FUNCTION
+        description: >
+          Infix notation (e.g., "a + b"), postfix notation (e.g., "a b + "),
+          prefix notation (e.g., "+ a b"), or function notation (func(a, b)).
       # Function-specific options that must be supported by the dialect.
       required_options:
         type: object
-      # Indicates if this is an aggregation function.
-      aggregate:
-        type: boolean
       # List of kernels (parameter type lists) supported by this function.
       # For example, i32 or str_str.  Variadic arguments are not included in
       # the kernel's list of parameters.
--- a/text/dialect_schema.yaml
+++ b/text/dialect_schema.yaml
@@ -72,9 +72,21 @@ properties:
     type: array
     items:
       $ref: "#/definitions/dialect_function"
+  # Relations that are supported by this dialect.  If not present or empty,
+  # all relations are supported.
+  supported_relations:
+    type: array
+    items:
+      $ref: "#/definitions/dialect_relation"
+  # Expressions that are supported by this dialect.  If not present or empty,
+  # all expressions are supported.
+  supported_expressions:
+    type: array
+    items:
+      $ref: "#/definitions/dialect_expression"
 
 definitions:
-  # Function definition used for all function types.
+  # Definition used for all function types.
   dialect_function:
     type: object
     additionalProperties: false
@@ -116,3 +128,78 @@ definitions:
       variadic:
         type: integer
         minimum: 1
+  dialect_relation:
+    type: object
+    additionalProperties: false
+    required: [rel_type]
+    properties:
+      # rel_type corresponds to the Rel rel_type oneof field names (in uppercase).
+      # Only relation types specified here are supported in this dialect.  As
+      # there is no way to tell one user-defined relation from another here, it
+      # is not recommended to include those types within a dialect.  If a
+      # relation is listed here, all features of that relation must be supported
+      # for the dialect's requested version of Substrait (in other words if
+      # there is a deprecated feature at that Substrait version it does not need
+      # to be supported).
+      rel_type:
+        type: string
+        enum:
+          - READ
+          - FILTER
+          - FETCH
+          - AGGREGATE
+          - SORT
+          - JOIN
+          - PROJECT
+          - SET
+          - EXTENSION_SINGLE
+          - EXTENSION_MULTI
+          - EXTENSION_LEAD
+          - CROSS
+          - REFERENCE
+          - WRITE
+          - DDL
+          - UPDATE
+          - HASH_JOIN
+          - MERGE_JOIN
+          - NESTED_LOOP_JOIN
+          - WINDOW
+          - EXCHANGE
+          - EXPAND
+      # sub_type corresponds to the primary oneof in any given relation type.
+      # For example, in ReadRel it is the field name inside read_type and
+      # in JoinRel it is the join_type.  If there is no oneof in a relation
+      # sub_type should be empty.  Functionality not controlled by the
+      # sub_type needs to be supported for the relation to be included in the
+      # dialect.
+      sub_type:
+        type: string
+  dialect_expression:
+    type: object
+    additionalProperties: false
+    required: [rex_type]
+    properties:
+      # rex_type corresponds to the Expression rex_type oneof field names.
+      rex_type:
+        type: string
+        enum:
+          - LITERAL
+          - SELECTION
+          - SCALAR_FUNCTION
+          - WINDOW_FUNCTION
+          - IF_THEN
+          - SWITCH_EXPRESSION
+          - SINGULAR_OR_LIST
+          - MULTI_OR_LIST
+          - CAST
+          - SUBQUERY
+          - NESTED
+          - DYNAMIC_PARAMETER
+      # sub_type corresponds to the primary oneof in any given expression type.
+      # For example, in ReferenceSegment it is the field numbers inside
+      # reference_type and in Nested it is the nested_type.  If there is no
+      # oneof in an expression sub_type should be empty.  Functionality not
+      # controlled by the sub_type needs to be supported for the expression to
+      # be included in the dialect.
+      sub_type:
+        type: string
--- a/text/simple_extensions_schema.yaml
+++ b/text/simple_extensions_schema.yaml
@@ -26,6 +26,8 @@ properties:
           type: string
         description:
           type: string
+        # An optional list of test files and/or directories associated with
+        # these function definitions.
         tests:
           type: array
           minItems: 1
--- a/text/simple_extensions_schema.yaml
+++ b/text/simple_extensions_schema.yaml
@@ -26,6 +26,11 @@ properties:
           type: string
         description:
           type: string
+        tests:
+          type: array
+          minItems: 1
+          items:
+            type: string
         structure:
           $ref: "#/$defs/type"
         parameters: # parameter list for compound types
--- a/.github/workflows/pr.yml
+++ b/.github/workflows/pr.yml
@@ -65,6 +65,20 @@ jobs:
             ajv validate -s ../text/simple_extensions_schema.yaml --strict=true --spec=draft2020 -d "$i"
           done
         working-directory: ./extensions
+      - run: |
+          set -euo pipefail
+          for i in $(ls);
+          do
+            ajv validate -s ../text/dialect_schema.yaml --strict=true --spec=draft2020 -d "$i"
+          done
+        working-directory: ./dialects/tests
+      - run: |
+          set -euo pipefail
+          for i in $(ls);
+          do
+            ajv validate -s ../text/dialect_schema.yaml --strict=true --spec=draft2020 -d "$i"
+          done
+        working-directory: ./dialects
   dry_run_release:
     name: Dry-run release
     runs-on: ubuntu-latest
--- a/.github/workflows/pr.yml
+++ b/.github/workflows/pr.yml
@@ -74,7 +74,7 @@ jobs:
         working-directory: ./dialects/tests
       - run: |
           set -euo pipefail
-          for i in $(ls);
+          for i in $(ls *.yaml);
           do
             ajv validate -s ../text/dialect_schema.yaml --strict=true --spec=draft2020 -d "$i"
           done
--- a/.github/workflows/pr.yml
+++ b/.github/workflows/pr.yml
@@ -60,14 +60,14 @@ jobs:
       - run: npm install -g ajv-cli
       - run: |
           set -euo pipefail
-          for i in $(ls);
+          for i in $(ls *.yaml);
           do
             ajv validate -s ../text/simple_extensions_schema.yaml --strict=true --spec=draft2020 -d "$i"
           done
         working-directory: ./extensions
       - run: |
           set -euo pipefail
-          for i in $(ls);
+          for i in $(ls *.yaml);
           do
             ajv validate -s ../../text/dialect_schema.yaml --strict=true --spec=draft2020 -d "$i"
           done
--- a/.github/workflows/pr.yml
+++ b/.github/workflows/pr.yml
@@ -69,16 +69,16 @@ jobs:
           set -euo pipefail
           for i in $(ls);
           do
-            ajv validate -s ../text/dialect_schema.yaml --strict=true --spec=draft2020 -d "$i"
+            ajv validate -s ../../text/dialect_schema.yaml --strict=true --spec=draft2020 -d "$i"
           done
-        working-directory: ../dialects/tests
+        working-directory: ./dialects/tests
       - run: |
           set -euo pipefail
           for i in $(ls);
           do
             ajv validate -s ../text/dialect_schema.yaml --strict=true --spec=draft2020 -d "$i"
           done
-        working-directory: ../dialects
+        working-directory: ./dialects
   dry_run_release:
     name: Dry-run release
     runs-on: ubuntu-latest
--- a/.github/workflows/pr.yml
+++ b/.github/workflows/pr.yml
@@ -71,14 +71,14 @@ jobs:
           do
             ajv validate -s ../text/dialect_schema.yaml --strict=true --spec=draft2020 -d "$i"
           done
-        working-directory: ./dialects/tests
+        working-directory: ../dialects/tests
       - run: |
           set -euo pipefail
           for i in $(ls);
           do
             ajv validate -s ../text/dialect_schema.yaml --strict=true --spec=draft2020 -d "$i"
           done
-        working-directory: ./dialects
+        working-directory: ../dialects
   dry_run_release:
     name: Dry-run release
     runs-on: ubuntu-latest
--- a/dialects/errors/test5.yaml
+++ b/dialects/errors/test5.yaml
@@ -1,7 +1,0 @@
-type: sql
-dependencies:
-supported_types:
-window_functions:
-- name: arithmetic.ntile
-  supported_kernels:
-  - i32
--- a/dialects/errors/test5.yaml
+++ b/dialects/errors/test5.yaml
@@ -1,0 +1,7 @@
+type: sql
+dependencies:
+supported_types:
+window_functions:
+- name: arithmetic.ntile
+  supported_kernels:
+  - i32
--- a/dialects/errors/test6.yaml
+++ b/dialects/errors/test6.yaml
@@ -1,0 +1,6 @@
+type: sql
+dependencies:
+supported_types:
+  i32:
+   sql_type_name: TINYINT,
+   supported_as_column: true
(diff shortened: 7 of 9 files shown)


dialects/tests/test5.yaml
```diff
@@ -0,0 +1,36 @@
...
+    sub_type: [ICEBERG_TABLE]
+  - rel_type: FILTER
+  - rel_type: FETCH
+  - rel_type: AGGREGATE
+  - rel_type: SORT
+  - rel_type: JOIN
+  - rel_type: PROJECT
+    sub_type:
+      - LEFT
+      - RIGHT
```
vbarua, line 23: I assume that LEFT and RIGHT should be JOIN subtypes?

text/dialect_schema.yaml
```diff
@@ -0,0 +1,221 @@
...
+    type: object
+    additionalProperties: false
+    required: [rel_type]
+    properties:
+      rel_type:
+        type: string
+        description: >
+          rel_type corresponds to the Rel rel_type oneof field names (in uppercase).
+          Only relation types specified here are supported in this dialect.  As
+          there is no way to tell one user-defined relation from another here, it
```
vbarua, line 149: > there is no way to tell one user-defined relation from another here

I could see doing this via the sub_type mechanism you've introduced, were the subtype here is the
> URL that describes the type of the serialized message.

from https://protobuf.dev/reference/java/api-docs/com/google/protobuf/Any.html

So like
```proto
  - rel_type: EXTENSION_LEAF 
    sub_type:
      - type.googleapis.com/google.profile.Person
      - <your extenstion URL>
```

dialects/tests/test5.yaml
EpsilonPrime: Yep, that highlights the lack of checking for the sub_type.  It is technically possible to check rel_type/sub_type combinations if we want the added protection.

text/dialect_schema.yaml
EpsilonPrime: Done.

dialects/tests/test1.yaml
```diff
@@ -0,0 +1,35 @@
...
+  i32:
+    sql_type_name: TINYINT
+    supported_as_column: true
+  i64:
+    sql_type_name: TINYINT
+    supported_as_column: true
+scalar_functions:
+- name: comparison.is_null
+  local_name: IS NULL
+  notation: POSTFIX
```
Blizzara, line 17: what is the point of these? they seem like they relate to how this plan would be written as sql, but why is that relevant here?

dialects/tests/test1.yaml
```diff
@@ -0,0 +1,35 @@
...
+  supported_kernels:
```
Blizzara, line 18: the extensions files call these `impls`, if I understood right what this means? I don't think there is anything in Substrait called "kernels"

```suggestion
  supported_impls:
```

text/dialect_schema.yaml
```diff
@@ -0,0 +1,220 @@
...
+      supported_kernels:
+        type: array
+        description: >
+          List of kernels (parameter type lists) supported by this function.
+          For example, i32 or str_str.  Variadic arguments are not included in
+          the kernel's list of parameters.
+        minItems: 1
+        items:
+          type: string
+      variadic:
```
Blizzara, line 133: the variadicness is part of the impls (ie. "kernel" here), taking it higher may cause issues at some point

dialects/tests/test1.yaml
```diff
@@ -0,0 +1,35 @@
+name: testdb
+dependencies:
+  arithmetic: 
+    https://github.com/substrait-io/substrait/blob/main/extensions/functions_arithmetic.yaml
+  comparison: 
+    https://github.com/substrait-io/substrait/blob/main/extensions/functions_comparison.yaml
+supported_types:
+  i32:
+    sql_type_name: TINYINT
```
Blizzara, line 9: what's the point of this? Not every system has anything to do with SQL

dialects/tests/test1.yaml
EpsilonPrime: These aid in conversion to and from a specific SQL dialect.  They are a feature of the existing dialect files supported in substrait-go.

text/dialect_schema.yaml
EpsilonPrime: This also comes from Substrait Go   See #815 where the Substrait Go compatible version will be introduced.

EpsilonPrime: There are notes in the Go code that variadic isn't well defined.  Additionally the implementation is based on the non-variadic portion of the arguments.  I would guess that making a variadic block separate from a non-variadic block (duplicating the function) would be necessary.

dialects/tests/test2.yaml
```diff
@@ -0,0 +1,66 @@
...
+  local_name: +
+  notation: INFIX
+  required_options:
+    overflow: ERROR
+    rounding: TIE_TO_EVEN
+  supported_kernels:
+  - i32_i32
+  - i64_i64
+  - fp32_fp32
+  - fp64_fp64
```
vbarua, line 38: This entry for `artithmetic.add` is a little weird because `overflow` is only an option for the integer variants, and `rounding` is only an option for the float variants.

Is the meaning that the option is required only if the impl declares it?

dialects/tests/test2.yaml
```diff
@@ -0,0 +1,66 @@
...
+  - bool
+- name: comparison.is_null
+  local_name: IS NULL
+  notation: POSTFIX
+  supported_kernels:
+  - any1
+- name: string.concat
+  local_name: '||'
+  required_options:
+    null_handling: ACCEPT_NULLS
```
vbarua, line 61: When declaring a required option, is the meaning that this system only supports `string.concat` with the null handling options set to `ACCEPT_NULLS`?

text/dialect_schema.yaml
```diff
@@ -0,0 +1,220 @@
...
+# This schema defines the structure for Substrait dialect files, which specify how
+# functions and types are supported in different execution environments.
+
+$id: http://substrait.io/schemas/dialect
+$schema: https://json-schema.org/draft/2020-12/schema
+title: Dialect
+additionalProperties: false
+type: object
+
+# A valid dialect must have at least one defined function (of any type).
```
vbarua, line 11: Do we want to allow dialects with _no_ functions? Imagine a system that only supports ReadRels for fetching data. It would still be useful to have a dialect to indicated support types.

text/dialect_schema.yaml
```diff
@@ -0,0 +1,220 @@
...
+      all relations are supported.
+    items:
+      $ref: "#/definitions/dialect_relation"
+  supported_expressions:
+    type: array
+    description: >
+      Expressions that are supported by this dialect.  If not present or empty,
+      all expressions are supported.
+    items:
+      $ref: "#/definitions/dialect_expression"
```
vbarua, line 90: minor: convention-wise I would almost suggest that we lay out* the properties as
* `supported_types`
* `supported_relations`
* `supported_expression`
* `scalar_functions`
* `aggregate_functions`
* `window_functions`

The declarations of the first 3 properties are generally going to be short compares to the tens/hundreds of function entries.

\* I know that the schema doesn't enforce an ordering of these for end-users.

dialects/tests/test1.yaml
Blizzara: Interesting, so kinda like defining the consumer directly in the dialect yaml, I guess that could be nice. How far do we want to take it though? Sometimes the target functions take the same args but in different order, should we allow for e.g. that as well?

Nb. maybe the "sql_type_name" above should be "local_name" as well?

dialects/tests/test2.yaml
Blizzara: I wonder if these should just be structurally full copies of the Substrait extension yml files. They could then include only the impls and options that are supported. The name could just use the same format as here to provide a link to the original extension yaml. Feels like any simpler subset of it is going to be lacking in some respect?

dialects/tests/test1.yaml
EpsilonPrime: Done.

EpsilonPrime: (Order difference is likely a different function implementation.)

dialects/tests/test1.yaml
EpsilonPrime: Done.

text/simple_extensions_schema.yaml
```diff
@@ -26,6 +26,13 @@ properties:
           type: string
         description:
           type: string
+        # An optional list of test files and/or directories associated with
+        # these function definitions.
+        tests:
+          type: array
+          minItems: 1
+          items:
+            type: string
```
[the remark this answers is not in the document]
EpsilonPrime: Removed in favor of #817.

dialects/tests/test1.yaml
```diff
@@ -0,0 +1,35 @@
...
+    supported_as_column: true
+  i64:
+    sql_type_name: TINYINT
+    supported_as_column: true
```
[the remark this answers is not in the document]
EpsilonPrime: done.

text/dialect_schema.yaml
EpsilonPrime: Removed the restriction.

text/dialect_schema.yaml
EpsilonPrime: Rearranged the schema and tests.

dialects/tests/test2.yaml
EpsilonPrime: Yes, the list of options is exactly this.

dialects/tests/test1.yaml
```diff
@@ -0,0 +1,35 @@
+name: testdb
+dependencies:
+  arithmetic:
+    https://github.com/substrait-io/substrait/blob/main/extensions/functions_arithmetic.yaml
```
vbarua, line 4: meta: Something I'm thinking about, having just look at updating the schema definition for retrieving tests in repos, is that this also suffers from the same problem in that `https://github.com/substrait-io/substrait/blob/main/extensions/functions_arithmetic.yaml` isn't really a link to the raw extension file but the GitHub UI for it.

I'm not proposing we do anything about it now, but it might make sense to actually host these extensions somewhere.

dialects/tests/test2.yaml
vbarua: > If the option is there, it must be provided. That function always is associated with those options. If no options are provided, no options should be present.

By this do you mean if the option is present for a specific variant? So like `add:i8_i8` must set `overflow:ERROR` to match with this, but not `rounding` because that variant doesn't declare a `rounding` options?

This would be good to document on the website itself.

> I wonder if these should just be structurally full copies of the Substrait extension yml files.

What do you mean by this @Blizzara?

text/dialect_schema.yaml
```diff
@@ -0,0 +1,203 @@
...
+      The key is the Substrait name for the type.
+    additionalProperties:
+      type: object
+      properties:
+        local_name:
+          type: string
+          description: The SQL name for this type in the dialect.
+        supported_as_column:
+          type: boolean
+          description: Whether this type can be used as a column type.
```
vbarua, line 39: Possibly something better left for a follow-up, but `extension_uri` could also be a useful field here for user-defined types in extensions.

text/dialect_schema.yaml
```diff
@@ -0,0 +1,203 @@
...
+          - HASH_JOIN
+          - MERGE_JOIN
+          - NESTED_LOOP_JOIN
+          - WINDOW
+          - EXCHANGE
+          - EXPAND
+      sub_type:
+        type: array
+        description: >
+          sub_type corresponds to the primary oneof in any given relation type.
```
vbarua, line 162: This both make sense and feels underspecified. Specifically `primary` is a little bit "know it if you see it".

Your examples make it clear what the intended idea is, and I think this is a reasonable starting point, but I can see ourselves needing to come back and making explicit sub-properties for specific relations like:
```yaml
supported_relations:
  rels: [PROJECT, FILTER]
  join_rel:
    supported_join_types: [INNER, LEFT, RIGHT]
  read_rel:
    supported_read_types: [VIRTUAL_TABLE, NAMED_TABLE]
```

to make this truly unambiguous in the near future.

text/dialect_schema.yaml
```diff
@@ -0,0 +1,203 @@
...
+          For example, in ReadRel it is the field name inside read_type and
+          in JoinRel it is the join_type.  If there is no oneof in a relation
+          sub_type should be empty.  Functionality not controlled by the
+          sub_type needs to be supported for the relation to be included in the
+          dialect.  For user-defined types, use the URL that describes the type
+          of the message (for example type.googleapis.com/google.profile.Person).
```
vbarua, line 168: minor suggestions
`user-defined type` is a separate thing in Substrait. Something like
```
For user-defined relations, use the URL (e.g. type.googleapis.com/google.profile.Person) that describes the type of the relation message.
```
might be a little less ambiguous.

vbarua (dismissed):
I think this looks pretty good overall, and is compatible with some of the goals I am interested in from https://docs.google.com/document/d/11zxu19BMiDqmw9En0h_n1ZhKcLhlr1vz_Mlg5VBQTJ8/edit?tab=t.0

One thing that I think would be useful would be a page on the website talking about dialects to give folks context as to what problem we want to solve with them, and also document some of the matching logic. This can probably be fleshed out after this PR, but I think even a minimal explanation would be good to include at the start.

I'm curious to hear what other folks think about this as well, but I'm happy to approve with some minor comments.

dialects/tests/test1.yaml
wackywendell: It is possible to link to the raw files hosted on Github; in this case:
<https://raw.githubusercontent.com/substrait-io/substrait/refs/heads/main/extensions/functions_arithmetic.yaml>

Or if you want to specify a particular tag or commit:
- `v0.73.0`: <https://raw.githubusercontent.com/substrait-io/substrait/refs/tags/v0.73.0/extensions/functions_arithmetic.yaml>
- `d430e521`: <https://raw.githubusercontent.com/substrait-io/substrait/d430e521f203aec6a4e06731d4bfd68cdf61f443/extensions/functions_arithmetic.yaml>

Would it make sense to use one of those links here?

text/dialect_schema.yaml
jacques-n: I'm wondering whether this is going to immediately be deprecated and we should start with objects rather than strings.

```
rels:
  - type: join
  - type: project
  - type: filter
```

Which will allow us to expand specific properties in future:
```
rels:
  - type: join
    join_types: ['LEFT', 'RIGHT']
  - type: project
  - type: filter
```

Of source, we can support oneof for instead. E.g.

```
rels:
  - type: join
    join_types: ['LEFT', 'RIGHT']
  - project
  - filter
  - ...
```

but that increases burden on implementers

proto/substrait/plan.proto
```diff
@@ -28,6 +28,13 @@ message Plan {
   // versions.
   Version version = 6;
 
+  // An URN pointing to a specific dialect that this plan conforms to.  The
+  // plan's producer is responsible for ensuring conformity.  Dialect URNs
+  // may include a version in the anchor.  If it does, the consumer should
+  // exactly match the URN (there are no semantics defined for at least
+  // this version).
+  optional string dialect_version = 8;
```
jacques-n: the _version part here seems weird. why not just dialect?

text/dialect_schema.yaml
```diff
@@ -0,0 +1,203 @@
...
+type: object
+
+properties:
+  name:
+    type: string
+    description: Optional name of the dialect (e.g., "MySQL", "PostgreSQL", "Spark").
+  substrait_version:
+    type: string
+    description: The version of Substrait that this dialect applies for.  Optional.
+  dependencies:
```
jacques-n, line 15: I suggest adding which properties are required explicitly throughout this schema and blocking extra properties. I find it is better to have a strict schema so people don't realize pieces are being ignored.

text/dialect_schema.yaml
jacques-n: I'm not following why we would include any variadic info here. I suspect that the only thing we can do is further contrain an existing implementation. For example, the impl supports 20 args but this dialect only supports 10. We should never be broadening semantics here.

text/dialect_schema.yaml
```diff
@@ -0,0 +1,203 @@
...
+  dialect_expression:
+    type: object
+    additionalProperties: false
+    required: [rex_type]
+    properties:
+      rex_type:
+        type: string
+        description: rex_type corresponds to the Expression rex_type oneof field names.
+        enum:
+          - LITERAL
```
jacques-n: same comment as relations about strings versus objects. I may support only certain subquery type for example.

jacques-n (commented):
High level note, this is the structure we used at Dremio:

https://github.com/jasonhughes248/dremio-salesforce-connector/blob/master/src/main/resources/arp/implementation/salesforce-arp.yaml

Specific comments throughout.

text/dialect_schema.yaml
vbarua: > I'm wondering whether this is going to immediately be deprecated

When you put it that way, we probably should just do the Right Thing (TM) to begin with. I like the object idea. It's more work up front, but I think it's more precise and clearer in the long-term.

dialects/tests/test1.yaml
yongchul: Since this is test, using a link to the head (`refs/heads/`) makes sense to me.

text/dialect_schema.yaml
yongchul: What does `column` mean here? Is it tied to storage (i.e., serialization and deserialization are defined so can appear in read, write, query result)?

dialects/tests/test1.yaml
EpsilonPrime: While being able to point to a specific version is nice this particular approach means that the file becomes invalid if github changes their website.  I suppose any such change would happen over time and could be addressed if needed.

Another approach would be to use the same repository/subdirectory form that test files are going to use in #817 .

text/dialect_schema.yaml
EpsilonPrime: Changed to allow both the minimum and maximum.  The minimum is also useful for narrowing of the allowed semantics.

dialects/tests/test2.yaml
EpsilonPrime: I've split this Substrait Go example into two sections (one for integers and one for floats).  That way the rounding option isn't there when it shouldn't be.

text/dialect_schema.yaml
```diff
@@ -0,0 +1,203 @@
...
+          - EXPAND
+      sub_type:
+        type: array
+        description: >
+          sub_type corresponds to the primary oneof in any given relation type.
+          For example, in ReadRel it is the field name inside read_type and
+          in JoinRel it is the join_type.  If there is no oneof in a relation
+          sub_type should be empty.  Functionality not controlled by the
+          sub_type needs to be supported for the relation to be included in the
+          dialect.  For user-defined types, use the URL that describes the type
```
yongchul: Since `user-defined type` (or `user-defined relation` as @vbarua suggested) is mentioned here, I am wondering whether those supported `user-defined X` should be added to `supported_X` in terms of discoverability.

EpsilonPrime: User defined relations (extension single/multi/leaf) now have a message_type list for this purpose.  There are no user defined expressions so there's no supported added there.  I've added the message_types field for supported_types but there is no check yet to prevent use on other types.

text/dialect_schema.yaml
EpsilonPrime: It's not particularly clear how this is used in the Substrait Go but it appears to control whether the type can appear in a table.

dialects/tests/test1.yaml
yongchul: If the github links are perma links, I hope github keeps the promise. There is definitely value for link to official yaml files so that machine can read and deal with it, so definitely worth to host somewhere... For tests, repository solution also makes sense -- just whatever makes life easier. :)

text/dialect_schema.yaml
vbarua: I don't know the original intent, but I know that in some systems there are types that can only be used in literals and not columns. An example of this is the INTERVAL type in Snowflake.

See: https://docs.snowflake.com/en/sql-reference/data-types-datetime#interval-constants

This might be what this is related to. If that's the case, we should document it.

EpsilonPrime: Documented.

text/dialect_schema.yaml
```diff
@@ -0,0 +1,536 @@
...
+            Whether this type can be used as a column in a table.  Types listed
+            here are always supported as literals.
+        message_types:
+          type: array
+          description: >
+            The list of supported message types.  Use the URI (e.g.
+            type.googleapis.com/google.profile.Person) that is related to the
+            detail field.  Only valid for user defined types.
+          items:
+            type: string
```
vbarua: Could you add an example of what it would look like to indicate support for a user-defined type like point in https://github.com/substrait-io/substrait/blob/d430e521f203aec6a4e06731d4bfd68cdf61f443/extensions/extension_types.yaml#L2-L10?

For user-defined types, I'm not sure if there is a message URI that can be used as the canonical type, because they are defined in extensions.

Thinking about the key for this as well, we might want to include the dependency alias in the key for user-define functions, because there's nothing stopping people from defining a type with the same name in different extensions.

text/dialect_schema.yaml
```diff
@@ -0,0 +1,536 @@
...
+          For example, i32 or str_str.  Variadic arguments are not included in
+          the impl's list of parameters.
+        minItems: 1
+        items:
+          type: string
+      variadic:
+        type: object
+        description: >
+          If present, indicates the function accepts a variable number of arguments.
+          Specifies the minimum and/or maximum number of arguments allowed.
```
vbarua, line 223: I assume this exists to allow systems to further the restrict the number of variadic arguments beyond what might be declared in extensions like: https://github.com/substrait-io/substrait/blob/d430e521f203aec6a4e06731d4bfd68cdf61f443/extensions/functions_comparison.yaml#L265-L266

text/dialect_schema.yaml
```diff
@@ -0,0 +1,536 @@
...
+    required: [relation]
+  exchange_relation:
+    type: object
+    additionalProperties: false
+    properties:
+      relation:
+        const: EXCHANGE
+      kinds:
+        type: array
+        description: The list of valid exchange an exchange relation can perform.
```
vbarua: minor pluralization:
`The list of valid exchange an ...` to
`The list of valid exchanges an ...`

text/dialect_schema.yaml
```diff
@@ -0,0 +1,536 @@
...
+        const: MULTI_OR_LIST
+    required: [expression]
+  # embedded_function_expression not included due to insufficient proto support.
+  cast_expression:
+    type: object
+    additionalProperties: false
+    properties:
+      expression:
+        const: CAST
+    required: [expression]
```
vbarua, line 874: What do think about making indicating the failure options that are supported (i.e RETURN_NULL and THROW_EXCEPTION)

dialects/tests/test1.yaml
vbarua: Following up on #817, at the last sync we decide to stick to the implicit test lookup for the core tests, for now, and see how far that gets us.

With regards to my original comments, I think retaining the links as they are is fine. It's relatively straightforward to derive the machine readable link from it, and there are advantages to linking to the nicer and more human readable version when exploring the dialects.

With regards to versioning, that could also be a good idea. There's some implicit versioning happening in that the dialect refers to functions and types in extension explicitly, so even if new functions are added the dialects won't have access to them immediately. If functions get removed, that could be a problem, but that's also a breaking change so we should be avoiding anyways.

There's actually nothing stopping one dialect from using a link like
`https://github.com/substrait-io/substrait/blob/main/extensions/functions_arithmetic.yaml`,
and another from using a link like
`https://github.com/substrait-io/substrait/blob/v0.73.0/extensions/functions_arithmetic.yaml`
so we sort of already have versioning baked in.

Anyways, for me, no strong feelings about this.

dialects/tests/test2.yaml
```diff
@@ -0,0 +1,74 @@
...
+dependencies:
+  arithmetic:
+    https://github.com/substrait-io/substrait/blob/main/extensions/functions_arithmetic.yaml
+  boolean:
+    https://github.com/substrait-io/substrait/blob/main/extensions/functions_boolean.yaml
+  comparison:
+    https://github.com/substrait-io/substrait/blob/main/extensions/functions_comparison.yaml
+  string:
+    https://github.com/substrait-io/substrait/blob/main/extensions/functions_string.yaml
+supported_types:
```
vbarua: Friday use case musings. For types like PrecisionTimestamp, would we also want to be able to restrict the values parameters could take. For example, some systems might only supports up to microsecond timestamp precision `PT<6>`, while others may only support nanosecond precision `PT<9>`.

If that's a restriction worth encoding, do we also need something similar for types as you did for rels?

text/dialect_schema.yaml
EpsilonPrime: Right.  One may want to use a binary operator for max of 2 and a function for a minimum of 3 arguments.

text/dialect_schema.yaml
EpsilonPrime: Done.

text/dialect_schema.yaml
EpsilonPrime: There are two ways of defining user defined types in the proto.  The current method uses the anyof (because there isn't a type anchor to refer to outside of a plan).  Given that user defined types have some definition issues right now it's probably better to punt on this one until there is more clarity and/or a solid use case.

vbarua: Strictly speaking, the user-defined stuff in the protos is for [user-defined _literals_](https://github.com/substrait-io/substrait/blob/d430e521f203aec6a4e06731d4bfd68cdf61f443/proto/substrait/algebra.proto#L1119-L1134). The type is defined in the YAML extensions.

It's possible to define a type in the YAML without defining literal representations for it, and you can query for columns of that type with only the type defined in the YAML.

text/dialect_schema.yaml
```diff
@@ -0,0 +1,541 @@
...
+          The name used in the specific dialect implementation
+          (e.g., "+" might be the local_name for "add").
+      notation:
+        type: string
+        enum:
+          - INFIX
+          - POSTFIX
+          - PREFIX
+          - FUNCTION
+        default: FUNCTION
```
vbarua: minor: I've been playing around with generating dialects from this schema, and I don't think we should set a default value here.

`notation` only really makes sense for systems that accept some kind of query string (i.e. SQL), which to be fair is most systems, but part of the appeal of Substrait (to me) is to enable systems with plan first APIs

EpsilonPrime: It may be a difference in definition.  To me FUNCTION applies to the 80+% of functions which aren't operator related.  Not having to provide this makes sense especially when you have a dialect which isn't dealing with local syntax.

A potential approach to make this clear that it is local syntax related would be to rename this local_notation.  Other alternatives would be to rename function to something else.

text/dialect_schema.yaml
EpsilonPrime: Changed to use the same naming scheme for functions (a reference to a dependency followed by its name).

dialects/tests/test2.yaml
EpsilonPrime: This would likely apply to all parameterized types (like decimal).

Given that these types are used to deal with determining which function signature to use I'd prefer to leave that out.  The functions themselves can further restrict what can be done (helped along with the language for deriving output parameters).

proto/substrait/plan.proto
```diff
@@ -28,6 +28,13 @@ message Plan {
   // versions.
   Version version = 6;
 
+  // An URN pointing to a specific dialect that this plan conforms to.  The
+  // plan's producer is responsible for ensuring conformity.  Dialect URNs
+  // may include a version in the anchor.  If it does, the consumer should
+  // exactly match the URN (there are no semantics defined for at least
+  // this version).
+  optional string dialect = 8;
```
tokoko: iiuc, dialects define what part of the spec is or isn't supported by a specific consumer, plus additional mapping information that might be used by a consumer, is that right?

I'm not sure I understand the point of this field, then. dialects themselves don't alter plans, each plan might (and should) conform to multiple dialects. Some sort of a utility that takes plan and a dialect definition as an input and checks conformity sounds more logical to me instead of embedding this into the plan itself.

If the goal is to allow for the producer to signal conformity w/o the need for the consumer to double-check, I think the field should be repeated instead a single dialect string: `repeated string supported_dialects` or something similar.

vbarua: I agree that this is a little ambiguous at the moment. As @/tokoko points out, a plan could conform to multiple dialects.

What's your intended usecase for this?

text/dialect_schema.yaml
vbarua: What's an operator in this context, and what's a function which isn't an operator?

In pure Substrait world, everything is a function (or a relation). Some systems might care to distinguish between functions and infix, postfix and prefix operators, which is what I understand `notation` to be used for.

My annoyance had more to do with the fact that because this sets _a_ default here, code generated by things like [typify](https://github.com/oxidecomputer/typify/tree/main) will actually set it every time in the YAML, even in cases where you don't want to have a `notation` at all. That's actually not a good enough reason not to include something like this.

proto/substrait/plan.proto
westonpace: I could see this being useful as an informative field.  It's similar to the producer string.  However, we should expand the comment to mention that this is not required and has no impact on plan execution.  Maybe we could put it in the `Version` message?

text/dialect_schema.yaml
```diff
@@ -0,0 +1,575 @@
...
+      supported_impls:
+        type: array
+        description: >
+          List of implementations (parameter type lists) supported by this function.
+          For example, i32 or str_str.  Variadic arguments are not included in
+          the impl's list of parameters.
+        minItems: 1
+        items:
+          type: string
+      variadic:
```
westonpace, line 219: How does one know the type of the variadic argument?

westonpace (approved):
Now that I look at this it all comes back to me :laughing:.  Nicely crafted schema!  I appreciate the addition of relations and relation options.  I had thought that too daunting a task but you've kept it reasonable concise with a focus on what I think are major options.

text/dialect_schema.yaml
vbarua: For a function that is variadic, that last argument should be the one that is variadic and hence they type of the that should be the type of the variadic argument.

text/dialect_schema.yaml
```diff
@@ -0,0 +1,575 @@
...
+        description: >
+          Function-specific options that must be supported by the dialect.  These
+          are the exact list of options that must be used.  If no options are
+          listed here, no options must be present.
+      supported_impls:
+        type: array
+        description: >
+          List of implementations (parameter type lists) supported by this function.
+          For example, i32 or str_str.  Variadic arguments are not included in
+          the impl's list of parameters.
```
vbarua: I just noticed this. Why isn't the variadic argument included? As I understand it the signature of something like
```yaml
  -
    name: concat
    impls:
      - args:
          - value: "string"
            name: "input"
        variadic:
          min: 1
        options:
          null_handling:
            values: [ IGNORE_NULLS, ACCEPT_NULLS ]
        return: "string"
```
would be `concat:str`

EpsilonPrime: For multiple impls wouldn't the second impl have different arguments and thus a different argument signature?  I believe the argument signature can be computed so there's no need to make the name more noisy.

proto/substrait/plan.proto
EpsilonPrime: I'm okay dropping this for now.  We can add this when needed.

text/dialect_schema.yaml
vbarua: I'm realizing that my assumption was that the items in `supported_impls` were effectively the RHS of the [Function Signature Compound Names](https://substrait.io/extensions/#function-signature-compound-names)
```
               | -> this part
```

which would make it easy to associate the functions as defined in the extensions with those referenced in the dialects. Your schema, as is, is effectively the Function Signature Compound Names _except_ for variadic functions.

What do you think about specifying that the supported impls strings should be computed in the same way as the argument portion of function compound name? I feel that what would be more consistent.

EpsilonPrime: I guess this is as a good of a time as any to bring this up, but could we modify the function signature definition to make more sense?  It seems quite odd that you can't tell the difference between a two arg function and a two arg function with variadics.  Is this something we can define now for 1.0 and then make this forward compatible?

vbarua: > you can't tell the difference between a two arg function and a two arg function with variadics.

Is that something that comes up? Do you have an example of it? I would expect that the call to the 2 arg function would be equivalent to calling the variadic version with those same arguments.

EpsilonPrime: Let's take plus.  In order to implement conversion from an operator and still handle the function implementation you need two definitions.  They could go to the same implementation but one might go to a more optimized version.

Now let's add an overload for different types.  How do you tell any vs int if the variadics are missing?

If it's a name intended to differentiate the instance then it should be a differentiator.  If we are expected to use the provided function signature then there's no issue.  But we end up referring to them by name here.  Maybe we should be referring to name + signature and avoid the name issue altogether.

vbarua: > It seems quite odd that you can't tell the difference between a two arg function and a two arg function with variadics. Is this something we can define now for 1.0 and then make this forward compatible?

I think this is a reasonable goal. It is somewhat accomplishable now by giving the functions different names.
```
1: plus(i64, i64) -> plus:i64_i64
2.plus_variadic(i64, ... i64) -> plus_variadic:i64_i64 
```
though this is definitely hacky, and having the variadicness be part of the signature could be useful. For this specific example, you could make the case that the variadic plus should just be
```
plus(... i64) -> plus:i64
```

> Let's take plus. In order to implement conversion from an operator and still handle the function implementation you need two definitions.

I'm assuming that when your talking about conversion from operator and function implementations, you're talking about going from some system _to_ Substrait. Do you necessarily need 2 different definitions in Substrait for this? Substrait doesn't distinguish between operator and functions, and when mapping _into_ the target system, it could choose to use the systems operator or function implementation based on the number of arguments.

> Now let's add an overload for different types. How do you tell any vs int if the variadics are missing?

I'm not sure I understand the specific scenario you're discussing here. Do you mean if no variadic arguments are included in a specific _invocation_? From a plan perspective, the signature is baked into function reference:
https://github.com/substrait-io/substrait/blob/17de0366dc20e2c0c33f5668d410bf15aacc2572/proto/substrait/extensions/extensions.proto#L65-L66
The producing system would need to pick the specific variant to use in that case.

Taking a step back. If we want to consider changes to the function signatures, as well as handling of the additional problems you've identified, that probably warrants a dedicated issue. Do you think that's something that needs to be decided before we merge in your dialect work. Changing the signature in any way will probably require a migration anyways as it's baked into function resolution in the libraries.

text/dialect_schema.yaml
```diff
@@ -0,0 +1,575 @@
...
+additionalProperties: false
+type: object
+
+properties:
+  name:
+    type: string
+    description: Optional name of the dialect (e.g., "MySQL", "PostgreSQL", "Spark").
+  substrait_version:
+    type: string
+    description: The version of Substrait that this dialect applies for.  Optional.
```
yongchul: nit: two minor things.

1. `Version` proto message include a lot of fields and the intent of this is `<major>(.<minor>(.<patch>)?)?`. Not sure whether it is worth to mention what is expected here.
2. Is the model dialect schema per concrete Substrait version or there is some sort of compatibility level? For instance, if this field is saying, 1.2, would the dialect work for 1.3? How about 1.4 with a breaking change?

text/dialect_schema.yaml
```diff
@@ -0,0 +1,575 @@
...
+  dependencies:
+    type: object
+    description: >
+      A map of function YAML files depended on by this dialect.
+      Each key is an alias used by functions with the value being a URI.
```
yongchul: nit: URI or URN? :-)

text/dialect_schema.yaml
```diff
@@ -0,0 +1,575 @@
...
+    type: object
+    description: >
+      A map of types supported by this dialect.
+      The key is the Substrait name for the type.
+    additionalProperties:
+      type: object
+      properties:
+        local_name:
+          type: string
+          description: The SQL name for this type in the dialect.
```
yongchul: Should we use `SQL` here or something generic like `system preferred name` or simply `preferred name`?

text/dialect_schema.yaml
```diff
@@ -0,0 +1,575 @@
...
+        supported_as_column:
+          type: boolean
+          description: >
+            Whether this type can be used as a column in a table.  Types listed
+            here are always supported as literals.
+        type_definition:
+          type: string
+          description: >
```
yongchul: can we have a test for this just to show as an example?

text/dialect_schema.yaml
```diff
@@ -0,0 +1,575 @@
...
+          - EXTENSION_LEAF
+          - CROSS
+          - REFERENCE
+          - WRITE
+          - DDL
+          - UPDATE
+          - HASH_JOIN
+          - MERGE_JOIN
+          - NESTED_LOOP_JOIN
+          - CONSISTENT_PARTITION_WINDOW
```
yongchul, line 111: Do we need a gate checker to  catch not updating this file when we adding a new rel or a new expression? Perhaps, we should create a issue to track it.

text/dialect_schema.yaml
yongchul: Isn't `notation` here somewhat tangential to "substrait" dialect, given that everything is function? To me, this looks like just preferred stringify on printing (or unparsing).

Or, is this dialect meant to do something like `MySQL SQL text from/to Substrait` or `Pandas code from/to Substrait`?

text/dialect_schema.yaml
yongchul: +1 to change signature, which also requires a revision to incorporate type variation :)

vbarua: Updating the PR based on the sync discussion, and also have a seperate change to clarify the function signature for variadic functions: https://github.com/substrait-io/substrait/pull/863

text/dialect_schema.yaml
vbarua: Removing `substrait_version ` for now based on prior discussions.

text/dialect_schema.yaml
vbarua: Heh, good catch. Updated.

proto/substrait/plan.proto
vbarua: I've dropped this for now. Part of this is realizing we probably need to define the URN scheme for dialects in order to reference them 😅

text/dialect_schema.yaml
vbarua: Good point. Updated this from `SQL` to `system`.

text/dialect_schema.yaml
vbarua: I think `local_name` may not be precise enough to describe this. Made some updates in [`1a91386` (#816)](https://github.com/substrait-io/substrait/pull/816/commits/1a91386928593e92f0b1f7a525e68f5a05ddf295) to try and clarify this.

The intent with this is to _assist_ with mapping to and from the system the dialect is modelling. The dialect indicates which Substrait capabilities a system supports, and the `system_name` and `system_notation` help with translation to and from the system as you describe.

text/dialect_schema.yaml
vbarua: Now that've pointed this out, I'm not sure how it's meant to be used either. I'll try and come up with an example, or update it.

vbarua:
@yongchul questions made me go back, revisit and simplify (hopefully) a number of things. Notable changes:

* Converted `supported_types` into an `oneOf` that is either the base name of a type (i.e. BOOL, I64) or an object. Most of the objects have limited functionality (for now), except for the user-defined type object. This is what we had already done for expression and relations.
* Declared arrays of options as unique when possible. We had done this in a number of places, but figured we should do it everywhere it made sense to.
* Split the `dialect_function` combined `name` field into explicit `source` and `name` fields. Instead of having `name: arithmetic.add` we now just have `source: arithmetic, name: add`. This avoids having yet another way of referencing functions. `source` is also useful because it's used in user-defined types as well.
* Updated and consolidated tests. Instead of having test1 through test6, each which tested/demonstrated random bits of functionality, we now have dedicated tests files for functions, expressions, types and relations.

text/dialect_schema.yaml
```diff
@@ -0,0 +1,888 @@
...
+    type: array
+    description: Regular functions that operate on scalar values.
+    items:
+      $ref: "#/definitions/dialect_function"
+  aggregate_functions:
+    type: array
+    description: Functions that aggregate multiple rows into a single result.
+    items:
+      $ref: "#/definitions/dialect_function"
+  window_functions:
```
westonpace: Minor nit: we have **supported**_types, **supported**_expressions, **supported**_relations and then scalar_functions, aggregate_functions, and window_functions.  Should we add `supported_` in front of the functions?  Or should we (my preference) drop `supported_` from the others?

Maybe I've just been traumatized by https://rust-lang.github.io/rust-clippy/master/index.html#enum_variant_names

text/dialect_schema.yaml
```diff
@@ -0,0 +1,888 @@
...
+            const: PRECISION_TIMESTAMP
+  type_precision_timestamp_tz:
+    type: object
+    allOf:
+      - $ref: "#/definitions/type_base"
+      - type: object
+        required: [type]
+        properties:
+          type:
+            const: PRECISION_TIMESTAMP_TZ
```
westonpace: If a dialect lists support for precision timestamp does that just mean the system supports nanosecond precision or does it just mean the substrait<->system translation layer can handle all 12 digits of precision without error.

In other words, if a system with microsecond support only parses a substrait plan to compare values against a literal with nanosecond precision by truncating/rounding the nanoseconds from the literals and applying the plan does that system list this as a supported type?

westonpace (approved):
Some nits but I'm perfectly happy merging without them being addressed.

text/dialect_schema.yaml
vbarua: At the moment, it means that you support all instances of that type, yes.

Part of my intent with making all the types have objects is to allow us to add future refinements on a type basis for things like this. You can imagine have a `supported_precisions` array for all of the precisions timestamp types to further restrict what's supported. I'm not planning on making that part of this PR. For now it's either all or nothing.

text/dialect_schema.yaml
vbarua: I like this a lot actually. It make all the fields consistent, and makes it extra clear that we're not declaring functions. Updated.

text/dialect_schema.yaml
vbarua: > In other words, if a system with microsecond support only parses a substrait plan to compare values against a literal with nanosecond precision by truncating/rounding the nanoseconds from the literals and applying the plan does that system list this as a supported type?

To answer your second question, I think that if you can't consume a nanosecond timestamp _without_ truncation/rounding you can't claim to support it.

text/dialect_schema.yaml
yongchul: Thank you. `system_` seem better to me. I still believe these `system_` could go somewhere else but I also see the case how they could be useful for something like `Substrait2Sql -dialect postgresql` where code is generic but consumes the dialect specs as in this yaml schema. Oh, well, this is optional anyway so no harm to keep it around. It serves as an annotation at least so still being useful.

For the default, I don't have a strong opinion but I'd prefer to leave it completely optional. Perhaps, we can lift the **default** to global scope like **default_system_notation** under definitions.

text/dialect_schema.yaml
yongchul: I have a feeling that these domain of "supported" type arguments will be a thing pretty soon. :) Even if the system does not support nano second precision but do the right thing with respect to its coarser timestamp values stored in the system, it is okay to say it supports timestamp with nanosecond precision IMO (i.e., as long as the result is correct, that's fine).

text/dialect_schema.yaml
```diff
@@ -0,0 +1,888 @@
...
+      name:
+        type: string
+        description: >
+          The name of the function as declared in the extension it is defined in.
+      system_name:
+        type: string
+        description: >
+          The name for this function in the system modelled by the dialect.
+          For example, "+" might be the system name for "add".
+      system_notation:
```
EpsilonPrime: Consider renaming this local_notation for consistency.

EpsilonPrime:
I can't approve this logistically but I approve the changes.

text/dialect_schema.yaml
vbarua: I believe I've renamed all instances of `local` to `system`.

text/dialect_schema.yaml
vbarua: > that these domain of "supported" type arguments will be a thing pretty soon.

Same here, but this PR has been open long enough 😅

text/dialect_schema.yaml
vbarua: Good idea: https://github.com/substrait-io/substrait/issues/867

vbarua:
@yongchul's suggestion in https://github.com/substrait-io/substrait/pull/816#discussion_r2361275607 made me consider what I think is an even better approach for handling system specific metadata. I've implemented this in the latest commit [`15dd643` (#816)](https://github.com/substrait-io/substrait/pull/816/commits/15dd6439e9637a3a23a861bf0494693939bce782)

To summarize, we can encapsulate system specific information into `system_metadata` fields like
```yaml
supported_scalar_functions:
- source: arithmetic
  name: add
  system_metadata:
    name: +
    notation: INFIX
    # Calcite specific metadata
    SqlStdOperatorTable: CHECKED_PLUS
  required_options:
    overflow: ERROR
  supported_impls:
    - i32_i32
    - i64_i64

supported_types:
  - type: I32
    system_metadata:
      name: INTEGER
      supported_as_column: true
```

For the core fields on functions and types we can enable `additionalProperties: false`, so that the common fields are truly common across dialects. They must match exactly what is in the schema. For the `system_metadata` object though, we include common fields to help guide usage but don't restrict the presence of additional fields.

This actually better enables one of my original use cases for dialects, which is driving transformation from Substrait to Calcite. The additional metadata in the example can be used to perform lookups of Calcite classes during the mapping process. This is the kind of information that would only ever be useful to Calcite, so it would not make sense to have in the core schema, but it's useful to have a location to put information like this.

Unless YongChul snipes me with yet another great idea, this should be the last modification 😅 

cc: @EpsilonPrime @westonpace

text/dialect_schema.yaml
```diff
@@ -0,0 +1,910 @@
...
+          min:
+            type: integer
+            description: "Minimum number of arguments required"
+            minimum: 1
+          max:
+            type: integer
+            description: "Maximum number of arguments allowed"
+            minimum: 1
+        additionalProperties: false
+  system_function_metadata:
```
yongchul, line 234: Okay, this is a very reasonable middle ground and I like it. :-)

One quick question -- this is due to my ignorance of YAML -- so in the example you had a calcite specific metadata but I don't see anything like that in this schema. Is this the default mode (i.e., if you have some properties that are not here, it is okay to have them)? Or, we need some sort of extension mechanism or some generic key-value pairs or attributes to define?

EpsilonPrime: One potential issue is there is less checking for mistakes if you don't make unknown fields an error.  If you want unknown key value pairs you could put them in an arbitrary key/value data section.

vbarua: > so in the example you had a calcite specific metadata but I don't see anything like that in this schema.

For objects, there is an `additionalProperties` field that you can configure in the schema. By default it is set to false so you can have fields not in the schema. If it is set to true, you can have additional fields.

> One potential issue is there is less checking for mistakes if you don't make unknown fields an error. 

Good point, the way it's structured at this moment means that if you typo something like
```
system_metadata:
    nam: +
    notatio: INFIX
```
it won't be detected as a validation failure.

I've updated the PR to have make the structure
```
supported_scalar_functions:
- source: arithmetic
  name: add
  system_metadata:
    name: +
    notation: INFIX
    # System specific metadata
    extra:
      SqlStdOperatorTable: CHECKED_PLUS
```

I set additionalProperties to `true` for the fields under `system_metadata`, so that we can validate commons fields, and then the `extra` field is an object is where additional arbitrary information can be handled. Let me know if that makes sense, or if you have a better name than `extra` for it 😅

yongchul: How about `extensions` 🤣 ? `annotations` is another candidate.

vbarua: You know what, I'm going to remove the `extra` stuff for now and propose it in a follow-up PR with it's own discussion in order to close out the current changes.

`system_metadata` now looks like

```
supported_scalar_functions:
- source: arithmetic
  name: add
  system_metadata:
    name: +
    notation: INFIX
```
and we have checks for the fields in it.

If this seems reasonable, I'll merge it in later today as we've got 3 PMC approvales (Me, David and Weston)

yongchul: @vbarua Yes, we can definitely leave out whatever `extra` for follow up. :)

EpsilonPrime:
Later is good.  Approved

vbarua:
Merged based on approval from 3 PMCs (myself, David and Weston)

PR #817, opened 2025-05-21
feat: reference tests within extension files

vbarua:


vbarua:
Talking w/ @EpsilonPrime about dialects, we realized that it would be helpful to have the function tests be referenced by extensions explicitly, as this would allow Substrait tooling to access the tests programatically.

In this PR I've pulled out a small section of David's work as this is separate from the dialect work, and useful in it's own regard.

extensions/functions_arithmetic.yaml
```diff
@@ -1,5 +1,8 @@
 %YAML 1.2
 ---
+tests:
+  - "https://github.com/substrait-io/substrait/tree/main/tests/cases/arithmetic/"
```
vbarua, line 4: I've only updated a handful of extension files for now. Once we've agreed that something like this make sense, I will update alll of them.

extensions/functions_boolean.yaml
```diff
@@ -1,5 +1,8 @@
 %YAML 1.2
 ---
+tests:
+  - "https://github.com/substrait-io/substrait/tree/main/tests/cases/boolean/"
```
vbarua, line 4: I wonder if this should actually be something like:
```
tests:
  - github:
     - repo: https://github.com/substrait-io/substrait.git
       directory: tests/cases/boolean
```
because "https://github.com/substrait-io/substrait/tree/main/tests/cases/boolean/" isn't really a directory.

EpsilonPrime: The lack of a way to enumerate the files underneath a web page is annoying.  Listing multiple files would work but it would typically require adding a test file for each defined function which could be unwieldy.  Providing both git and local files test types would be reasonable and would avoid verbosity.

vbarua: I've updated this with just a git file type for now.

text/simple_extensions_schema.yaml
```diff
@@ -304,3 +311,23 @@ $defs:
             window_type:
               type: string
               enum: [STREAMING, PARTITION]
+
+  git_repo:
```
EpsilonPrime, line 315: git_repo isn't used in the examples above.

extensions/functions_arithmetic.yaml
```diff
@@ -1,5 +1,10 @@
 %YAML 1.2
 ---
+tests:
+  - repository_url: https://github.com/substrait-io/substrait.git
```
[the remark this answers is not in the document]
vbarua: Good idea. Made repository_url optional to allow this.

EpsilonPrime (dismissed):

jacques-n (commented):
I'm confused by the changes here. Aren't these relationships already derivable? Why do we need to make humans write them explicit?

vbarua:
> I'm confused by the changes here. Aren't these relationships already derivable?

They're derivable in the sense that for the core spec we implicitly know that the tests for extensions in https://github.com/substrait-io/substrait/blob/main/extensions/functions_arithmetic.yaml are found in the associated directory https://github.com/substrait-io/substrait/tree/main/tests/cases/arithmetic.

Another option would be to document this more explicitly on [substrait.io](https://substrait.io/), but that would only apply for tests in the core spec.

The intent with this is to make the tests more machine discoverable, but also more human discoverable because the extension tells you were you can find them.

This was originally part of https://github.com/substrait-io/substrait/pull/816, but we pulled it out because it could be proposed as its own change but is highly synergistic as part of testing dialects. The flow I'm imagining is:
1. A system (i.e. DataFusion) defines a dialect which points to both core extensions and system specific extensions.
2. The dialect tells the system which extensions are in use, allowing the system to fetch the extensions.
3. The extensions tell the system which tests are available, allowing the system to fetch them.
4. The system can now execute the tests for functionality it claims to support to help validate that it's support is correct.

vbarua:
Going to hold of these changes for now and see how far we can get with the implicit lookup strategy.

PR #825, opened 2025-07-14
fix: typo on window functions site page

lorenarosati:
The window functions page on the site refers to binding aggregate functions instead of window functions.


westonpace:
Can merge once CLA check is passing

vbarua (approved):

PR #833, opened 2025-07-25
docs: clarify advanced extensions

wackywendell:
The Advanced Extensions section of the docs seemed somewhat mismatched from what was in the protobufs, so I tried to clarify it, based on what was in the protobuf definitions.

Thoughts welcome! Also could use some sanity-checking on the details!

#### Naming

I also included custom relations (`ExtensionLeafRel`, …) and custom reads and writes (`ExtensionTable`, …) all under the heading "Advanced Extensions", even though there is an `AdvancedExtension` message that doesn't cover those. The name "Advanced Extension" seems a bit ambiguous here - does it cover all of the above, or only the enhancements and optimizations in the `AdvancedExtension` message? - but it seems to be what we have, so I went with it. Thoughts?

#### Guidance

This is a bit low on guidance on when to use which - e.g. `ExtensionLeafRel`, `ExtensionTable`, or `ReadRel.advanced_extension.enhancement` could potentially all be used for an unusual kind of read. I'm not sure what the guidance here should be, or if there should be any, so I left it out; I'm not sure any guidance here would be that helpful.

### DIFF
--- a/site/docs/extensions/index.md
+++ b/site/docs/extensions/index.md
@@ -159,7 +159,7 @@ The simplest forms of advanced extensions use the `AdvancedExtension` message, w
 #### Enhancements
 
 - Modify the semantic behavior of operations
-- Must be understood by consumers or the plan cannot be executed correctly
+- Must be understood by consumers, or else the plan cannot be executed correctly
 - Only one enhancement per message
 - Examples: specialized join conditions (e.g. fuzzy matching, geospatial)
 


site/docs/extensions/index.md
```diff
@@ -132,15 +132,78 @@ The `any[\d]` types (i.e. `any1`, `any2`, ..., `any9`) impose an additional rest
...
-| Plan Extensions                      | Semantic and/or optimization based additions at the plan level. |
-
-Because extension mechanisms are different for each serialization format, please refer to the corresponding serialization sections to understand how these extensions are defined in more detail.
+Advanced extensions provide a way to embed custom functionality that goes beyond the standard YAML-based simple extensions. Unlike simple extensions, advanced extensions use Protocol Buffer's `google.protobuf.Any` type to embed arbitrary extension data directly into Substrait messages.
+
+### How Advanced Extensions Work
+
+Advanced extensions come in several main forms, discussed below:
+
+1. Embedded extensions: These use the `AdvancedExtension` message for adding custom data to existing Substrait elements
```
yongchul: I'm not sure or don't remember whether `element` is commonly used across the doc. Now that I check again, advanced extensions are generously sprinkled over the proto file... :) That said, it would be clearer to mention the scope of the embedded extension like `adding custom data to enclosing Substrait message (or element)`.

The above mentioned protobuf and call out `Any`. So I'm wondering whether we can stick to `Substraite message` rather than using `element`.

site/docs/extensions/index.md
```diff
@@ -132,15 +132,78 @@ The `any[\d]` types (i.e. `any1`, `any2`, ..., `any9`) impose an additional rest
...
+
+The `AdvancedExtension` message can be attached to various parts of a Substrait plan:
+
+| Location                          | Usage                                       |
+| --------------------------------- | ------------------------------------------- |
+| **`Plan`**                        | Global extensions affecting the entire plan |
+| **`RelCommon`**                   | Extensions for any relational operator      |
+| **Relations** (e.g. `ProjectRel`) | Extensions for a specific relation type     |
+| **Hints**                         | Extensions within optimization hints        |
+| **`ReadRel.NamedTable`**          | Add custom metadata to named table references |
```
yongchul: nit: remove `Add` to be consistent with the rest of Usages? (i.e., either all starts with verbs or nouns)

yongchul: looks like `LocalFiles` also has advanced extension.

site/docs/extensions/index.md
```diff
@@ -132,15 +132,78 @@ The `any[\d]` types (i.e. `any1`, `any2`, ..., `any9`) impose an additional rest
...
+| **`WriteRel.NamedObjectWrite`**   | Add custom metadata to write targets        |
+| **`DdlRel.NamedObjectWrite`**     | Add custom metadata to DDL targets          |
```
yongchul: noticed some of the advanced extension tags are not consistent in these Ddl related messages. Just note to self.

site/docs/extensions/index.md
```diff
@@ -132,15 +132,78 @@ The `any[\d]` types (i.e. `any1`, `any2`, ..., `any9`) impose an additional rest
...
+  // Enhancements alter semantics. Cannot be ignored by a consumer.
+  google.protobuf.Any enhancement = 2;
+}
+```
+
+#### Optimizations
+
+- Provide hints to improve performance but don't change the meaning of operations
+- Can be safely ignored by consumers that don't understand them
+- Examples: memory usage hints, preferred algorithms, caching strategies
```
yongchul, line 157: nit: example goes last?

site/docs/extensions/index.md
```diff
@@ -132,15 +132,78 @@ The `any[\d]` types (i.e. `any1`, `any2`, ..., `any9`) impose an additional rest
...
+- Multiple optimizations can be attached to a single element
+
+#### Enhancements
+
+- Modify the semantic behavior of operations
+- Must be understood by consumers or the plan cannot be executed correctly
+- Examples: custom aggregation logic, specialized join conditions, new relation types
```
yongchul: relational operators? or just custom operators?

site/docs/extensions/index.md
```diff
@@ -132,15 +132,78 @@ The `any[\d]` types (i.e. `any1`, `any2`, ..., `any9`) impose an additional rest
...
-
-| Extension Type                       | Description                                                  |
-| ------------------------------------ | ------------------------------------------------------------ |
-| Relation Modification (semantic)     | Extensions to an existing relation that will alter the semantics of that relation. These kinds of extensions require that any plan consumer understand the extension to be able to manipulate or execute that operator. Ignoring these extensions will result in an incorrect interpretation of the plan. An example extension might be creating a customized version of Aggregate that can optionally apply a filter before aggregating the data. <br /><br />Note: Semantic-changing extensions shouldn't change the core characteristics of the underlying relation. For example, they should *not* change the default direct output field ordering, change the number of fields output or change the behavior of physical property characteristics. If one needs to change one of these behaviors, one should define a new relation as described below. |
```
yongchul, line 170: These old content looks good to me (at least the guideline parts). Perhaps, we could reorganize them in each section and call out as notes.

site/docs/extensions/index.md
wackywendell: Makes sense, `message` makes more sense than `element` for consistency!

site/docs/extensions/index.md
wackywendell: Yes, I can clarify a bit here!

wackywendell:
@yongchul - thanks for the review! Good comments - I've updated to address them. Can you give them a look?

Thanks!

site/docs/extensions/index.md
```diff
@@ -132,15 +132,91 @@ The `any[\d]` types (i.e. `any1`, `any2`, ..., `any9`) impose an additional rest
...
+  repeated google.protobuf.Any optimization = 1;
+
+  // Enhancements alter semantics. Cannot be ignored by a consumer.
+  google.protobuf.Any enhancement = 2;
+}
+```
+
+!!! note "Enhancements vs Optimizations"
+
+Use **optimizations** for performance hints that don't change semantics and can be safely ignored. Use **enhancements** for semantic changes that must be understood by consumers or the plan cannot be executed correctly.
```
yongchul:
```suggestion
* Use **optimizations** for performance hints that don't change semantics and can be safely ignored.
* Use **enhancements** for semantic changes that must be understood by consumers or the plan cannot be executed correctly.
```

site/docs/extensions/index.md
```diff
@@ -132,15 +132,91 @@ The `any[\d]` types (i.e. `any1`, `any2`, ..., `any9`) impose an additional rest
...
+
+    Custom relations are the most flexible but least interoperable option. In most cases it is better to use enhancements to existing relations rather than defining new custom relations, as it means existing code patterns can easily be extended to work with the additional properties.
+
+### Custom Read and Write Types
+
+The third form of advanced extensions allows you to define extension data sources and destinations:
+
+| Extension Type                 | Description                                   | Examples                     |
+| ------------------------------ | --------------------------------------------- | ---------------------------- |
+| **`ReadRel.ExtensionTable`**   | Define entirely new table source types        | APIs, specialized formats    |
```
yongchul: nit: is `entirely` necessary?

site/docs/extensions/index.md
```diff
@@ -132,15 +132,91 @@ The `any[\d]` types (i.e. `any1`, `any2`, ..., `any9`) impose an additional rest
...
+These extension relations are first-class relation types in Substrait and can be used anywhere a standard relation would be used.
+
+!!! note "Interoperability Guidance"
+
+    Custom relations are the most flexible but least interoperable option. In most cases it is better to use enhancements to existing relations rather than defining new custom relations, as it means existing code patterns can easily be extended to work with the additional properties.
+
+### Custom Read and Write Types
+
+The third form of advanced extensions allows you to define extension data sources and destinations:
+
```
yongchul: This section can also have the `note` like above sections that checking with the community before going down the extension path. If the scenario turns out to be common enough, it may go directly to the specification.

yongchul (commented):
Thank you for taking suggestions! Looks good to me :)

yongchul (approved):

site/docs/extensions/index.md
wackywendell: 👍  - also realized I needed to indent these so they would be in the note.

site/docs/extensions/index.md
wackywendell: No! Dropped.

wackywendell:
Thanks again! Incorporated. I also realized I had the `!!! note` sections incorrect (they needed indenting), so I fixed that, and ran `mkdocs` locally to verify it worked - and it did!

wackywendell:
^ Thanks, @yongchul for the merge! I'm wondering - what do we need to do next to advance this? Do we need additional reviewers so we can get it approved and merged?

yongchul:
> ^ Thanks, @yongchul for the merge! I'm wondering - what do we need to do next to advance this? Do we need additional reviewers so we can get it approved and merged?

Thank you for your patience! We will need another SMC review I believe. I'll bring this up in the sync up.

site/docs/extensions/index.md
```diff
@@ -126,15 +126,95 @@ The `any[\d]` types (i.e. `any1`, `any2`, ..., `any9`) impose an additional rest
...
+3. Custom read/write types: for defining new ways to read from or write to data sources
+
+### Embedded Extensions via `AdvancedExtension`
+
+The simplest forms of advanced extensions use the `AdvancedExtension` message, which contains two types of extensions:
+
+```proto
+message AdvancedExtension {
+  // Optimizations are helpful information that don't influence semantics.
+  // May be ignored by a consumer.
```
yongchul: There is a new syntax along the line below. Could you update it? See https://github.com/substrait-io/substrait/pull/876/files#diff-d70112026895153901e9800a631fb22423a53b1c27caa25db4be85ffabe49556R238 for example

```proto
%%% proto.extensions.AdvancedExtension %%%
```

site/docs/extensions/index.md
```diff
@@ -126,15 +126,95 @@ The `any[\d]` types (i.e. `any1`, `any2`, ..., `any9`) impose an additional rest
...
-Because extension mechanisms are different for each serialization format, please refer to the corresponding serialization sections to understand how these extensions are defined in more detail.
+Advanced extensions provide a way to embed custom functionality that goes beyond the standard YAML-based simple extensions. Unlike simple extensions, advanced extensions use Protocol Buffer's `google.protobuf.Any` type to embed arbitrary extension data directly into Substrait messages.
+
+### How Advanced Extensions Work
+
+Advanced extensions come in several main forms, discussed below:
+
+1. Embedded extensions: These use the `AdvancedExtension` message for adding custom data to existing Substrait messages
+2. Custom relation types: For defining entirely new relational operations
+3. Custom read/write types: for defining new ways to read from or write to data sources
```
benbellick: Any thoughts on perhaps putting the custom read/write types before the custom relation types, both in this list and in sections below? 

My thinking is that we want to discourage people from using custom relations unless they really can't express what they want to do any other way. 

I worry that people trying to do a very quick pass of the documentation will just assume that the order in which these items are presented reflect the preferred order of extending substrait.

site/docs/extensions/index.md
```diff
@@ -126,15 +126,88 @@ The `any[\d]` types (i.e. `any1`, `any2`, ..., `any9`) impose an additional rest
...
-
-Because extension mechanisms are different for each serialization format, please refer to the corresponding serialization sections to understand how these extensions are defined in more detail.
+Advanced extensions provide a way to embed custom functionality that goes beyond the standard YAML-based simple extensions. Unlike simple extensions, advanced extensions use Protocol Buffer's `google.protobuf.Any` type to embed arbitrary extension data directly into Substrait messages.
+
+### How Advanced Extensions Work
+
+Advanced extensions come in several main forms, discussed below:
+
+1. Embedded extensions: These use the `AdvancedExtension` message for adding custom data to existing Substrait messages
+2. Custom read/write types: for defining new ways to read from or write to data sources
```
benbellick:
```suggestion
2. Custom read/write types: For defining new ways to read from or write to data sources
```
nit: consistent capitalization with the other items.

site/docs/extensions/index.md
```diff
@@ -126,15 +126,88 @@ The `any[\d]` types (i.e. `any1`, `any2`, ..., `any9`) impose an additional rest
...
+The simplest forms of advanced extensions use the `AdvancedExtension` message, which contains two types of extensions:
+
+```proto
+%%% proto.extensions.AdvancedExtension %%%
+```
+
+!!! note "Enhancements vs Optimizations"
+
+    * Use **optimizations** for performance hints that don't change semantics and can be safely ignored.
+    * Use **enhancements** for semantic changes that must be understood by consumers or the plan cannot be executed correctly.
```
benbellick, line 150: nit: perhaps we could move this section to after the two sections below? It feels like it makes sense to compare them only after explaining what each thing is.

site/docs/relations/physical_relations.md
```diff
@@ -34,12 +34,12 @@ The hash equijoin join operator will build a hash table out of one input (defaul
...
-| Inputs               | 2                                                            |
-| Outputs              | 1                                                            |
-| Property Maintenance | Distribution is maintained. Orderedness is eliminated.       |
-| Input Order   | Same as the [Join](logical_relations.md#join-operator) operator. |
+| Signature            | Value                                                             |
+| -------------------- | ----------------------------------------------------------------- |
+| Inputs               | 2                                                                 |
+| Outputs              | 1                                                                 |
+| Property Maintenance | Distribution is maintained. Orderedness is eliminated.            |
+| Input Order          | Same as the [Join](logical_relations.md#join-operation) operator. |
```
benbellick, line 38: Thank you for doing all of these changes here 🚀

site/docs/extensions/index.md
```diff
@@ -126,15 +126,88 @@ The `any[\d]` types (i.e. `any1`, `any2`, ..., `any9`) impose an additional rest
...
+| ---------------------- | ----------------------------------------------- | -------- |
+| **`ExtensionLeafRel`**   | Custom relations with no inputs | Custom table sources   |
+| **`ExtensionSingleRel`** | Custom relations with one input | Custom transforms      |
+| **`ExtensionMultiRel`**  | Custom relations with multiple inputs | Custom joins     |
+
+These extension relations are first-class relation types in Substrait and can be used anywhere a standard relation would be used.
+
+!!! note "Interoperability Guidance"
+
+    Custom relations are the most flexible but least interoperable option. In most cases it is better to use enhancements to existing relations rather than defining new custom relations, as it means existing code patterns can easily be extended to work with the additional properties.
```
benbellick: Could be worth considering breaking this out into a proper section titled:
> When to Use What

or something like that. I think explicit guidance could be useful. All of that being said, I think this is great!

benbellick (approved):
Left some minor comments but this is a welcome change!

benbellick:
@yongchul @wackywendell As I understand it, the governance rules dictate [that only one SMC approval is necessary to get documentation changes in.](https://substrait.io/governance/#substrait-voting-process)

Since @yongchul is an SMC, do we have to wait for any other approvals? Or can we just merge it?

site/docs/extensions/index.md
wackywendell: Hm - I see your point, although there's already a note there about Enhancement Constraints. Should we put the two notes one after the other, or leave this where it is?

yongchul:
> @yongchul @wackywendell As I understand it, the governance rules dictate [that only one SMC approval is necessary to get documentation changes in.](https://substrait.io/governance/#substrait-voting-process)
> 
> Since @yongchul is an SMC, do we have to wait for any other approvals? Or can we just merge it?

I'm going to merge after re-reading the changes one last time. :smile

site/docs/extensions/index.md
```diff
@@ -126,15 +126,88 @@ The `any[\d]` types (i.e. `any1`, `any2`, ..., `any9`) impose an additional rest
...
-| Plan Extensions                      | Semantic and/or optimization based additions at the plan level. |
-
-Because extension mechanisms are different for each serialization format, please refer to the corresponding serialization sections to understand how these extensions are defined in more detail.
+Advanced extensions provide a way to embed custom functionality that goes beyond the standard YAML-based simple extensions. Unlike simple extensions, advanced extensions use Protocol Buffer's `google.protobuf.Any` type to embed arbitrary extension data directly into Substrait messages.
+
+### How Advanced Extensions Work
+
+Advanced extensions come in several main forms, discussed below:
+
+1. Embedded extensions: These use the `AdvancedExtension` message for adding custom data to existing Substrait messages
```
yongchul, line 135: not in this PR but it would be great if we can just easily reference this to the actual proto message definition. perhaps, onetime clean up is necessary.

site/docs/extensions/index.md
benbellick: Yeah that's a good point. 🤷 lets just leave it as it is for now then.

site/docs/extensions/index.md
```diff
@@ -126,15 +126,88 @@ The `any[\d]` types (i.e. `any1`, `any2`, ..., `any9`) impose an additional rest
...
+- Can be safely ignored by consumers that don't understand them
+- Multiple optimizations can be attached to a single message
+- Examples: memory usage hints, preferred algorithms, caching strategies
+
+#### Enhancements
+
+- Modify the semantic behavior of operations
+- Must be understood by consumers or the plan cannot be executed correctly
+- Only one enhancement per message
+- Examples: specialized join conditions (e.g. fuzzy matching, geospatial) or sorting (e.g. clustering)
```
yongchul: `clustering` is unclear. can we have more detailed example or expand `clustering` a bit? Fetch with tie is a good example but I have a pending PR #797 for this 😆

site/docs/extensions/index.md
```diff
@@ -126,15 +126,88 @@ The `any[\d]` types (i.e. `any1`, `any2`, ..., `any9`) impose an additional rest
...
+
+| Extension Type                 | Description                          | Examples                     |
+| ------------------------------ | ------------------------------------ | ---------------------------- |
+| **`ReadRel.ExtensionTable`**   | Define new table source types        | APIs, specialized formats    |
+| **`WriteRel.ExtensionObject`** | Define new write destination types   | APIs, specialized formats    |
+| **`DdlRel.ExtensionObject`**   | Define new DDL destination types     | Catalogs, schema registries  |
+
+!!! note "Consider Core Specification First"
+
+    Before implementing custom read/write types as extensions, consider checking with the Substrait community. If your scenario turns out to be common enough, it may be more appropriate to add it directly to the specification rather than as an extension.
```
yongchul: link to https://substrait.io/community/#get-in-touch

yongchul (changes requested):

yongchul:
@wackywendell thank you for the changes and sorry about taking it so long! Please resolve the two minor changes and it looks good to me to go!

site/docs/extensions/index.md
wackywendell: Good point. I was thinking something like similarity matching, but that's a bit of a complicated example to explain in a few words, so I switched for locale-aware string ordering; does that make sense to you?

wackywendell:
Thanks, @yongchul! I updated, let me know how it looks to you - especially the sorting example, clustering was unclear as you noted!

site/docs/extensions/index.md
yongchul: locale aware sort is possible already as you can specify a custom comparison operation per field. Shall we just drop it for now and add an example later once we have a good one?

wackywendell: Oh, you're right - I hadn't seen the custom comparison operator before.

I've dropped the sorting example.

jacques-n:
The one issue I see with the extensions updates here is that they act like the only way to express something is using protobuf. Our goal was to spec first, with proto being an implementation. That being said, the reality is a bit different than that.

wackywendell:
@jacques-n - good point. I was looking at it, and while I think most of the protobuf references are hard to get away from - e.g. field names, hierarchy - the top-level description did clearly treat the Protobuf representation as the only one, so I changed that; see https://github.com/substrait-io/substrait/pull/833/commits/5fce28307f936b719c18e89df0a26abc8b6b13c7.

For the other places where it references Protobuf messages directly, that seems to be in line with the rest of the docs, and also seems like a good place to show implementations; and in theory, they could also be updated to multiple 'dropdowns' showing the representations in various forms if other implementations existed.

Does that make sense to you, or were there more specific ways we could address this?

wackywendell:
Note - CI failure seems to be an authentication / rate-limiting error:
```
Warning: No github_token supplied, API requests will be subject to stricter rate limiting
Setting up buf version "1.50.0"
Resolving the download URL for the current platform...
Error: API rate limit exceeded for 52.159.247.226. (But here's the good news: Authenticated requests get a higher rate limit. Check out the documentation for more details.) - https://docs.github.com/rest/overview/resources-in-the-rest-api#rate-limiting
```

site/docs/relations/physical_relations.md
```diff
@@ -2,111 +2,103 @@
 
 There is no true distinction between logical and physical operations in Substrait. By convention, certain operations are classified as physical, but all operations can be potentially used in any kind of plan. A particular set of transformations or target operators may (by convention) be considered the "physical plan" but this is a characteristic of the system consuming substrait as opposed to a definition within Substrait.
```
vbarua, line 4: For other reviewers, this file only has whitespace changes and one typo fix.

site/docs/extensions/index.md
```diff
@@ -126,15 +126,88 @@ The `any[\d]` types (i.e. `any1`, `any2`, ..., `any9`) impose an additional rest
...
+#### Enhancements
+
+- Modify the semantic behavior of operations
+- Must be understood by consumers or the plan cannot be executed correctly
+- Only one enhancement per message
+- Examples: specialized join conditions (e.g. fuzzy matching, geospatial)
+
+!!! note "Enhancement Constraints"
+
+    Semantic-changing extensions shouldn't change the core characteristics of the underlying relation. For example, they should *not* change the default direct output field ordering, change the number of fields output or change the behavior of physical property characteristics. If one needs to change one of these behaviors, one should define a new relation as described below.
```
vbarua: I might advocate for changing the language here from `should not change` to `avoid changing`.

It feels a little strong to say that enhancements _should not_ change core characteristics, especially because it's hard to define what a "core characteristic" is or is not. Your examples of
> change the default direct output field ordering,

and
> change the number of fields output

are concrete enough to enforce, but I'm not sure what
> change the behavior of physical property characteristics.

would entail.

site/docs/extensions/index.md
```diff
@@ -126,15 +126,88 @@ The `any[\d]` types (i.e. `any1`, `any2`, ..., `any9`) impose an additional rest
...
+
+### Custom Relations
+
+The third form of advanced extensions provides entirely new relational operations via dedicated extension relation types. These allow you to define custom relations while maintaining proper integration with the type system:
+
+| Relation Type          | Description                                     | Examples |
+| ---------------------- | ----------------------------------------------- | -------- |
+| **`ExtensionLeafRel`**   | Custom relations with no inputs | Custom table sources   |
+| **`ExtensionSingleRel`** | Custom relations with one input | Custom transforms      |
+| **`ExtensionMultiRel`**  | Custom relations with multiple inputs | Custom joins     |
```
vbarua: minor/pedantic: You went out of your way to nicely align all the other tables, but not this one 😅

site/docs/extensions/index.md
```diff
@@ -126,15 +126,88 @@ The `any[\d]` types (i.e. `any1`, `any2`, ..., `any9`) impose an additional rest
...
+
+These extension relations are first-class relation types in Substrait and can be used anywhere a standard relation would be used.
+
+#### When to Use What
```
vbarua:
```suggestion
### When to Use What
```

We should pop this out 1 level as it's currently nested under `Custom Relations`, when I think it should be at the same level in the nav:

Image

site/docs/extensions/index.md
```diff
@@ -126,15 +126,88 @@ The `any[\d]` types (i.e. `any1`, `any2`, ..., `any9`) impose an additional rest
...
+
+Custom relations are the most flexible but least interoperable option. In most cases it is better to use enhancements to existing relations rather than defining new custom relations, as it means existing code patterns can easily be extended to work with the additional properties.
```
vbarua: Another piece of guidance that might make sense is to push for users to use the least powerful extension mechanism that they need. i.e Don't add a custom relation when a enhancment of an existing relation would work.

vbarua (commented):
Overall, this really helps provide guidance as to how to utilize extensions.

With regards to the protos, this does lean on them a fair bit, but so does most of our documentation at this point and I'd rather deal with that as its own thing because this content is super useful to have.

I did leave a couple of comments, but this does feel like something we could get in this week.

benbellick:
@wackywendell I reopened this PR. I'm not sure you are aware that there are some unaddressed comments from @vbarua. 

If you don't have capacity to address those comments, I'm happy to push the commits for it myself. 

But I want to see this PR get in! I think this is a valuable contribution and we should see it through 👍

site/docs/extensions/index.md
wackywendell: Yes, makes sense!

wackywendell:
@benbellick @vbarua - thanks for the comments, and sorry for the delay here @benbellick - thanks for the nudge. I've addressed the comments, let me know what you think!

site/docs/extensions/index.md
```diff
@@ -144,34 +123,91 @@ The `any` type indicates that the argument can take any possible type. In the `f
...
+#### Enhancements
+
+- Modify the semantic behavior of operations
+- Must be understood by consumers or the plan cannot be executed correctly
+- Only one enhancement per message
+- Examples: specialized join conditions (e.g. fuzzy matching, geospatial)
+
+!!! note "Enhancement Constraints"
+
+    Semantic-changing extensions shouldn't change the core characteristics of the underlying relation. For example, they should avoid changing the default direct output field ordering or the number of fields output. If one needs to change one of these behaviors, one should define a new relation as described below.
```
yongchul: described `below` --> link to the Custom Relations

site/docs/extensions/index.md
```diff
@@ -144,34 +123,91 @@ The `any` type indicates that the argument can take any possible type. In the `f
...
+    Before implementing custom read/write types as extensions, consider [checking with the Substrait community](https://substrait.io/community/#get-in-touch). If your scenario turns out to be common enough, it may be more appropriate to add it directly to the specification rather than as an extension.
+
+### Custom Relations
+
+The third form of advanced extensions provides entirely new relational operations via dedicated extension relation types. These allow you to define custom relations while maintaining proper integration with the type system:
+
+| Relation Type            | Description                                     | Examples               |
+| ------------------------ | ----------------------------------------------- | ---------------------- |
+| **`ExtensionLeafRel`**   | Custom relations with no inputs                 | Custom table sources   |
+| **`ExtensionSingleRel`** | Custom relations with one input                 | Custom transforms      |
```
yongchul: nit: transforms -> relational transformations

site/docs/extensions/index.md
```diff
@@ -144,34 +123,91 @@ The `any` type indicates that the argument can take any possible type. In the `f
...
-| New Write Types                      | Similar to a read type but for writes. As with reads, the community recommends that interested extenders first discuss with the community about developing new write types in the community before using the extension mechanisms. |
-| Plan Extensions                      | Semantic and/or optimization based additions at the plan level. |
+### When to Use What
 
-Because extension mechanisms are different for each serialization format, please refer to the corresponding serialization sections to understand how these extensions are defined in more detail.
+Custom relations are the most flexible option, but also the least interoperable. Prefer enhancements to existing relations when they can express your use case, since this preserves existing patterns and compatibility. As a general rule, choose the least powerful extension mechanism that solves the problem.
```
yongchul, line 213: This is more of note to myself.

Now that I read again "compatibility"-wise, I don't see much difference between enhancement and custom relations because you can't execute the plan anyway in the end if the consumer is unaware of.

With enhancement however, the "validation" of Substrait message could be done better by standard tooling as it preserve the direct outputs and so on while custom relation can't. Is it a big deal? Probably not... you get a better message when you receive illformed plan with enhancement but you can't execute anyway.

The implementation effort from other system wants to support the thing is probably the main factor. It is probably at most that of custom relation.

I think we can phrase this part better but not in this PR. 😄

wackywendell: Yes - this might not be the place to go into it, but perhaps the difference comes in two ways:

- Semantically, an `enhancement` should follow the same 'invariants' as the thing it is extending; e.g. an `Aggregate` should still have the same columns output, the same uniqueness-per-grouping, etc. So using an enhancement does tell the recipient something an extension might be less clear about.
- Practically, this does seem a bit fuzzier. Validation might still be easier, as you noted; perhaps planning as well, if you separate logical planning from physical planning, as there may be details you can elide / pass-through - e.g. a `ReadRel` extension with some enhancement like `read_after: timestamp` or `disable_cache: true` or something like that.

> The implementation effort from other system wants to support the thing is probably the main factor. 

Perhaps that along with 'how well does it follow the invariants of the enhanced relation?' - that might be the other question that's important?

Also just thinking aloud a bit here; I don't feel like I've thought this through well enough for this PR either!

yongchul: Still, technically you can't optimize or plan with unknown enhancement. So I think the only value added is really just syntactic checking... anything related to semantic, you can't do it with unknown enhancement -- no planning, no execution. That's my 2 cents. 😄 

We can follow up on this topic in a separate PR. Thank you for the thorough contribution!

@benbellick I would like you to go over the PR one last time before merging this PR.

site/docs/extensions/index.md
```diff
@@ -126,15 +126,88 @@ The `any[\d]` types (i.e. `any1`, `any2`, ..., `any9`) impose an additional rest
...
+
+- Provide hints to improve performance but don't change the meaning of operations
+- Can be safely ignored by consumers that don't understand them
+- Multiple optimizations can be attached to a single message
+- Examples: memory usage hints, preferred algorithms, caching strategies
+
+#### Enhancements
+
+- Modify the semantic behavior of operations
+- Must be understood by consumers or the plan cannot be executed correctly
```
benbellick:
```suggestion
- Must be understood by consumers, or else the plan cannot be executed correctly
```
I'm not wed to this particular change, but I think it makes it a bit easier to read as it prevents the language parse error of `Must be understood by (consumers or the plan) cannot be executed correctly`

site/docs/extensions/index.md
```diff
@@ -126,15 +126,88 @@ The `any[\d]` types (i.e. `any1`, `any2`, ..., `any9`) impose an additional rest
...
+
+#### Where `AdvancedExtension` Messages Can Be Used
+
+The `AdvancedExtension` message can be attached to various parts of a Substrait plan:
+
+| Location                          | Usage                                       |
+| --------------------------------- | ------------------------------------------- |
+| **`Plan`**                        | Global extensions affecting the entire plan |
+| **`RelCommon`**                   | Extensions for any relational operator      |
+| **Relations** (e.g. `ProjectRel`) | Extensions for a specific relation type     |
```
benbellick, line 178: I agree that the current state of the protos allows for this, but this actually feels like a strange decision. Lets not fix it in this PR, but would it be worth adding a ticket to drop the per-relation `advanced_extension` field and only allow the `RelCommon` location? 

If we don't make that change, what determines whether an enhancement/optimization belongs in the `RelCommon` field or the dedicated `advanced_extension` field?

Issue #775, opened 2025-01-17
JSON schema doesn't exist at specified URL


wackywendell:
The `simple_extensions_schema.yaml` gives an [id that references](https://github.com/substrait-io/substrait/blob/7e3c1394fe6c5a0cce1c1afcc6db9731db2dfc52/text/simple_extensions_schema.yaml#L1 a [URL that doesn't exist](http://substrait.io/schemas/simple_extensions).

Would it make sense to publish that schema at that URL?

PR #848, opened 2025-08-06
fix: make simple extensions schema accessible at specified URL

wackywendell:
## Summary

This addresses [#775](https://github.com/substrait-io/substrait/issues/775).

The `simple_extensions_schema.yaml` file defines its `$id` as `http://substrait.io/schemas/simple_extensions`, but this URL was not accessible on the substrait.io website. This PR fixes the issue by copying the schema file to the correct location during the CI site build process.

## Changes Made

- GitHub Workflow: Updated `.github/workflows/site.yml` to copy the schema file over
- Documentation: Added a README file to document how the schema system works. And configured mkdocs to ignore it.

## Testing

From local testing with `mkdocs serve`, this works - the schema file is available if you copy it in the same way as the github workflow, and the README is not.

## Possible Concerns

- This schema file should perhaps be versioned, and at a versioned URL. That seems out of scope for this, though…
- The copying-in-CI was a straightforward, small change, but it does mean its less obvious how to replicate locally. We could probably do this through a mkdocs gen-files plugin if we wanted full consistency (like as in `docs/extensions/generate_function_docs.py`), but that seemed overly complicated.

### DIFF
--- a/.github/workflows/site.yml
+++ b/.github/workflows/site.yml
@@ -6,6 +6,7 @@ on:
     paths:
       - "site/**"
       - "extensions/**"
+      - "text/**"
 
 jobs:
   site:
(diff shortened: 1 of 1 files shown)


vbarua (approved):
This should be quite helpful when working with schemas. Thanks for fixing this.

vbarua:
As this is not a specification change, I will bypass the 2 vote restriction.

PR #852, opened 2025-08-07
feat: Use URN for all extension URIs

### LINKED ISSUE #828: Pick a canonical URI for the default extensions
(1/2 suggested by @vbarua [here](https://github.com/substrait-io/substrait/issues/827#issuecomment-3105100063))

There is no settled on URI to reference default extensions. As a result, there is ad-hoc usage of different URIs. 

For example: 
- substrait-java: `/functions_boolean.yaml`
- substrait-go: `https://github.com/substrait-io/substrait/blob/main/extensions/functions_boolean.yaml`
- substrait-rs: `https://github.com/substrait-io/substrait/raw/main/extensions/functions_boolean.yaml`

As a consequence, substrait plans from one of the above libraries cannot naturally be consumed by plans from downstream libraries. 

To resolve this issue, let's settle on a single canonical representation. This can be altered later, but it would be useful to make this consistent in the substrait implementors. 

Considering that substrait already uses the format e.g. `https://github.com/substrait-io/substrait/blob/main/extensions/functions_<category>.yaml` in [the tutorial](https://github.com/substrait-io/substrait/blob/793c64ba26e337c22f5e91b658be58b1eea7efd3/site/docs/tutorial/sql_to_substrait.md?plain=1#L899), I propose that we go with that one for now. 

Note that this is separate from the question of how to handle versioning. Since versioning doesn't to be consequential _yet_, I suggest we push it off to another ticket.

benbellick:
closes #828

 - `feat: add feature X`
 - `fix: X in case of Y`


site/docs/extensions/index.md
```diff
@@ -44,6 +43,18 @@ scalar_functions:
...
+
+Substrait provides a set of default simple extensions included in the repository. These extensions use URNs (Uniform Resource Names) as their canonical URIs following the pattern `urn:substrait:<filename_without_extension>`. While downstream users can use any URI they prefer to identify these extensions, library authors **must** use the canonical URIs to ensure compatibility between Substrait plans produced and consumed across different libraries.
+
+All default extension files in the repository follow this naming convention. For example:
+- `functions_boolean.yaml` -> `urn:substrait:functions_boolean`
+- `functions_arithmetic.yaml` -> `urn:substrait:functions_arithmetic`
+- `functions_string.yaml` -> `urn:substrait:functions_string`
+- `extension_types.yaml` -> `urn:substrait:extension_types`
+
+This pattern applies to all extension files in the Substrait repository, including function extensions (for arithmetic, comparison, datetime, etc.) and type extensions.
```
yongchul: nit: do not need `(for arithmetic, comparison, datetime, etc.)`. remove?

site/docs/extensions/index.md
```diff
@@ -44,6 +43,18 @@ scalar_functions:
 
 Here, the choice for the name `ext` is arbitrary, as long as it does not conflict with anything else in the YAML file.
 
+### Canonical URIs for Default Extensions
+
+Substrait provides a set of default simple extensions included in the repository. These extensions use URNs (Uniform Resource Names) as their canonical URIs following the pattern `urn:substrait:<filename_without_extension>`. While downstream users can use any URI they prefer to identify these extensions, library authors **must** use the canonical URIs to ensure compatibility between Substrait plans produced and consumed across different libraries.
```
yongchul: To be slightly future proof, perhaps add *relative path from extension directory* or something along that line? This does not change whatever described below, but leaves the door open to naturally incorporate  subdirectories when we introduce.

But as-is is also fine and we can extend this when we actually have the cases.

yongchul (approved):
Looks good to me. Thank you for quick follow up!

EpsilonPrime (approved):
This approach looks pretty reasonable to me.  Obviously there are a few consumers that will need to change as a result of this.  One of them is Acero which tries to implement this field properly (even allowing for its own namespace).  See https://arrow.apache.org/docs/cpp/acero/substrait.html

benbellick:
> This approach looks pretty reasonable to me. Obviously there are a few consumers that will need to change as a result of this. One of them is Acero which tries to implement this field properly (even allowing for its own namespace). See https://arrow.apache.org/docs/cpp/acero/substrait.html

@EpsilonPrime Thanks for raising that. As a next issue, I'm going to look into upstreaming these into some of the other libraries.

site/docs/extensions/index.md
benbellick: I personally think adding that bit here may add some slight confusion and is an easy thing to add later. So if its alright with you, I'll leave as is. Thanks!

yongchul: Yeah, it is fine as-is. We can come back later when we have too many standard extensions perhaps. :)

bvolpato (dismissed):

site/docs/extensions/index.md
yongchul: nit: perhaps link to wikipedia? https://en.wikipedia.org/wiki/Uniform_Resource_Name

yongchul (approved):

site/docs/extensions/index.md
benbellick: I added a link for the first mention of `URI` as well just for good measure 🙂

yongchul (approved):

site/docs/extensions/index.md
```diff
@@ -25,11 +25,10 @@ A Substrait plan can reference one or more YAML files via URI for extension. In
 
 A YAML file can also reference types and type variations defined in another YAML file. To do this, it must declare the YAML file it depends on using a key-value pair in the `dependencies` key, where the value is the URI to the YAML file, and the key is a valid identifier that can then be used as an identifier-safe alias for the URI. This alias can then be used as a `.`-separated namespace prefix wherever a type class or type variation name is expected.
 
-For example, if the YAML file at `file:///extension_types.yaml` defines a type called `point`, a different YAML file can use the type in a function declaration as follows:
-
+For example, if a YAML file needs to reference a type from Substrait's default extensions (which use URNs as their preferred URI format, see [Canonical URIs](#canonical-uris-for-default-extensions) below), it would look like this. If the YAML file `extension_types.yaml` with URI `urn:substrait:extension_types` defines a type called `point`, a different YAML file can use the type in a function declaration as follows:
```
westonpace:
```suggestion
For example, if a YAML file needs to reference a type from Substrait's default extensions (which use URNs as their preferred URI format, see [Canonical URIs](#canonical-uris-for-default-extensions) below). If we have a YAML file `extension_types.yaml` with URI `urn:substrait:extension_types` defines a type called `point`, a different YAML file can use the type in a function declaration as follows:
```

westonpace (approved):
Small grammar suggestion but good otherwise

site/docs/extensions/index.md
benbellick: > For example, if a YAML file needs to reference a type from Substrait's default extensions.

I think with this suggestion it would no longer be a complete sentence. 

Perhaps even better would be 
```suggestion
For example, consider a YAML file that needs to reference a type from Substrait's default extensions (which use URNs as their preferred URI format, see [Canonical URIs](#canonical-uris-for-default-extensions) below). If the YAML file `extension_types.yaml` with URI `urn:substrait:extension_types` defines a type called `point`, a different YAML file can use the type in a function declaration as follows:
```

westonpace: Yes, that looks good to me!

benbellick:
Sorry that I keep having to re-request everyone! Whenever I push a commit to address a comment, everyone's approval is rendered stale. Not sure if the correct etiquette is to keep requesting but please let me know if not. Thanks for the reviews 🚀

westonpace (approved):

vbarua (approved):
Changes LGTM. Thanks for following up on this 🙇

jacques-n:
We're mixing the use of URI and URN. I understand that one contains the other but am wondering if we'd better to consistent use URN everywhere?

benbellick:
> We're mixing the use of URI and URN. I understand that one contains the other but am wondering if we'd better to consistent use URN everywhere?

@jacques-n I do think still using URI throughout the documentation is appropriate, as we still allow URIs (even if they are not URNs) to be used in general for references to extensions. We would only forbid non-URN URIs for the default extensions. 

The first mention of [URN](https://github.com/substrait-io/substrait/pull/852/files#diff-717c2bcd817e44ead94946142d8fb2b482f0b9b1dddd506e9428e3bc116afc92R28) makes note of the fact that it is the URI for default extensions in particular (and then links to the more detailed section). If you think it is unclear in that section that URNs are just being used as a kind of URI, I would be happy to rewrite it.

jacques-n:
> > We're mixing the use of URI and URN. I understand that one contains the other but am wondering if we'd better to consistent use URN everywhere?
> 
> @jacques-n I do think still using URI throughout the documentation is appropriate, as we still allow URIs (even if they are not URNs) to be used in general for references to extensions. We would only forbid non-URN URIs for the default extensions.
> 
> The first mention of [URN](https://github.com/substrait-io/substrait/pull/852/files#diff-717c2bcd817e44ead94946142d8fb2b482f0b9b1dddd506e9428e3bc116afc92R28) makes note of the fact that it is the URI for default extensions in particular (and then links to the more detailed section). If you think it is unclear in that section that URNs are just being used as a kind of URI, I would be happy to rewrite it.

I guess I took the discussion differently. I've posted a [question there](https://github.com/substrait-io/substrait/issues/828#issuecomment-3177134658) for clarification.

EpsilonPrime (approved):

benbellick:
Closing this in favor of [this PR](https://github.com/substrait-io/substrait/pull/859)

PR #857, opened 2025-08-22
feat: type aliases to save reference to heavy type definitions

yongchul:
feat: type aliases to save reference to heavy type definitions

## Motivation

In Substrait message, we specify full type definition wherever a type is expected. With user-defined types or some complex nested types, the type definition can be quite heavy and can seriously bloat the serialized plan size when "generously" repeated throughout the plan.

Substrait has already embraced anchors to combat this problem and used to encode URI (now URN?), names, and types. Now it is turn of `types`.

## Proposal

This PR introduces `type_aliases` at `Plan` level so that one can spell out heavy type definitions once. The declared alias can be referenced via `Type.alias` oneof field, which has an option to override nullability.

### Why in this way?

This is not just user-defined types but any type. That's why this is not part of extensions rather serves as a "generic" dictionary to encode the plan. Any string type parameter has a potential to be heavy and this maybe true even in standard type in the future like string with collation.

# Backward compatibility

The change is backward compatible. The usage of compact encoding is completely optional and does not alter encoding.

# Maybe follow up?

* We may consider start to looking into do something similar for literals.
* Generic scalar expression requires "binding" field references, which requires more spec-ing.


proto/substrait/plan.proto
```diff
@@ -50,6 +51,11 @@ message Plan {
   // An optional list of bindings for dynamic parameters used in this plan.
   // Each binding maps a parameter_anchor to its corresponding runtime value.
   repeated DynamicParameterBinding parameter_bindings = 7;
+
+  // An optional list of concrete types (i.e., no alias!). Parameterized types
```
jcamachor: Maybe reword the comment for clarity, it says "no alias!" but the field is a `repeated TypeAlias`.

proto/substrait/type.proto
```diff
@@ -52,6 +52,8 @@ message Type {
     // encountered, treat it as being non-nullable and having the default
     // variation.
     uint32 user_defined_type_reference = 31 [deprecated = true];
+
+    TypeAliasReference alias = 37;
```
jcamachor, line 57: Please add comment here.

jcamachor (commented):
+1, this would help reduce the plan bloat we've seen, especially when dealing with user-defined types with multiple parameters.

jcamachor:
@jacques-n , @cpcloud , @westonpace , @EpsilonPrime , @vbarua , we'd love to hear your thoughts on this proposal when you have a chance. Thanks

proto/substrait/type.proto
```diff
@@ -243,6 +246,22 @@ message Type {
...
+    Nullability nullability = 2;
+  }
+}
+
+message TypeAlias {
+  // A surrogate key used in the context of a single plan to reference a
+  // specific type alias.
+  uint32 type_alias_anchor = 1;
+
+  // A concrete type (no alias!) to be aliased. All type parameters must be specified.
```
jacques-n: is it composable with sub-aliases? E.g. 

eg:
- foo => some giant alias one
- bar => mymap<foo>

jacques-n (commented):
Should we add some docs to the site? This is format as opposed to spec really and I'm not sure we doc that well.

EpsilonPrime:
Two questions:

How does this work with type variations?

Is there a name that can be used to leverage these types in the YAML or are these plan only (which appears to be the case).

proto/substrait/type.proto
yongchul: @jacques-n it is doable as long as we prescribe the resolution policy in the spec. Using alias in type parameter is perfectly fine IMO. Let's hear what others think about this.

Regarding resolution, do we want to prescribe the order (i.e., you can use a type alias as type parameter if it is already declared in Plan.type_aliases)? Either way, either consumer or producer needs to be smarter.

vbarua: +1 to this. Something like:
```
fancystring => ...
fancystruct = > struct<...,>
fancymap => map<fancystring, fancystruct>
```

seems like it could be reasonable. Put another way, is there a reasons to disallow nested type aliases?

proto/substrait/type.proto
```diff
@@ -243,6 +246,22 @@ message Type {
...
+  // Nullability of the type may be overridden per reference.
+  Type type = 2;
```
vbarua, line 269: What do you think about ignoring the nullability as it is declared here, and always requiring it at the alias site. Instead of having to check the nullability of `TypeAliasReference.nullability` and then falling back conditionally, we can just ignore it here and always require it there. We can replace the nullability text here with something like:
```
The nullability of the declared type here is ignored. When an aliased typed is reference, its nullability is determined by the value of TypeAliasReference.nullability
```

proto/substrait/plan.proto
```diff
@@ -50,6 +51,12 @@ message Plan {
   // An optional list of bindings for dynamic parameters used in this plan.
   // Each binding maps a parameter_anchor to its corresponding runtime value.
   repeated DynamicParameterBinding parameter_bindings = 7;
+
+  // An optional list of type aliases. Parameterized types of which definitions
+  // are large (e.g., user-defined types) can be spelled out with
+  // specific parameter values and referenced anywhere in the plan using `Type.alias`
+  // where a concrete type is expected.
+  repeated TypeAlias type_aliases = 8;
```
vbarua: Potential simplification:
> An optional list of type aliases. Types can be specified here once, and then referenced within the plan using the TypeAliasReference type. This feature is intended to assist with the usability of parameterized types, which require that all parameters be specified when they are declared. This can bloat plans with redundant redaclarations, especially if the parameterized types being declared have many parameters (i.e big structs or nested parameterized types).

vbarua (commented):
Some minor suggestions, but overall this seems like a reasonable feature that will help reduce plan size bloat and also improve readability.

proto/substrait/type.proto
yongchul: @vbarua a forgot to publish my comments while I was jumping around meetings. No, I don't see a particular reason to block nested references. But probably need to say something about the order either explicit or implicit...

Either way works and probably okay without implicit -- means putting the burden to consumer. Any thoughts on this? This is probably a constant theme, which one gets the added complexity.

yongchul:
> Two questions:
> How does this work with type variations?

By design, this is meant to be used within a plan and always refers "concrete type", means type variation is part of concrete type.
 
> Is there a name that can be used to leverage these types in the YAML or are these plan only (which appears to be the case).

No, this is not meant to be used in the YAML. This is strictly for improving plan encoding.

jcamachor:
@yongchul , I just realized we'll need to extend `Literal` with type alias support as well.

jcamachor:
> @yongchul , I just realized we'll need to extend `Literal` with type alias support as well.

It can be done in a follow-up PR as well.

proto/substrait/type.proto
vbarua: (Heads up for the future, slight mistag I'm @vbarua)

> But probably need to say something about the order either explicit or implicit...

What do you mean about explicit or implicit here? 

I assume you're talking about the issue that if we have multiple type aliases that depend on each other in some nested way, depended on how they are declared they might appear out-of-order to the consumer if it processes them linearly. I don't think we should constrain the ordering in any way, and consumers should be able to handle references in any order, which does increase the burden on the consumer a bit.

One thing that might be worth a callout is that cycles should not be permitted in aliases. We probably have to guard against them in the consumers because they may appear in badly formed or malicious plans, but they should be considered erroneous.

yongchul: Fixed the comment. :)

> What do you mean about explicit or implicit here?

You got it right. I meant the order of aliases appearing in the plan. If the general theme is keeping the producer as light weight as possible, that's fine.

> ... cycle ...

Yes, let's spell out cyclic dependencies among type aliases are not allowed.

yongchul: Updated the comments and added documentation. @vbarua @jacques-n PTAL.

site/docs/types/type_aliases.md
```diff
@@ -0,0 +1,35 @@
+# Type Aliases
+
+In a Substrait plan, types are spelled out whenever and wherever they are needed. For parameterized types, all type parameters are spelled out per type reference. For a parameterized type with a large number of parameters, complex nested type parameters, or string parameters, this can significantly bloat the size of the plan proportional to the number of such type references in the plan as such type parameters are repeatedly spelled out.
+
+To alleviate the problem, Substrait offers type alias mechanism.
```
vbarua:
```suggestion
To alleviate the problem, Substrait offers a type alias mechanism.
```

site/docs/types/type_aliases.md
```diff
@@ -0,0 +1,35 @@
...
+
+Type alias allows a plan can declare a type once and reference it multiple times within a plan like extensions; you can reference an aliased type where a type is expected.
```
vbarua:
```suggestion
Type aliases allow a plan to declare a type once and reference it multiple times within a plan. A type alias can be used wherever a type is expected.
```

site/docs/types/type_aliases.md
```diff
@@ -0,0 +1,35 @@
...
+
+## Type Alias
+
+A type alias is simply a mapping from anchor to a concrete Substrait type. A valid type alias is described below.
```
vbarua:
```suggestion
A type alias is a mapping from an anchor to a concrete Substrait type. A valid type alias is described below.
```

site/docs/types/type_aliases.md
```diff
@@ -0,0 +1,35 @@
...
+
+* All type parameters must be spelled out.
```
vbarua:
```suggestion
* All type parameters must be specified.
```

site/docs/types/type_aliases.md
```diff
@@ -0,0 +1,35 @@
...
+* Can not alias another alias.
```
vbarua:
```suggestion
* Cannot directly be another alias.
```

site/docs/types/type_aliases.md
```diff
@@ -0,0 +1,35 @@
...
+* Type parameters can reference other aliased types as long as there are no circular dependencies among involved type aliases.
```
vbarua:
```suggestion
* Type parameters can reference other aliased types as long as no circular dependcies are introduced.
```

proto/substrait/type.proto
```diff
@@ -243,6 +246,28 @@ message Type {
...
+  uint32 type_alias_anchor = 1;
+
+  // A concrete type to be aliased.
+  //
+  // * Can not be another type alias.
+  // * All type parameters must be specified. Type parameters can refer other type
+  //   aliases as long as there are no circular dependencies among type aliases.
+  // * Nullability of TypeAlias.type is ignored. When an aliased type is referenced,
+  //   its nullability is determined by the value of TypeAliasReference.nullability.
+  // * TypeAlias.type.type_variation_reference can be specified.
```
vbarua: I would suggest the same updates here as in the README.

proto/substrait/type.proto
```diff
@@ -243,6 +246,27 @@ message Type {
...
+message TypeAlias {
+  // A surrogate key used in the context of a single plan to reference a
+  // specific type alias.
+  uint32 type_alias_anchor = 1;
+
+  // A concrete type to be aliased.
+  //
+  // * All type parameters must be specified.
+  // * Cannot directly be another alias.
+  // * Type parameters can reference other aliased types as long as no circular dependcies are introduced.
```
vbarua: minor: typo fix
```suggestion
  // * Type parameters can reference other aliased types as long as no circular dependencies are introduced.
```

site/docs/types/type_aliases.md
```diff
@@ -0,0 +1,35 @@
...
+
+Type aliases allow a plan to declare a type once and reference it multiple times within a plan. A type alias can be used wherever a type is expected.
+
+## Type Alias
+
+A type alias is a mapping from an anchor to a concrete Substrait type. A valid type alias is described below.
+
+* All type parameters must be specified.
+* Cannot directly be another alias.
+* Type parameters can reference other aliased types as long as no circular dependcies are introduced.
```
vbarua: minor: typo fix
```suggestion
* Type parameters can reference other aliased types as long as no circular dependencies are introduced.
```

site/docs/types/type_aliases.md
```diff
@@ -0,0 +1,35 @@
...
+* Nullability of aliased type is **ignored**. Nullability must be specified when the aliased type is referenced.
+* Type variation may be specified in the aliased type.
+
+## Type Alias Reference
+
+Type alias reference is a Substrait type thus can appear wherever a Substrait type is expected. The reference must specify nullability of the aliased type.
```
vbarua:
```suggestion
A type alias reference is a Substrait type and can appear wherever a Substrait type is expected. The reference must specify the nullability of the aliased type.
```

site/docs/types/type_aliases.md
```diff
@@ -0,0 +1,35 @@
...
+
+## Examples
+
+```
+type alias 1 --> VARCHAR<100> // OK to alias Substrait VARCHAR.
+type alias 2 --> UserDefined<i8, type alias ref (2, NULLABLE)> // OK to reference other type alias 2. UserDefined<i8, VARCHAR<100>?>.
```
vbarua: I assume the intent is for TA 2 to reference TA1, and not itself? Otherwise you would have a circular reference.
```suggestion
type alias 2 --> UserDefined<i8, type alias ref (1, NULLABLE)> // OK to reference other type alias 1. UserDefined<i8, VARCHAR<100>?>.
```

site/docs/types/type_aliases.md
```diff
@@ -0,0 +1,35 @@
...
+type alias 3 --> UserDefined<i8, type alias ref (2, REQUIRED)> // OK to reference other type alias 2. UserDefined<i8, VARCHAR<100>>.
+
+type alias 4 --> type alias ref (1, NULLABLE)  // NOT OK to alias another alias.
```
vbarua:
```suggestion
type alias 4 --> type alias ref (1, NULLABLE)  // NOT OK to alias another alias directly.
```

site/docs/types/type_aliases.md
```diff
@@ -0,0 +1,35 @@
...
+## Type Alias Reference
+
+Type alias reference is a Substrait type thus can appear wherever a Substrait type is expected. The reference must specify nullability of the aliased type.
+
+## Examples
+
+```
+type alias 1 --> VARCHAR<100> // OK to alias Substrait VARCHAR.
+type alias 2 --> UserDefined<i8, type alias ref (2, NULLABLE)> // OK to reference other type alias 2. UserDefined<i8, VARCHAR<100>?>.
+type alias 3 --> UserDefined<i8, type alias ref (2, REQUIRED)> // OK to reference other type alias 2. UserDefined<i8, VARCHAR<100>>.
```
vbarua: I assume the intent is for this to also reference TA1, but override the nullability?
```suggestion
type alias 3 --> UserDefined<i8, type alias ref (2, REQUIRED)> // OK to reference other type alias 1. UserDefined<i8, VARCHAR<100>>.
```

vbarua (commented):
Overall looks good. Left some wording suggestions, and also some possible corrections for your examples. If you can take a look at those, I'd be happy to approve.

site/docs/types/type_aliases.md
yongchul: Ah, yes. Thanks!

PR #859, opened 2025-08-27
feat!: Canonicalize extension URNs

### LINKED ISSUE #856: Define Standard Identifier for Extensions (and Dialects)
See #828 for context, but basically, we need to come up with a canonical representation for the default substrait extensions's URLs. This opened up a broader discussion on what should be used as identifiers for **all** extensions, not just the default ones. 

--------------

After having a conversation with @wackywendell, I realize that some more care should be taken for this particular issue. I want to write up a summary of the goals we’re trying to achieve, the difficulties at hand, and my current suggestion. I think if we can get this right once and for all then it can be something to stop worrying about (and the ticket can finally complete). 

# TLDR
Let's use urns + reverse DNS for ensuring globally unique extensions.

# Problem

We need **identifiers** for Substrait extensions (and later dialects). The ideal state is that these identifiers will be:

- **Unique**
- **Human-readable**
- **Offline-first**: no need to download files from the web at runtime
- **Optional version support in the future**
- **No dependency on a Substrait-managed central registry**

In an ideal world, we would also offer some security strategy to ensure that the loaded extensions are what we think they are (e.g., content-hash verification), but we can deal with this later.

# Possible Strategies

We have two fundamental approaches: enforce uniqueness from the start, or accept potential collisions initially and address them later.

## Asserting Uniqueness

If we want to ensure uniqueness without a central registry, we need to delegate that guarantee to an existing, global naming system. A few possible approaches:

### GitHub URLs

Identifiers could simply be URLs to the source file on GitHub, e.g. `https://github.com/substrait-io/substrait/blob/main/extensions/functions_boolean.yaml`.

**Pros:**
- Uniqueness “for free” (GitHub handles namespace).
- Version information can be embedded in tags or commit SHAs.

**Cons:**
- Tied to GitHub (other hosts would require bespoke handling).
- Sensitive to file location within the repo (renames/moves break references).

### Generic Git SSH

We could use the Git SSH form to identify a repository, e.g. `git@github.com:substrait-io/substrait.git`.

But we’d then need a way to reference a specific file within the repository. Rust uses a similar repo-level approach, but solves the “which file” problem with a top-level `Cargo.toml` manifest.

**Pros:**
- Unique at the repository level.
- Works across any git hosting service.

**Cons:**
- Requires an additional manifest convention to locate files.

### Reverse DNS

Adopt [reverse DNS notation](https://en.wikipedia.org/wiki/Reverse_domain_name_notation). Since Substrait owns `substrait.io`, we could use e.g. `urn:substrait:io.substrait:extension:functions_boolean`.

**Pros:**
- Globally unique without a central registry (uniqueness via DNS).
- Precedent in Java, Maven, Android, and other ecosystems.
- Independent of repository layout — files can move without changing the identifier.
- No dependency on GitHub or any specific host.

**Cons:**
- Requires extension/dialect authors to control a domain.
- Domain ownership changes could create conflicts if new owners reuse identifiers.
- Domain owners must prevent naming clashes within their own namespace.

## Dropping Uniqueness (Short-term Alternative)

We could start by not enforcing uniqueness, reserving the `urn:substrait:` prefix and encouraging authors to choose collision-resistant names. This would be easiest to adopt now, but comes with significant risks:

**What happens with collisions:**
- Runtime errors when loading extensions with identical names
- Manual intervention required to resolve conflicts

**Migration challenges:**
- Breaking changes when we later enforce uniqueness
- Existing deployments may need to update identifiers
- Potential ecosystem fragmentation if some adopt unique naming while others don't

While tempting for immediate adoption, this approach likely creates more work long-term than solving uniqueness properly from the start.

# Proposed Solution

## Identifier Format

Reverse DNS is the best long-term approach. It is:
- Globally unique without a Substrait-managed registry.
- Untethered from any hosting provider.
- Familiar to many developers.

Additionally, it allows us to defer versioning until we solve #829. 

**Final goal format (once versioning is defined and required):**
```
urn:substrait:<Reverse-DNS>:<Kind>:<ID>:<Version>
```
Where:
- `<Reverse-DNS>`: domain name you control, reversed (e.g., `io.substrait`, `com.datadoghq`, `org.apache.arrow`).
- `<Kind>`: `extension` or `dialect`.
- `<ID>`: one or more colon-separated tokens for the artifact name.
- `<Version>`: whatever version format we define once we solve #829

**Current initial format (before versioning is defined):**
```
urn:substrait:<Reverse-DNS>:<Kind>:<ID>
```

**Examples (final goal format):**
- `urn:substrait:io.substrait:extension:functions_set:1.0.0`
- `urn:substrait:io.substrait:dialect:postgresql:1.2.12`
- `urn:substrait:org.apache.arrow:extension:relations:aggregates:29.32.1`

We omit `<Version>` entirely for now to avoid prematurely committing to version semantics. Once the versioning policy is finalized, we make it a required segment separated by a colon, consistent with Java/Maven-style conventions.

## Identifier Declaration

Extensions and dialects declare their own identifiers in their YAML definitions. For example:

```yaml
---
id: urn:substrait:io.substrait:extension:types
types:
  - name: point
    structure:
      latitude: i32
      longitude: i32
  - name: line
    structure:
      start: point
      end: point
```

This means:
- The identifier is defined by the artifact itself, not by the loader.
- Implementations simply read the declared `id` instead of inventing their own naming scheme.

# Conclusion

Uniquely identifying extensions and dialects without a central registry is **hard** (i.e. [naming is hard](https://martinfowler.com/bliki/TwoHardThings.html)), but Reverse DNS gives us:
- Maximal readability
- Clear precedent from large ecosystems + usage of valid URNs
- Global uniqueness
- Flexibility to later add integrity verification and versioning rules

This proposal offers a clean, future-proof foundation. The colon-separated `<Version>` segment will be required in the future once a clear versioning strategy is established. For now, we omit version entirely to keep the format simple and avoid committing prematurely to version semantics.

I’m opening the discussion now to hear feedback and alternative ideas. :)

benbellick:
Closes #856 

This PR introduces the notion of URNs to uniquely identify extensions. As a consequence, all mentions of `Uri`s have been switched to `Urn`s. This will result in quite a bit of a breaking change. None of the binary encoded substrait is changing (we have just changed field names), but consumers will have to be altered to handle the new naming convention, and rendered JSON plans will be quite different. 

If there is a better migration strategy to adopt here, please let me know. 

BREAKING CHANGE: Alteration of protos and extension yaml schema to reflect usage of URNs for Thank you for submitting a PR!


benbellick:
Two things I wanted to bring up: 

First, I realize that the changes introduced here will involve some breaking changes to all of the libraries implementing substrait (e.g. `substrait-go`, `substrait-java`). **I will take responsibility for updating all of these downstream consumers to make this transition as easy as possible**

Second, this is a bit of an aggressively breaking change. Should some kind of more careful strategy be taken to ease adoption (other than just breaking it, bumping the version, and declaring that it is breaking)? Remember that these changes will not affect what is being sent on the wire, except for the fact that the extension URNs themselves will change. So, a few options here:
~~- leave the PR as is, just break things and bump the versions, and make consumers handle the transition themselves, probably updating downstream consumers to handle both the old and the new messages, then sending the new messages from producers, and then updating downstream to only handle the new messages~~
~~- leave the PR as is just like before, but have the downstream libraries implement a migration where they temporarily accept both the old and the new identifier for the built in extensions~~
~~- Revert all of the proto changes and leave them all referencing `URN`s, but this will be a bit confusing because of the intermixing of `URN`s and `URI`s~~
- Perform a protobuf-level migration, where we first introduce the new fields and mark the current ones as deprecated, and then in a later release drop the deprecated ones. 

EDIT: Changed after the fact to reflect @vbarua's below comment. We will go with the proto-level migration.

site/docs/extensions/generate_function_docs.py
```diff
@@ -189,6 +191,12 @@ def write_markdown(file_obj: dict, file_name: str) -> None:
             )
         )
 
+        if 'urn' in yaml_file_object:
+            urn = yaml_file_object['urn']
+            description += f". The extension URN is `{urn}`."
+
+        mdFile.new_paragraph(description)
```
benbellick, line 195: Now displays like this: 

image

site/docs/extensions/index.md
```diff
@@ -13,23 +13,30 @@ Some kinds of primitives are so frequently extended that Substrait defines a sta
...
+To extend these items, developers can create one or more YAML files that describe the properties of each of these extensions. Each YAML file must include a required `urn` field that uniquely identifies the extension. This URN ([Uniform Resource Name](https://en.wikipedia.org/wiki/Uniform_Resource_Name)) uses the format `<KIND>:<OWNER>:<ID>`, where:
 
-A Substrait plan can reference one or more YAML files via URI for extension. In the places where these entities are referenced, they will be referenced using a URI + name reference. The name scheme per type works as follows:
+- `KIND` is currently only `extension` (additional kinds may be introduced in the future)
```
jacques-n: should kind be function or type?

benbellick: This URN is for the unique name of the whole file, so `<KIND>` is `extension` (as opposed to something like `dialect`). 

It's confusing at this moment to acknowledge the possibility of other kinds of files which need to have unique URNs until the dialects PR is in, so I think it's better to just do the below:
```
To extend these items, developers can create one or more YAML files that describe the properties of each of these extensions. Each YAML file must include a required `urn` field that uniquely identifies the extension. This URN ([Uniform Resource Name](https://en.wikipedia.org/wiki/Uniform_Resource_Name)) uses the format `extension:<OWNER>:<ID>`, where:
- `OWNER` represents the organization or entity providing the extension and should follow [reverse domain name convention](https://en.wikipedia.org/wiki/Reverse_domain_name_notation) (e.g., `io.substrait`, `com.example`, `org.apache.arrow`) to prevent name collisions
- `ID` is the specific identifier for the extension (e.g., `functions_arithmetic`, `custom_types`)
```

benbellick: I've altered the documentation to be as I mentioned above. Now the notion of `<KIND>` is dropped, and we just write that users should use `"extension"`. Later, we can reintroduce when there is a need for a separate namespace, e.g. in the dialects PR.

benbellick:
I think the right approach here is to just go ahead and bump the version, leaving this PR as-is. Downstream consumers should then handle the migration by accepting both the original URL form and the new URN form for a transitional period. In a subsequent version, we can remove support for the URL form and require only URNs.

This keeps the change within Substrait itself simple — there’s no need to complicate the schema or introduce intermediate fields, and the wire format remains unchanged. Also, given that downstream libraries were already using URLs in a bespoke way that was never declared as an official part of the specification, it seems reasonable for the migration responsibility to sit with them[^1].

[^1]: which is really me, since I will shepard all of the migrations through.

site/docs/extensions/index.md
```diff
@@ -13,23 +13,29 @@ Some kinds of primitives are so frequently extended that Substrait defines a sta
 * Window Functions
 * Table Functions
 
-To extend these items, developers can create one or more YAML files at a defined URI that describes the properties of each of these extensions. The YAML file is constructed according to the [YAML Schema](https://github.com/substrait-io/substrait/blob/main/text/simple_extensions_schema.yaml). Each definition in the file corresponds to the YAML-based serialization of the relevant data structure. If a user only wants to extend one of these types of objects (e.g. types), a developer does not have to provide definitions for the other extension points.
+To extend these items, developers can create one or more YAML files that describe the properties of each of these extensions. Each YAML file must include a required `urn` field that uniquely identifies the extension. This URN ([Uniform Resource Name](https://en.wikipedia.org/wiki/Uniform_Resource_Name)) uses the format `extension:<OWNER>:<ID>`, where:
```
vbarua: A suggestion. Because we aren't strictly following the official URN schema anymore, what do you think using Extension URN as a name / noun for these to try and make it clearer that they're not straight URNs. So here we can replace the last sentence with:

> This Extension URN uses the format `extension:<OWNER>:<ID>`

benbellick: Done! I also added a small comment ack'ing that these are not technically urns.

benbellick:
Per today's sync, I am going to:
1. Add a comment in the protobuf clarifying that if both `uri` and `urn` are present in the message, `urn` takes precedence, and 
2. update the commit history to remove items about breaking changes. This change is technically not a breaking change, but the later PR to drop support for `uri` will be.

vbarua (approved):
Changes look great. Thanks for pushing for this in the community, it should really help with interoperability.

PR #863, opened 2025-09-11
docs: clarify function signatures for variadic functions

vbarua:


site/docs/extensions/index.md
```diff
@@ -44,19 +44,28 @@ scalar_functions:
 
 Here, the choice for the name `ext` is arbitrary, as long as it does not conflict with anything else in the YAML file.
 
-### Function Signature Compound Names
```
vbarua, line 55: I don't think there's that much value in referring to these as `Function Signature Compound Names` over just `Function Signature`. It's a mouthful every time I have to say it, and most of the time we just shorten it to `Function Signature` anyways. Not strongly wedded to this, but figured it might be a good cleanup alongside this.

site/docs/extensions/index.md
```diff
@@ -44,19 +44,28 @@ scalar_functions:
...
+* The name of the function
+* The defined arguments for the implementation
 
+which are combined as follows:
 ```
 ```
 
-Rather than using a full data type representation, the input argument types (`short_arg_type`) are mapped to single-level short name. The mappings are listed in the table below.
+Argument types (`short_arg_type`) are encoded using the Type Short Names given below.
```
yongchul, line 70: I'm throwing another UDT axe. So, let's say you have two UDTs both are named as X but belong to different name space N and M (i.e., N.X and M.X). Suppose you have a function that can handle both types. What will be the signature for that?

Or, should type short name actually need to spell out the fully qualified UDT name (e.g., `u:<URN>.<name>`)? Or instead of URN, we use the alias defined in the YAML... but then the consumer should be aware of those aliases of the dependent YAMLs.

Or, do we just let it happen (i.e., without namespace qualification) and the consumer figures out the correct implementation based on the actual argument types regardless of name collision?

Or, one should not put the implementations with the different UDTs of the same name in one YAML from the beginning?

This is not something to address in this PR but we should follow up and clarify  -- give a clear guideline.

site/docs/extensions/index.md
```diff
@@ -44,19 +44,28 @@ scalar_functions:
...
 
-!!! note
+#### Variadic Functions
 
-    Every compound function signature must be unique.  If two function implementations in a YAML file would generate the same compound function signature, then the YAML file is invalid and behavior is undefined.
+For variadic functions, the variadic argument is included *once* in the function signature.
+
+#### Uniqueness Constraint
+
+A function signature uniquely identifies a function implementation within a single YAML file. As such, every function implementation within a YAML **must** have a distinct function signature in order for references to the implementation to remain unambiguous. A YAML file in which this is not the case is invalid.
```
yongchul, line 78: do we have a validator for this?

yongchul (commented):
overall, looks good to me. we will need to revisit the short name for UDT and type variation handling in next PRs.

site/docs/extensions/index.md
vbarua: That's a good axe 😅 

Created an issue to track it https://github.com/substrait-io/substrait/issues/864

site/docs/extensions/index.md
```diff
@@ -44,19 +44,32 @@ scalar_functions:
 
 Here, the choice for the name `ext` is arbitrary, as long as it does not conflict with anything else in the YAML file.
 
-### Function Signature Compound Names
+### Function Signature
 
-A YAML file may contain one or more functions by the same name. The key used in the function extension declaration to reference a function is a combination of the name of the function along with a list of the required input argument types. The format is as follows:
+A YAML file may contain one or more functions with the same name, each with one or more implementations (impls). A specific function implementation within a YAML file can be identified using a Function Signature which consists of two components
+* Function Name: the name of the function
+* Argument Signature: a signature based on the defined arguments of the function
```
vbarua, line 59: I was updating the dialect PR, and realized that it would be helpful to give the argument side of this signature an explicit name.

site/docs/extensions/index.md
vbarua: Not in the core spec we don't, and we have introduced changes that have this problem, which we've noticed when substrait-java blows up because it does check this.

Having a check for this in CI is worth an issue https://github.com/substrait-io/substrait/issues/865

yongchul (approved):

EpsilonPrime (approved):

site/docs/extensions/index.md
jacques-n: I thought we added ci check for this but someone would need to confirm. (As part of the test case work.)

jacques-n (approved):

PR #868, opened 2025-09-19
feat: support type alias in user defined literal

yongchul:
# Motivation

#857 introduced type alias concept but not plumbed all the way through. User defined literal inlines substrait.Type thus can't use type alias where it probably needs the most. Because of how the current message is defined, it is controversial how we add the type alias support, thus here is the PR for that to discuss and finalize.

# Proposal

I see at least three options.

1. Create a new user defined message that only supporting type alias. This is the cleanest at the cost leaving two ways of specify user defined literal.
2. Introduce `type_alias_anchor` as oneof with `type_anchor`. Again, comment says "hey! `type_parameters` are ignored when this is set!".
3. Introduce `type_alias_anchor` to existing UD literal message. wave hands in the comment to say that "hey! this is actually oneof of `type_anchor` and when set `type_parameters` are ignored!"

My preference is 1 ~ 2 >> 3. So I start the PR with option 1. Let's discuss whether there are better ways to do this.

### DIFF
--- a/proto/substrait/algebra.proto
+++ b/proto/substrait/algebra.proto
@@ -1030,6 +1030,7 @@ message Expression {
       Type.List empty_list = 31;
       Type.Map empty_map = 32;
       UserDefined user_defined = 33;
+      UserDefinedWithTypeAlias user_defined_with_type_alias = 38;
     }
 
     // Whether the literal_type above should be treated as a nullable type.
@@ -1132,6 +1133,19 @@ message Expression {
         google.protobuf.Any value = 2;
         // the value of the literal, serialized using the structure definition in its declaration
         Literal.Struct struct = 4;
+      }
+    }
+
+    message UserDefinedWithTypeAlias {
+      // points to a type_alias_anchor defined in this plan.
+      uint32 type_alias_reference = 1;
+
+      // a user-defined literal can be encoded in one of two ways
+      oneof val {
+        // the value of the literal, serialized using some type-specific protobuf message
+        google.protobuf.Any value = 2;
+        // the value of the literal, serialized using the structure definition in its declaration
+        Literal.Struct struct = 3;
       }
     }
   }


proto/substrait/algebra.proto
```diff
@@ -1134,6 +1135,19 @@ message Expression {
         Literal.Struct struct = 4;
       }
     }
+
+    message UserDefinedWithTypeAlias {
```
jcamachor: `UserDefinedWithTypeAlias` -> `AliasDefined` ? Since this message doesn't necessarily have to be limited to user-defined type literals, it could also more generally reference any type that has been aliased such as varchar, etc.

yongchul: That's a good point. I like `TypeAliasDefined` to be consistent. WDYT?

jcamachor (approved):

yongchul:
> I believe 2 >>> 1, so your current solution is great.
> 
> As I understand it, you could want to use a type alias on a type that is not really a `UserDefined` type but a compound type built directly into substrait, e.g. `Map<String,int32>`, etc. in which case `TypeAliasDefined` seems like a truly new `Literal` type as opposed to an alternative case of `UserDefined` (which is option 1 of your proposal).
> 
> Please let me know if I misunderstood anything. Thanks!

For Substrait native complex types (e.g., `Map<string,int32>`), you can still use the literal message as-is. In theory, you could go with the new type alias literal but that means you are doing it with an unnecessary nesting of Literal itself. Do you see a case?

benbellick:
> > I believe 2 >>> 1, so your current solution is great.
> > As I understand it, you could want to use a type alias on a type that is not really a `UserDefined` type but a compound type built directly into substrait, e.g. `Map<String,int32>`, etc. in which case `TypeAliasDefined` seems like a truly new `Literal` type as opposed to an alternative case of `UserDefined` (which is option 1 of your proposal).
> > Please let me know if I misunderstood anything. Thanks!
> 
> For Substrait native complex types (e.g., `Map<string,int32>`), you can still use the literal message as-is. In theory, you could go with the new type alias literal but that means you are doing it with an unnecessary nesting of Literal itself. Do you see a case?

Ah yes, that is a good point. Hard to think of a reason you would want to do that. One more proposal for how things could be done: 

```proto
 message Literal {
    oneof literal_type {
      bool boolean = 1;
      int32 i8 = 2;
      ...
      UserDefined user_defined = 33;
      // No new TypeAliased value field
    }

    message UserDefined {
      // points to a type_anchor defined in this plan
      uint32 type_reference = 1;
      // The parameters to be bound to the type class, if the type class is
      // parameterizable.
      repeated Type.Parameter type_parameters = 3;

      oneof locality {
         Local local = 5;
         TypeAliased type_aliased = 6;
      }

       message Local {
         uint32 type_reference = 1;
         repeated Type.Parameter type_parameters = 2;
       }
       message Aliased {
         uint32 type_alias_reference = 1;
       }

      // a user-defined literal can be encoded in one of two ways
      oneof val {
        // the value of the literal, serialized using some type-specific protobuf message
        google.protobuf.Any value = 2;
        // the value of the literal, serialized using the structure definition in its declaration
        Literal.Struct struct = 4;
      }
    }
}
```

This would have the benefit of keeping the types as one and wouldn't require a special comment saying not to set the `type_parameters` field if type alias reference is set, but it would involve a migration. I'm okay with either approach though, and I know the migrations can be a pain :)




One more consideration: is it possible that we will in the future want to be able to offer type aliases to types which are not fully concrete / still have parameters? For example, suppose we have `MyComplexType<T1,T2,T3,T4>`. We may later decide we want a type alias of the form `MySimplerType<T> = MyComplexType<int32,T,String,float>`. If that does turn out to be the case, then approach 2 seems most flexible to allowing this behavior in the future.

yongchul:
> Ah yes, that is a good point. Hard to think of a reason you would want to do that. One more proposal for how things could be done:
> 
> ```proto
>  message Literal {
>     oneof literal_type {
>       bool boolean = 1;
>       int32 i8 = 2;
>       ...
>       UserDefined user_defined = 33;
>       // No new TypeAliased value field
>     }
> 
>     message UserDefined {
>       // points to a type_anchor defined in this plan
>       uint32 type_reference = 1;
>       // The parameters to be bound to the type class, if the type class is
>       // parameterizable.
>       repeated Type.Parameter type_parameters = 3;
> 
>       oneof locality {
>          Local local = 5;
>          TypeAliased type_aliased = 6;
>       }
> 
>        message Local {
>          uint32 type_reference = 1;
>          repeated Type.Parameter type_parameters = 2;
>        }
>        message Aliased {
>          uint32 type_alias_reference = 1;
>        }
> 
>       // a user-defined literal can be encoded in one of two ways
>       oneof val {
>         // the value of the literal, serialized using some type-specific protobuf message
>         google.protobuf.Any value = 2;
>         // the value of the literal, serialized using the structure definition in its declaration
>         Literal.Struct struct = 4;
>       }
>     }
> }
> ```
> 
> This would have the benefit of keeping the types as one and wouldn't require a special comment saying not to set the `type_parameters` field if type alias reference is set, but it would involve a migration. I'm okay with either approach though, and I know the migrations can be a pain :)

Yeah, this is a variation of option 2 in the description, involving some migration (not much but just we are policing the type parameters are not used with type alias at the moment). It is better to be able to specify inline or type aliased within the literal but I'm not sure. Considering the potential extension of partial binding, type parameters can actually stay, it is just matter of where to look (i.e., type reference or type alias reference). So if we were to have oneof, we just need one oneof around the anchor. 

So below is my concrete incarnation of option 2.

```proto
message UserDefind {
  oneof type_anchor_type {
    uint32 type_reference = 1;
    uint32 type_alias_reference = 5;
  }
  // the rest remains as-is today.
  ...

}
```

I don't recall whether protobuf supports oneof of same primitive types... I hope it does.

> One more consideration: is it possible that we will in the future want to be able to offer type aliases to types which are not fully concrete / still have parameters? For example, suppose we have `MyComplexType<T1,T2,T3,T4>`. We may later decide we want a type alias of the form `MySimplerType<T> = MyComplexType<int32,T,String,float>`. If that does turn out to be the case, then approach 2 seems most flexible to allowing this behavior in the future.

Thought about it but I am not fully convinced whether the partial binding is really necessary at the moment so for the simplicity, leaving it like this. But yeah, why not? :-)


So do you @benbellick  like option 2?

benbellick:
@yongchul Thanks for the conversation. Considering all of the above above, I think that yes, I have indeed settled on 2 :) 

To me, it feels like a more consistent abstraction to consider TypeAliases as just being UserDefined types which (at the moment) take zero type parameters. And just as passing in the wrong number of type parameters is incorrect to the standard type, passing in a non-zero amount of type parameters to a type-alias is (for now at least) incorrect. 

All of that being said, you have gotten all of the sufficient approvals and my preference is relatively minor. Thanks for bearing with my flip-flopping!

yongchul:
> @yongchul Thanks for the conversation. Considering all of the above above, I think that yes, I have indeed settled on 2 :)
> 
> To me, it feels like a more consistent abstraction to consider TypeAliases as just being UserDefined types which (at the moment) take zero type parameters. And just as passing in the wrong number of type parameters is incorrect to the standard type, passing in a non-zero amount of type parameters to a type-alias is (for now at least) incorrect.
> 
> All of that being said, you have gotten all of the sufficient approvals and my preference is relatively minor. Thanks for bearing with my flip-flopping!

Thank you for sharing your thoughts! As I said in the description, my preference for option 1 and 2 are similar but I started with option 1 because it is simple and back compat. I don't want to introduce something only to be cleaned up later... 

Now that I think again, option 2 is also probably better for the implementers -- one less message to handle.

I'll update the PR and reset the vote and get other SMCs take a second look.

proto/substrait/algebra.proto
```diff
@@ -1123,8 +1123,13 @@ message Expression {
     }
 
     message UserDefined {
-      // points to a type_anchor defined in this plan
-      uint32 type_reference = 1;
+      oneof type_anchor_type {
```
yongchul, line 1126: Looks like this is a breaking change but I disagree that this is a necessarily breaking change... unless you actually use type reference value 0.

This is the right direction in the long term and wondering whether it is worth to go through this minor breaking change -- assuming user defined literal usage is relatively rare. Want to hear your thoughts on this.

benbellick: I also personally wonder why we didn't disallow usage of `0` as a type / function anchor in the first place? It certainly would make all of the client library implementations easier.

westonpace: I did confirm that Datafusion will use 0 as a type reference.  It also does utilize user defined literals for fp16 data.  However, the fp16 feature was added very recently and I don't think has been released yet.  Perhaps we can upgrade Substrait in DF before that happens.  Either way, the blast radius does seem small.

yongchul:
@westonpace  thank you for confirmation! Two things.

1. shall we go for breaking change as-is or remove the breaking change from the merge message?
2. it would be great if we can bump DF before its release. How does this work? open an issue there or just send a PR bumping the version?

proto/substrait/algebra.proto
yongchul: (I don't like the github commenting system...) @westonpace thank you for confirmation! Two things. 

1. Shall I remove the breaking change from the merge message?
2. It would be great if we can bump DF before its release. How does this work? open an issue there or just send a PR bumping the version?

westonpace: Let's merge it and I will make a PR to bump DF when the Substrait release is out.

PR #871, opened 2025-10-08
docs: clarify meaning of LocalFiles

### LINKED ISSUE #869: Clarify LocalFiles semantics and blob storage support
## Context

At my job, we are internally representing blob storage reads using advanced extensions. However, I thought that this is a common enough use case that it would be beneficial to have in the main substrait repository.

It looks like ReadRel could in theory hold blob storage access if we took advantage of the `LocalFiles` read_type:

```proto
message ReadRel {
   ...
  // Definition of which type of scan operation is to be performed
  oneof read_type {
    LocalFiles local_files = 6;
    ...
  }
```

`LocalFiles` contains a list of the following message:

```proto
message FileOrFiles {
  oneof path_type {
    // A URI that can refer to either a single folder or a single file
    string uri_path = 1;
    // A URI where the path portion is a glob expression that can
    // identify zero or more paths.
    // Consumers should support the POSIX syntax.  The recursive
    // globstar (**) may not be supported.
    string uri_path_glob = 2;
    // A URI that refers to a single file
    string uri_file = 3;
    // A URI that refers to a single folder
    string uri_folder = 4;
  }

  // The start position in byte to read from this item
  uint64 start = 7;
  // The length in byte to read from this item
  uint64 length = 8;

  oneof file_format {
    ParquetReadOptions parquet = 9;
    // ... other formats
  }
}
```

Note that `FileOrFiles` already has byte-range slicing fields (`start`, `length`) that are useful for distributed reads. Also note that `IcebergTable.MetadataFileRead` already uses remote URIs as an example: `s3://mybucket/mytable/<ver>-<uuid>.metadata.json`.

## Problem

What exactly *is* a "local" file? If it is meant to be a file local on a machine, then having a `uri` field seems too permissive, since the only valid URI in this case should be `file://path/to/file`.

This ambiguity means that implementations either:
- Use extensions for blob storage (like we do)
- Make assumptions about whether `LocalFiles` accepts remote URIs

## Proposed solutions

**Option 1: Restrict `LocalFiles` to local filesystem only**
- Clarify that only `file://` URIs are valid for `LocalFiles`
- Add a new `BlobStorage` or `RemoteFiles` read type for object stores (`s3://`, `gs://`, `abfs://`, etc.)
- The new type would reuse the existing byte-range slicing and format fields

**Option 2: Generalize `LocalFiles` to all file-based access**
- Rename `LocalFiles` to `Files` or `FileBasedTable`
- Clarify that URI schemes distinguish local (`file://`) from remote (`s3://`, `gs://`, `abfs://`)
- This would align with `IcebergTable` already accepting remote URIs

## Summary

There is great benefit in standardizing our representation of blob storage reads, as it is common operation in deconstructed databases. Standardizing would make plans more portable. Furthermore, the ambiguity in the meaning of `LocalFiles` is worth clarifying, regardless of whether or not there is broad support for introducing standard plan representations for blob storage. 

There is of course discussion that will have to be had about backwards compatibility if `LocalFiles` are altered. (Perhaps a reason to favor Option 1, which preserves existing behavior while adding new functionality!)

benbellick:
Closes #869 

This change clarifies the documentation surrounding `LocalFiles`. It will be clearer that `LocalFiles` can also include references to blob storage etc. as specified by the URI's protocol. 

It also moves the duplication of the `ReadRel` proto in the documentation to the end of the section to be consistent with other `*Rel`s documetation.

### DIFF
--- a/site/docs/relations/logical_relations.md
+++ b/site/docs/relations/logical_relations.md
@@ -100,12 +100,6 @@ happens the consumer can determine which slice a particular chunk belongs to. Fo
 possible approach is that a chunk should only be read if the midpoint of the chunk (dividing by
 2 and rounding down) is contained within the asked-for byte range.
 
-=== "ReadRel Message"
-
-    ```proto
-%%% proto.algebra.ReadRel %%%
-    ```
-
 #### Iceberg Table Type
 
 A Iceberg Table is a table built on [Apache Iceberg](https://iceberg.apache.org/). Iceberg tables can be read by either directly reading a [metadata file](https://iceberg.apache.org/spec/#table-metadata) or by consulting a [catalog](https://iceberg.apache.org/concepts/catalog/). 
@@ -119,6 +113,12 @@ Points to an [Iceberg metadata file](https://iceberg.apache.org/spec/#table-meta
 | metadata_uri    | A URI for an Iceberg metadata file. This current snapshot will be read from this file.  | Required |
 | snapshot_id    | The snapshot that should be read using id. If not provided, the current snapshot is read. Only one of snapshot_id or snapshot_timestamp should be set. | Optional |
 | snapshot_timestamp    | The snapshot that should be read using timestamp. If not provided, the current snapshot is read. | Optional |
+
+=== "ReadRel Message"
+
+    ```proto
+%%% proto.algebra.ReadRel %%%
+    ```
 
 
 ## Filter Operation


site/docs/relations/logical_relations.md
```diff
@@ -70,13 +70,24 @@ provide the consumer with more information on how to retrieve the data.
...
-#### Files Type
+#### Local Files
+
+Local files are references to files which are locally-accessible. This may include files located in the receiving engine's file system, or files accessible as blob storage (e.g., S3, GCS, Azure Blob Storage). The method of file retrieval is determined by the URI's protocol (e.g. `file://`, `s3://`, `gs://`).
 
 | Property                    | Description                                                       | Required |
 | --------------------------- | ----------------------------------------------------------------- | -------- |
 | Items                       | An array of Items (path or path glob) associated with the read.   | Required |
-| Format per item             | Enumeration of available formats. Only current option is PARQUET. | Required |
-| Slicing parameters per item | Information to use when reading a slice of a file.                | Optional |
```
benbellick, line 79: I removed these two lines because they are not technically a property of LocalFiles. Instead, they are properties of the individual Items themselves. Thus, I added a section on File Formats, similar to Slicing Files.

site/docs/relations/logical_relations.md
```diff
@@ -109,6 +114,12 @@ Points to an [Iceberg metadata file](https://iceberg.apache.org/spec/#table-meta
 | snapshot_id    | The snapshot that should be read using id. If not provided, the current snapshot is read. Only one of snapshot_id or snapshot_timestamp should be set. | Optional |
 | snapshot_timestamp    | The snapshot that should be read using timestamp. If not provided, the current snapshot is read. | Optional |
 
+=== "ReadRel Message"
+
+    ```proto
+%%% proto.algebra.ReadRel %%%
+    ```
+
```
benbellick, line 122: I moved these here because all other relations have a copy of the proto at the **end** of the documentation. Prior to my change, the duplication of the proto was randomly sandwiched between File Types and Iceberg Table Type.

Issue #753, opened 2024-12-06
Case Study to design logical TableFunction operator


drin:
I am interested in helping to design the spec for TableFunction in substrait and am hoping to track my own development in a way that provides a reference for discussion. As I mentioned in https://github.com/substrait-io/substrait/issues/745#issuecomment-2524165647, I will eventually try to have an implementation of 3 different types of table functions and explore what feels natural and what feels like an antipattern (if anyone already has strong opinions feel free to share, I have very little experience).

The 3 types of table functions:
1. A `TableFunction` acting as a leaf operator, e.g. `scan_arrow_ipc` [in duckdb][doc-scanarrow].
2. A `TableFunction` acting as a transformation of an input table to an output table. For this, I want to implement something a bit unique: a function that maintains the cardinality of the table (output rows == input rows) but applies a function across the columns, thereby changing the schema (like a projection, e.g. [<id>, <datacol1>, <datacol2>, ..., <datacolN>] -> [<id>, <avg across columns>].
3. A `TableFunction` acting like a fused operator, e.g. "GroupJoin" as in [Accelerating Queries with... Join by GroupJoin][paper-groupjoin].

I'll add more information to this issue as I prototype, if anyone has recommendations on a different way to track or reference the work, let me know and I can adjust as we like.

[doc-scanarrow]: https://duckdb.org/docs/extensions/arrow.html#functions

[paper-groupjoin]: https://www.vldb.org/pvldb/vol4/p843-moerkotte.pdf

jacques-n:
I think about it more like two distinct behaviors as opposed to three. 

A. Generator table function. Takes in only constant arguments and produces 0..N records. Operates as leaf in a tree.
B. Set-based table function: takes in a set of records and adds one or more additional columns to each. 

For your type 2, I think of that as a window function which possibly excludes certain (or all) input columns

Type A (your type 1) feels pretty simple and that we have most of the low-level concepts to build against. I could see:
* Add a new table function extension type.
* Declare input arguments for table function as a collection of constant arguments.
* Define the output to be a collection of output columns/fields rather than a single column

Type B (likely matches your type 3), we need to likely introduce a lateral operator or similar. Can you remind me how some tools like Calcite represent this? The lateral would accept a table function extension type that could be constants or field references. First use would probably be FLATTEN/UNNEST.

benbellick:
Hi, wondering if there is any update on the status of this? I'm interested in spending some time exploring solutions and wanted to make sure I have all available information.

[...]

Issue #193, opened 2022-05-11
Should we do something about protobuf IDLs not reachable from a Plan object?


jvanstraten:
I've been wondering for a while now what the purpose is of [function.proto](https://github.com/substrait-io/substrait/blob/main/proto/substrait/function.proto), [parameterized_types.proto](https://github.com/substrait-io/substrait/blob/main/proto/substrait/parameterized_types.proto), and [type_expressions.proto](https://github.com/substrait-io/substrait/blob/main/proto/substrait/type_expressions.proto). The messages in them are not reachable from `substrait.Plan`, and it seems like they're mostly concerned with things that are communicated via YAML rather than via the plans. Are these files simply dead code that no one has yet removed, or do they still serve a purpose?

Note that capabilities.proto also isn't reachable, but I believe it is at least referred to from the website in relation to a gRPC-based handshake.

jacques-n:
The purpose of these is to be able to communicate function definitions on the wire. For example, if an IDE probes a substrait consumer and asks "What are all the function signatures you support." I definitely was relying on these at one point in the substrait-java project (but don't remember offhand if that is still the case). 

One of the things that makes capabilities and the function definitions disconnected is that we don't have a "substrait protocol" if you will. Some way via rest or other rpc method a consumer and producer could interact. The handshake, function listings, capabilities, etc all fall into these categories. 

I'm inclined to leave as is for a little longer and see if we can better clarify the protocol side of things. Thoughts?

jvanstraten:
Isn't that the purpose of the YAML files, though? `substrait.Capabilities.SimpleExtension` currently also refers to YAML URIs and function names, rather than listing all the function prototypes outright. The way I've been picturing the handshake-y protocol, based on what already exists, is:

 1) producer asks consumer for `substrait.Capabilities`;
 2) if needed (probably not, because in all likelihood it already needs to know what's in it in order to meaningfully use it, though I suppose a producer that merely parses some text format like SQL and doesn't do any form of optimization, like Isthmus, might not need to know what the functions do), the producer can download and parse (and cache) the YAML files for details;
 3) producer sends consumer a `substrait.Plan`;
 4) consumer sends producer the result data in [TBD] format.

Or the responsibility for compatibility could be shifted to the consumer:

 1) producer sends consumer a `substrait.Plan`;
[...]

PR #876, opened 2025-10-25
feat: introduction of simple table functions

### LINKED ISSUE #823: Where can Table Function be used?
I'm trying to understand how to use Table Function in Substrait. I looked at the docs at https://substrait.io/expressions/table_functions/ , but there are no explanation of how to use these. Is there some other place I can check?

I'd like to find out whether I could use Table Function to implement Presto's UNNEST: https://prestodb.io/docs/current/sql/select.html#unnest

Thanks.

benbellick:
- feat: introduces a notion of YAML-defined table functions which are represented as a new `TableFunctionRel` (take in no relations)

Closes #823

### DIFF
--- a/extensions/functions_table.yaml
+++ b/extensions/functions_table.yaml
@@ -1,19 +1,17 @@
 %YAML 1.2
 ---
 # Table functions: Functions that produce relations (zero or more records).
-# Currently, only 0-input functions are supported - these take constant arguments
-# and generate data as leaf operators.
 urn: extension:io.substrait:functions_table
 table_functions:
   - name: "generate_series"
     description: >-
       Generates a series of integer values from start to stop, incrementing by step.
 
-      Takes constant arguments and produces zero or more records containing a single
-      integer value. The series includes both the start and stop values if they fall
-      on a step boundary. If step is positive, stops when the value exceeds stop.
-      If step is negative, stops when the value is less than stop. Returns empty if
-      step is zero or if the step direction doesn't allow reaching stop from start.
+      Produces zero or more records containing a single integer value. The series
+      includes both the start and stop values if they fall on a step boundary.
+      If step is positive, stops when the value exceeds stop. If step is negative,
+      stops when the value is less than stop. Returns empty if step is zero or if
+      the step direction doesn't allow reaching stop from start.
     impls:
       - args:
           - name: start
@@ -25,7 +23,6 @@ table_functions:
           - name: step
             value: i64
             description: The increment between values
-            constant: true
         return:
           names:
             - value
@@ -42,7 +39,6 @@ table_functions:
           - name: step
             value: i32
             description: The increment between values
-            constant: true
         return:
           names:
             - value
--- a/extensions/functions_table.yaml
+++ b/extensions/functions_table.yaml
@@ -1,0 +1,102 @@
+%YAML 1.2
+---
+# Table functions: Functions that produce relations (zero or more records).
+# Currently, only 0-input functions are supported - these take constant arguments
+# and generate data as leaf operators.
+urn: extension:io.substrait:functions_table
+table_functions:
+  - name: "generate_series"
+    description: >-
+      Generates a series of integer values from start to stop, incrementing by step.
+
+      Takes constant arguments and produces zero or more records containing a single
+      integer value. The series includes both the start and stop values if they fall
+      on a step boundary. If step is positive, stops when the value exceeds stop.
+      If step is negative, stops when the value is less than stop. Returns empty if
+      step is zero or if the step direction doesn't allow reaching stop from start.
+    impls:
+      - args:
+          - name: start
+            value: i64
+            description: The starting value of the series
+          - name: stop
+            value: i64
+            description: The ending value of the series (inclusive)
+          - name: step
+            value: i64
+            description: The increment between values
+            constant: true
+        deterministic: true
+        sessionDependent: false
+        schema:
+          names:
+            - value
+          struct:
+            types:
+              - i64
+      - args:
+          - name: start
+            value: i32
+            description: The starting value of the series
+          - name: stop
+            value: i32
+            description: The ending value of the series (inclusive)
+          - name: step
+            value: i32
+            description: The increment between values
+            constant: true
+        deterministic: true
+        sessionDependent: false
+        schema:
+          names:
+            - value
+          struct:
+            types:
+              - i32
+      - args:
+          - name: start
+            value: i64
+            description: The starting value of the series
+          - name: stop
+            value: i64
+            description: The ending value of the series (inclusive)
+        deterministic: true
+        sessionDependent: false
+        schema:
+          names:
+            - value
+          struct:
+            types:
+              - i64
+      - args:
+          - name: start
+            value: i32
+            description: The starting value of the series
+          - name: stop
+            value: i32
+            description: The ending value of the series (inclusive)
+        deterministic: true
+        sessionDependent: false
+        schema:
+          names:
+            - value
+          struct:
+            types:
+              - i32
+  - name: "unnest"
+    description: Expands a list literal into a set of rows, one row per element.
+    impls:
+      - args:
+          - name: input
+            value: "list<T>"
+            description: The list to unnest
+        deterministic: true
+        sessionDependent: false
+        # Schema references type parameter T from list<T>
+        # The field type is derived from the list element type
+        schema:
+          names:
+            - element
+          struct:
+            types:
+              - T
--- a/extensions/functions_table.yaml
+++ b/extensions/functions_table.yaml
@@ -28,7 +28,7 @@ table_functions:
             constant: true
         deterministic: true
         sessionDependent: false
-        schema:
+        return:
           names:
             - value
           struct:
@@ -47,7 +47,7 @@ table_functions:
             constant: true
         deterministic: true
         sessionDependent: false
-        schema:
+        return:
           names:
             - value
           struct:
@@ -62,7 +62,7 @@ table_functions:
             description: The ending value of the series (inclusive)
         deterministic: true
         sessionDependent: false
-        schema:
+        return:
           names:
             - value
           struct:
@@ -77,7 +77,7 @@ table_functions:
             description: The ending value of the series (inclusive)
         deterministic: true
         sessionDependent: false
-        schema:
+        return:
           names:
             - value
           struct:
@@ -94,7 +94,7 @@ table_functions:
         sessionDependent: false
         # Schema references type parameter T from list<T>
         # The field type is derived from the list element type
-        schema:
+        return:
           names:
             - element
           struct:
--- a/extensions/functions_table.yaml
+++ b/extensions/functions_table.yaml
@@ -84,11 +84,11 @@ table_functions:
             types:
               - i32
   - name: "unnest"
-    description: Expands a list literal into a set of rows, one row per element.
+    description: Expands a list expression into a set of rows, one row per element.
     impls:
       - args:
           - name: input
-            value: "list<T>"
+            value: list<T>
             description: The list to unnest
         deterministic: true
         sessionDependent: false
--- a/extensions/functions_table.yaml
+++ b/extensions/functions_table.yaml
@@ -26,8 +26,6 @@ table_functions:
             value: i64
             description: The increment between values
             constant: true
-        deterministic: true
-        sessionDependent: false
         return:
           names:
             - value
@@ -45,8 +43,6 @@ table_functions:
             value: i32
             description: The increment between values
             constant: true
-        deterministic: true
-        sessionDependent: false
         return:
           names:
             - value
@@ -60,8 +56,6 @@ table_functions:
           - name: stop
             value: i64
             description: The ending value of the series (inclusive)
-        deterministic: true
-        sessionDependent: false
         return:
           names:
             - value
@@ -75,8 +69,6 @@ table_functions:
           - name: stop
             value: i32
             description: The ending value of the series (inclusive)
-        deterministic: true
-        sessionDependent: false
         return:
           names:
             - value
@@ -90,8 +82,6 @@ table_functions:
           - name: input
             value: list<T>
             description: The list to unnest
-        deterministic: true
-        sessionDependent: false
         # Schema references type parameter T from list<T>
         # The field type is derived from the list element type
         return:
--- a/extensions/functions_table.yaml
+++ b/extensions/functions_table.yaml
@@ -72,17 +72,22 @@ table_functions:
             types:
               - i32
   - name: "unnest"
-    description: Expands a list expression into a set of rows, one row per element.
+    description: >-
+      Expands one or more list expressions into a set of rows.
+
+      When multiple lists are provided, produces rows by expanding the lists in parallel (like a zip operation).
+      For example, unnest([1,2], [3,4]) produces rows: (1,3), (2,4).
+      If lists have different lengths, shorter lists are padded with nulls.
     impls:
       - args:
           - name: input
-            value: list<T>
-            description: The list to unnest
-        # Schema references type parameter T from list<T>
-        # The field type is derived from the list element type
+            value: list<any1>
+            description: The list(s) to unnest
+        variadic:
+          min: 1
         return:
           names:
             - element
           struct:
             types:
-              - T
+              - any1
--- a/proto/substrait/algebra.proto
+++ b/proto/substrait/algebra.proto
@@ -558,7 +558,6 @@ message ExpandRel {
 // Table functions produce a table with a schema that can be:
 //   - Statically defined (concrete types in YAML)
 //   - Type-parameterized (derived from argument types, e.g., unnest(list<T>) -> {element: T})
-//   - Runtime-dependent (determined by inspecting data content, use derived: false)
 //
 // Future extensions may add an optional input field to support transformation
 // table functions that operate on input relations.
@@ -574,8 +573,6 @@ message TableFunctionRel {
   // number of arguments specified in the function definition from the YAML file,
   // and the argument types must also match exactly:
   //  - Value arguments must be bound using FunctionArgument.value.
-  //    Currently (0-input functions only), expressions must be constants
-  //    (literals or expressions evaluable without input data).
   //  - Type arguments must be bound using FunctionArgument.type.
   //  - Enum arguments must be bound using FunctionArgument.enum with a
   //    string that case-insensitively matches one of the allowed options.
@@ -586,25 +583,12 @@ message TableFunctionRel {
   // cases.
   repeated FunctionOption options = 5;
 
-  // Indicates whether the schema was derived from the YAML function definition:
-  // - If true: the table_schema was computed from the ExpressionNamedStruct in the
-  //   YAML file by evaluating it with the bound argument types. This includes:
-  //   * Static schemas (concrete field types)
-  //   * Type-parameterized schemas (e.g., T from list<T>)
-  // - If false: the table_schema was determined by the plan producer based on runtime
-  //   data inspection (YAML omits the return field)
-  //
-  // This value must be true if and only if a return field is provided in the YAML
-  // definition of this function.
-  bool derived = 6;
+  // The concrete output schema of the relation. For schemas which can be derived
+  // from their YAML specification this must match the schema computed by evaluating the
+  // YAML ExpressionNamedStruct with the bound argument types.
+  NamedStruct table_schema = 6;
 
-  // The concrete output schema of the relation. For derived schemas (derived=true),
-  // this must match the schema computed by evaluating the YAML ExpressionNamedStruct
-  // with the bound argument types. For runtime-dependent schemas (derived=false),
-  // this is provided by the plan producer.
-  NamedStruct table_schema = 7;
-
-  substrait.extensions.AdvancedExtension advanced_extension = 8;
+  substrait.extensions.AdvancedExtension advanced_extension = 7;
 }
 
 // A relation with output field names.
--- a/proto/substrait/algebra.proto
+++ b/proto/substrait/algebra.proto
@@ -555,9 +555,10 @@ message ExpandRel {
 // Invokes a table-valued function that produces a relation (zero or more records).
 //
 //
-// Table functions produce a table with either:
-//   - A schema that can be derived based on argument types (type-parameterized functions)
-//   - A schema that depends on runtime data (use derived: false)
+// Table functions produce a table with a schema that can be:
+//   - Statically defined (concrete types in YAML)
+//   - Type-parameterized (derived from argument types, e.g., unnest(list<T>) -> {element: T})
+//   - Runtime-dependent (determined by inspecting data content, use derived: false)
 //
 // Future extensions may add an optional input field to support transformation
 // table functions that operate on input relations.
@@ -572,7 +573,6 @@ message TableFunctionRel {
   // The arguments to be bound to the function. This must have exactly the
   // number of arguments specified in the function definition from the YAML file,
   // and the argument types must also match exactly:
-  //
   //  - Value arguments must be bound using FunctionArgument.value.
   //    Currently (0-input functions only), expressions must be constants
   //    (literals or expressions evaluable without input data).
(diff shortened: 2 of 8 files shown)


proto/substrait/algebra.proto
```diff
@@ -552,6 +552,51 @@ message ExpandRel {
...
+  //
+  //  - Value arguments must be bound using FunctionArgument.value.
+  //    Currently (0-input functions only), expressions must be constants
+  //    (literals or expressions evaluable without input data).
+  //  - Type arguments must be bound using FunctionArgument.type.
+  //  - Enum arguments must be bound using FunctionArgument.enum with a
+  //    string that case-insensitively matches one of the allowed options.
+  repeated FunctionArgument arguments = 3;
+
+  // The derived fields indicates whether or not the YAML file produced the schema:
```
benbellick: Every other function type has a `FunctionOption` field around here. Do we need that for table functions as well? 
```protobuf
message TableFunctionRel {
    ...
    repeated FunctionOption options = 4;
    ...
}
```

text/simple_extensions_schema.yaml
```diff
@@ -307,3 +311,33 @@ $defs:
...
+          type: object
+          additionalProperties: false
+          properties:
+            args:
+              $ref: "#/$defs/arguments"
+            sessionDependent:
+              $ref: "#/$defs/sessionDependent"
+            deterministic:
+              $ref: "#/$defs/deterministic"
+            schema:
```
benbellick: Would it be better to come up with a more explicit way to mark the inability of deriving an output schema for a particular table function? Or is it sufficient to just let the absence of the `schema` field indicate this?

proto/substrait/algebra.proto
```diff
@@ -552,6 +552,51 @@ message ExpandRel {
...
+// Invokes a table-valued function that produces a relation (zero or more records).
+//
+//
+// Table functions produce a table with either:
+//   - A schema that can be derived based on argument types (type-parameterized functions)
+//   - A schema that depends on runtime data (use derived: false)
+//
+// Future extensions may add an optional input field to support transformation
+// table functions that operate on input relations.
+message TableFunctionRel {
```
benbellick, line 564: Is a more appropriate name for this `GeneratorTableFunctionRel` as introduced by 
@jacques-n [here](https://github.com/substrait-io/substrait/issues/753#issuecomment-2530029519)? 


In my mind there are two possible paths we can go down which determine the appropriate name. 
1. We could explore expanding this proto (by e.g. adding a relations input field) so that it appropriately models both kinds of table functions. In this case, keeping the name as `TableFunctionRel` makes sense. 
2. We could introduce a brand new relation to represent table functions which take relations as input. In this case, it would make sense to rename this relation to `GeneratorTableRel` or some name which distinguishes it from the other kind of table function.

proto/substrait/function.proto
```diff
@@ -106,6 +106,19 @@ message FunctionSignature {
     }
   }
 
+  message TableFunction {
```
benbellick: I added this to be consistent with the other function definitions, though AFAICT, this isn't really used anywhere. I found #193 from a few years ago which suggests that this should continue to be updated, but I'm not sure if that is still the prevailing philosophy.

proto/substrait/function.proto
```diff
@@ -106,6 +106,19 @@ message FunctionSignature {
...
+    repeated Argument arguments = 2;
+    repeated string name = 3;
+    Description description = 4;
+
+    bool deterministic = 7;
+    bool session_dependent = 8;
+
+    NamedStruct schema = 9;
+
+    repeated Implementation implementations = 10;
```
benbellick: Other functions have variadic final argument logic like: 
```protobuf
message TableFunction {
    ...
    oneof final_variable_behavior {
      FinalArgVariadic variadic = 10;
      FinalArgNormal normal = 11;
    }
    ...
}
```

However, I skipped it for now for simplicity. Is it necessary on a first pass?

site/docs/expressions/table_functions.md
```diff
@@ -1,8 +1,265 @@
...
+}
+```
+
+## Future Extensions
+
+The current specification focuses on 0-input (generator/leaf) table functions. Future versions may support:
+
+- **Transformation table functions**: Functions that take an input relation and transform it (by adding an optional `input` field to `TableFunctionRel`)
+- **Set-returning functions**: Functions that process input records and produce multiple output records per input
+- **Lateral joins**: Applying table functions to each row of an input relation
```
benbellick, line 286: Ideally, we would be able to model something like:
```sql
SELECT unnest(col) FROM table;
```
However, my understanding is that this is really syntactic sugar for: 
```sql
SELECT u.value 
FROM table t 
CROSS JOIN LATERAL unnest(t.col) AS u(value);
```

If that is correct, then we really need `LateralJoinRel` to properly model this usage of `unnest`. 

It is probably worth then clarifying somewhere in the documentation introduced in this PR that the arguments to a table function can't be field references.

site/docs/extensions/generate_function_docs.py
```diff
@@ -125,19 +125,27 @@ def write_markdown(file_obj: dict, file_name: str) -> None:
                 # If the return value for the function implementation is multiple lines long,
                 # print each line separately. This is the case for some functions in
                 # functions_arithmetic_decimal.yaml
-                if "\n" in impl["return"]:
```
benbellick, line 128: TODO: this could probably use a little work to make it a bit nicer. I'll work on that later once we have some general agreement on the spec changes. 

The change here is for now just to get CI/CD to pass.

site/docs/expressions/table_functions.md
```diff
@@ -1,8 +1,265 @@
...
+## Function Signatures
+
+Table functions are defined in YAML extension files, similar to scalar, aggregate, and window functions. A table function signature specifies:
+
+- **Arguments**: The parameters the function accepts (must be constant expressions)
+- **Schema**: The output schema of the generated relation
+- **Determinism**: Whether the function produces the same output for the same inputs
+- **Session Dependency**: Whether the function depends on session state
+
+## Schema Determination
```
benbellick, line 25: TODO: I think this section could use a little bit of work: 
- add some examples with YAML files to demonstrate the difference between derived / explicit
- Make the fact that there are these two kinds a bit clearer.

extensions/functions_table.yaml
```diff
@@ -0,0 +1,102 @@
...
+        deterministic: true
+        sessionDependent: false
+        return:
+          names:
+            - value
+          struct:
+            types:
+              - i32
+  - name: "unnest"
+    description: Expands a list literal into a set of rows, one row per element.
```
yongchul: list expression? we don't have to restrict it to a literal

proto/substrait/algebra.proto
yongchul: From my biased view, it is better to categorize them... roughly

* Scan like (0 relational input)
* Project or Filter like (1 relational input)
* (N-ary) Join like (2+ relational inputs optional grouping support)
* Group like (1 relational input with grouping support)

This gives way more modular and systematic way to describe a table function.

proto/substrait/function.proto
yongchul: Because you have unnest, I think it is actually makes sense to have variadic support. 😆

yongchul (commented):
Flushing some comments as my meetings getting longer and have on call duty coming up in the afternoon. Will come back later today.

proto/substrait/algebra.proto
yongchul: For the sake of consistency, maybe it is worth it? I don't understand what this FunctionOption is for... (code archaeology time!)

extensions/functions_table.yaml
```diff
@@ -0,0 +1,102 @@
...
+    impls:
+      - args:
+          - name: input
+            value: "list<T>"
```
yongchul: perhaps we can follow other extensions. for instance, in decimal rounding functions, we just write type rather than wrap it in a quote (i.e., decimal<P,S> rather than "decimal<P,S>").

Also, I'm wondering whether we can follow the same pattern as others (i.e., allow type expression).

proto/substrait/algebra.proto
```diff
@@ -552,6 +552,51 @@ message ExpandRel {
...
+  //    string that case-insensitively matches one of the allowed options.
+  repeated FunctionArgument arguments = 3;
+
+  // The derived fields indicates whether or not the YAML file produced the schema:
+  // - If true, the table_schema was produced purely from the type expressions in the
+  //   YAML file + the types of the provided arguments
+  // - If false, the table_schema was produced by the plan producer
+  //
+  // This value is required to be true if and only if a schema is provided in the YAML
+  // definition of this function.
```
yongchul: Can we simply have `table_schema` only? If it is empty, it must be from the YAML and should be able to be derived. Otherwise, we can just use the `table_schema`. If both are present, we'd better validate.

site/docs/expressions/table_functions.md
```diff
@@ -1,8 +1,319 @@
...
+Future extensions may add support for transformation table functions that consume and transform input relations by adding an optional input field to `TableFunctionRel`.
+
+## Function Signatures
+
+Table functions are defined in YAML extension files, similar to scalar, aggregate, and window functions. A table function signature specifies:
+
+- **Arguments**: The parameters the function accepts (must be constant expressions)
+- **Schema**: The output schema of the generated relation (may or may not be specified in YAML)
+- **Determinism**: Whether the function produces the same output for the same inputs
+- **Session Dependency**: Whether the function depends on session state
```
yongchul: I believe I understand what `session` here means but have we defined any of that in Substrait?

site/docs/expressions/table_functions.md
```diff
@@ -1,8 +1,319 @@
...
+
+Future extensions may add support for transformation table functions that consume and transform input relations by adding an optional input field to `TableFunctionRel`.
+
+## Function Signatures
+
+Table functions are defined in YAML extension files, similar to scalar, aggregate, and window functions. A table function signature specifies:
+
+- **Arguments**: The parameters the function accepts (must be constant expressions)
+- **Schema**: The output schema of the generated relation (may or may not be specified in YAML)
+- **Determinism**: Whether the function produces the same output for the same inputs
```
yongchul: I'm not sure whether we want to define these somewhat advanced properties (determinism and session dependency) for the table functions now... It's probably better to move those as hints?

extensions/functions_table.yaml
benbellick: Ah yes, I am operating under the assumption that we will indeed need to allow type expressions. I altered the `functions.proto` to accept a     `DerivationExpression.ExpressionNamedStruct`. I'm not entirely sure, but I believe this is meant to capture type expressions.

proto/substrait/algebra.proto
benbellick: Is your suggestion to have a different relation for each one of those? Or one shared relation? Maybe something like the following would be preferable...
```proto
message TableFunctionRel {
  ...
  oneof {
    ScanLike scan_like = 1; // PR implements this one
    //ProjectLike project_like = 2;
    //JoinLike join_like = 3;
    //GroupLike group_like = 4;
  }
  // implementations below
  ...
}
```

proto/substrait/function.proto
benbellick: I like the idea of introducing variadic argument handling, but as I understand it, it is a bit complicated in the current state to introduce variadics for things like `unnest`, since each argument of `unnest` could potentially be a different list type, e.g. `unnest(ARRAY[1,2,3], ARRAY["a","b","c"], ARRAY[true,true,false])`. 

So unlike the typical usage of variadic arguments for other function types, where the last argument is the same type, this usage is a bit more complicated and I'm not sure yet how it could be encoded. An approach is to instead keep the implementation as non-variadic and then introduce another tool to stitch relations together (something like a zip relation).

site/docs/expressions/table_functions.md
benbellick: I was following the pattern as in [scalar functions](https://substrait.io/expressions/scalar_functions/):
> Whether this function is influenced by the session context it is invoked within. For example, a function may be influenced by a user who is invoking the function, the time zone of a session, or some other non-obvious parameter. This can inform caching systems on whether a particular function is cacheable.

site/docs/expressions/table_functions.md
benbellick: I dropped both in a follow-up commit. More than happy to keep things simple!

site/docs/expressions/table_functions.md
benbellick: As mentioned in your lower comment, I just had them removed.

proto/substrait/algebra.proto
benbellick: I wanted to be consistent with the behavior of `ScalarFunctions`, which includes return types in the plans even though it is known from yaml files: 
```proto
  message ScalarFunction {
    ...
    // Must be set to the return type of the function, exactly as derived
    // using the declaration in the extension.
    Type output_type = 3;
    ...
  }
```

I figured that it makes the plans more explicit to always include the `table_schema`, regardless of if things were derived or not. However, I still could drop the `derived` field. Then we just specify that the schema _must_ conform to the implied schema in the yaml, if one is present. Otherwise, it can be anything set by the producer. What do you think?

extensions/functions_table.yaml
```diff
@@ -0,0 +1,92 @@
...
+          - name: start
+            value: i64
+            description: The starting value of the series
+          - name: stop
+            value: i64
+            description: The ending value of the series (inclusive)
+          - name: step
+            value: i64
+            description: The increment between values
+            constant: true
```
yongchul: Is this `constant` defined somewhere? What does it mean?

extensions/functions_table.yaml
```diff
@@ -0,0 +1,92 @@
...
+      If step is negative, stops when the value is less than stop. Returns empty if
+      step is zero or if the step direction doesn't allow reaching stop from start.
+    impls:
+      - args:
+          - name: start
+            value: i64
+            description: The starting value of the series
+          - name: stop
+            value: i64
+            description: The ending value of the series (inclusive)
```
yongchul, line 22: Looks like there is an interesting spilit between `stop` being `inclusive` and `exclusive`.

* python, pandas and snowflake exclude the stop value.
* PostgreSQL, T-SQL, ZetaSQL include the stop value.

I guess either way is fine... Another problem is that some systems do support generating series for other numeric types (e.g., fp32, fp64, decimal) but we can add those later.

extensions/functions_table.yaml
```diff
@@ -0,0 +1,92 @@
...
+# and generate data as leaf operators.
+urn: extension:io.substrait:functions_table
+table_functions:
+  - name: "generate_series"
+    description: >-
+      Generates a series of integer values from start to stop, incrementing by step.
+
+      Takes constant arguments and produces zero or more records containing a single
+      integer value. The series includes both the start and stop values if they fall
+      on a step boundary. If step is positive, stops when the value exceeds stop.
```
yongchul: Do we want to say something about the default value of `step`?

proto/substrait/algebra.proto
```diff
@@ -552,6 +552,61 @@ message ExpandRel {
...
+  // Points to a function_anchor defined in this plan, which must refer
+  // to a table function in the associated YAML file. Avoid using
+  // anchor/reference zero.
+  uint32 function_reference = 2;
+
+  // The arguments to be bound to the function. This must have exactly the
+  // number of arguments specified in the function definition from the YAML file,
+  // and the argument types must also match exactly:
+  //  - Value arguments must be bound using FunctionArgument.value.
+  //    Currently (0-input functions only), expressions must be constants
```
yongchul: We do not need to restrict ourselves to `constants`. As long as it is expression, it should be fine as long as producer implement those and future proof. It is likely many engines just support scalar expressions in the beginning but we do not necessarily restrict implementers on this.

Curious about how others think.

proto/substrait/algebra.proto
```diff
@@ -552,6 +552,61 @@ message ExpandRel {
...
+  // cases.
+  repeated FunctionOption options = 5;
+
+  // Indicates whether the schema was derived from the YAML function definition:
+  // - If true: the table_schema was computed from the ExpressionNamedStruct in the
+  //   YAML file by evaluating it with the bound argument types. This includes:
+  //   * Static schemas (concrete field types)
+  //   * Type-parameterized schemas (e.g., T from list<T>)
+  // - If false: the table_schema was determined by the plan producer based on runtime
+  //   data inspection (YAML omits the return field)
```
yongchul: `based on runtime data inspection` is not necessary.

proto/substrait/algebra.proto
```diff
@@ -552,6 +552,61 @@ message ExpandRel {
...
+  //
+  // This value must be true if and only if a return field is provided in the YAML
+  // definition of this function.
+  bool derived = 6;
+
+  // The concrete output schema of the relation. For derived schemas (derived=true),
+  // this must match the schema computed by evaluating the YAML ExpressionNamedStruct
+  // with the bound argument types. For runtime-dependent schemas (derived=false),
+  // this is provided by the plan producer.
+  NamedStruct table_schema = 7;
```
yongchul: I am wondering whether we can omit `derived`. Always having `table_schema` is fine and make it self-contained to some degree. Then, `derived=true` is implied and meaningful if and only if the engine is capable of handling YAML and YAML entry has struct there.

site/docs/expressions/table_functions.md
yongchul: I don't understand `SELECT unnest(col) FROM table`. What is the type of that statement?

Below makes sense.

```sql
SELECT LIST(UNNEST(col)) FROM table
```

If you want to flatten all the lists in the table.col, then your lateral join is way to go. "Table" can only appear in the FROM clause in SQL (I believe 😄 ).

proto/substrait/function.proto
yongchul: Indeed above case requires implicit type coercion rule which Substrait chose not to. So strictly speaking, your example is an invalid Substrait -- the user must have put cast (cough).

proto/substrait/algebra.proto
yongchul: I prefer `Rel` for each type, clearly expressing the degree of input at message level. But I'm open to exploring `oneof` definition.

site/docs/expressions/table_functions.md
benbellick: At least in postgresql, `SELECT unnest(col) FROM table;` is legal (see example [here](https://www.db-fiddle.com/f/cEaSifnLbWpZESAvx3xtJP/1)). This is documented [here](https://www.postgresql.org/docs/current/xfunc-sql.html#XFUNC-SQL-FUNCTIONS-RETURNING-SET). 
> Functions returning sets can also be called in the select list of a query. For each row that the query generates by itself, the set-returning function is invoked, and an output row is generated for each element of the function's result set. 

and also
> PostgreSQL's behavior for a set-returning function in a query's select list is almost exactly the same as if the set-returning function had been written in a LATERAL FROM-clause item instead. For example,
```sql
SELECT x, generate_series(1,5) AS g FROM tab;
```
> is almost equivalent to

```sql 
SELECT x, g FROM tab, LATERAL generate_series(1,5) AS g;
```
> It would be exactly the same, except that in this specific example, the planner could choose to put g on the outside of the nested-loop join, since g has no actual lateral dependency on tab. That would result in a different output row order. Set-returning functions in the select list are always evaluated as though they are on the inside of a nested-loop join with the rest of the FROM clause, so that the function(s) are run to completion before the next row from the FROM clause is considered.

Though of course, postgresql is just one DB in a sea of DBs 😅

extensions/functions_table.yaml
benbellick: https://substrait.io/expressions/scalar_functions/#argument-types
> Value arguments: arguments that refer to a data value. These could be constants (literal expressions defined in the plan) or variables (a reference expression that references data being processed by the plan).

My understanding is that `constant` means that a value can only be a literal and not a reference to a value unknown at plan construction e.g. a column reference. Although I'm unsure why we wouldn't allow these values to be expressions which lack references. 

All that being said, postgresql does in fact accept column references in the arguments to generate_series (example [here](https://www.db-fiddle.com/f/cgG7YG1UWeFxEL2pACzNkT/0)), so I think it makes sense to lift this restriction.

extensions/functions_table.yaml
benbellick: We could add this as a function option which specifies if they include or exclude the stop value. What do others think?

proto/substrait/algebra.proto
benbellick: Upon thinking about it a bit, I agree. I will push a fix for that now.

proto/substrait/algebra.proto
benbellick: I'm okay with omitting `derived`. I included it initially because I thought it made the dependency a little clearer between the YAML and the plan. But ultimately, libraries implementing the substrait spec will have to check consistency between the table_struct and the yaml anyways, so it doesn't matter that much.

extensions/functions_table.yaml
benbellick: [Even step size can be dynamic from table data](https://www.db-fiddle.com/f/cgG7YG1UWeFxEL2pACzNkT/3)!

extensions/functions_table.yaml
benbellick: Would it be better to instead drop the implementations where `step` is missing? That way, we force the plans to be more explicit.

proto/substrait/algebra.proto
benbellick: Leaving this conversation unresolved if others have something to say, but my most recent commit has dropped `derived` from the proto. Thanks!

site/docs/expressions/table_functions.md
yongchul: Hmm... I see. Looks like PostgreSQL chose to allow the syntactic sugar (to me) which I disagree to have but oh well. 😆

proto/substrait/algebra.proto
benbellick: I think that these options are actually [these bits of the yaml](https://github.com/substrait-io/substrait/blob/29844f1b58a666f36493c27def553854e81328a7/extensions/functions_logarithmic.yaml#L12-L14): 
```yaml
%YAML 1.2
---
urn: extension:io.substrait:functions_logarithmic
scalar_functions:
  -
    name: "ln"
    description: "Natural logarithm of the value"
    impls:
      - args:
          - name: x
            value: i64
        options: # all of these parts below
          rounding:
          on_domain_error:
            values: [ NAN, "NULL", ERROR ]
          on_log_zero:
            values: [NAN, ERROR, MINUS_INFINITY]
        return: fp64
```

So given that, it would make sense to include them.

proto/substrait/function.proto
benbellick: Ah sorry I think I misunderstood the point you were making initially. I thought it was about wanting to call unnest across an arbitrary number of arrays all of potentially different types (e.g. `list<X>`, `list<Y>`, `list<Z>`, $X\neq Y \neq Z$) but it sounds like instead you mean calling unnest on arbitrary of lists, all with the **same type**. 

If that is the case, then use it makes sense to me to include variadic handling. 

[Postgresql does have some handling for `unnest` across different list types](https://www.postgresql.org/docs/18/queries-table-expressions.html#QUERIES-TABLEFUNCTIONS), but they make a point in the documentation to say that this is syntactic sugar: 
> The special table function UNNEST may be called with any number of array parameters, and it returns a corresponding number of columns, as if UNNEST ([Section 9.19](https://www.postgresql.org/docs/18/functions-array.html)) had been called on each parameter separately and combined using the ROWS FROM construct.

extensions/functions_table.yaml
yongchul: We shouldn't unnecessarily restrict ourselves to literals and this is another place we shouldn't IMO. Implementation may not support generic expression in certain places but we shouldn't prevent it from the spec. In fact, this is another potential information to encode in **dialect** whether an expression has some restrictions.

proto/substrait/algebra.proto
yongchul: It is fine to have `table_schema` always but there is consequence -- it can be large especially with names. So I prefer to be able to omit the `table_schema` so that let consumer derive it from YAML. This is another point to add to **dialect** whether the implementation supports YAML based derivation. If the implementation does not support derivation, the producer must send explicit `table_schema`.

extensions/functions_table.yaml
```diff
@@ -0,0 +1,93 @@
...
+      For example, unnest([1,2], [3,4]) produces rows: (1,3), (2,4).
+      If lists have different lengths, shorter lists are padded with nulls.
+    impls:
+      - args:
+          - name: input
+            value: list<any1>
+            description: The list(s) to unnest
+        variadic:
+          min: 1
+        return:
```
benbellick, line 88: I have made this a draft PR again. I need to have a proper think about how we can represent this dynamic return type if we really want to have these functions be variadic 😰

benbellick:
I'm going to close this one for now. I am not actively working on it, and I hope that @EpsilonPrime's PR will sufficiently replace the work here to introduce the unnest operator (https://github.com/substrait-io/substrait/pull/917)

PR #881, opened 2025-11-03
docs: clarify valid URNs

benbellick:
As discovered in [this](https://github.com/substrait-io/substrait-rs/pull/419#issue-3576398920) discussion with @mbrobbel, there is a need to clarify a URN ambiguity. Current `urn` implementation across java, python, and go assumes that there are exactly two colons, i.e. they all are using the regex `^extension:[^:]+:[^:]+$`.

~~Instead, we clarify that urns as defined here are exactly as in [rfc 8141](https://www.rfc-editor.org/rfc/rfc8141.html) but with the `urn:` prefix cut off.~~

We update the documentation to enforce regex `^extension:[^:]+:[^:]+$`.

- - -
This change is [Reviewable](https://reviewable.io/reviews/substrait-io/substrait/881)

### DIFF
--- a/proto/substrait/extensions/extensions.proto
+++ b/proto/substrait/extensions/extensions.proto
@@ -16,9 +16,11 @@ message SimpleExtensionURN {
   // 0 is a valid anchor/reference, but prefer non-zero values for ergonomics.
   uint32 extension_urn_anchor = 1;
 
-  // The extension URN that uniquely identifies this extension. This must follow the
-  // format extension:<OWNER>:<ID> and serves as the "namespace" of this extension.
-  // This must conform to the following regex: ^extension:[a-z0-9_.-]+:[a-z0-9_.-]+$
+  // The extension URN that uniquely identifies this extension. The canonical format
+  // is urn:substrait:extension:<NAMESPACE>:<ID>. For backwards compatibility, URNs
+  // beginning with "extension:" have "urn:substrait:" prepended before validation.
+  // The canonical form must be a valid RFC 8141 URN and conform to the regex:
+  // ^urn:substrait:extension:[a-z0-9_.-]+:[a-z0-9_.-]+$
   string urn = 2;
 }
 
--- a/site/docs/extensions/index.md
+++ b/site/docs/extensions/index.md
@@ -13,14 +13,16 @@ Some kinds of primitives are so frequently extended that Substrait defines a sta
 * Window Functions
 * Table Functions
 
-To extend these items, developers can create one or more YAML files that describe the properties of each of these extensions. Each YAML file must include a required `urn` field that uniquely identifies the extension. While these identifiers are URN-like but not technically URNs (they lack the `urn:` prefix), and they will be referred to as `extension URNs` for clarity.
+To extend these items, users can create one or more YAML files that describe the properties of each of these extensions. Each YAML file must include a required `urn` field that uniquely identifies the extension.
 
-This extension URN uses the format `extension:<OWNER>:<ID>`, where:
+The canonical form of an extension URN is `urn:substrait:extension:<NAMESPACE>:<ID>`, where:
 
-- `OWNER` represents the organization or entity providing the extension and should follow [reverse domain name convention](https://en.wikipedia.org/wiki/Reverse_domain_name_notation) (e.g., `io.substrait`, `com.example`, `org.apache.arrow`) to prevent name collisions
+- `NAMESPACE` represents the organization or entity providing the extension and should follow [reverse domain name convention](https://en.wikipedia.org/wiki/Reverse_domain_name_notation) (e.g., `io.substrait`, `com.example`, `org.apache.arrow`) to prevent name collisions
 - `ID` is the specific identifier for the extension (e.g., `functions_arithmetic`, `custom_types`)
 
-These URNs must match the regex `^extension:[a-z0-9_.-]+:[a-z0-9_.-]+$`.
+For backwards compatibility, if a URN begins with `extension:` instead of `urn:substrait:extension:`, the prefix `urn:substrait:` is prepended before validation. This means `extension:io.substrait:functions_list` is equivalent to `urn:substrait:extension:io.substrait:functions_list`.
+
+Extension URNs in canonical form must be valid [RFC 8141](https://www.rfc-editor.org/rfc/rfc8141.html) URNs (with NID `substrait`) and must match the regex `^urn:substrait:extension:[a-z0-9_.-]+:[a-z0-9_.-]+$`.
 
 The YAML file is constructed according to the [YAML Schema](https://github.com/substrait-io/substrait/blob/main/text/simple_extensions_schema.yaml). Each definition in the file corresponds to the YAML-based serialization of the relevant data structure. If a user only wants to extend one of these types of objects (e.g. types), a developer does not have to provide definitions for the other extension points.
 
--- a/site/docs/serialization/binary_serialization.md
+++ b/site/docs/serialization/binary_serialization.md
@@ -22,7 +22,7 @@ For simple extensions, a plan references the extension URNs associated with the 
 
 Simple extensions within a plan are split into three components: an extension URN, an extension declaration and a number of references.
 
-* **Extension URN**: A unique identifier for the extension following the format `extension:<OWNER>:<ID>` that identifies a YAML document specifying one or more specific extensions. Declares an anchor that can be used in extension declarations. The extension URN must conform to the regex `^extension:[a-z0-9_.-]+:[a-z0-9_.-]+$`.
+* **Extension URN**: A unique identifier for the extension in the canonical format `urn:substrait:extension:<NAMESPACE>:<ID>` that identifies a YAML document specifying one or more specific extensions. Declares an anchor that can be used in extension declarations. For backwards compatibility, URNs beginning with `extension:` have `urn:substrait:` prepended before validation. See [Simple Extensions](../extensions/index.md#simple-extensions) for details.
 * **Extension Declaration**: A specific extension within a single YAML document. The declaration combines a reference to the associated extension URN along with a unique key identifying the specific item within that YAML document (see [Function Signature](../extensions/index.md#function-signature)). It also defines a declaration anchor. The anchor is a plan-specific unique value that the producer creates as a key to be referenced elsewhere.
 * **Extension Reference**: A specific instance or use of an extension declaration within the plan body.
 
--- a/text/simple_extensions_schema.yaml
+++ b/text/simple_extensions_schema.yaml
@@ -7,7 +7,7 @@ required: [urn]
 properties:
   urn:
     type: string
-    pattern: "^extension:[a-z0-9_.-]+:[a-z0-9_.-]+$"
+    pattern: "^urn:substrait:extension:[a-z0-9_.-]+:[a-z0-9_.-]+$"
   dependencies:
     # For reusing type classes and type variations from other extension files.
     # The keys are namespace identifiers that you can then use as dot-separated
--- a/extensions/extension_types.yaml
+++ b/extensions/extension_types.yaml
@@ -1,5 +1,5 @@
 ---
-urn: extension:io.substrait:extension_types
+urn: urn:substrait:extension:io.substrait:extension_types
 types:
   - name: point
     structure:
--- a/extensions/functions_aggregate_approx.yaml
+++ b/extensions/functions_aggregate_approx.yaml
@@ -1,6 +1,6 @@
 %YAML 1.2
 ---
-urn: extension:io.substrait:functions_aggregate_approx
+urn: urn:substrait:extension:io.substrait:functions_aggregate_approx
 aggregate_functions:
   - name: "approx_count_distinct"
     description: >-
--- a/extensions/functions_aggregate_decimal_output.yaml
+++ b/extensions/functions_aggregate_decimal_output.yaml
@@ -1,6 +1,6 @@
 %YAML 1.2
 ---
-urn: extension:io.substrait:functions_aggregate_decimal_output
+urn: urn:substrait:extension:io.substrait:functions_aggregate_decimal_output
 aggregate_functions:
   - name: "count"
     description: Count a set of values. Result is returned as a decimal instead of i64.
--- a/extensions/functions_aggregate_generic.yaml
+++ b/extensions/functions_aggregate_generic.yaml
@@ -1,6 +1,6 @@
 %YAML 1.2
 ---
-urn: extension:io.substrait:functions_aggregate_generic
+urn: urn:substrait:extension:io.substrait:functions_aggregate_generic
 aggregate_functions:
   - name: "count"
     description: Count a set of values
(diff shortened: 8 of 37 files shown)


site/docs/extensions/index.md
```diff
@@ -13,7 +13,7 @@ Some kinds of primitives are so frequently extended that Substrait defines a sta
 * Window Functions
 * Table Functions
 
-To extend these items, developers can create one or more YAML files that describe the properties of each of these extensions. Each YAML file must include a required `urn` field that uniquely identifies the extension. While these identifiers are URN-like but not technically URNs (they lack the `urn:` prefix), they will be referred to as `extension URNs` for clarity.
+To extend these items, developers can create one or more YAML files that describe the properties of each of these extensions. Each YAML file must include a required `urn` field that uniquely identifies the extension. While these identifiers are URN-like but not technically URNs (they lack the `urn:` prefix), they will be referred to as `extension URNs` for clarity. (These URNs are otherwise valid [RFC 8141](https://www.rfc-editor.org/rfc/rfc8141.html) with the `"urn:` prefix removed from the front.)
```
yongchul: nit: should period go right to the right parenthesis?

site/docs/serialization/binary_serialization.md
```diff
@@ -22,7 +22,7 @@ For simple extensions, a plan references the extension URNs associated with the
 
 Simple extensions within a plan are split into three components: an extension URN, an extension declaration and a number of references.
 
-* **Extension URN**: A unique identifier for the extension following the format `extension:<OWNER>:<ID>` that identifies a YAML document specifying one or more specific extensions. Declares an anchor that can be used in extension declarations.
+* **Extension URN**: A unique identifier for the extension following the format `extension:<OWNER>:<ID>` that identifies a YAML document specifying one or more specific extensions. Declares an anchor that can be used in extension declarations. The URN is neccessarily a valid [RFC 8141](https://www.rfc-editor.org/rfc/rfc8141.html) but with the `"urn:"` prefix cutoff.
```
yongchul: nit: The URN with `"urn:"` prefix should conform [RFC 8141].?

proto/substrait/extensions/extensions.proto
```diff
@@ -29,6 +29,7 @@ message SimpleExtensionURN {
 
   // The extension URN that uniquely identifies this extension. This must follow the
   // format extension:<OWNER>:<ID> and serves as the "namespace" of this extension.
+  // Otherwise, the urn is a RFC 8141 valid URN with the "urn:" prefix taken off.
```
yongchul: nit: without the "urn:" prefix.

site/docs/extensions/index.md
benbellick: Yeah that is a bit of a confusing one. I ended up just removing the parentheses.

site/docs/extensions/index.md
```diff
@@ -13,13 +13,15 @@ Some kinds of primitives are so frequently extended that Substrait defines a sta
 * Window Functions
 * Table Functions
 
-To extend these items, developers can create one or more YAML files that describe the properties of each of these extensions. Each YAML file must include a required `urn` field that uniquely identifies the extension. While these identifiers are URN-like but not technically URNs (they lack the `urn:` prefix), they will be referred to as `extension URNs` for clarity.
+To extend these items, developers can create one or more YAML files that describe the properties of each of these extensions. Each YAML file must include a required `urn` field that uniquely identifies the extension. While these identifiers are URN-like but not technically URNs (they lack the `urn:` prefix), they will be referred to as `extension URNs` for clarity. These URNs must be valid [RFC 8141](https://www.rfc-editor.org/rfc/rfc8141.html) format without the `urn:` prefix.
```
vbarua: Minor adjustment for clarity.
```suggestion
To extend these items, users can create one or more YAML files that describe the properties of each of these extensions. Each YAML file must include a required `urn` field that uniquely identifies the extension. These identifiers are URN-like but not technically URNs (they are prefixed with `extension:` instead of `urn:`), and will be referred to as `extension URNs` for clarity.

Extension URNs must be valid [RFC 8141](https://www.rfc-editor.org/rfc/rfc8141.html) URNs when replacing `extension:` with `urn:`.
```

site/docs/serialization/binary_serialization.md
```diff
@@ -22,7 +22,7 @@ For simple extensions, a plan references the extension URNs associated with the
 
 Simple extensions within a plan are split into three components: an extension URN, an extension declaration and a number of references.
 
-* **Extension URN**: A unique identifier for the extension following the format `extension:<OWNER>:<ID>` that identifies a YAML document specifying one or more specific extensions. Declares an anchor that can be used in extension declarations.
+* **Extension URN**: A unique identifier for the extension following the format `extension:<OWNER>:<ID>` that identifies a YAML document specifying one or more specific extensions. Declares an anchor that can be used in extension declarations. The URN with the `urn:` prefix added must conform to [RFC 8141](https://www.rfc-editor.org/rfc/rfc8141.html).
```
vbarua: > The URN with the `urn:` prefix added must conform to [RFC 8141](https://www.rfc-editor.org/rfc/rfc8141.html).

It's a bit weird to say this. The way we've structured them now maps to 
```
         | <NID> | <NSS>
extension:<owner>:<id>
```

I guess they do conform to the RFC if we prefix `urn`, but the interpretation would be different technically:
```
   | <NID>   |   <NSS>
urn:extension:<owner>:<id>
```

What do you think about:
> The Extension URN with the `extension:` replaced with `urn:` must conform to RFC 8141

benbellick: Hmm... I see the point you are making. However, we don't actually enforce that the structure of the `<owner>` part is reverse DNS. Why don't we just loosen the restriction on the `urn` entirely and say: 
> The `urn` is required to be a valid URN when `urn:` is prepended to the string. The format must conform to `urn:extension:<Identifier>`. The recommended format for the identifier is `<Reverse-DNS-Name>:<any-valid-name>`. This is consistent with the default substrait extensions and prevents name collisions. 

To me, this feels more consistent with the `urn` spec. Maybe its just how my brain works, but saying "`urn:` added to the front makes it a valid URN" makes more sense to me than saying "`urn:` replacing `extension:` makes it a valid URN".

benbellick:
~~After a discussion with @vbarua, it makes sense to just enforce that the "URN" is something compliant with the regex `^extension:[^:]+:[^:]+$`. We can say it is URN-like, but we may as well be overly restrictive.~~

~~This regex is exactly what is implemented in the substrait libs for java, go, and python.~~

@jacques-n I have altered the regex to be `^extension:[a-z0-9_.-]+:[a-z0-9_.-]+$`. 

[playground](https://regex101.com/r/Qe8DgC/1)

site/docs/extensions/index.md
```diff
@@ -13,13 +13,15 @@ Some kinds of primitives are so frequently extended that Substrait defines a sta
...
+To extend these items, developers can create one or more YAML files that describe the properties of each of these extensions. Each YAML file must include a required `urn` field that uniquely identifies the extension. While these identifiers are URN-like but not technically URNs (they lack the `urn:` prefix), and they will be referred to as `extension URNs` for clarity.
 
 This extension URN uses the format `extension:<OWNER>:<ID>`, where:
 
 - `OWNER` represents the organization or entity providing the extension and should follow [reverse domain name convention](https://en.wikipedia.org/wiki/Reverse_domain_name_notation) (e.g., `io.substrait`, `com.example`, `org.apache.arrow`) to prevent name collisions
 - `ID` is the specific identifier for the extension (e.g., `functions_arithmetic`, `custom_types`)
 
+These URNs must match the regex `^extension:[a-zA-Z0-9_.-]+:[a-zA-Z0-9_.-]+$`.
```
yongchul: Now that I see it again, why do we allow upper case if we were to allow only narrow set of characters?

proto/substrait/extensions/extensions.proto
```diff
@@ -29,6 +29,7 @@ message SimpleExtensionURN {
 
   // The extension URN that uniquely identifies this extension. This must follow the
   // format extension:<OWNER>:<ID> and serves as the "namespace" of this extension.
+  // This must conform to the following regex: ^extension:[^:]+:[^:]+$
```
yongchul: but the regex we are enforcing is much stricter than the regex here.

benbellick: That was a mistake that I forgot to update in my last commit. Thanks for catching.

site/docs/extensions/index.md
benbellick: Do you recommend an even more restrictive urn? How about: 
`^extension:[a-z0-9_.-]+:[a-z0-9_.-]+$`

i.e same thing without capital letters.

benbellick: I updated it to be without capital letters, let me know what you think!

mbrobbel (commented):
Looking at this again I'm wondering if it wouldn't be easier to just use URNs as defined in RFC 8141?

benbellick:
> Looking at this again I'm wondering if it wouldn't be easier to just use URNs as defined in RFC 8141?

@mbrobbel The problem is that defining in terms of RFC 8141 is inherently clunky because of the decision to not include `urn:` in the string. This leaves us with three choices: 
1. clarify that an extension urn called `extension:<rest>` is only valid if `urn:extension:<rest>` is RFC 8141 vaild,
2. clarify that an extension urn called `extension:<rest>` is only valid if `urn:<rest>` is RFC 8141 valid, or
3. give our own format which we describe as URN-like and then formally give a regex. 

The problem with the first approach above is then `extension` becomes the `NID`, and so we have to put restrictions on what the `NSS` can be anyways.

The second approach is technically fine, but a bit clunky IMO. 

The third approach seems simpler all in all as you can check the string against a regex directly. It also gives us more flexibility in the future to add things like versioning to the end when we are ready. 

Also, the inspiration for this approach to using URN-like things came from [java's maven](https://maven.apache.org/pom.html#maven-coordinates), which is also not a general purpose URN. 


I am open to relying on the RFC 8141 spec, but it doesn't seem to me that that is necessarily the simplest solution. In hindsight I wish that we had included `urn:` at the beginning 😅. Another possible solution is to migrate to using `urn:` at the start and taking option 1 (with an extra regex for the `<NSS>`) but that means an extra migration. In which case, we might as well go with 3 for now and tackle 1 later soas not to have two urn-related migrations happening at once.

nielspardon:
I would also prefer to stay closer to the RFC.

> The problem with the first approach above is then `extension` becomes the `NID`, and so we have to put restrictions on what the `NSS` can be anyways.

Wikipedia says that the [NID should be registered with IANA](https://en.wikipedia.org/wiki/Uniform_Resource_Name#Namespaces) according to the RFC which probably makes `extension` not a good choice for a unique namespace identifier and something like `substrait` would be a better choice so you could have something like:

```
urn:substrait:extension:<rest>
```

benbellick:
> I would also prefer to stay closer to the RFC.
> 
> > The problem with the first approach above is then `extension` becomes the `NID`, and so we have to put restrictions on what the `NSS` can be anyways.
> 
> Wikipedia says that the [NID should be registered with IANA](https://en.wikipedia.org/wiki/Uniform_Resource_Name#Namespaces) according to the RFC which probably makes `extension` not a good choice for a unique namespace identifier and something like `substrait` would be a better choice so you could have something like:
> 
> ```
> urn:substrait:extension:<rest>
> ```

@nielspardon I do think that that is a good approach. What if we then said that valid URNs are `urn:substrait:extension:<rest>` where the `urn:substrait` portion is optional? That way we don't have to do a migration yet, but we could later transition to the fully explicit URN. I would prefer not to do any migration at the moment, considering we are in the middle of the uri -> urn migration.

nielspardon:
> That way we don't have to do a migration yet, but we could later transition to the fully explicit URN. I would prefer not to do any migration at the moment, considering we are in the middle of the uri -> urn migration.

sure, we can do the change as a 1.0 item. Just saying if we want to give this another go we probably should consider that NID should be something we could register with IANA if we wanted to.

benbellick:
Sounds good. We will still need to introduce some sort of regex to validate the `<NSS>` component. [This would be required to register with IANA anyways](https://www.rfc-editor.org/rfc/rfc8141.html#section-6.1). We also may want to withhold from doing any registration until 1.0, when we have a stable idea of what these should look like (e.g. we will want to include version in the string eventually).

benbellick:
@yongchul I would be happy with that decision. Thanks for keeping tabs on this one. I'll update the PR to reflect that today.  


So, would weconsider the following two urns equivalent for now? 
- `extension:io.substrait:functions_list`
- `urn:substrait:extension:io.substrait_functions_list`

What recommendation should we give for which of these go in the YAML file?

nielspardon:
> * `extension:io.substrait:functions_list`
> 
> * `urn:substrait:extension:io.substrait_functions_list`

I would use this for the second one:

`urn:substrait:extension:io.substrait:functions_list`

> What recommendation should we give for which of these go in the YAML file?

I would use the second one.

benbellick:
> > * `extension:io.substrait:functions_list`
> 
> > 
> 
> > * `urn:substrait:extension:io.substrait_functions_list`
> 
> 
> 
> I would use this for the second one:
> 
> 
> 
> `urn:substrait:extension:io.substrait:functions_list`
> 
> 
> 
> > What recommendation should we give for which of these go in the YAML file?
> 
> 
> 
> I would use the second one.

Ah my bad, that was a typo! Agreed :)

text/simple_extensions_schema.yaml
```diff
@@ -7,6 +7,7 @@ required: [urn]
 properties:
   urn:
     type: string
+    pattern: "^urn:substrait:extension:[a-z0-9_.-]+:[a-z0-9_.-]+$"
```
benbellick, line 10: This is technically a breaking change, as it would force extensions to update their URNs to the new format. However, we can make it so that all of the implementing libraries (go, python, rs, java) do a proper migration, where they can accept either old and new, and emit old. 

Later we emit new, and then finally we can consider dropping the old if we would like.

yongchul: I am probably being paranoid and overdoing here but if we do this, I suggest we register `substrait` with IANA ASAP. We don't have to make this a break change right now (simply make the regex treat `urn:substrait:` optional) but I don't have a strong opinion on this as I do not know how many things will actually break because of this change...

benbellick: I have no idea what the process looks like or how long it would take to register with IANA, but sounds good to me :)

benbellick: As for making it optional, it will break peoples YAML files, but that would be an easy fix. Then I would imagine that in all of the libraries, we make it so that they can accept both:
- `urn:substrait:extension:io.substrait:functions_list`, and
- `extension:io.substrait:functions_list`

And then they always emit
- `extension:io.substrait:functions_list`

Then one day we can switch them all to emitting
- `urn:substrait:extension:io.substrait:functions_list`

And then we can consider dropping support for the old URN.

yongchul (approved):
I don't have an issue with this but it would be great if we can register `substrait` as official namespace in IANA registry. Created #1016 to track the registration.

PR #882, opened 2025-11-05
feat: add enable_row_group_maxmin_index to ParquetReadOptions

EpsilonPrime:
Add enable_row_group_maxmin_index field to ParquetReadOptions to support row group min/max index-based filtering. When enabled, readers can use min/max statistics from Parquet row groups to skip row groups that do not contain data matching filter predicates, improving query performance.

This brings the Gluten tweak into Substrait as part of the Gluten informing process.


westonpace:
Why would this ever be disabled?

EpsilonPrime:
> Why would this ever be disabled?

Good question.  I'll check to see if I can eliminate this field in Gluten altogether.

EpsilonPrime:
Filed https://github.com/apache/incubator-gluten/pull/11277 to address this instead.

PR #883, opened 2025-11-05
feat(proto): add advanced_extension field to RelRoot

EpsilonPrime:
Add advanced_extension field to RelRoot message for consistency with other relation types. All other major relation types include this field at position 10 to allow producers and consumers to attach implementation-specific metadata or enhancements.

### DIFF
--- a/proto/substrait/algebra.proto
+++ b/proto/substrait/algebra.proto
@@ -558,6 +558,7 @@ message RelRoot {
   Rel input = 1;
   // Field names in depth-first order
   repeated string names = 2;
+  substrait.extensions.AdvancedExtension advanced_extension = 10;
 }
 
 // A relation (used internally in a plan)


proto/substrait/algebra.proto
```diff
@@ -558,6 +558,7 @@ message RelRoot {
   Rel input = 1;
   // Field names in depth-first order
   repeated string names = 2;
+  substrait.extensions.AdvancedExtension advanced_extension = 10;
```
yongchul, line 561: Any concrete scenario to have advanced extension at this level?

EpsilonPrime:
This was constructed automatically as part of my automated unfork Gluten efforts but I don't see the corresponding usage.  Closing.

EpsilonPrime:
Ah, it was to deal with output_schema which Gluten has added to RelRoot.  It doesn't belong anywhere higher (PlanRel) and although it could go in the Rel directly under the RelRoot it is semantically close to the list of names already at the root.

I was going to put it into an advanced extension to unfork that part of Gluten (which apparently is only used with Clickhouse) without having to fix ClickHouse.

Issue #879, opened 2025-10-29
introduce a standard JSON type


benbellick:
`JSON` has become a common datatype in many databases. I propose we add a definition of `JSON` as an extension type to standardize its usage in plans across many different query engines.

EpsilonPrime:
Gluten has added a definition for JSON in local files.  It's useful as a reference but probably not exactly what we need.

JSON has a few issues, one being that Substrait does not yet have great support for structured types (maybe a few engines can load it but transformation is light).  Another issue is compatibility.  I've heard stories about incompatibility with invalid JSON.

westonpace:
> JSON has a few issues, one being that Substrait does not yet have great support for structured types (maybe a few engines can load it but transformation is light). Another issue is compatibility. I've heard stories about incompatibility with invalid JSON.

I don't know that Substrait needs support for structured types.  From Substrait's perspective it is "just a type".  I think we would need to define the type class (what strings are legal JSON?), literals (can just be a string I assume), and ideally add some new functions.

I agree it would be pretty nice to add.

PR #887, opened 2025-11-11
feat: add simple extension for JSON

benbellick:
- introduce a simple extension for representing JSON (#879)

- - -
This change is [Reviewable](https://reviewable.io/reviews/substrait-io/substrait/887)

### DIFF
--- a/extensions/json.yaml
+++ b/extensions/json.yaml
@@ -12,19 +12,20 @@ types:
 scalar_functions:
   - name: "parse_json"
     description: >-
-      Parse a JSON string into a JSON value.
+      Parses a JSON string into a JSON value.
     impls:
       - args:
           - name: json_string
             value: string
         options:
           on_error:
-            description: Controls behavior when input is not valid JSON
+            description: Controls behavior when input is not valid JSON.
             values: [ ERROR, "NULL" ]
         return: u!json?
 
   - name: "to_string"
-    description: "Convert a JSON value to its string representation"
+    description: >-
+      Converts a JSON value to its string representation.
     impls:
       - args:
           - name: json_value
@@ -33,7 +34,7 @@ scalar_functions:
 
   - name: "json_extract"
     description: >-
-      Extract a value from JSON using a JSONPath expression.
+      Extracts a value from JSON using a JSONPath expression.
       JSONPath expressions should follow RFC 9535 (https://datatracker.ietf.org/doc/html/rfc9535).
     impls:
       - args:
@@ -43,10 +44,10 @@ scalar_functions:
             value: string
         options:
           on_invalid_path:
-            description: Controls behavior when the JSONPath expression is syntactically invalid
+            description: Controls behavior when the JSONPath expression is syntactically invalid.
             values: [ ERROR, "NULL", UNDEFINED ]
           on_path_not_found:
-            description: Controls behavior when the path does not exist in the JSON document
+            description: Controls behavior when the path does not exist in the JSON document.
             values: [ ERROR, "NULL" ]
         return: u!json?
 


extensions/json.yaml
benbellick, line 1: There are a few more functions I could imagine including, e.g. [json_each](https://neon.com/docs/functions/json_each), [json_keys](https://dev.mysql.com/doc/refman/8.4/en/json-search-functions.html#function_json-keys), etc. but I wanted to start with a small set of primitives.

extensions/json.yaml
```diff
@@ -0,0 +1,60 @@
+%YAML 1.2
+---
+urn: extension:io.substrait:json
+types:
+  - name: json
+    structure: string
```
benbellick: As I understand it, this bit here is necessary to be able to represent JSON literals within plans. E.g.: 
```
  Literal.UserDefined {
    type_reference: <json type anchor>
    struct: {
      fields: [
        { string: "{\"name\": \"Alice\", \"age\": 30}" }
      ]
    }
  }
```

extensions/json.yaml
```diff
@@ -0,0 +1,61 @@
...
+            values: [ ERROR, "NULL" ]
+        return: u!json?
+
+  - name: "to_string"
+    description: "Convert a JSON value to its string representation"
+    impls:
+      - args:
+          - name: json_value
+            value: u!json
+        return: string
```
yongchul, line 33: Can this return null somehow? Also do we want to have some sort of formatting options like indentation and etc.?

extensions/json.yaml
```diff
@@ -0,0 +1,61 @@
...
+
+  - name: "json_extract"
+    description: >-
+      Extract a value from JSON using a JSONPath expression.
```
yongchul: nit: Extracts? Or, be consistent whether to use third person plurals or not. My personal preference is using third person plurals (i.e., "this (function|type)" as an implicit subject)

extensions/json.yaml
```diff
@@ -0,0 +1,61 @@
+%YAML 1.2
+---
+urn: extension:io.substrait:json
+types:
+  - name: json
+    structure:
+      content: string
```
yongchul, line 7: Do you envision extending this into something like variant?

extensions/json.yaml
```diff
@@ -0,0 +1,61 @@
...
+  - name: "parse_json"
+    description: >-
+      Parse a JSON string into a JSON value.
+    impls:
+      - args:
+          - name: json_string
+            value: string
+        options:
+          on_error:
+            description: Controls behavior when input is not valid JSON
```
yongchul: nit: style. consistently put `.` or drop everywhere.

extensions/json.yaml
benbellick: These are good points, though I am unsure of the answer. On the one hand, I can see a case for assuming that values saved as JSON are already valid, but on the other hand, this may not actually be the case.

As for formatting options, that is definitely a good idea. I'm not sure what the right approach is to cover many general use cases without more research.

extensions/json.yaml
benbellick: I don't have a ton of experience with variants, only heard about them for the first time during the last sync. Would it be a better approach to wait to handle variants first?

extensions/json.yaml
benbellick: I am not particular about which style is used, though I will note that the usage is inconsistent in the other extensions. For example, `functions_set.yaml` uses "checks", whereas `functions_arithmetic_decimal.yaml` uses "Add two decimal values."

extensions/json.yaml
yongchul: Just checking whether you have thought about it. :smile Finding an intersection across popular implementations would be an interesting survey but I think we may end up with some sort of property bag that we define some standard options and the undefined are up to the implementations... Something along that line. 100% agree that no need to add at this point.

extensions/json.yaml
yongchul: Yeah, this is something to clean up as well. Hope we can set up some simple rules in the PR precheck.

extensions/json.yaml
yongchul: I can see that the actual systems will map it their own internal representations -- simply use this as a tag. So `content` is somewhat meaningless.

BTW, what's the difference between `to_string(json_value)` and `json_value.content`? Are they the same?

benbellick: The way I understand it, `structure` is just a means by which [to communicate literals in a plan (besides a custom `any` representation)](https://substrait.io/types/type_classes/#communicating-user-defined-types). So I think that `json_value.content` isn't a meaningful operation from a substrait perspective, even though it is would be identical to `to_string(json_value)` in the case of literals present in plans. But if we have a table of type `NSTRUCT<index::i32, data::JSON>` called `table`, then we can do: 
```sql
SELECT to_string(data) FROM table
```

But in this literal-less context, `json_value.content` doesn't mean anything. Let me know if I misunderstood anything :)

extensions/json.yaml
```diff
@@ -0,0 +1,62 @@
...
+    impls:
+      - args:
+          - name: json_value
+            value: u!json
+          - name: path
+            value: string
+        options:
+          on_invalid_path:
+            description: Controls behavior when the JSONPath expression is syntactically invalid.
+            values: [ ERROR, "NULL", UNDEFINED ]
```
EpsilonPrime, line 48: Velox once encountered three different JSON libraries.  One actually succeeded despite not checking validity because it only read the part of the JSON that mattered.

EpsilonPrime (commented):
So the purpose of this PR is to create a JSON type and some functions to manipulate it.

But it does not try to tackle ingestion as done here:  https://github.com/apache/incubator-gluten/blob/main/gluten-substrait/src/main/resources/substrait/proto/substrait/algebra.proto#L165

extensions/json.yaml
vbarua: IMO it might make sense to not include this in a first pass. On one hand `to_string` is probably a very common operation, but also I don't know if there _is_ a standard way to stringify JSON that is consistent across engines. Not to say it's not worth trying to define this, but if the intent is start small and expand, this doesn't feel like a small function to start with.

extensions/json.yaml
```diff
@@ -0,0 +1,62 @@
...
+    impls:
+      - args:
+          - name: json_value
+            value: u!json
+        return: string
+
+  - name: "json_extract"
+    description: >-
+      Extracts a value from JSON using a JSONPath expression.
+      JSONPath expressions should follow RFC 9535 (https://datatracker.ietf.org/doc/html/rfc9535).
```
vbarua, line 38: Oh yay, there's a standard for this, though it appears to be [in progress](https://mailarchive.ietf.org/arch/browse/jsonpath/?q=rfc9535%20OR%20%22draft-ietf-jsonpath-base%22) and not final. It does feel like the most _neutral_ path format to pick as I imagine most systems have their own flavours of it, which will likely represent a compatibility challenge much like regex has been.

extensions/json.yaml
```diff
@@ -0,0 +1,62 @@
...
+      Parses a JSON string into a JSON value.
+    impls:
+      - args:
+          - name: json_string
+            value: string
+        options:
+          on_error:
+            description: Controls behavior when input is not valid JSON.
+            values: [ ERROR, "NULL" ]
+        return: u!json?
```
vbarua, line 24: Including a `?` on the type is redundant when declaring a function with [nullability handling](https://substrait.io/expressions/scalar_functions/#nullability-handling) of MIRROR (which is the default when it is not declared). The nullability is entirely determined by the nullability of the inputs.

extensions/json.yaml
vbarua: Was that the validity of the JSON, or the validity of the JSON path? Though if I think on it, I can imagine both "working". Lazy evaluation of a valid path on malformed JSON that never hits the bad json, and lazy evaluation of an invalid path on JSON that short-circuits the bad parts of the path.

extensions/json.yaml
```diff
@@ -0,0 +1,62 @@
...
+          - name: path
+            value: string
+        options:
+          on_invalid_path:
+            description: Controls behavior when the JSONPath expression is syntactically invalid.
+            values: [ ERROR, "NULL", UNDEFINED ]
+          on_path_not_found:
+            description: Controls behavior when the path does not exist in the JSON document.
+            values: [ ERROR, "NULL" ]
+        return: u!json?
```
vbarua, line 52: The `?` is also redundant on here.

vbarua (commented):
I'm broadly in favour of introducing this. It doesn't solve the discussion we've had historically about whether we should standardize on json, variant or some combination thereof. That discussion though has usually been in the context of adding JSON as a _core_ type in the protos.

I think that adding JSON support in a YAML extension like this is much lower stakes way to experiment with it. In practical reality, a bunch of systems have a JSON type, it's unlikely to go away anytime soon, and people seem to love putting JSON in their data systems so being able to encode and capture these capabilities should help us bring in more users.

I do suspect that if we go down this path, we will eventually run into portability/compatibility issues around "I want to communicate JSON functions to a system that only has VARIANT" or vice-versa. At that point though, we can hopefully harness the power of a bunch of developers wanting it work for their system / use case to help us figure out a grand unified theory of JSON/VARIANT interop, and it will help to already have a model for JSON operations when we start that.

PR #893, opened 2025-11-19
chore: remove invalid test

benbellick:
There is a test in `equals.test` comparing two types which are not the same. Thus, this test is invalid. This PR just removes this test.


yongchul (approved):

jacques-n:
agree can delete but we should fix the underlying missing function problem.

benbellick:
@jacques-n Added an issue for it here: https://github.com/substrait-io/substrait/issues/894

vbarua (approved):

PR #900, opened 2025-11-24
docs: consistently mark 0 as valid anchor/reference

### LINKED ISSUE #899: consistently clarify that 0 is allowed value in anchors / references
This explicitly allowance is currently missing for at least UDTs.

benbellick:
There are lots of places where the validity of anchor/reference 0 is not explicitly mentioned. This PR adds explicitly clarification in the proto comments and the site docs to clarify that anchor/reference 0 is valid. 

Closes #899


proto/substrait/algebra.proto
```diff
@@ -1189,8 +1196,8 @@ message Expression {
   // A scalar function call.
   message ScalarFunction {
     // Points to a function_anchor defined in this plan, which must refer
-    // to a scalar function in the associated YAML file. Required; avoid
-    // using anchor/reference zero.
+    // to a scalar function in the associated YAML file.
+    // Required; 0 is a valid anchor/reference.
```
benbellick: This is the only one I am not sure about... I find the language here a bit confusing. Is it saying the two separate things:
1. this field is required, and
2. you must not use anchor/reference 0
?

Or is it saying "it is required not to use anchor/reference 0"? 

I think it is the former. In which case there is nothing special about `ScalarFunction.function_reference` that prevents it from being 0.

proto/substrait/type.proto
```diff
@@ -10,6 +10,9 @@ option go_package = "github.com/substrait-io/substrait-protobuf/go/substraitpb";
 option java_multiple_files = true;
 option java_package = "io.substrait.proto";
 
+// Note: type_variation_reference fields within Type messages reference a
+// type_variation_anchor defined in the plan's extension declarations. The value
+// 0 represents the system-preferred variation and is a valid reference value.
```
benbellick: Wasn't sure where a good place to put this was considering `type_variation_reference` shows up a ton of times.

nielspardon:
Given that the proposed documentation changes say "non-negative" would it make sense to always say `>= 0` in the proto comments to be super specific?

I do remember the substrait-validator having a check for the 0 anchors reporting a warning. Should probably be also updated at some point in time.

proto/substrait/algebra.proto
vbarua: I can confirm that the intent of this is:
1. It is required that this field be set.
2. 0 is a valid anchor / reference

vbarua:
I can provide some guidance on why the non-zero language exists for references. Effectively, if a `uint32` field in a protobuf message is not set explicilty, the value defaults to 0. This makes it difficult to distinguish unset references from those intentionally set to 0. When serializing protobuf for debugging/logging, by default `uint32` fields set to 0 are not output either which can make plans difficult read.

The stance of the project has generally been
```
0 is a valid anchor/reference, but prefer non-zero values for ergonomics
```

What do you think about including the explicit guidance to avoid zero values?

proto/substrait/type.proto
```diff
@@ -10,6 +10,10 @@ option go_package = "github.com/substrait-io/substrait-protobuf/go/substraitpb";
 option java_multiple_files = true;
 option java_package = "io.substrait.proto";
 
+// Note: type_variation_reference fields within Type messages reference a
+// type_variation_anchor defined in the plan's extension declarations. The value
+// 0 represents the system-preferred variation and is a valid reference value,
+// but prefer non-zero values for ergonomics.
```
vbarua: minor
```
// Note: type_variation_reference fields within Type messages reference a
// type_variation_anchor defined in the plan's extension declarations.
// 
// The value 0 represents the system-preferred variation and is a valid reference value.
```

It's a bit weird to include the
> but prefer non-zero values for ergonomics.

because we are saying to use 0 to communicate the use of the system-preferred variation. It's also redundant to include, because 0 is reserved in this way.

vbarua (approved):
Thanks for updating this. Left on minor comment, but otherwise looks good to me.

proto/substrait/type.proto
benbellick: Good catch, what I wrote is incorrect!

drin:
Should we also update tutorials for URN anchors to start from 0?
https://github.com/substrait-io/substrait/blob/main/site/docs/tutorial/final_plan.json

Edit: I just reread this to understand:
> 0 is a valid anchor/reference, but prefer non-zero values for ergonomics

I guess the idea is that anchors are unique, unordered values and they're all valid as long as they're unique (within a plan) at any given time?

proto/substrait/type.proto
yongchul: For type variation anchor defined in the extension, it is better to not use `0` because it is *reserved* IMO. Perhaps, we can document this in the type extension proto message as well as type variation documentation?

proto/substrait/extensions/extensions.proto
```diff
@@ -63,15 +68,18 @@ message SimpleExtensionDeclaration {
...
     // references the extension_urn_anchor defined for a specific extension URN.
     // If both extension_urn_reference and extension_uri_reference are present,
     // extension_urn_reference takes precedence.
+    // 0 is a valid anchor/reference, but prefer non-zero values for ergonomics.
     uint32 extension_urn_reference = 4;
 
     // A surrogate key used in the context of a single plan to reference a
-    // specific type variation
+    // specific type variation.
+    // 0 is a valid anchor/reference, but prefer non-zero values for ergonomics.
```
yongchul: shall we be explicit about using non-zero here?

proto/substrait/algebra.proto
```diff
@@ -388,6 +388,8 @@ message ConsistentPartitionWindowRel {
   // sorts, and bounds, since those must be consistent across the various functions in this rel.  Refer
   // to the `WindowFunction` message for a description of these fields.
   message WindowRelFunction {
+    // References a function_anchor defined in this plan.
```
yongchul, line 391: Looking this PR again, I'm wondering whether we can remove all the updates on the `.*_reference` because as long as you don't have `anchor` bearing non-zero, reference just follows. What do you think?

proto/substrait/extensions/extensions.proto
yongchul: Here I don't see a point to allow zero here because `0` is reserved but it might be a breaking change indeed.

proto/substrait/algebra.proto
benbellick: I totally agree actually 😅 I removed it!

yongchul (approved):

PR #904, opened 2025-12-03
docs: bad mkdoc link to logical_relations.md#join-operator

vbarua:


vbarua:
When running `mkdoc serve` locally, I noticed that it complained about
```
Doc file 'relations/physical_relations.md' contains a link 'logical_relations.md#join-operator', but the doc
```

I've updated the links
```
from: logical_relations.md#join-operator
to:   logical_relations.md#join-operation
```

benbellick (approved):
These changes are actually just a subset of @wackywendell's proposed PR (#833). I am approving, but I think we should focus on getting his change in. 

As an aside: I'm surprised we don't catch things like bad links in CI. Opened an issue here: #905

PR #889, opened 2025-11-13
feat: introduce lambda expressions

benbellick:
- add lambda expression proto message
- expand type parsing to include lambda types (e.g. `lambda<int32 -> string>`)
- update documentation to reflect new lambda expression
- check textformat proto examples of lambdas in CI/CD
- expand test case grammar to allow dependencies on extensions which aren't focus of test

closes #349

PR #910, opened 2025-12-03
fix: remove redundant nullability markers in test case grammar

### LINKED ISSUE #909: incorrect nullability in test case grammar
The test case grammar currently allows things like `[1, 2, 3]::List?<i32>?`. What is that meant to mean? I imagine the intention was to only allow `[1, 2, 3]::List?<i32>`. Furthermore, `tests/coverage/visitor.py` makes no reference to the `isnull` field, suggesting that this is in fact dead code.

benbellick:
Consistently allow types to declare their own nullable syntax in the test case antlr grammar. An organizational principle is adopted in which all types have a rule called `<x>Type`, which is responsible for managing the location of the `?` character to indicate nullability. This prevents the double `?` problem where syntax like `[1,2,3]::List?<i8>?` was previously allowed.

Closes #909


benbellick:
This is going to need to fix the problem for lambdas as well once #889 is in.

tests/coverage/test_coverage.py
```diff
@@ -499,3 +499,53 @@ def test_uri_match_in_get_function(
...
+        test_file = parse_string(header + case + "\n")
+        assert len(test_file.testcases) == 1
+
+
+def test_double_nullable_rejected():
+    """Verify that double-nullable syntax (e.g., Type??) is rejected by the parser."""
+    invalid_cases = [
+        "add([1, 2]::List?<i8>?, [3]::List?<i8>) = [1]::List?<i8>",
+        "add(1.5::dec??, 2.5::dec?) = 4.0::dec?",
+        "add('abc'::fchar?<3>?, 'def'::fchar?<3>) = 'abcdef'::fchar?<6>",
```
westonpace, line 540: Can you add a test case for `fchar<3?>`

westonpace: (that being said, given my amateurish reading of the grammar, I think we're ok)

westonpace (approved):

tests/coverage/test_coverage.py
benbellick: done!

tests/coverage/test_coverage.py
```diff
@@ -499,3 +499,54 @@ def test_uri_match_in_get_function(
...
+
+def test_nullable_types():
+    valid_cases = [
+        # Integer types
+        "add(1::i8?, 2::i8?) = 3::i8?",
+        "add(1::i16?, 2::i32?) = 3::i64?",
+        # Float types
+        "add(1.0::fp32?, 2.0::fp64?) = 3.0::fp64?",
+        # Decimal
+        "add(1.5::dec?, 2.5::dec?) = 4.0::dec?",
```
vbarua, line 512: I don't think you changed anything around the decimal type, but I'm surprised that an unparameterized decimal type is valid. 

Not a byproduct of this PR though.

vbarua (approved):
Your changes here look good and bring the parser inline with the written spec.

PR #913, opened 2025-12-08
fix(extensions): nullif output should always be nullable

### LINKED ISSUE #912: nullif function has incorrect nullabiltity
The [nullif](https://github.com/substrait-io/substrait/blob/5f031b69ed211e1ec307be3db7989d64c65d33a2/extensions/functions_comparison.yaml#L246-L255) function is defined as follows. From the description, it can return null even if both inputs are not null. As such, it should set the output nullability explicitly and use the DECLARED_NULLABILITY output mode.

Slimsammylim:
* Updated `nullif` function to return  `any1?` with `nullability: DECLARED_OUTPUT`
* Added a couple test cases to `nullif.test`

Closes #912

### DIFF
--- a/tests/cases/comparison/nullif.test
+++ b/tests/cases/comparison/nullif.test
@@ -2,7 +2,10 @@
 ### SUBSTRAIT_INCLUDE: '/extensions/functions_comparison.yaml'
 
 # basic: Basic examples without any special cases
+nullif(10::i8, 2::i8) = 10::i8
+nullif(3::i8, 3::i8) = null::i8
 nullif(1::i16, 5::i16) = 1::i16
+nullif(1::i16, 1::i16) = null::i16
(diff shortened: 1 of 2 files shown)


tests/cases/comparison/nullif.test
```diff
@@ -2,7 +2,10 @@
 ### SUBSTRAIT_INCLUDE: '/extensions/functions_comparison.yaml'
 
 # basic: Basic examples without any special cases
+nullif(10::i8, 2::i8) = 10::i8
+nullif(3::i8, 3::i8) = null::i8
```
vbarua, line 6: This also demonstrates why we need to use DECLARED_OUTPUT `any1?` for this. Both inputs are not-nullable, but the output is (literally) null.

`MIRROR` (the default mode) would require the output to be not-null.

vbarua:
Merging as this is a function update with 2 PMC votes.

PR #916, opened 2025-12-11
fix: month not extracting

mbwhite:
For SQL queries such as 

```
select month(o_orderdate) from orders;
```
 where `o_orderdate` is a SQL `DATE` type, an assertion error is thrown.  For Substrait-Java this is

```
java.lang.AssertionError
   at io.substrait.expression.EnumArg.of(EnumArg.java:29)
   ....
```

Compare the function for extract on a date, with the extract on a date, but with the offset argument.  Just the year fields are inaccurate, and should include the others.

### DIFF
--- a/extensions/functions_datetime.yaml
+++ b/extensions/functions_datetime.yaml
@@ -108,7 +108,8 @@ scalar_functions:
         return: i64
       - args:
           - name: component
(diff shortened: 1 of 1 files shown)


mbwhite:
Going to close this... discussion with @nielspardon suggested that the actual substrait function doesn't need to be altered - as the function that accepts the `indexing` and `date` argument can be used instead.

Issue #903, opened 2025-11-25
Clarify `any1` vs `T` in yaml files


benbellick:
There is a PR [here](https://github.com/substrait-io/substrait/pull/629) from 04/18/2024 which suggests that the usage of `T` is always incorrect in yaml files.

OTOH, there is [this PR](https://github.com/substrait-io/substrait/pull/713) from 03/13/2025 (and thus is more recent) which _adds_ functions which use parameters like `P`, `S`, etc.

This may lead one to believe that the distinction is that parameters like `T` can only be used as a stand-in for integer parameters, but then if that is the case, [the documentation currently says things like](https://substrait.io/expressions/scalar_functions/#parameterized-types):
> Types are parameterized by two types of values: by inner types (e.g. `List<K>`) and numeric values (e.g. `DECIMAL<P,S>`). Parameter names are simple strings (frequently a single character). There are two types of parameters: integer parameters and type parameters.

Which suggests that `K` can be used as the parameter for `List`.

Furthermore, the compound types section of the documentation uses non-any1 type variables quite liberally (e.g., `add(List<T>, T) => List<T>` in the examples).

However, searching through all YAML files shows that no functions actually use `value: T`, `value: K`, etc. - they all use `any1`, `any2`, etc. The only single-letter parameters used are integer parameters like `P`, `S` in `decimal<P,S>`.

### Suggestions

- Clarify in the documentation that `T`, `K`, `V` are conceptual examples for types, not valid YAML syntax
- Show that `any1`, `any2`, etc. should be used in actual YAML function definitions for types
- Distinguish between type parameters (use `any1-any9`) and integer parameters (use `P`, `S`, `N`, etc.)

PR #917, opened 2025-12-13
feat: introduce GenerateRel for lateral view and unnest operations

EpsilonPrime:
Add GenerateRel, a new relational operator that applies table functions to produce zero or more output rows per input row. This enables SQL LATERAL VIEW, EXPLODE, and UNNEST operations for flattening nested data structures.

## Changes

- Adds TableFunction message based on original plans for such a message
- Adds GenerateRel message which is the first use of table functions

## Key Design Decisions

1. **Dedicated TableFunction message**: Provides type safety and clear semantics for multi-row producing functions, following the pattern of ScalarFunction, WindowFunction, and AggregateFunction.

2. **preserve_on_empty flag**: Controls NULL handling when generators produce zero rows. Named to reflect behavior (row preservation) rather than SQL join terminology ("outer"). Equivalent to Spark's LATERAL VIEW OUTER or standard SQL's LEFT JOIN LATERAL.

3. **No automatic ordinal column**: Unlike ExpandRel which appends an i32 ordinal, GenerateRel relies on table functions to include position explicitly (e.g., posexplode includes pos field in output) when needed.

4. **TableFunction not in Expression**: Table functions produce multiple rows, incompatible with Expression semantics (single value). Only used in relational operators like GenerateRel.

## Comparison to ExpandRel

- ExpandRel: Fixed cardinality (plan time), for grouping sets/cube
- GenerateRel: Variable cardinality (runtime), for lateral view/unnest

## Related Work

Will be used to unfork the changes made in Apache Gluten PR [#574](https://github.com/apache/incubator-gluten/pull/574).

- - -
This change is [Reviewable](https://reviewable.io/reviews/substrait-io/substrait/917)


proto/substrait/algebra.proto
```diff
@@ -556,6 +556,43 @@ message ExpandRel {
...
+  //   This is the standard "explode" or "unnest" behavior.
+  //
+  // When true:
+  //   Input rows that produce zero output rows still generate one output row
+  //   with the input fields populated and the generator output fields set to NULL.
+  //   This is equivalent to Spark's "LATERAL VIEW OUTER" or can be achieved in
+  //   standard SQL via LEFT JOIN LATERAL.
+  //
+  // Optional; defaults to false.
+  bool preserve_on_empty = 4;
```
benbellick, line 660: This is a similar idea to this PR here from @yongchul: https://github.com/substrait-io/substrait/pull/890

I wonder if there is some way to unify these ideas?

proto/substrait/algebra.proto
```diff
@@ -556,6 +556,43 @@ message ExpandRel {
...
+
+  // For compatibility and optimization hints.
+  substrait.extensions.AdvancedExtension advanced_extension = 10;
```
```suggestion
  substrait.extensions.AdvancedExtension advanced_extension = 10;
```

None of the other usages of `advanced_extension` has a comment, so I think lets just drop it to be consistent. 

----------
Also, I understand wanting to set the field number to 10 for consistency, but there are actually usages of `advanced_extension` which are `10`, `9`, `5`, and `4` in this file. Any reason now to just make it 5 so that it is the next field? TBH though this is not particularly important to me.

proto/substrait/algebra.proto
```diff
@@ -592,6 +629,7 @@ message Rel {
     ConsistentPartitionWindowRel window = 17;
     ExchangeRel exchange = 15;
     ExpandRel expand = 16;
+    GenerateRel generate = 23;
```
benbellick: This feels more of a general relation, rather than a physical relation. Does it make sense to move it after `UpdateRel` above?

proto/substrait/algebra.proto
```diff
@@ -1219,6 +1257,48 @@ message Expression {
...
+    // Must be set to the output type of the table function. For table functions,
+    // this must be a Struct type where each field represents a column in the
+    // generated output rows.
+    //
+    // Examples:
+    //   - explode(array<i32>) -> Struct{value: i32}
+    //   - posexplode(array<string>) -> Struct{pos: i64, value: string}
+    //   - explode(map<string, i32>) -> Struct{key: string, value: i32}
+    //
+    // Required.
```
benbellick: Rather than embedding a bunch of examples in the proto file, could we introduce these functions formally in the yaml file and then just demonstrate their usage inside of the documentation?

proto/substrait/algebra.proto
```diff
@@ -1219,6 +1257,48 @@ message Expression {
...
+    Type output_type = 4;
+  }
```
benbellick, line 1363: FYI this PR has a lot of overlap with this one I wrote before: https://github.com/substrait-io/substrait/pull/876

Though what you have here is more general than I what I did, and so I think your approach is better. I just wanted to share in case any of the discussion there is useful. One thing in particular that was discussed there: have you at all thought about variadic handling for table functions?

site/docs/expressions/table_functions.md
```diff
@@ -1,8 +1,137 @@
...
+| Field | Type | Description | Required |
+|-------|------|-------------|----------|
+| function_reference | uint32 | Points to a function_anchor in the plan that references a table function in the extension YAML. 0 is a valid anchor/reference. | Required |
+| arguments | repeated FunctionArgument | Arguments bound to the table function (value, type, or enum arguments) | Required |
+| options | repeated FunctionOption | Behavior options for corner cases | Optional |
+| output_type | Type | The output type of the table function (must be a Struct type where each field is a generated column) | Required |
+
+### Output Type
+
+The `output_type` field must be a `Struct` type where each field represents a column in the generated output rows. This explicitly defines the schema of rows produced by the table function.
```
benbellick: Can the struct be nullable? Can it be a named struct?

site/docs/expressions/table_functions.md
```diff
@@ -1,8 +1,137 @@
...
+
+While specific table functions are defined in extension YAML files, common table functions include:
+
+### Array Functions
+
+| Function | Signature | Output | Description |
+|----------|-----------|--------|-------------|
+| `explode` | `array<T>` → `Struct{value: T}` | One row per array element | Flattens an array by producing one output row per element |
+| `posexplode` | `array<T>` → `Struct{pos: i64, value: T}` | One row per array element with position | Like explode but includes 0-indexed position |
+| `unnest` | `array<T>` → `Struct{value: T}` | One row per array element | Standard SQL synonym for explode |
```
benbellick, line 56: I mentioned it in a previous comment, but can we formally define these in a yaml file?

site/docs/relations/logical_relations.md
```diff
@@ -559,3 +559,106 @@ The operator that defines modifications of a database schema (CREATE/DROP/ALTER
...
+  Output: (no rows - filtered out)
+
+With preserve_on_empty=true:
+  Output: {user_id: 1, tags: [], tag: NULL}
+```
+
+### Examples
+
+**Array Explode:**
+```
```
benbellick, line 654: The infrastructre isn't really there until the [lambda PR is in](https://github.com/substrait-io/substrait/pull/889), but it would be great to have some proper proto examples showing how these things would look, like: 
https://github.com/substrait-io/substrait/blob/35101020d961eda48f8dd1aafbc794c9e5cac077/site/docs/expressions/lambda_expressions.md?plain=1#L55-L61

benbellick (commented):
Very excited for this one, as it would be useful for modeling `Unnest` at my job. 

Couple things that I think are high priority: 
1. Could we introduce the schema changes for table functions in [simple_extension_schema.yaml](https://github.com/substrait-io/substrait/blob/main/text/simple_extensions_schema.yaml)? 
2. Could we introduce some proper examples into something like `functions_explode.yaml` to show these definitions? This also helps ensure that downstream libraries are properly handling these things. 
3. For my particular use case, I would love to be able to represent [queries like](https://www.db-fiddle.com/f/c12oDn2H6WknH1NtTXoVx8/0) `SELECT UNNEST(ARRAY[1,2]), UNNEST(ARRAY[4,5,6])` which has a zipping behavior rather than cartesian product. What would be the best way to represent these? (One way I can think of is as variadic unnest call which has zipping behavior.)

All in all this will be great and I'm happy to help implement them in the downstream libraries.

westonpace (commented):
I have a general question.  There are (at least) two ways in which table functions are utilized.  I'm not entirely sure if these are two different concepts or the same concept.

# Explode / lateral join / cross apply

I believe `GenerateRel` captures these cases.  The input to the table function is the output of another relation.  The arguments will (typically) include field references (or scalar functions applied to field references).

# From

The other way in which table functions are applied is by using them directly in the FROM clause.  For example `SELECT * FROM myfunc(7)`.  For example see [postgres](https://www.postgresql.org/docs/7.3/xfunc-tablefunctions.html).  In this case the arguments are (typically? always?) all literals.

Is the `TableFunction` message here intended to eventually be usable in both contexts?  This PR itself only adds support for the first (which is fine).  Still, it seems like message itself wouldn't have to change.  I think the only change I can imagine is adding a new source rel that has no input rels and specifies the table function invocation.

proto/substrait/algebra.proto
EpsilonPrime: Having them all the same makes it easier for some proto uses (but I can't name any at the moment).  As part of Substrait 1.0 I'd want to see these moved up to 100 or higher.  Lower field numbers compress better so it'd be nice to reserve some space (not that we really need 100 reserved field numbers).

proto/substrait/algebra.proto
EpsilonPrime: done

site/docs/expressions/table_functions.md
EpsilonPrime: Nullable yes, named struct no.  Named structs are a different kind of object (i.e. not a type).  If you want name hints you can use that feature to provide them.  Substrait explicitly is ignoring names.

proto/substrait/algebra.proto
EpsilonPrime: An alternative would be to make this part of TableFunction but if there are other table expressions they might need this functionality.

site/docs/expressions/table_functions.md
EpsilonPrime: Actually, not nullable structs since you have to return some number of rows.

site/docs/expressions/table_functions.md
EpsilonPrime: done

site/docs/relations/logical_relations.md
EpsilonPrime: done

proto/substrait/algebra.proto
EpsilonPrime: The output needs to be determinable from the input.  So for more complicated structures we likely need to upgrade the grammar.  For now we can achieve most of what we need with function overloading.

EpsilonPrime:
> I have a general question. There are (at least) two ways in which table functions are utilized. I'm not entirely sure if these are two different concepts or the same concept.
>
> ...
>
> Is the `TableFunction` message here intended to eventually be usable in both contexts? This PR itself only adds support for the first (which is fine). Still, it seems like message itself wouldn't have to change. I think the only change I can imagine is adding a new source rel that has no input rels and specifies the table function invocation.

I believe it would work as long as you had a relation that worked just fine without requiring an input (which is currently the case here).  I've expanded the implementation to create a TableExpression which gives us even more future capability but we will still have the same requirement -- we need to have a relation that allows us to provide no input.

json_decode is one such function.  Getting a little meta, I could also see a function (say substrait_execute) that takes a substrait plan as an argument using this.  :)

benbellick:
@EpsilonPrime the link in your description to #574 is linking to an unrelated substrait issue. I think the intention is to link to https://github.com/apache/incubator-gluten/pull/574. Is that correct?

extensions/functions_table_generic.yaml
benbellick, line 1: Throughout the YAML file here, you are using `T`, etc. as type parameters. However, other examples / some documentation seem to indicate that we should only be using `any1` etcs. I have an issue open to discuss this here: https://github.com/substrait-io/substrait/issues/903

extensions/functions_table_generic.yaml
```diff
@@ -0,0 +1,109 @@
+%YAML 1.2
+---
+urn: extension:io.substrait:functions_table_generic
```
benbellick, line 3: Very minor nit, but IMO adding `_generic` on the end doesn't really provide any extra information. Might as well just keep the URN shorter. Same goes with the file name. Feel free to ignore this one though
```suggestion
urn: extension:io.substrait:functions_table
```

extensions/functions_table_generic.yaml
```diff
@@ -0,0 +1,109 @@
...
+table_functions:
+  - name: explode
```
benbellick, line 5: You have two `explodes` in the yaml file here. I don't believe that this is _technically_ incorrect, though I opened up an issue (https://github.com/substrait-io/substrait/issues/931) proposing making it the case that function names must be unique. 

Is it possible / sensible to make this one function with two `impls`? And then just have the description describe both? I think this is simpler to handle for both implementers and easier for humans to read.

extensions/functions_table_generic.yaml
```diff
@@ -0,0 +1,109 @@
...
+          - name: input2
+            value: list<T2>
+            description: Second array to unnest.
+          - name: input3
+            value: list<T3>
+            description: Third array to unnest.
+          - name: input4
+            value: list<T4>
+            description: Fourth array to unnest.
+        return: struct<T1, T2, T3, T4>
```
benbellick: This was kind of the exact difficulty I was dealing with in my version of the unnest PR. I think this is a fine solution for now, but it will be great in the future to figure out a more flexible grammar for describing this behavior for an arbitrarily large number of inputs.

tests/cases/table_generic/explode_array.test
benbellick, line 1: Considering that these are a new test format, it would be great if we could explain the format a bit in https://github.com/substrait-io/substrait/blob/main/tests/README.md?

proto/substrait/algebra.proto
```diff
@@ -1690,6 +1726,54 @@ message Expression {
   }
 }
 
+// TableExpression represents expressions that produce zero or more rows,
```
benbellick, line 1837: If `TableExpression` only makes sense in the context of `GenerateRel`, should we move this proto definition to be within the scope of `GenerateRel`?

proto/substrait/algebra.proto
```diff
@@ -1690,6 +1726,54 @@ message Expression {
...
+// as opposed to Expression which produces a single scalar value.
+// Table expressions are used in relational contexts where multi-row output
+// is expected (e.g., GenerateRel for lateral views and unnest operations).
+//
+// Since they have different cardinality than Expression expressions,
+// TableExpression expressions cannot be composed with other expressions.
+message TableExpression {
+  oneof table_expr_type {
+    TableFunction table_function = 1;
```
benbellick, line 1846: What are the other things here that you anticipate adding in the future?

benbellick (commented):
I haven't done a complete review yet but just dumping my comments so far. Hopefully will get a chance to do rest later. Thanks!

proto/substrait/algebra.proto
EpsilonPrime: Table expressions are definitely possible elsewhere.  I could see read relations being redesigned using table expressions.

proto/substrait/algebra.proto
EpsilonPrime: It's possible that table expressions could have their own operators (perhaps implementing joins, projects, or fetches).  I'm using Expression as a guide here.

extensions/functions_table_generic.yaml
EpsilonPrime: This is mirroring the aggregate functions.  I would go with table_functions over functions_table.  One could functions is also devoid of information but table by itself fails to have enough information to make it clear what it is on its own.

extensions/functions_table_generic.yaml
```diff
@@ -0,0 +1,109 @@
...
+      Produces one row per map entry.
+
+      For each input row, this function generates one output row for each key-value
+      pair in the input map. The output is a struct with two fields: 'key' and 'value'.
+      Note that map entries may be visited in any order.
+
+      Empty maps produce zero output rows. For outer semantics (producing a row
+      with NULLs when the map is empty), use GenerateRel's preserve_on_empty option.
+
+      Some systems call this unnest.
```
yongchul: This also applies to the other `explode`

extensions/functions_table_generic.yaml
```diff
@@ -0,0 +1,109 @@
...
+            description: The map to explode.
+        return: struct<any1, any2>
+
+  - name: posexplode
+    description: >-
+      Produces one row per array element with position.
+
+      For each input row, this function generates one output row for each element
+      in the input array. The output is a struct with two fields:
+      - 'pos': 0-indexed position of the element in the array (i64)
```
yongchul, line 40: silly thing. some systems do use 1-indexed. do we want to have it as a function option?

extensions/functions_table_generic.yaml
```diff
@@ -0,0 +1,109 @@
...
+          - name: input2
+            value: list<any2>
+            description: Second array to unnest.
+          - name: input3
+            value: list<any3>
+            description: Third array to unnest.
+          - name: input4
+            value: list<any4>
+            description: Fourth array to unnest.
+        return: struct<any1, any2, any3, any4>
```
yongchul, line 102: can this be expressed using inconsistent variadic arguments? or we need to extend the type grammar?

proto/substrait/algebra.proto
```diff
@@ -556,6 +556,41 @@ message ExpandRel {
...
+  // When false (default):
+  //   Input rows that produce zero output rows are excluded from the result.
+  //
+  // When true:
+  //   Input rows that produce zero output rows still generate one output row
+  //   with the input fields populated and the generator output fields set to NULL.
+  //   This is equivalent to Spark's "LATERAL VIEW OUTER" or can be achieved in
+  //   standard SQL via LEFT JOIN LATERAL.
+  //
+  // Optional; defaults to false.
```
yongchul, line 659: not: Optional`:`?

proto/substrait/algebra.proto
yongchul: I plan to send a proposal to add `ApplyRel` by next week, which will behave like a join at high-level but taking a RelOp tree on the subquery part. We could consolidate that with this `GeneratorRel` with following extensions to GenerateRel.

* TableExpression should have `RelOp subquery` in its oneof.
* `subquery` should reference the input field by scoped field reference (outer reference).
* There is join type equivalent options (cross, outer, semi, and anti-semi). These will determine the behavior of the apply as well as output schema.

@EpsilonPrime do you see we need a seperate `ApplyRel` or extend `GeneratorRel`?

proto/substrait/algebra.proto
yongchul: Based on my other comment on `ApplyRel`, this can be actually an enum of `CROSS`, `OUTER`, `SEMI`, and `ANTISEMI`, default to `CROSS` and stay in `GenerateRel`.

site/docs/expressions/table_functions.md
```diff
@@ -1,8 +1,154 @@
...
 
+## Overview
 
+Table functions are similar to scalar functions in their definition and reference mechanism, but differ fundamentally in their output:
 
-to be completed...
+- **Scalar functions:** Produce exactly one value per input row
+- **Window functions:** Produce one value per input row (with access to window of rows)
+- **Aggregate functions:** Produce one value per group of input rows
+- **Table functions:** Produce zero or more rows per input record
```
yongchul: record -> row?

site/docs/expressions/table_functions.md
```diff
@@ -1,8 +1,154 @@
...
+
+## TableFunction Message
+
+The `TableFunction` message is used to invoke table functions within relational operators (primarily `GenerateRel`).
+
+| Field | Type | Description | Required |
+|-------|------|-------------|----------|
+| function_reference | uint32 | Points to a function_anchor in the plan that references a table function in the extension YAML. 0 is a valid anchor/reference. | Required |
+| arguments | repeated FunctionArgument | Arguments bound to the table function (value, type, or enum arguments) | Required |
```
yongchul: we could have a table function requiring no arguments -- just doing fixed stuff. `Optional`?

site/docs/expressions/table_functions.md
```diff
@@ -1,8 +1,154 @@
...
+| output_type | Type | The output type of the table function (must be a Struct type where each field is a generated column) | Required |
+
+### Output Type
+
+The `output_type` field must be a non-nullable `Struct` type where each field represents a column in the generated output rows. This explicitly defines the schema of rows produced by the table function.
+
+**Examples:**
+- `explode(array<i32>)` → `Struct{value: i32}`
+- `posexplode(array<string>)` → `Struct{pos: i64, value: string}`
+- `explode(map<string, i32>)` → `Struct{key: string, value: i32}`
```
yongchul, line 32: wait.. do we have named struct? 😄

site/docs/expressions/table_functions.md
```diff
@@ -1,8 +1,154 @@
...
+
+## Differences from Other Function Types
+
+| Aspect | Scalar Function | Window Function | Aggregate Function | Table Function |
+|--------|----------------|-----------------|-------------------|----------------|
+| **Invocation context** | Expressions (ProjectRel) | Window expressions (WindowRel) | Aggregate expressions (AggregateRel) | Relational operators (GenerateRel) |
+| **Input scope** | Single row | Window of rows | Group of rows | Single row |
+| **Output cardinality** | Exactly 1 value | Exactly 1 value | Exactly 1 value | 0 or more rows |
+| **Output type** | Scalar type | Scalar type | Scalar type | Struct type |
+| **Can appear in Expression** | Yes | Yes (WindowFunction) | No | No |
```
yongchul, line 120: it could appear in the `make_array` or `make_struct` like functions if we have.. no?

site/docs/relations/logical_relations.md
```diff
@@ -559,3 +559,105 @@ The operator that defines modifications of a database schema (CREATE/DROP/ALTER
...
+- **Exploding maps:** Turn one row with map {k1:v1, k2:v2} into two rows
+- **Position-indexed explode:** Generate (position, value) pairs from arrays
+- **JSON parsing:** Extract multiple fields from JSON strings
+
+| Signature            | Value                                                                                                                                                          |
+|----------------------|----------------------------------------------------------------------------------------------------------------------------------------------------------------|
+| Inputs               | 1                                                                                                                                                              |
+| Outputs              | 1                                                                                                                                                              |
+| Property Maintenance | Distribution and orderedness are not maintained. The variable number of output rows per input breaks distribution properties, and row multiplication breaks ordering. |
```
yongchul: input distribution is preserved as long as the input fields are intact and available -- which should be the case by definition.

site/docs/relations/logical_relations.md
```diff
@@ -559,3 +559,105 @@ The operator that defines modifications of a database schema (CREATE/DROP/ALTER
...
+| Direct Output Order  | Input fields followed by generated fields. The output schema is: `[input_field_1, ..., input_field_N, generated_field_1, ..., generated_field_M]` where generated fields come from the table function's output type. Unlike ExpandRel which appends an i32 ordinal column, GenerateRel does NOT automatically add an ordinal column. Table functions that need position information (like posexplode) include it explicitly in their output_type. The RelCommon.emit field can be used to reorder columns or project out unwanted fields. |
```
yongchul, line 615: isn't output type is STRUCT? So we automatically flatten the struct in direct output order?

site/docs/relations/logical_relations.md
```diff
@@ -559,3 +559,105 @@ The operator that defines modifications of a database schema (CREATE/DROP/ALTER
...
+
```
yongchul, line 616: can the position information be part of optional behavior of generator rel? In that way, we don't need extra function and implementation can opt to pick which mode it wants... something like

```proto
enum OffsetField 
{
    NONE, // not generating offset value (default)
    ZERO_BASED, // 0-based index
    ONE_BASED // 1-based index
}
```

when the option is there, the last field is the offset.

proto/substrait/algebra.proto
EpsilonPrime: One needs preserve on empty, the other RelOp.  That feels different enough to me.  Fortunately there's enough room for ApplyRel in TableExpression.

extensions/functions_table_generic.yaml
EpsilonPrime: Could conversion from 0-indexed to 1-indexed is something that could be handled by their Substrait layer?  If so, then we always canonicalize one way.

extensions/functions_table_generic.yaml
EpsilonPrime: We need to extend the type grammar.  I left that to a future project.

site/docs/relations/logical_relations.md
EpsilonPrime: Returning a struct is possible but it will be harder to use.  We have columns we might as well use them.

site/docs/expressions/table_functions.md
benbellick: [Section on named structs here](https://substrait.io/types/named_structs/#named-structs)

[Some more info here](https://substrait.io/types/type_classes/)

My understanding is that they are actually just `STRUCTS` always, but you can use them when it is convenient to pretend something has names 🤷

nielspardon:
@EpsilonPrime — #917 looks like it was closed by the stale-bot rather than on review grounds. I'd like to help carry `GenerateRel` forward as part of the Gluten↔Substrait consolidation. Would you prefer to reopen/continue it yourself, or are you happy for me to pick it up crediting your PR? (Minor: `generate = 23` now collides with `TopNRel`, so it'd become `= 25`.)

EpsilonPrime:
> @EpsilonPrime — #917 looks like it was closed by the stale-bot rather than on review grounds. I'd like to help carry `GenerateRel` forward as part of the Gluten↔Substrait consolidation. Would you prefer to reopen/continue it yourself, or are you happy for me to pick it up crediting your PR? (Minor: `generate = 23` now collides with `TopNRel`, so it'd become `= 25`.)

I can pick this back up tomorrow but if you have time to work on it today feel free to reverse the process.  Finishibg the Gluten cleanup would be nice.

nielspardon:
Alright, let me reopen this one. I'm going through the list of Gluten modifications currently and I'm in touch internally at IBM with the Gluten team on this. Let me leave this one to you and I'll carry on with the other items.

extensions/functions_table_generic.yaml
EpsilonPrime: Good call — merged the array and map forms into a single `explode` function with two `impls` (0cb8d3c). That also drops the `explode_dup1` rename from the coverage registry.

extensions/functions_table_generic.yaml
EpsilonPrime: Done — with the two explode forms now merged into one function, the "some systems call this unnest" note lives in the shared description and applies to both (0cb8d3c).

site/docs/relations/logical_relations.md
EpsilonPrime: You're right — the input fields are preserved unchanged on every output row, so distribution is maintained; only orderedness is broken by the row multiplication. Updated the Property Maintenance text and the ExpandRel comparison table accordingly (0cb8d3c).

site/docs/expressions/table_functions.md
EpsilonPrime: Agreed — a table function can take no arguments. Marked `arguments` as Optional (it's a `repeated` field, so it may be empty) (0cb8d3c).

site/docs/expressions/table_functions.md
EpsilonPrime: Fixed — "input record" → "input row", and I applied the same wording fix to the matching GenerateRel proto comment (0cb8d3c).

site/docs/expressions/table_functions.md
EpsilonPrime: No named structs — Substrait structs are positional. The field names in those examples are illustrative only. I've added a note to the Output Type section spelling this out (356eb3c): generated columns map to the output schema by position, and names are just optional hints with no semantic meaning.

site/docs/expressions/table_functions.md
EpsilonPrime: Right — same point: any "names" here (or in make_struct-style functions) are positional hints, not part of the type. Clarified in the Output Type note added in 356eb3c.

proto/substrait/algebra.proto
EpsilonPrime: Leaving this as `Optional; defaults to false.` — it matches the existing convention used elsewhere in the file (e.g. the WindowFunction bound comments), so changing the punctuation here would make it less consistent.

nielspardon (commented):
Overall this looks like the right shape — an additive relation plus table functions declared through the normal extension mechanism rather than built-in generators. Four things to fix before merge, and three design questions worth settling while field numbers are still cheap to change.

## Blocking

**Nullability is self-contradictory.** No impl sets `nullability:`, so all seven default to MIRROR — and MIRROR makes `explode(list?<i32>)` derive `struct?<i32>`, which `table_functions.md` forbids ("must be a non-nullable `Struct` type"). A nullable collection column therefore has no legal `output_type`, and that's the common case.

**`preserve_on_empty` can't produce the NULLs it promises.** `posexplode` returns `struct<i64, any1>`, and this PR's own `posexplode.textproto` marks `pos` as `NULLABILITY_REQUIRED`. One rule fixes it: when `preserve_on_empty` is true, every field of `output_type` is nullable.

**The new kind isn't wired into existing machinery.** `tests/baseline.py` got its `num_table_functions` counter; six siblings didn't:

| Where | Missing |
|---|---|
| `text/dialect_schema.yaml` | `GENERATE`, `generate_relation`, `supported_table_functions` |
| `text/simple_extensions_schema.yaml` | `deprecated`, `metadata` — so a table function can never be deprecated *(details inline)* |
| `tests/coverage/coverage.py` | `_result_is_nullable` not extended for `MultiRowResult` — no nullable-input table test can pass |
| `tests/type/test_type_grammar.py` | `table_functions` in `iter_type_expressions` — new type strings are never grammar-checked |
| `site/docs/extensions/index.md` | two "scalar, aggregate, and window" enumerations (lines 164, 175) |
| `site/examples/proto-textformat/README.md` | `table_function/` missing from the Directories index |

I'd rather these land with the feature. The dialect schema has drifted before (`TOP_N`, `EXTENSION_*`), but that argues for not extending the drift — and the function kind has no precedent at all: nothing anywhere can express table-function support until this lands.

<details>
<summary>Concrete edits for the five that aren't in the diff</summary>

**`text/dialect_schema.yaml`**

```yaml
# supported_relations enum + the parallel oneOf list
- GENERATE
- $ref: "#/definitions/generate_relation"

# definitions, alongside expand_relation
  generate_relation:
    type: object
    additionalProperties: false
    properties:
      relation: {const: GENERATE}
      metadata: {type: object}
    required: [relation]

# alongside supported_scalar_functions
  supported_table_functions:
    type: array
    items: {$ref: "#/definitions/dialect_function"}
```

`expand_relation` carries `field_types` as its per-relation capability; the analogue here would be whether the dialect supports `preserve_on_empty`, which engines genuinely differ on. One open question, not a request: does `supported_expressions` need a table-expression member? Probably not if `supported_table_functions` exists — but if `TableExpression` keeps its wrapper shape, a dialect can't say it supports the container but not a given function.

**`tests/coverage/coverage.py`** — `TestCase.result` gained `MultiRowResult`, but `_result_is_nullable` still only handles `CaseLiteral`, so it returns False for every table result regardless of what the file declares. That makes any nullable input unsatisfiable — both spellings fail identically:

```
explode([1, 2]::list?<i32>) = [(1), (2)]::struct?<i32>
→ MIRROR violation: output should be nullable because an input is nullable
```

The grammar accepts `structType`'s `isnull` and `visitMultiRowResult` keeps the `?` — it's just dropped here, since `MultiRowResult` has no `nullable`/`is_nullable()`. DECLARED_OUTPUT and DISCRETE have the same gap. Once the YAML sets nullability explicitly, this is what stands between us and a test for the nullable case.

**`tests/type/test_type_grammar.py`** — one-line fix: `iter_type_expressions` walks `scalar_functions`/`aggregate_functions`/`window_functions` but not `table_functions`. The JSON schema can't cover it either (`$defs/type` is `oneOf: [string, object]` with no pattern, so any string validates). Latent rather than live, since the current strings do parse — but `return: struct<value: T>`, the form the new docs recommend, would have been caught immediately.

**`site/docs/extensions/index.md`** — line 14 has listed "Table Functions" as a simple-extension kind for a while, but 164 still says "(scalar, aggregate, and window functions)" and 175 still scopes `deprecated` to "(scalar, aggregate, and window)". This is the PR that makes the kind real; 175 can only become true once the schema def gains `deprecated`.

**`site/examples/proto-textformat/README.md`** — the new `table_function/` directory isn't in the Directories index, which is also the only place a reader learns the one-message-class-per-directory convention.
</details>

**The test surface should arrive complete.** `explode(map)` ships with no test, and the new format can't express either an expected error or a nested value in a result row — both of which these functions specifically need. Details inline.

## Design questions

**Relationship to `LateralJoinRel`?** It's already on `main`, evaluates its right input once per left row, and its `JoinType` INNER/LEFT is exactly `preserve_on_empty` — but the new docs compare only against `ExpandRel`, twenty lines below `LateralJoinRel`'s own section. A producer facing `LATERAL VIEW OUTER EXPLODE(tags)` now has two encodings and no rule for choosing. A paragraph saying which to emit would settle it. Two knock-ons: a bool can't grow the way the enum can (LEFT_SEMI/ANTI/SINGLE/MARK all have Generate analogues), and `LateralJoinRel` defines a `rel_anchor`/`OuterReference` scope where `GenerateRel` says nothing about the generator's reference scope.

**`TableExpression` shape.** Single-member oneof, one reference in the tree, and the non-composability it exists to express is prose rather than enforced. `AggregateFunction` is the precedent for the same situation — a call that can't appear inside `Expression`, modelled as a bare top-level message. As written, every SDK adds a whole expression-kind axis to reach one leaf, and `TableExpression{}` with no generator is schema-valid. Hoisting `TableFunction`, or putting the oneof on `GenerateRel` directly (house style: `ReadRel.read_type`, `WriteRel`, `DdlRel`), avoids both.

**Variadic return arity.** `returnValue` can't express a struct whose arity tracks the argument count, so `unnest` enumerates impls and stops at four. Real generators exceed that — Spark `stack(n, v1..vk)` yields `ceil(k/n)` columns unbounded. Fine to defer the general fix, but the docs should say the arity list is an enumeration, not a model limit.

<details>
<summary><b>Gluten migration check</b> — nothing inexpressible; items 1 and 2 are what block it</summary>

Gluten ships a vendored `GenerateRel`, so I checked what a migration would need:

| Gluten | This PR |
|---|---|
| `repeated Expression child_output` | `RelCommon.emit` — it's only ever field selections (producer emits `makeSelection(childIndex)`; the Velox consumer asserts field-only), so `emit` covers it and the field shouldn't be carried over |
| `bool outer` | `preserve_on_empty` |
| `GenerateParameters:` string hints in `advanced_extension` | derivable from the function anchor — also retires a positional convention where the consumer assumes the generator input is the child projection's last column |

So the migration is a simplification. The two blockers: Spark array columns are nullable (item 1), and `LATERAL VIEW OUTER POSEXPLODE` needs a NULL position (item 2).
</details>

Worth a short version of that table in the PR body — AGENTS.md asks for the per-library note on `proto/` changes.

extensions/functions_table_generic.yaml
```diff
@@ -0,0 +1,102 @@
...
+      row with NULLs when the collection is empty), use GenerateRel's
+      preserve_on_empty option.
+
+      Some systems call this unnest.
+    impls:
+      - args:
+          - name: input
+            value: list<any1>
+            description: The array to explode.
+        return: struct<any1>
```
nielspardon, line 27: No impl here sets `nullability:`, so all seven default to MIRROR — which makes `explode(list?<i32>)` derive `struct?<i32>`, a nullable struct that `table_functions.md` forbids. So a nullable collection column has no legal `output_type`.

```yaml
        return: struct<any1>
        nullability: DECLARED_OUTPUT
```

Separately: the descriptions say what an *empty* collection does, never a NULL one. Stating that NULL behaves as empty would match what engines do and let `preserve_on_empty` cover it.

extensions/functions_table_generic.yaml
```diff
@@ -0,0 +1,102 @@
...
+        return: struct<i64, any1>
+
+  - name: unnest
+    description: >-
+      Expands one or more lists into rows by zipping them in parallel (PostgreSQL semantics).
+
+      When multiple lists are provided, produces rows by expanding the lists in parallel,
+      similar to a zip operation. For example, unnest([1,2], [3,4]) produces rows: (1,3), (2,4).
+
+      If lists have different lengths, shorter lists are padded with NULL values to match
```
nielspardon, line 59: "shorter lists are padded with NULL values" can't be satisfied by `struct<any1, any2>` under MIRROR over non-nullable lists — the padded position can't hold NULL. Same fix as above.

Two notes: no test exercises unequal lengths (every case zips equal-length lists), and it's worth saying the four arities are an enumeration rather than a model limit, since `returnValue` can't yet express argument-dependent arity.

proto/substrait/algebra.proto
```diff
@@ -627,6 +627,41 @@ enum FetchMode {
...
+  bool preserve_on_empty = 4;
```
nielspardon, line 660: This sets the generated fields to NULL, but nothing promotes them to nullable — and `posexplode.textproto` in this PR declares `pos` as `NULLABILITY_REQUIRED`, so `preserve_on_empty: true` writes NULL into a field whose type forbids it. Under MIRROR over non-nullable lists that's the derived default, not an edge case.

Could we state the rule here and in Empty Collection Handling — when `preserve_on_empty` is true, every field of `output_type` is nullable? Without it, two consumers derive different schemas for the same plan.

text/simple_extensions_schema.yaml
```diff
@@ -376,3 +380,37 @@ $defs:
             window_type:
               type: string
               enum: [STREAMING, PARTITION]
+
+  tableFunction:
```
nielspardon, line 384: This def omits `deprecated` (function *and* impl level), `metadata`, and impl-level `description`, all of which the other three function defs carry. With `additionalProperties: false` those are hard failures, so a table function can never be deprecated — awkward with the `functions_aggregate_decimal_output` deprecation in flight. The missing impl `description` also makes `generate_function_docs.py`'s per-impl rendering unreachable here.

Since all four defs share an identical impl block, a shared `#/$defs/functionImpl` would put the next added property in one place instead of four.

grammar/FuncTestCaseParser.g4
```diff
@@ -78,6 +103,18 @@ aggFuncTestCase
     : aggFuncCall ( OBracket funcOptions CBracket )? Eq result
     ;
 
+tableFuncTestCase
+    : functionName=identifier OParen arguments CParen ( OBracket funcOptions CBracket )? Eq multiRowResult
+    ;
+
+multiRowResult
```
nielspardon, line 110: Two gaps versus the other two formats, both cheaper to close now than once files depend on the shape:

- **No expected errors.** `multiRowResult` has no `substraitError` alternative, so `is_return_type_error()` is permanently False here. That's exactly where these functions are least specified — `explode` on NULL, `posexplode` position overflow, `unnest` on unequal lengths.
- **No nested result values.** `rowTuple` uses `literal` rather than `compoundLiteral`, so `explode([[1,2],[3,4]]::list<list<i32>>) = [([1, 2]), ([3, 4])]::struct<list<i32>>` doesn't parse — the nested-flattening case the feature exists for. `literalStruct`/`literalMap` already use `compoundLiteral`.

`rowTuple` also duplicates `columnValues`; `multiRowResult : literalList DoubleColon structType ;` reuses both and lets `MultiRowResult` collapse into `CaseLiteral`.

tests/baseline.json
```diff
@@ -1,16 +1,17 @@
...
+    "num_table_functions": 3,
+    "num_function_overloads": 509
   },
   "coverage": {
-    "total_test_count": 1177,
-    "num_function_variants": 498,
-    "num_covered_function_variants": 245
+    "total_test_count": 1257,
+    "num_function_variants": 509,
+    "num_covered_function_variants": 255
```
nielspardon, line 15: `explode(map<any1, any2>)` is the one uncovered variant of the seven. It slips through because the accepted gap is rewritten here (253 → 254) and `validate_against` compares with `>`, so an equal gap passes — the exact case its own message targets ("New functions should be added along with test cases that illustrate their behavior").

It's also the overload whose iteration order the YAML declares unspecified, so the most worth pinning down. I know the baseline on `main` was already stale — which is why the deltas look larger than they are — but I'd rather a legacy gap didn't absorb a new one.

tests/coverage/extensions.py
```diff
@@ -81,6 +81,9 @@ def get_short_type(long_type):
         short_type = type_to_short_type.get(long_type, None)
 
         if short_type is None:
+            # For struct types, preserve the full parameterized type
+            if long_type.startswith("struct<"):
```
nielspardon, line 85: I think both hunks can come out. `struct` is already in `type_to_short_type` (`test_build_type_to_short_type` pins it), so the existing param-stripping path normalizes `struct<i64, any1>` → `struct` and the original `is_same_type` matched it:

| Reverted | Result |
|---|---|
| `struct<` early return only | green |
| `is_same_type` widening only | 18 failures |
| **both** | green, byte-identical `baseline.json` |

So the pair adds no capability — but the widening isn't table-scoped, it drops parameters for all 509 overloads. `is_same_type('dec<38,0>', 'dec<10,2>')` and `('list<i32>', 'list<str>')` were False before and are True now; unreachable today only because struct is the sole type that keeps its parameters.

Live cost: `posexplode(...) = [(1, 2, 3), ('a', 'b', 'c')]::struct<i64,i32>` — wrong arity *and* element types — is accepted and counted as coverage of the `struct<i64, any1>` variant.

tests/coverage/visitor.py
```diff
@@ -546,6 +614,22 @@ def visitSubstraitError(self, ctx: FuncTestCaseParser.SubstraitErrorContext):
             return SubstraitError("undefined")
         return SubstraitError("unknown_error")
 
+    def visitMultiRowResult(self, ctx: FuncTestCaseParser.MultiRowResultContext):
```
nielspardon, line 617: `rows` and `type` are stored independently and never cross-checked, so `[(1, 2, 3)]::struct<i32>`, `[()]::struct<i32>`, and rows of differing widths in one result all validate clean. `visitRowTuple` also drops each literal's type (`value, _ = self.visitLiteral(literal)`), so the rows are untyped strings. Combined with the `is_same_type` change, the format currently asserts only function name, argument base types and argument count.

Also unguarded: `[(P1Y2M3DT4H5M6S)]::struct<icompound>` parses, then raises `TypeError: cannot unpack non-iterable NoneType object` — `visitLiteral` has no `IntervalCompoundLiteral` branch. Since `load_all_testcases` walks the whole tree, one such file aborts the run with no file or line attribution.

site/docs/expressions/table_functions.md
```diff
@@ -1,8 +1,156 @@
...
+
+Example function registration (conceptual):
+```yaml
+table_functions:
+  - name: explode
+    impls:
+      - args:
+          - name: input
+            value: list<T>
+        return: struct<value: T>
```
nielspardon, line 138: Five things here will send implementers the wrong way:

- `return: struct<value: T>` — named fields are `nstruct<>`, parameters are `any1` not `T`; this fails the repo's own type grammar, and contradicts line 34 ("structs are positional, not named ... illustrative only").
- The snippet is inline, so neither `check-jsonschema` nor the type-grammar test sees it — an `--8<--` include of the real YAML can't drift.
- "Must be marked as table-valued in function metadata" — no such marker; the `table_functions:` key is the marker, and `metadata` is in fact rejected by the new def.
- `array<T>` here and in the relation docs — the type class is `list<T>` (`array<i32>` doesn't parse).
- The `unnest` row calls it a synonym for `explode` over one array; the YAML defines parallel zip over up to four lists with NULL padding. Implementing from the docs would mis-execute the multi-argument cases this PR ships tests for.

tests/README.md
```diff
@@ -339,3 +372,22 @@ In this example:
 - `add(2::i32, 3::i32)` is a nested function call used as an argument
 - The result side can also be a function call, not just a literal
 - Both sides of `=` are evaluated; the test asserts they produce the same value
+
+### Example of a test file for table functions
+
+```code
+### SUBSTRAIT_TABLE_TEST: v1.0
+### SUBSTRAIT_INCLUDE: '/extensions/functions_table_generic.yaml'
```
nielspardon, line 380: Looks like a rebase leftover — this example uses the pre-URN include form and doesn't parse: `mismatched input ''/extensions/functions_table_generic.yaml'' expecting ExtensionUrn`. Every other include here uses the URN form, and the three real `.test` files get it right. `### SUBSTRAIT_TABLE_TEST: V1` above it is also invalid — `FormatVersion` is lowercase.

tests/test_proto_example_validator.py
```diff
@@ -64,3 +64,18 @@ def test_validate_field_references():
...
+    examples_dir = Path("site/examples/proto-textformat/table_function")
+    # Examples are named after the message they represent: the GenerateRel
+    # wrapper or the bare TableExpression.TableFunction generator.
+    message_by_file = {
+        "generate_rel_explode.textproto": algebra_pb2.GenerateRel,
+        "posexplode.textproto": algebra_pb2.TableExpression.TableFunction,
+        "simple_explode.textproto": algebra_pb2.TableExpression.TableFunction,
+    }
+    for textproto_file in examples_dir.glob("*.textproto"):
+        message_class = message_by_file[textproto_file.name]
```
nielspardon, line 80: The filename→message dict raises `KeyError` for any new `.textproto` in `table_function/`, and silently does less work if one is renamed. Root cause is that this directory mixes `GenerateRel` and `TableExpression.TableFunction` where the siblings hold one class each — splitting into `generate_rel/` and `table_function/` restores that and lets all four tests collapse into one parametrized case.

grammar/FuncTestCaseParser.g4
```diff
@@ -6,15 +6,33 @@ options {
 }
 
 doc
```
nielspardon, line 8: Curiosity, not a request: was splitting `doc`/`header`/`testGroup` per-kind deliberate? Extending the existing token alternation would have been a much smaller diff — but the split also makes the corpus parse noticeably faster, presumably because the per-kind `doc` resolves on two tokens instead of ALL(*) lookahead at every group boundary. If that's why, worth a comment in the grammar, since the obvious future refactor would undo it.

PR #922, opened 2025-12-18
feat: add optional metadata to extension yaml files

### LINKED ISSUE #921: feat: allow arbitrary extension metadata to be added to extension files
One of substrait's virtues its its extensibility. To enable extension files to also be extensible, lets add a field that allows embedding arbitrary data. For example: 
```yaml
urn: extension:io.substrait:functions_comparison
scalar_functions:
  -
    name: "not_equal"
    description: >
      Whether two values are not_equal.

      `not_equal(x, y) := (x != y)`

      If either/both of `x` and `y` are `null`, `null` is returned.
    impls:
      - args:
          - value: any1
            name: x
          - value: any1
            name: y
        return: boolean
    user_metadata: # new addition to schema
      some_field: some_data
      another_field: 12344
```

A similar field would exist across all of the items in the schema.

benbellick:
Introduces extension mechanism in YAML files to introduce an arbitrary object as metadata. 

Closes #921

### DIFF
--- a/site/examples/extensions/metadata_example.yaml
+++ b/site/examples/extensions/metadata_example.yaml
@@ -3,6 +3,13 @@ urn: extension:io.substrait:functions_comparison
 metadata:
   version: 2.0
   maintainer: example-team
+types:
+  - name: point
+    structure:
+      latitude: i32
+      longitude: i32
+    metadata:
+      coordinate_system: "WGS84"
 scalar_functions:
   - name: "not_equal"
     impls:
--- a/site/examples/extensions/metadata_example.yaml
+++ b/site/examples/extensions/metadata_example.yaml
@@ -1,5 +1,5 @@
 # Example showing the metadata field at multiple levels
-urn: extension:io.substrait:functions_comparison
+urn: extension:io.substrait:metadata_examples
 metadata:
   version: 2.0
   maintainer: example-team
--- a/site/examples/extensions/metadata_example.yaml
+++ b/site/examples/extensions/metadata_example.yaml
@@ -1,0 +1,17 @@
+# Example showing the metadata field at multiple levels
+urn: extension:io.substrait:functions_comparison
+metadata:
+  version: 2.0
+  maintainer: example-team
+scalar_functions:
+  - name: "not_equal"
+    impls:
+      - args:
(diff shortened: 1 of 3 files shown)


site/examples/extensions/metadata_example.yaml
```diff
@@ -0,0 +1,17 @@
...
+    impls:
+      - args:
+          - value: any1
+            name: x
+          - value: any1
+            name: y
+        return: boolean
+    metadata:
+      performance_hint: "vectorized"
+      cost_estimate: 1
```
vbarua, line 24: minor: could you add examples for types, aggregate function and window functions for completeness.

text/simple_extensions_schema.yaml
```diff
@@ -221,6 +225,8 @@ $defs:
               $ref: "#/$defs/returnValue"
             implementation:
               $ref: "#/$defs/implementation"
+      metadata: # arbitrary data created by extension author
```
vbarua: minor: for readability, could you move the `metadata` definition to be just under `description`, to make it clearer that it's attached to the scalar function object, and do the same for the other function types.

vbarua (approved):
A request for a few more examples and moving some of the schema fields around for readability, but overall looks reasonable for me.

vbarua:
I thought there was prior art for this in the dialect files, but I'd forgotten that I'd left the addition of arbitrary metadata as a follow-up: https://github.com/substrait-io/substrait/pull/816#discussion_r2363787250

site/examples/extensions/metadata_example.yaml
benbellick: Happy to add for types. IMO it seems excessive to add it for all if we are going to embed it inside of the docs themselves. There is really nothing interesting or fundamentally different about window functions or aggregate functions in this case. So I think in that case it just becomes noise 🤷

yongchul (approved):

vbarua (approved):
Looks good to me. I'll leave this open until the end-of-day in case any of the other PMCs want to weigh-in, but we've already got the 2 votes we need to merge.

EpsilonPrime (approved):

PR #923, opened 2025-12-19
fix: function signatures should have correct nullabilitiy

Broshen:
Some function signatures in `functions_comparison.yaml` describe that their response can be null, but the type in `return` doesn't have a `?`, indicating that it is non-nullable. This PR fixes the function signature to match the description that responses can be NULL.


extensions/functions_comparison.yaml
```diff
@@ -16,7 +16,7 @@ scalar_functions:
             name: x
           - value: any1
             name: y
-        return: boolean
+        return: boolean?
```
vbarua, line 20: So the `?` at the end here doesn't mean anything unless you set the [Nullability Handling](https://substrait.io/expressions/scalar_functions/#nullability-handling) to DECLARED_OUTPUT or DISCRETE. MIRROR, the default, ignores it because the output nullability is determined by the input type nullabilities.

tokoko: @Broshen the default mode (MIRROR) implies that output is nullable if any of the inputs is nullable which is the intention here I think (?).

Setting it to `DECLARED_OUTPUR` + `?` is not correct because function output will have to be always treated as nullable, even if inputs are known not to contain nulls.

Broshen: does that mean the correct way is to have arguments be declared as e.g.
```
      - args:
          - value: any1?
            name: x
          - value: any1?
            name: y
```
to declare that the arguments can be nullable? does **that** syntax mean anything? or is that meaningless too, bc... we don't know whether arguments would contain nulls or not until function call time?

if `?` syntax in `args` is valid... does that mean we'd need an overload with nullable inputs + an overload with non-nullable inputs?
e.g.
```
    impls:
      - args:
          - value: any1?
            name: x
          ...
       return ... // nullable overload
      - args:
          - value: any1
            name: x
          ...
       return ... // non-nullable overload
```

vbarua: > to declare that the arguments can be nullable? does that syntax mean anything?

Argument nullability is ignored EXCEPT for when the nullability handling is set to [DISCRETE](https://substrait.io/expressions/scalar_functions/#nullability-handling), which is the only mode that restricts that allowed argument nullabilities.

vbarua:
Closing this as this isn't something to be fixed in the spec. There is room here for clarifying when nullability declarations are redundant, but that should be it's own change without extension changes.

extensions/functions_comparison.yaml
vbarua: To add, for the specific example of `not_equals` here, if both inputs are non-nullable, the output will never be nullable. If either input is nullable, then the output _may_ be nullable. This is the defintion of MIRROR nullability.

PR #927, opened 2026-01-05
feat: expanded timestamp arithmetic precision

### LINKED ISSUE #926: Timestamp and interval precision synchronization
## Problem

The current `add` function signature in [`functions_datetime.yaml`](https://github.com/substrait-io/substrait/blob/main/extensions/functions_datetime.yaml) requires the precision parameter `P` to be synchronized between `precision_timestamp` and `interval_day`:

```yaml
- args:
    - name: x
      value: precision_timestamp<P>
    - name: y
      value: interval_day<P>
  return: precision_timestamp<P>
```

This creates a constraint where `precision_timestamp<9>` can only be combined with `interval_day<9>`, not `interval_day<0>` or `interval_day<6>`. As expected, it manifests in `substrait-go` as overload resolution failure when generating plans:

```sql
SELECT '2023-05-09 10:00:00'::timestamp + INTERVAL '1 hour'
```
```
error building substrait function add:pts_iday: invalid type
```
## SQL Standard Behavior

**Trino** (following SQL standards) docs ([issue #1284](https://github.com/trinodb/trino/issues/1284)):

> "The result of any operation on two timestamps will result with a timestamp that is of **higher precision**, with the precision decimals of the lower precision timestamp assumed to be 0 if the digits do not exist."

**Postgres** ([docs](https://www.postgresql.org/docs/current/datatype-datetime.html)):
- `timestamp(P)` and `interval(Q)` specify precision independently (0-6 digits)
- No documented requirement for matching precisions in arithmetic

**DuckDB** ([docs](https://duckdb.org/docs/stable/sql/data_types/interval)):
- Intervals stored as independent basis units with no precision synchronization requirement

So I think the standard behavior allows independent precisions. Arithmetic operations promote to the higher precision, zero-padding the lower-precision operand.

## Potential Solution

Based on the SQL standard, I think using the **maximum precision** from both operands would make sense:

```yaml
precision_timestamp<P> + interval_day<Q> -> precision_timestamp<max(P,Q)>
```
- `timestamp<9>` + `interval_day<3>` → `timestamp<9>` (interval milliseconds zero-padded to nanoseconds)

This preserves all information from both operands without requiring explicit casts. However, I'm not sure if the Substrait type system can actually express `max(P,Q)` in return types. Maybe another approach would be better? I'm also not familiar with the history of the current signature. Is it a limitation of what the type system can express or is it a semantic reason? Would love to hear suggestions or alternatives

benbellick:
Closes #926

### DIFF
--- a/extensions/functions_datetime.yaml
+++ b/extensions/functions_datetime.yaml
@@ -289,8 +289,8 @@ scalar_functions:
           - name: x
             value: precision_timestamp<P>
           - name: y
-            value: interval_day<P>
-        return: precision_timestamp<P>
+            value: interval_day<Q>
+        return: precision_timestamp<max(P, Q)>
       - args:
           - name: x
             value: timestamp_tz
@@ -301,8 +301,8 @@ scalar_functions:
           - name: x
             value: precision_timestamp_tz<P>
           - name: y
-            value: interval_day<P>
-        return: precision_timestamp_tz<P>
+            value: interval_day<Q>
+        return: precision_timestamp_tz<max(P, Q)>
       - args:
           - name: x
             value: date
@@ -369,8 +369,8 @@ scalar_functions:
           - name: x
             value: interval_day<P>
           - name: y
-            value: interval_day<P>
-        return: interval_day<P>
+            value: interval_day<Q>
+        return: interval_day<max(P, Q)>
       - args:
           - name: x
             value: interval_year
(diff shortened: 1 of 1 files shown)


benbellick:
Note that these changes are backwards compatible. All previously valid invocations of these functions have identical output types. All that has changed is that we are allowing _additional_ combinations of input types to be considered valid.

vbarua (commented):
I'm not the biggest fan of these changes. See https://github.com/substrait-io/substrait/issues/926#issuecomment-3740922905 for details.

benbellick:
@vbarua thanks for the detailed write up. I agree and am closing this PR.

PR #929, opened 2026-01-07
feat: default value for parameterConsistency

### LINKED ISSUE #928: clarify default behavior of `parameterConsistency` or require parameter
### Problem
The JSON schema defines these parameters but doesn't mark them as required:

https://github.com/substrait-io/substrait/blob/92d2e757a330f9c973bb566817dc92afd1badcb2/text/simple_extensions_schema.yaml#L158-L168

[The docs](https://substrait.io/expressions/scalar_functions/#type-parameter-resolution-in-variadic-functions) define the meaning of the parameters when set, but doesn't specify the behavior when the consistency parameter _isn't_ set. 

### Fixes
One fix is to just require the parameter. Requiring the parameter could in theory break backwards compatibility. However, the breaking of backwards-compat is a minor issue because implementers will just throw an error at YAML-load time. This approach has the benefit of being explicit

The other option is to just define `CONSISTENT` as the default behavior when nothing is set.

benbellick:
closes #928

- - -
This change is [Reviewable](https://reviewable.io/reviews/substrait-io/substrait/929)

### DIFF
--- a/extensions/functions_comparison.yaml
+++ b/extensions/functions_comparison.yaml
@@ -266,6 +266,7 @@ scalar_functions:
           - value: any1
         variadic:
           min: 2
+          parameterConsistency: CONSISTENT
         return: any1
   -
     name: "least"
@@ -277,6 +278,7 @@ scalar_functions:
           - value: any1
         variadic:
           min: 2
+          parameterConsistency: CONSISTENT
         return: any1
         nullability: MIRROR
   -
@@ -289,6 +291,7 @@ scalar_functions:
           - value: any1
         variadic:
           min: 2
+          parameterConsistency: CONSISTENT
         return: any1
         # NOTE: The return type nullability as described above cannot be expressed currently
         # See https://github.com/substrait-io/substrait/issues/601
@@ -304,6 +307,7 @@ scalar_functions:
           - value: any1
         variadic:
           min: 2
+          parameterConsistency: CONSISTENT
         return: any1
         nullability: MIRROR
   -
@@ -316,6 +320,7 @@ scalar_functions:
           - value: any1
         variadic:
           min: 2
+          parameterConsistency: CONSISTENT
         return: any1
         # NOTE: The return type nullability as described above cannot be expressed currently
         # See https://github.com/substrait-io/substrait/issues/601
--- a/extensions/functions_comparison.yaml
+++ b/extensions/functions_comparison.yaml
@@ -266,7 +266,6 @@ scalar_functions:
           - value: any1
         variadic:
           min: 2
-          parameterConsistency: CONSISTENT
         return: any1
   -
     name: "least"
@@ -278,7 +277,6 @@ scalar_functions:
           - value: any1
         variadic:
           min: 2
-          parameterConsistency: CONSISTENT
         return: any1
         nullability: MIRROR
   -
@@ -291,7 +289,6 @@ scalar_functions:
           - value: any1
         variadic:
           min: 2
-          parameterConsistency: CONSISTENT
         return: any1
         # NOTE: The return type nullability as described above cannot be expressed currently
         # See https://github.com/substrait-io/substrait/issues/601
@@ -307,7 +304,6 @@ scalar_functions:
           - value: any1
         variadic:
           min: 2
-          parameterConsistency: CONSISTENT
         return: any1
         nullability: MIRROR
   -
@@ -320,7 +316,6 @@ scalar_functions:
           - value: any1
         variadic:
           min: 2
-          parameterConsistency: CONSISTENT
         return: any1
         # NOTE: The return type nullability as described above cannot be expressed currently
         # See https://github.com/substrait-io/substrait/issues/601
--- a/site/docs/expressions/scalar_functions.md
+++ b/site/docs/expressions/scalar_functions.md
@@ -84,7 +84,10 @@ When the same parameter name is used multiple times in a function definition, th
 
 #### Type Parameter Resolution in Variadic Functions
 
-When the last argument of a function is variadic and declares a type parameter e.g. `fn(A, B, C...)`, the C parameter can be marked as either consistent or inconsistent. If marked as consistent, the function can only be bound to arguments where all the C types are the same concrete type. If marked as inconsistent, each unique C can be bound to a different type within the constraints of what T allows.
+When the last argument of a function is variadic and declares a type parameter e.g. `fn(A, B, C...)`, the C parameter can be marked as either consistent or inconsistent. If marked as consistent, the function can only be bound to arguments where all the C types are the same concrete type. If marked as inconsistent, each unique C can be bound to a different type within the constraints of what T allows. The default behavior is `CONSISTENT`.
+
+!!! note
+    The `parameterConsistency` field only has an effect when the variadic argument contains a type parameter (like `any1`, `T`, etc.). For concrete types (like `boolean?`, `i32`, `string`), it has no effect. When the variadic argument does contain a type parameter, it is recommended that extension authors explicitly specify `parameterConsistency` to make the intended behavior clear.
 
  
 
--- a/site/docs/expressions/scalar_functions.md
+++ b/site/docs/expressions/scalar_functions.md
@@ -84,19 +84,21 @@ When the same parameter name is used multiple times in a function definition, th
 
 #### Type Parameter Resolution in Variadic Functions
 
-When the last argument of a function is variadic and declares a type parameter e.g. `fn(A, B, C...)`, the C parameter can be marked as either consistent or inconsistent. If marked as consistent, the function can only be bound to arguments where all the C types are the same concrete type. If marked as inconsistent, each unique C can be bound to a different type within the type parameter's constraints. The default behavior is `CONSISTENT`.
+Two types have the same *base type* if they are identical modulo their outermost nullability. For example, `i32` and `i32?` have the same base type, but `list<i32>` and `list<i32?>` do not.
+
+When the last argument of a function is variadic and declares a type parameter e.g. `fn(A, B, C...)`, the `C` parameter can be marked as either consistent or inconsistent. If marked as consistent, the function can only be bound to arguments where all the `C` types have the same base type. If marked as inconsistent, each unique `C` can be bound to a different type within the type parameter's constraints. Nullability follows the [nullability handling](#nullability-handling) option. The default behavior is `CONSISTENT`.
 
 ##### Examples
 
 ```yaml
 --8<-- "examples/extensions/variadic_consistent_example.yaml"
 ```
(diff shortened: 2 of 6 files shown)


site/docs/expressions/scalar_functions.md
```diff
@@ -84,7 +84,10 @@ When the same parameter name is used multiple times in a function definition, th
...
+When the last argument of a function is variadic and declares a type parameter e.g. `fn(A, B, C...)`, the C parameter can be marked as either consistent or inconsistent. If marked as consistent, the function can only be bound to arguments where all the C types are the same concrete type. If marked as inconsistent, each unique C can be bound to a different type within the constraints of what T allows. The default behavior is `CONSISTENT`.
+
+!!! note
+    The `parameterConsistency` field only has an effect when the variadic argument contains a type parameter (like `any1`, `T`, etc.). For concrete types (like `boolean?`, `i32`, `string`), it has no effect. When the variadic argument does contain a type parameter, it is recommended that extension authors explicitly specify `parameterConsistency` to make the intended behavior clear.
```
vbarua: I think now would be a good time to add examples for this, to make sure everyone is on the same page about what CONSISTENT and INCONSISTENT mean.

extensions/functions_comparison.yaml
```diff
@@ -266,6 +266,7 @@ scalar_functions:
           - value: any1
         variadic:
           min: 2
+          parameterConsistency: CONSISTENT
```
benbellick: I added this to the docs in the note part of `scalar_functions.md`, but I believe that consistent vs inconsistent doesn't matter here. This is because `any1` in the return type pins the argument.

benbellick: So I actually ended up removing this line. It falsely communicates to the user that `parameterConsistency` set to `CONSISTENT` is an important attribute here. In actuality, the `any1` value is pinned by the return type, so all of them MUST be fixed, regardless of the `parameterConsistency` setting. 

As part of this PR, I wanted to try and come up with some useful and standard scalar functions to add to the core extension set that demonstrate the use of both values of `parameterConsistency`, but I couldn't think of any or find any off the internet. [This was introduced in 2021.](https://github.com/substrait-io/substrait/commit/4e5866afe264cf991655ef3230b95062d39674be) Perhaps we should discuss if this feature is worth supporting?

benbellick: Okay actually, I found [postgresql's JSON_BUILD_OBJECT](https://neon.com/docs/functions/json_build_object), which is an example of a function where variadic INCONSISTENT behavior is desirable.

site/docs/expressions/scalar_functions.md
```diff
@@ -84,7 +84,32 @@ When the same parameter name is used multiple times in a function definition, th
 
 #### Type Parameter Resolution in Variadic Functions
 
-When the last argument of a function is variadic and declares a type parameter e.g. `fn(A, B, C...)`, the C parameter can be marked as either consistent or inconsistent. If marked as consistent, the function can only be bound to arguments where all the C types are the same concrete type. If marked as inconsistent, each unique C can be bound to a different type within the constraints of what T allows.
+When the last argument of a function is variadic and declares a type parameter e.g. `fn(A, B, C...)`, the C parameter can be marked as either consistent or inconsistent. If marked as consistent, the function can only be bound to arguments where all the C types are the same concrete type. If marked as inconsistent, each unique C can be bound to a different type within the type parameter's constraints. The default behavior is `CONSISTENT`.
```
yongchul: since in  #949 we are talking about nullability, maybe we should change some wordings in this PR as well to be consistent especially `concrete type` was including nullability in the description of `DISCRETE`. We can say that the consistency is matter of `base type` (not sure the term we want to use here) and nullability follows `nullability handling` option -- and link to the section. what do you think?

yongchul: nit: use `code` style to refer type parameter `C`?

benbellick: I opted to define `base type` right at the beginning of this section.

yongchul (approved):

nielspardon:
reopening

site/docs/expressions/scalar_functions.md
```diff
@@ -51,114 +51,67 @@ There are three main types of arguments: value arguments, type arguments, and en
...
+#### Type Parameter Resolution in Variadic Functions
 
-#### `any` Type Binding
+Two types have the same *base type* if they are identical modulo their outermost nullability. For example, `i32` and `i32?` have the same base type, but `list<i32>` and `list<i32?>` do not.
 
-The following examples show how `any` type parameters bind across the different nullability modes. Note that only the outermost nullability is stripped for binding (and only under `MIRROR` or `DECLARED_OUTPUT`); nested nullability within compound types (e.g. `list<i32?>`) is preserved and must match structurally.
+When the last argument of a function is variadic and declares a type parameter e.g. `fn(A, B, C...)`, the `C` parameter can be marked as either consistent or inconsistent. If marked as consistent, the function can only be bound to arguments where all the `C` types have the same base type. If marked as inconsistent, each unique `C` can be bound to a different type within the type parameter's constraints. Nullability follows the [nullability handling](#nullability-handling) option. The default behavior is `CONSISTENT`.
```
nielspardon, line 89: > The default behavior is `CONSISTENT`.

We probably should set a `default: CONSISTENT` in the extension schema mirroring this spec clarification:
https://github.com/substrait-io/substrait/blob/de9e6328f4f94abc0809977f57dfde7262a7258b/text/simple_extensions_schema.yaml#L176-L178

nielspardon: Should we also cleanup function definitions that clearly should be marked `INCONSISTENT` like `concat:vchar`? We then would also need a way in the return type expression language to operate over the list of type parameters in order to allow something like `max(L1)` or `sum(L1)` for a `varchar<L1>` argument.

site/examples/extensions/variadic_consistent_example.yaml
```diff
@@ -0,0 +1,12 @@
+urn: extension:example:variadic_consistent
+scalar_functions:
+- name: all_equal
+  description: Returns true if all arguments are equal. All arguments must be the same type.
+  impls:
+  - args:
+    - name: first
+      value: any
```
nielspardon, line 8: Would this need to be `any1` due to:

> The any type indicates that the argument can take any possible type. In the foo function above, arguments a and b can be of any type, even different ones in the same function invocation.

https://github.com/substrait-io/substrait/blob/de9e6328f4f94abc0809977f57dfde7262a7258b/site/docs/extensions/index.md?plain=1#L139

```suggestion
      value: any1
```

site/examples/extensions/variadic_inconsistent_example.yaml
```diff
@@ -0,0 +1,12 @@
+urn: extension:example:variadic_inconsistent
+scalar_functions:
+- name: count_args
+  description: Returns the number of arguments. Each argument can be a different type.
+  impls:
+  - args:
+    - name: first
+      value: any
```
nielspardon, line 8: Here I guess it would probably be cleaner to also use `any1` to mirror `site/examples/extensions/variadic_consistent_example.yaml`.

site/docs/expressions/scalar_functions.md
nielspardon, line 89: This restates the rule #960 landed on main: "All type components except the outermost nullability must match structurally" (L111), whose binding tables already use this same `i32`/`i32?` vs `list<i32>`/`list<i32?>` pair (L138-139). After a rebase it's a second name for an existing rule.

Worth cross-referencing rather than dropping outright, though, because it's a different comparison than #960's — that one binds a declared `any` against one concrete argument, this one compares arguments to each other across repetitions. And unqualified it reads against the `DISCRETE` bullet (L114): there each argument's outermost nullability must match the declaration at its position, and a variadic argument has exactly one declared position, so nullability is already uniform and the relaxation is a no-op. "Nullability follows the nullability handling option" is carrying that caveat implicitly.

```suggestion
When the last argument of a function is variadic and declares a type parameter e.g. `fn(A, B, C...)`, the `C` parameter can be marked as either consistent or inconsistent. If marked as consistent, the function can only be bound to arguments where all the `C` types bind to the same type, with outermost nullability treated as described in [Nullability and `any` Type Binding](#nullability-and-any-type-binding). If marked as inconsistent, each unique `C` can be bound to a different type within the type parameter's constraints. The default behavior is `CONSISTENT`.
```

Same wording change applies to "all variadic arguments must resolve to the same base type" below.

🤖 This comment was assisted by AI.

site/docs/expressions/scalar_functions.md
```diff
@@ -51,114 +51,67 @@ There are three main types of arguments: value arguments, type arguments, and en
...
 
-When the last argument of a function is variadic and declares a type parameter e.g. `fn(A, B, C...)`, the C parameter can be marked as either consistent or inconsistent. If marked as consistent, the function can only be bound to arguments where all the C types are the same concrete type. If marked as inconsistent, each unique C can be bound to a different type within the constraints of what T allows.
+!!! note
+    The `parameterConsistency` field only has an effect when the variadic argument contains a type parameter (like `any1`, `T`, etc.). For concrete types (like `boolean?`, `i32`, `string`), it has no effect. Additionally, if the type parameter appears in non-variadic positions (such as the return type), those occurrences pin the parameter and all variadic arguments must match that type regardless of the `parameterConsistency` setting. When the variadic argument does contain a type parameter, it is recommended that extension authors explicitly specify `parameterConsistency` to make the intended behavior clear.
```
nielspardon, line 114: The middle sentence here is the rule that decides every real case in the catalog, and it's sitting in a note as a parenthetical. 

Every parameterized variadic impl in the standard extensions puts the parameter in the return type — `coalesce`, `least`, `least_skip_null`, `greatest`, `greatest_skip_null` as `any1 -> any1`; `concat` and `concat_ws` vchar as `varchar<L1> -> varchar<L1>` — so as written, `parameterConsistency` is inert for all of them, which sits oddly with the next sentence recommending authors set it explicitly.

It also merges two different mechanisms:

- A parameter in a **non-variadic argument** does constrain binding — that argument's concrete
  type fixes it, and the same-name rule above (L82) already covers that.
- A **return type** is derived from the binding, so it can't constrain the input. Treating it as pinning makes the return type an input constraint, which is a new rule rather than a restatement of an existing one.

And implementations don't do this today: substrait-java binds `f(any1...) -> any1` with `INCONSISTENT` and arguments `(i32, string)` to `i32` — first repetition wins — so this would make the entire catalog's current behavior non-conforming.

Suggest promoting it into the body, split into those two cases, and deciding the return-type case explicitly: either `INCONSISTENT` combined with the parameter in the return type is invalid (a rule the schema can carry), or the spec says how the return type derives from the several bindings. As written, the note picks a third answer — constrain the input — without saying what a producer should emit.

🤖 This comment was assisted by AI.

nielspardon:
@ryanmaka07-cmyk I'm not sure what you are trying to do on this PR but so far your contributions were not really helpful. May I ask you to reconsider what you are posting here. Thanks.

benbellick:
I'm going to close this PR for now. I wrote up an issue (https://github.com/substrait-io/substrait/issues/1192) outlining what I believe is a better approach, which is to drop `parameterConsistency` entirely. Lets continue the discussion there. If it seems useful to take the approach offered by this PR, we can reevaluate and open this back up. Thanks

PR #932, opened 2026-01-08
feat!: require window function bounds and support multiple order-by columns

### LINKED ISSUE #930: RANGE bounds spec is overly restrictive for UNBOUNDED/CURRENT ROW
### Problem

The proto spec at [`algebra.proto:1339-1340`](https://github.com/substrait-io/substrait/blob/92d2e757a330f9c973bb566817dc92afd1badcb2/proto/substrait/algebra.proto#L1339-L1340) states:

> "This bounds type is only valid if there is a single ordering column."

This blocks valid SQL queries like:
```sql
SUM(amount) OVER (ORDER BY col1, col2 RANGE BETWEEN UNBOUNDED PRECEDING AND CURRENT ROW)
```

### SQL Standard Behavior

Per ISO/IEC 9075-2 Section 7.11, General Rule 5b.i.2.A.II, the single ordering column restriction applies only to RANGE with numeric offsets:

 > "If WFB1 specifies `<window frame preceding>`, then... let SK be **the only `<sort key>`** contained in the window ordering clause"

For RANGE with UNBOUNDED/CURRENT ROW:
- No "only sort key" requirement in the standard
- Works with any number of ORDER BY columns

Knowing that the SQL standard is behind a paywall, I validated it against Postgres, so that I could share a concrete example and a freely available documentation, as detailed in the next section.

### PostgreSQL Validation

PostgreSQL correctly implements the SQL standard, the following query executes correctly (including the `range_sum` part which is forbidden by the current substrait spec):

 ```sql
 WITH test(id, col1, col2, val) AS (
   VALUES (1, 10, 100, 1), (2, 10, 100, 2), (3, 10, 100, 3), (4, 20, 200, 4), (5, 20, 200, 5)
 )
 SELECT
   id, col1, col2, val,
   SUM(val) OVER (ORDER BY col1, col2) AS default_sum,
   SUM(val) OVER (ORDER BY col1, col2 RANGE BETWEEN UNBOUNDED PRECEDING AND CURRENT ROW) AS range_sum,
   SUM(val) OVER (ORDER BY col1, col2 ROWS BETWEEN UNBOUNDED PRECEDING AND CURRENT ROW) AS rows_sum
 FROM test ORDER BY id;
 ```

A runnable version is available via [db-fiddle](https://www.db-fiddle.com/f/kzZ2toPAZT8cm14YgdeU2W/1) for Postgres 18.

The Postgres [doc](https://www.postgresql.org/docs/current/sql-expressions.html#SYNTAX-WINDOW-FUNCTIONS) is also clear that this limitation only applies when numeric `offsets` are being used:

> "In the offset PRECEDING and offset FOLLOWING frame options, the offset must be an expression not containing any variables, aggregate functions, or window functions. The meaning of the offset depends on the frame mode:"
> [...]
> "      - In RANGE mode, these options require that the ORDER BY clause specify *exactly one column*."

(emphasis is mine)

PostgreSQL correctly restricts only numeric offsets:
```
WITH test(id, col1, col2, val) AS (VALUES (1, 10, 100, 1))
      SELECT
        id, col1, col2, val,
        SUM(val) OVER (ORDER BY col1, col2 RANGE BETWEEN 1 PRECEDING AND CURRENT ROW) AS range_sum
      FROM test
      ORDER BY id;
```
fails with `Query Error: RANGE with offset PRECEDING/FOLLOWING requires exactly one ORDER BY column` ([db-fiddle](https://www.db-fiddle.com/f/S78EjoxziUPt2nuJQJCxh/0)).

### Proposed Fix

Update [`algebra.proto:1339-1340`](https://github.com/substrait-io/substrait/blob/92d2e757a330f9c973bb566817dc92afd1badcb2/proto/substrait/algebra.proto#L1339-L1340) comment to something along the line of:
```
// When using numeric offsets (Preceding/Following with offset > 0), this bounds type
// requires a single ordering column. UNBOUNDED and CURRENT ROW work with any number
// of ordering columns.
```

I noticed this while using substrait-go, which [implements the validation](https://github.com/substrait-io/substrait-go/blob/main/expr/functions.go#L628-L637) according to the spec, and which should be updated too (filed https://github.com/substrait-io/substrait-go/issues/181 to track this).

I couldn't find anything similar for substrait-java, but all implementations should be checked for consistency.

benbellick:
Closes #930 

BREAKING CHANGE: Window function bindings no longer have default values.

- - -
This change is [Reviewable](https://reviewable.io/reviews/substrait-io/substrait/932)

### DIFF
--- a/proto/substrait/algebra.proto
+++ b/proto/substrait/algebra.proto
@@ -1334,10 +1334,12 @@ message Expression {
       // The lower and upper bound specify how many rows before and after the current row
       // the window should extend.
       BOUNDS_TYPE_ROWS = 1;
-      // The lower and upper bound describe a range of values.  The window should include all rows
-      // where the value of the ordering column is greater than or equal to (current_value - lower bound)
-      // and less than or equal to (current_value + upper bound).  This bounds type is only valid if there
-      // is a single ordering column.
+      // The lower and upper bound describe a range of values. When using numeric offsets (Preceding
+      // or Following with offset > 0), the window includes all rows where the value of the ordering
+      // column is greater than or equal to (current_value - lower bound) and less than or equal to
+      // (current_value + upper bound). Numeric offsets require a single ordering column.
+      // UNBOUNDED and CURRENT ROW bounds work with any number of ordering columns. CURRENT ROW
+      // includes all rows with matching values across all ordering columns (peer rows).
       BOUNDS_TYPE_RANGE = 2;
     }
 
--- a/proto/substrait/algebra.proto
+++ b/proto/substrait/algebra.proto
@@ -1337,8 +1337,9 @@ message Expression {
       // The lower and upper bound describe a range of values. When using numeric offsets (Preceding
       // or Following with offset > 0), the window includes all rows where the value of the ordering
       // column is greater than or equal to (current_value - lower bound) and less than or equal to
-      // (current_value + upper bound). Numeric offsets require a single ordering column.
-      // UNBOUNDED and CURRENT ROW bounds work with any number of ordering columns. CURRENT ROW
+      // (current_value + upper bound). When ANY numeric offset is present as a bound, there must be
+      // EXACTLY ONE ordering column.
+      // UNBOUNDED and CURRENT ROW bounds work with 0 or more ordering columns. CURRENT ROW
       // includes all rows with matching values across all ordering columns (peer rows).
       BOUNDS_TYPE_RANGE = 2;
     }
--- a/site/docs/expressions/window_functions.md
+++ b/site/docs/expressions/window_functions.md
@@ -15,13 +15,13 @@ Window function signatures contain all the properties defined for [aggregate fun
 
 When binding a window function, the binding must include the following additional properties beyond the standard aggregate binding properties:
 
-| Property    | Description                                                  | Required                                                     |
-| ----------- | ------------------------------------------------------------ | ------------------------------------------------------------ |
-| Partition   | A list of partitioning expressions.                          | False, defaults to a single partition for the entire dataset |
-| Order By    | A list of ordering expressions with sort directions.         | False, defaults to unordered                                 |
-| Bounds Type | ROWS or RANGE. ROWS bounds count physical rows. RANGE bounds consider value equivalence based on ordering columns. | False, defaults to RANGE                                     |
-| Lower Bound | Preceding(int64), Following(int64), CurrentRow, or Unbounded. | False, defaults to Unbounded (start of partition)            |
-| Upper Bound | Preceding(int64), Following(int64), CurrentRow, or Unbounded. | False, defaults to Unbounded (end of partition)              |
+| Property    | Description                                                  | Required |
+| ----------- | ------------------------------------------------------------ | -------- |
+| Partition   | A list of partitioning expressions. Empty list means a single partition for the entire dataset. | True     |
+| Order By    | A list of ordering expressions with sort directions. Empty list means unordered. | True     |
+| Bounds Type | ROWS or RANGE. ROWS bounds count physical rows. RANGE bounds consider value equivalence based on ordering columns. | True     |
+| Lower Bound | Preceding(int64), Following(int64), CurrentRow, or Unbounded. | True     |
+| Upper Bound | Preceding(int64), Following(int64), CurrentRow, or Unbounded. | True     |
 
 ### RANGE Bounds with Multiple Ordering Columns
 
--- a/site/docs/expressions/window_functions.md
+++ b/site/docs/expressions/window_functions.md
@@ -18,8 +18,16 @@ When binding a window function, the binding must include the following additiona
 | Property    | Description                                                  | Required                                                     |
 | ----------- | ------------------------------------------------------------ | ------------------------------------------------------------ |
 | Partition   | A list of partitioning expressions.                          | False, defaults to a single partition for the entire dataset |
-| Lower Bound | Bound Following(int64), Bound Trailing(int64) or CurrentRow. | False, defaults to start of partition                        |
-| Upper Bound | Bound Following(int64), Bound Trailing(int64) or CurrentRow. | False, defaults to end of partition                          |
+| Order By    | A list of ordering expressions with sort directions.         | False, defaults to unordered                                 |
+| Bounds Type | ROWS or RANGE. ROWS bounds count physical rows. RANGE bounds consider value equivalence based on ordering columns. | False, defaults to RANGE                                     |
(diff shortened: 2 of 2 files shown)


site/docs/expressions/window_functions.md
```diff
@@ -18,8 +18,16 @@ When binding a window function, the binding must include the following additiona
 | Property    | Description                                                  | Required                                                     |
 | ----------- | ------------------------------------------------------------ | ------------------------------------------------------------ |
 | Partition   | A list of partitioning expressions.                          | False, defaults to a single partition for the entire dataset |
-| Lower Bound | Bound Following(int64), Bound Trailing(int64) or CurrentRow. | False, defaults to start of partition                        |
-| Upper Bound | Bound Following(int64), Bound Trailing(int64) or CurrentRow. | False, defaults to end of partition                          |
+| Order By    | A list of ordering expressions with sort directions.         | False, defaults to unordered                                 |
+| Bounds Type | ROWS or RANGE. ROWS bounds count physical rows. RANGE bounds consider value equivalence based on ordering columns. | False, defaults to RANGE                                     |
```
asolimando: There is also `GROUP`, even if not implemented at the moment, would it be worth mentioning it and specifying is not supported yet?

site/docs/expressions/window_functions.md
```diff
@@ -18,8 +18,16 @@ When binding a window function, the binding must include the following additiona
...
+| Lower Bound | Preceding(int64), Following(int64), CurrentRow, or Unbounded. | False, defaults to Unbounded (start of partition)            |
+| Upper Bound | Preceding(int64), Following(int64), CurrentRow, or Unbounded. | False, defaults to Unbounded (end of partition)              |
```
asolimando: For defaults, it actually depends on a few additional factors:

Quoting Postgres [doc](https://www.postgresql.org/docs/current/sql-expressions.html#SYNTAX-WINDOW-FUNCTIONS):
> The default framing option is RANGE UNBOUNDED PRECEDING, which is the same as RANGE BETWEEN UNBOUNDED PRECEDING AND CURRENT ROW. With ORDER BY, this sets the frame to be all rows from the partition start up through the current row's last ORDER BY peer. Without ORDER BY, this means all rows of the partition are included in the window frame, since all rows become peers of the current row.

asolimando (commented):
LGTM mostly but there are a couple of open questions

proto/substrait/algebra.proto
```diff
@@ -1334,10 +1334,12 @@ message Expression {
       // The lower and upper bound specify how many rows before and after the current row
       // the window should extend.
       BOUNDS_TYPE_ROWS = 1;
-      // The lower and upper bound describe a range of values.  The window should include all rows
-      // where the value of the ordering column is greater than or equal to (current_value - lower bound)
-      // and less than or equal to (current_value + upper bound).  This bounds type is only valid if there
-      // is a single ordering column.
+      // The lower and upper bound describe a range of values. When using numeric offsets (Preceding
+      // or Following with offset > 0), the window includes all rows where the value of the ordering
```
vbarua: https://www.db-fiddle.com/f/vWVwD7V711mRtd77EG4Crs/1

It appears that if _any_ offset is present its not valid, including 0.

site/docs/expressions/window_functions.md
vbarua: We don't have to match Postgres default behaviours here, especially given that we've already assigned default behaviours for these and changing these would have implications for systems that currently implement the defaults correctly as given.

Not for this PR, but my actual inclination for window functions would be to get rid of _all_ of the defaults. This stuff is already complicated enough, and trying to reason about it in the presence of defaults just makes it harder. Forcing plan producers to be explicit about this seems easier than trying to pick the correct defaults for everyone to use.

proto/substrait/algebra.proto
```diff
@@ -1334,10 +1334,12 @@ message Expression {
...
+      // column is greater than or equal to (current_value - lower bound) and less than or equal to
+      // (current_value + upper bound). Numeric offsets require a single ordering column.
```
vbarua:
```suggestion
      // (current_value + upper bound). When ANY numeric offset is present as a bound, there must be
      // EXACTLY ONE ordering column.
```

proto/substrait/algebra.proto
```diff
@@ -1334,10 +1334,12 @@ message Expression {
...
+      // UNBOUNDED and CURRENT ROW bounds work with any number of ordering columns. CURRENT ROW
```
vbarua:
```suggestion
      // UNBOUNDED and CURRENT ROW bounds work with 0 or more ordering columns. CURRENT ROW
```

vbarua:
I would normally finagle the title to pass the conventional commit CI check, but in this case it's complaining about the capitalization of `RANGE` and I think that actually is needed / useful for clarity.

site/docs/expressions/window_functions.md
yongchul: +1 to remove default.

site/docs/expressions/window_functions.md
yongchul: @asolimando  could you cite the reference of `GROUP` you are referring to?

asolimando: It's described in "Section 4.15.14 Windowed tables" of the "Part 2: Foundation (SQL/Foundation)" in my SQL standard 2011 (draft version):

```
A window ordering group is a maximal set of rows in a window partition that are peers according to the window
ordering. Although the ordering of rows within a window ordering group is implementation-dependent, it is
possible to totally order the window ordering groups of a window partition, as follows: if WOG1 and WOG2
are two window ordering groups contained in the same window partition P, then WOG1 precedes WOG2 if
and only if some row of WOG1 precedes some row of WOG2.

Optionally, a window may define a window frame for each row R. A window frame is always defined relative
to the current row. A window frame is specified by up to four syntactic elements:

- The choice of RANGE, to indicate a logical definition of the window frame by offsetting forward or
backward from the current row by an increment or decrement to the value of the <sort key>; GROUPS,
to indicate a logical definition of the window frame by counting forward or backward a number of window
ordering groups (as defined by the <sort key>s) from the window ordering group containing the current
row; or ROWS, to indicate a physical definition of the window frame, by counting rows forward or backward
from the current row.
- ...
[...]
```

For what I could check, and up to my best understanding, Postgres' follows the standard in that respect, so its [documentation](https://www.postgresql.org/docs/current/sql-expressions.html#SYNTAX-WINDOW-FUNCTIONS) can be referred to.

Please note that this is an optional feature, citing from conformance rules from the SQL:2011 standard: 
```
Without Feature T620, “WINDOW clause: GROUPS option”, conforming SQL language shall not contain
<window frame units> that specifies GROUPS.
```

benbellick:
This is still relevant, not stale

PR #934, opened 2026-01-08
docs: add func to type short name list

### LINKED ISSUE #933: missing short name for lambdas in documentation
@tokoko rightly pointed out that the `func` type is missing from type short names [here](https://substrait.io/extensions/#type-short-names). Lets add it! :)

benbellick:
Closes #933


site/docs/extensions/index.md
```diff
@@ -97,18 +97,20 @@ A function signature uniquely identifies a function implementation within a sing
 | struct&lt;T1,T2,...,TN&gt;      | struct         |
 | list&lt;T&gt;                   | list           |
 | map&lt;K,V&gt;                  | map            |
+| func&lt;T-&gt;R&gt;, func&lt;(T1,...,TN)-&gt;R&gt; | func |
```
tokoko, line 101: not important, but maybe we should have `func<(T1,...,TN), R>` here and also modify struct to `struct<(T1,T2,...,TN)>`. would be more in-line with proto fields.

benbellick: I don't quite follow. The above rendered becomes: `func<T->R>, func<(T1,...,TN)->R>`. This is consistent with the grammar defined in the antlr files.

Similarly, `struct<T1,T2,...,TN>` is valid according to the antlr files. 

Are you proposing altering the type grammar?

tokoko: oh.. never mind then. I was looking at protos only.

vbarua:
We should also include it in https://github.com/substrait-io/substrait/blob/main/site/docs/types/type_classes.md.

site/docs/extensions/index.md
```diff
@@ -97,18 +97,20 @@ A function signature uniquely identifies a function implementation within a sing
...
-| `sum(any1) => any1`                               | `sum:any`        |
-| `concat(str...) => str`                           | `concat:str`     |
+| Function Signature                                | Function Name       |
+| ------------------------------------------------- | ------------------- |
+| `add(optional enumeration, i8, i8) => i8`         | `add:i8_i8`         |
+| `avg(fp32) => fp32`                               | `avg:fp32`          |
+| `extract(required enumeration, timestamp) => i64` | `extract:req_ts`    |
+| `sum(any1) => any1`                               | `sum:any`           |
+| `concat(str...) => str`                           | `concat:str`        |
+| `transform(list<any1>, func<any1 -> any2>) => list<any2>` | `transform:list_func` |
```
yongchul, line 114: I need to do a lot of catch up but what does  this `list_func` mean and where is it defined?

benbellick: `transform:list_func` is the unique identifier of the function named `transform` which has two parameters, a `list` and a `func`. 

https://substrait.io/extensions/#function-signature

yongchul: ah... silly, yes. :)

yongchul (approved):

vbarua (approved):

PR #936, opened 2026-01-08
docs: clarify UDT references must begin with u!

### LINKED ISSUE #935: docs: clarify u! prefix requirement for user-defined types in YAML
The documentation is unclear about whether the `u!` prefix is required before user-defined types in YAML files.

For example, this file doesn't use the `u!` prefix:
https://github.com/substrait-io/substrait/blob/92d2e757a330f9c973bb566817dc92afd1badcb2/extensions/extension_types.yaml#L1-L11

But this file does:
https://github.com/substrait-io/substrait/blob/92d2e757a330f9c973bb566817dc92afd1badcb2/extensions/functions_geometry.yaml#L12-L21

benbellick:
- Fixes error in `extension_types.yaml` which incorrectly uses `point` instead of `u!point`
- Adds clarification to documentation ensuring that UDTs from same file must use `u!` prefix

Closes #935

### DIFF
--- a/site/docs/extensions/index.md
+++ b/site/docs/extensions/index.md
@@ -36,9 +36,15 @@ Within YAML extension files, user-defined types must be referenced with the `u!`
 --8<-- "examples/types/user_defined_point.yaml"
 ```
 
-Built-in types like `i32`, `fp64`, and `string` do not use any prefix.
+[Built-in types](../types/type_classes.md) like `i32`, `string`, or `list` (as in `list<fp64>`) do not use any prefix.
 
 A YAML file can also reference types and type variations defined in another YAML file. To do this, it must declare the extension it depends on using a key-value pair in the `dependencies` key, where the value is the extension URN, and the key is a valid identifier that can then be used as an identifier-safe alias for the extension URN. This alias can then be used as a `.`-separated namespace prefix wherever a type class or type variation name is expected. Note that user-defined types still require the `u!` prefix when referenced via namespace aliases (e.g., `ext.u!point`).
+
+!!! note "Grammar"
+    The grammar for referencing user-defined types is:
+    ```
+    <udt_reference> ::= [<dependency_alias> '.'] 'u!' <type_name>
+    ```
 
 For example, if the extension with extension URN `extension:io.substrait:extension_types` defines a user-defined type called `point`, a different YAML file can use the type in a function declaration as follows:
 
--- a/site/docs/extensions/index.md
+++ b/site/docs/extensions/index.md
@@ -36,7 +36,7 @@ Within YAML extension files, user-defined types must be referenced with the `u!`
 --8<-- "examples/types/user_defined_point.yaml"
 ```
 
-[Built-in types](../types/type_classes.md) like `i32`, `string`, or `list` (as in `list<fp64>`) do not use any prefix.
+[Built-in types](../types/type_classes.md#built-in-types) like `i32`, `string`, or `list` (as in `list<fp64>`) do not use any prefix.
 
 A YAML file can also reference types and type variations defined in another YAML file. To do this, it must declare the extension it depends on using a key-value pair in the `dependencies` key, where the value is the extension URN, and the key is a valid identifier that can then be used as an identifier-safe alias for the extension URN. This alias can then be used as a `.`-separated namespace prefix wherever a type class or type variation name is expected. Note that user-defined types still require the `u!` prefix when referenced via namespace aliases (e.g., `ext.u!point`).
 
--- a/site/docs/extensions/index.md
+++ b/site/docs/extensions/index.md
@@ -117,7 +117,7 @@ A function signature uniquely identifies a function implementation within a sing
 | list&lt;T&gt;                   | list           |
 | map&lt;K,V&gt;                  | map            |
 | any[\d]?                        | any            |
-| user defined type               | u!name         |
+| user-defined type &lt;name&gt;  | u!&lt;name&gt; |
 
 #### Examples
 
--- a/site/docs/extensions/index.md
+++ b/site/docs/extensions/index.md
@@ -30,6 +30,14 @@ A Substrait plan can reference one or more YAML files via their extension URN. I
 | Type Variation     | The name as defined on the type variation object.            |
 | Function Signature | A function signature as described below.       |
 
+Within YAML extension files, user-defined types must be referenced with the `u!` prefix followed by the type name (e.g., `u!point`) in function arguments and return types:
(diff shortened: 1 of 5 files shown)


site/docs/extensions/index.md
```diff
@@ -30,6 +30,14 @@ A Substrait plan can reference one or more YAML files via their extension URN. I
...
+
+```yaml
+--8<-- "examples/types/user_defined_point.yaml"
+```
+
+Built-in types like `i32`, `fp64`, and `string` do not use any prefix.
+
 A YAML file can also reference types and type variations defined in another YAML file. To do this, it must declare the extension it depends on using a key-value pair in the `dependencies` key, where the value is the extension URN, and the key is a valid identifier that can then be used as an identifier-safe alias for the extension URN. This alias can then be used as a `.`-separated namespace prefix wherever a type class or type variation name is expected.
 
 For example, if the extension with extension URN `extension:io.substrait:extension_types` defines a type called `point`, a different YAML file can use the type in a function declaration as follows:
```
benbellick: The example below shows reference a type via `ext.point`. But maybe it is worth considering if `ext.u!point` is more consistent?

vbarua: I think it's worth forcing the `u!` for consistency. It makes it easier to disambiguate for user-defined types both visually and for parsing purposes.

vbarua (approved):
LGTM. Good clarification.

site/docs/extensions/index.md
benbellick: I added it for UDTs from other files as well. Thanks!

site/docs/extensions/index.md
```diff
@@ -30,9 +30,17 @@ A Substrait plan can reference one or more YAML files via their extension URN. I
...
 
-A YAML file can also reference types and type variations defined in another YAML file. To do this, it must declare the extension it depends on using a key-value pair in the `dependencies` key, where the value is the extension URN, and the key is a valid identifier that can then be used as an identifier-safe alias for the extension URN. This alias can then be used as a `.`-separated namespace prefix wherever a type class or type variation name is expected.
+Within YAML extension files, user-defined types must be referenced with the `u!` prefix followed by the type name (e.g., `u!point`) in function arguments and return types:
 
-For example, if the extension with extension URN `extension:io.substrait:extension_types` defines a type called `point`, a different YAML file can use the type in a function declaration as follows:
+```yaml
+--8<-- "examples/types/user_defined_point.yaml"
+```
+
+Built-in types like `i32`, `fp64`, and `string` do not use any prefix.
```
yongchul: link to builtin types?

site/docs/extensions/index.md
```diff
@@ -30,9 +30,17 @@ A Substrait plan can reference one or more YAML files via their extension URN. I
...
+
+A YAML file can also reference types and type variations defined in another YAML file. To do this, it must declare the extension it depends on using a key-value pair in the `dependencies` key, where the value is the extension URN, and the key is a valid identifier that can then be used as an identifier-safe alias for the extension URN. This alias can then be used as a `.`-separated namespace prefix wherever a type class or type variation name is expected. Note that user-defined types still require the `u!` prefix when referenced via namespace aliases (e.g., `ext.u!point`).
```
yongchul, line 43: I'm wondering whether including or referring a grammar (perhaps type syntax parsing?) as well would help understand... either BNF or regular expression like below.

```
(<dependency extension key> '.')? 'u!' <UDT name>
```

benbellick: I ended up doing it in a style to match: 
https://github.com/substrait-io/substrait/blob/92d2e757a330f9c973bb566817dc92afd1badcb2/site/docs/extensions/index.md?plain=1#L51-L52

site/docs/extensions/index.md
benbellick: I also created a new section in `type_classes.md` called `#Built-in types` and moved `Simple Types` and `Compound Types` to be subsections of these. This way, we can link directly to the built in types section.

site/docs/types/type_parsing.md
```diff
@@ -66,6 +66,30 @@ Similar to structs, maps and lists can also have a type as one of their paramete
     map<i32?, list<map<i32, string?>>>
     ```
 
+### User-Defined Types
+
+User-defined types must be referenced using the `u!` prefix followed by the type name:
```
benbellick: Might as well link this section, too.

site/docs/types/type_classes.md
```diff
@@ -4,7 +4,9 @@ In Substrait, the "class" of a type, not to be confused with the concept from ob
 
 Implementations of a Substrait type must support *at least* this set of values, but may include more; for example, an `i8` could be represented using the same in-memory format as an `i32`, as long as functions operating on `i8` values within [-128..127] behave as specified (in this case, this means 8-bit overflow must work as expected). Operating on values outside the specified range is unspecified behavior.
 
-## Simple Types
+## Built-in Types
```
benbellick, line 7: This way we can make links to `Built-in Types`.

site/docs/extensions/index.md
yongchul: Use ':' and '_' to denote those are the literals.

yongchul: Here it is using `{}` to group terms. Yet the prefix syntax use `[]`. Should we stick to one notation? Or, we can just adapt https://datatracker.ietf.org/doc/html/rfc5234 for all these references.

yongchul (changes requested):

site/docs/extensions/index.md
yongchul: Just to be clear, if we were to adapt one notation, this PR should not change everything of course (can be other PR).

benbellick: I have altered to ABNF in this file, but I have left `type_parsing.md` untouched for now. LMK if you think I should alter that one in this PR as well!

yongchul: Can you also update this from "user defined type" to "user defined type `<name>`"?

yongchul (approved):

site/docs/extensions/index.md
yongchul: Thanks for doing this! Looks great to me. We can update other files in a separate PR. I believe @EpsilonPrime probably interested in this so let's get his opinion on this ABNF.

benbellick: I also added the `<name>` part to the second column to make it clear that it is a value to be inserted. It renders as follows:

image

EpsilonPrime (approved):

vbarua (approved):

Issue #849, opened 2025-08-07
what's an ExpressionNamedStruct


gabry-lab:
Hi, folks, I find a ExpressionNamedStruct message in DerivationExpression, since it is not a kind of DerivationExpression, then what it is?

nielspardon:
@gabry-lab can you give us a bit more detail on where exactly you are finding `ExpressionNamedStruct`?

tokoko:
The last time I asked about those protos (DerivationExpression and everything else related to it) I came to the conclusion that they are unused and out of date. Derivation expressions are defined with antlr right now, not sure what the intention of protos were originally (maybe udfs??)

nielspardon:
I see, it's defined here (if anyone was wondering): 

https://github.com/substrait-io/substrait/blob/c37d459e93875c33c7b728c3deb7c9bfe5fc7694/proto/substrait/type_expressions.proto#L13

It is only referenced in this file:
https://github.com/substrait-io/substrait/blob/c37d459e93875c33c7b728c3deb7c9bfe5fc7694/proto/substrait/function.proto

And the message types in `function.proto` do not seem to be used anywhere else.

nielspardon:
I'm proposing to remove the two files to avoid further confusion: https://github.com/substrait-io/substrait/pull/940

nielspardon:
#940 has been merged and the files no longer exist in the current spec.

PR #940, opened 2026-01-15
feat!: remove unused function.proto, parameterized_types.proto, type_expressions.proto

nielspardon:
The clarification question asked in https://github.com/substrait-io/substrait/issues/849 raised awareness to me about the following three proto files we include the Substrait specification which do not seem to be used:

- https://github.com/substrait-io/substrait/blob/c37d459e93875c33c7b728c3deb7c9bfe5fc7694/proto/substrait/function.proto
- https://github.com/substrait-io/substrait/blob/c37d459e93875c33c7b728c3deb7c9bfe5fc7694/proto/substrait/type_expressions.proto
- https://github.com/substrait-io/substrait/blob/c37d459e93875c33c7b728c3deb7c9bfe5fc7694/proto/substrait/parameterized_types.proto

BREAKING CHANGE: I would propose to remove these proto definitions if they are not used / no longer part of the spec.


benbellick (approved):
I can't find the discussion, but I believe those files were originally in because @jacques-n wanted to allow for the possibility of representing the YAML extensions in proto files so that systems can communicate them. 


I do think there is value in that idea, but I think in the current state it is a bit premature. Until then, including these protos just becomes another place to stub your toe since it is very easy to alter the YAML schema but forget to alter these protos. 


So, I agree with getting rid of them! We can always add them back :)

tokoko:
+1 to removal. I think parameterized types proto is used for derivations only as well

nielspardon:
> +1 to removal. I think parameterized types proto is used for derivations only as well

included that one in a follow-up commit

yongchul (approved):
+1 to removal but let's make sure @jacques-n or one more PMC agrees just in case.

jacques-n:
+1

nielspardon:
This has the required number of approvals by PMC members and would only require a PMC member to click the merge button.

PR #942, opened 2026-01-26
docs: clarify type equality and MIRROR nullability handling

### LINKED ISSUE #941: Clarify type equality definition
The type system documentation defines four components of a type:
https://github.com/substrait-io/substrait/blob/95f8f350328f63bf44be63f28eb67680983edbb1/site/docs/types/type_system.md?plain=1#L7-L12

However, it doesn't explicitly state that two types are equal only when all four components match. The docs should clarify that type equality requires all four components (class, nullability, variation, and parameters) to be identical, and that this applies recursively for compound types.

See #943  for a related question about how `any` type parameter binding interacts with nullability.

benbellick:
## Summary
- Explicitly define type equality as requiring all four components (class, nullability, variation, parameters) to match
- Clarify that MIRROR mode only considers outermost nullability, not inner/element nullability in compound types

Fixes #941


benbellick:
Closing for now as I have altered a bit what the individual tickets are about.

PR #945, opened 2026-01-28
feat: add current_date, current_timestamp and current_timezone variables

nielspardon:
This PR adds 3 new execution context variables:
- `current_date` for retrieving the current date
- `current_timestamp` for retrieving the timestamp with and without applying a timezone
- `current_timezone` for retrieving the local timezone

- - -
This change is [Reviewable](https://reviewable.io/reviews/substrait-io/substrait/945)

### DIFF
--- a/proto/buf.yaml
+++ b/proto/buf.yaml
@@ -1,7 +1,6 @@
 version: v1
 name: buf.build/substrait/substrait
 lint:
-  allow_comment_ignores: true
   use:
     - STANDARD
   ignore_only:
--- a/proto/substrait/plan.proto
+++ b/proto/substrait/plan.proto
@@ -65,7 +65,7 @@ message Plan {
   // many fields, nested parameterized types, string as a type parameter).
   repeated TypeAlias type_aliases = 9;
 
-  // Optional execution behavior settings for this plan.
+  // Execution behavior settings for this plan, required.
   ExecutionBehavior execution_behavior = 10;
 }
 
@@ -122,12 +122,10 @@ message ExecutionBehavior {
   // VariableEvaluationMode specifies the frequency and scope of execution
   // context variable evaluation.
   enum VariableEvaluationMode {
-    // Variables are evaluated once per Substrait plan execution (default).
-    // Skipping the ENUM_ZERO_VALUE_SUFFIX linting rule so we can have
-    // PER_PLAN as default.
-    // buf:lint:ignore ENUM_ZERO_VALUE_SUFFIX
-    VARIABLE_EVALUATION_MODE_PER_PLAN = 0;
+    VARIABLE_EVALUATION_MODE_UNSPECIFIED = 0;
+    // Variables are evaluated once per Substrait plan execution.
+    VARIABLE_EVALUATION_MODE_PER_PLAN = 1;
     // Variables are evaluated once per record during execution.
-    VARIABLE_EVALUATION_MODE_PER_RECORD = 1;
+    VARIABLE_EVALUATION_MODE_PER_RECORD = 2;
   }
 }
--- a/site/docs/expressions/specialized_record_expressions.md
+++ b/site/docs/expressions/specialized_record_expressions.md
@@ -85,7 +85,7 @@ The execution behavior settings in a Substrait Plan control how execution contex
 
 The execution behavior defines a `VariableEvaluationMode` that controls the scope and frequency of execution context variable evaluation:
 
-- **VARIABLE_EVALUATION_MODE_PER_PLAN** (default): Variables are evaluated once per Substrait plan execution. All records in a single plan execution will see the same values for execution context variables like `current_date` and `current_timestamp`.
+- **VARIABLE_EVALUATION_MODE_PER_PLAN**: Variables are evaluated once per Substrait plan execution. All records in a single plan execution will see the same values for execution context variables like `current_date` and `current_timestamp`.
 - **VARIABLE_EVALUATION_MODE_PER_RECORD**: Variables are evaluated once per record during execution. Each record may see different values for execution context variables if the execution context changes between records.
 
 This setting is particularly important for time-based functions where the evaluation time affects the returned value. For example, with `VARIABLE_EVALUATION_MODE_PER_PLAN` mode, all rows processed in a single plan execution will have the same `current_date` value, while with `VARIABLE_EVALUATION_MODE_PER_RECORD` mode, the date could potentially change between rows if the plan spans a date boundary during execution.


grammar/FuncTestCaseParser.g4
```diff
@@ -31,7 +31,7 @@ testGroupDescription
     ;
 
 testCase
-    : functionName=identifier OParen arguments CParen ( OBracket funcOptions CBracket )? Eq result
+    : functionName=identifier OParen arguments? CParen ( OBracket funcOptions CBracket )? Eq result
```
nielspardon: this change allows test cases with no arguments:

```
current_date() = '2026-01-28'::date
```

tests/coverage/antlr_parser/FuncTestCaseParser.py
nielspardon, line 1: this file was regenerated after the ANTLR grammar changes

tests/coverage/visitor.py
```diff
@@ -291,6 +293,18 @@ def visitArgument(self, ctx: FuncTestCaseParser.ArgumentContext):
...
+        if ctx.varCharArg() is not None:
+            return self.visitVarCharArg(ctx.varCharArg())
+        if ctx.fixedBinaryArg() is not None:
+            return self.visitFixedBinaryArg(ctx.fixedBinaryArg())
+        if ctx.precisionTimeArg() is not None:
+            return self.visitPrecisionTimeArg(ctx.precisionTimeArg())
+        if ctx.precisionTimestampArg() is not None:
+            return self.visitPrecisionTimestampArg(ctx.precisionTimestampArg())
+        if ctx.precisionTimestampTZArg() is not None:
+            return self.visitPrecisionTimestampTZArg(ctx.precisionTimestampTZArg())
```
nielspardon, line 305: this is what Claude identified as missing

tests/coverage/visitor.py
```diff
@@ -364,6 +378,38 @@ def visitIntervalYearArg(self, ctx: FuncTestCaseParser.IntervalYearArgContext):
...
+        )
+
+    def visitPrecisionTimestampTZArg(
+        self, ctx: FuncTestCaseParser.PrecisionTimestampTZArgContext
+    ):
+        type_str = ctx.precisionTimestampTZType().getText().lower()
+        return CaseLiteral(
+            value=ctx.TimestampTzLiteral().getText().strip("'"), type=type_str
+        )
+
```
nielspardon, line 410: this is what Claude identified as missing

extensions/functions_datetime.yaml
```diff
@@ -997,6 +997,33 @@ scalar_functions:
           - name: origin
             value: time
         return: time
+  - name: "current_date"
```
benbellick: I think you will want to set `deterministic` to false in all of these new functions (see [here](https://substrait.io/expressions/scalar_functions/#scalar-functions) for details on that setting). 

Perhaps it is worth altering the test framework so that non-deterministic functions aren't required to have tests? Otherwise, I don't really understand how we can write tests for these functions.

nielspardon: makes sense. I did set `deterministic: false` and removed the tests. seems that is possible already

nielspardon:
I also marked functions that depend on the local timezone as session dependent

extensions/functions_datetime.yaml
```diff
@@ -997,6 +997,39 @@ scalar_functions:
...
+    impls:
+      - args:
+          - name: timezone
+            description: Timezone string from IANA tzdb.
+            value: string
+        return: precision_timestamp_tz<P>
+        deterministic: false
+      - return: precision_timestamp<P>
+        deterministic: false
+        sessionDependent: true
```
yongchul: This is funky case because type parameter `P` here is unbound or context sensitive... Not sure whether we had a case like this before. One option is that we can accept the desired type argument and use that to return.

nielspardon: agreed, I also wasn't sure how that would work here

nielspardon: What do you think of always setting it to the maximum precision allowed by the spec? 

Seems to me that is what is being done for e.g. the `factorial` function for decimals:
https://github.com/substrait-io/substrait/blob/f43b87625eff1cbabdb77bd23cf09e79085621a2/extensions/functions_arithmetic_decimal.yaml#L160-L169

An example where the scale is provided as a parameter (but not used as the scale of the return type if I read it correctly) might be the decimal round function:
https://github.com/substrait-io/substrait/blob/f43b87625eff1cbabdb77bd23cf09e79085621a2/extensions/functions_rounding_decimal.yaml#L30-L83

extensions/functions_datetime.yaml
```diff
@@ -997,6 +997,59 @@ scalar_functions:
...
+    impls:
+      - args:
+          - name: timezone
+            description: Timezone string from IANA tzdb.
+            value: string
+        return: precision_timestamp_tz<12>
+        deterministic: false
+      - return: precision_timestamp<12>
+        deterministic: false
+        sessionDependent: true
```
nielspardon: for these two impls I'm setting the precision to the maximum allowed by the spec for precision timestamp: https://substrait.io/types/type_classes/#compound-types

extensions/functions_datetime.yaml
```diff
@@ -997,6 +997,59 @@ scalar_functions:
...
+          precision = integer_parameter(p)
+          precision_timestamp_tz<precision>
+        deterministic: false
+      - args:
+          - name: p
+            description: precision of the returned timestamp 0 <= p <= 12
+            value: i8
+        return: |-
+          precision = integer_parameter(p)
+          precision_timestamp<precision>
```
nielspardon: I'm adding two more impls that have an int8 value argument `p` which then I can retrieve in the return type expression using `integer_parameter(p)` and use as the precision of the timestmap. 

I gathered this is how this could work based on the information here: https://substrait.io/expressions/scalar_functions/#return-type-expressions

The documentation is a bit sparse and I couldn't find any other instance of `integer_parameter` being used.

extensions/functions_datetime.yaml
yongchul: That works but I prefer to take the precision as a parameter and resolve time per requested parameter. I don't know how many systems actually support pico second precision at the moment and this requires the implementations may need to have extra conversion step to native scale.

Let's hear about how others think.

benbellick: I don't love either of these solutions. Specifically because I worry that allowing unbound type parameters in return types will lead to unspecified behavior somehow. 

(I was actually working on a PR before to explicitly ban the use of unbound parameters in return types.)

Could this be an opportunity to use [type arguments](https://substrait.io/expressions/scalar_functions/#argument-types)? I think we could make the output precision configurable with that.

benbellick: Unfortunately, I don't see any example usages of `type_arg` anywhere, so it might not be quite so simple if it isn't entirely clear. This could be an opportunity to introduce the first example though :)

yongchul: @benbellick yes, either type argument or explicit precision as an argument -- just somehow determine the return type based on that rather than leaving it unbounded or default to max. It looks like Trino does support pico seconds and they do millisecond as default precision and have a function accepting desired precision.

nielspardon: I already added a variant with a precison argument. I can drop the one without arguments and just leave the one with the precision argument: https://github.com/substrait-io/substrait/pull/945#discussion_r2741604336

---

Some more examples of implementations:

PostgreSQL has a variant accepting a precision argument (0-6) and by default 6: 
- https://www.postgresql.org/docs/current/functions-datetime.html#FUNCTIONS-DATETIME-CURRENT

Db2 supports a precision argument (0-12) and 6 by default: 
- https://www.ibm.com/docs/en/db2/12.1.x?topic=registers-current-timestamp

Databricks does not support a precision argument for `CURRENT_TIMESTAMP` though supports 0-6 with the `TIMESTAMP` data type and seems to default to 3 (based on the examples): 
- https://docs.databricks.com/aws/en/sql/language-manual/functions/current_timestamp
- https://docs.databricks.com/aws/en/sql/language-manual/data-types/timestamp-type

Oracle supports a precision argument (0-9) and defaults to 6: 
- https://docs.oracle.com/en/database/oracle/oracle-database/26/sqlrf/CURRENT_TIMESTAMP.html
- https://docs.oracle.com/en/database/oracle/oracle-database/26/sqlrf/Data-Types.html#GUID-94A82966-D380-4583-9AF1-AEE681881E64

yongchul: @nielspardon thx! I missed those new variants. +1 to leaving no argument version from this PR until we figure out the better story around that but I'm not sure whether the return type derivation is valid...

Is `integer_parameter` is the supported type expression? As I skimmed the grammar, looks like there is no way to promote a value argument as a type argument. We definitely need to improve documentation and maybe add support promoting value argument as a type argument.

But the clearer way within current grammar looks like passing a type argument (i.e., `pts<P>`) rather than a value argument.

@EpsilonPrime am I reading the grammar right? 😄

vbarua:
`current_date` and `current_timestamp` have come up in the past:
* https://github.com/substrait-io/substrait/pull/524
* https://github.com/substrait-io/substrait/pull/548

I can try and summarize the discussion, which is effectively that many / most(?) SQL engines don't actually execute these as normal functions. In practice, they get executed _once_ in the query context and then values are reused. #548 tried to capture that distinction with the concept of BATCH (execute once per query) and STREAM (execute once per row), but it fizzled out. This is maybe(?) what `sessionDependent: true` is intended for in the spec, but that's also not a clearly define concept in practice.

> Further this adds a missing specified into the timezone string documentation sentence for existing function definitions.

If possible, could you split this out into its own PR because it adds noise to this PR. It's a simple enough change that I'm YongChul and I can basically auto-approve and merge as its a pure documentation improvement.

nielspardon:
> > Further this adds a missing specified into the timezone string documentation sentence for existing function definitions.
> 
> If possible, could you split this out into its own PR because it adds noise to this PR. It's a simple enough change that I'm YongChul and I can basically auto-approve and merge as its a pure documentation improvement.

I removed the docs change in a97bdce721cc889f775aa01f9bb3ea24a8d7c2c5 and created a new PR for it in #957

nielspardon:
I separated out the visitor changes https://github.com/substrait-io/substrait/pull/958 and removed the ANTLR and visitor changes in this PR in 0eb9fc085f717738c3dd7dd153f67cd486cb86fc

extensions/functions_datetime.yaml
nielspardon: > @nielspardon thx! I missed those new variants. +1 to leaving no argument version from this PR

I removed those in e645439367d93a753c77d5493d738d2456527594

nielspardon: > Is `integer_parameter` is the supported type expression? As I skimmed the grammar, looks like there is no way to promote a value argument as a type argument. We definitely need to improve documentation and maybe add support promoting value argument as a type argument.

When you say you skimmed the grammar were you looking in the same place that I referenced in this comment or is there another place the grammar is defined? https://github.com/substrait-io/substrait/pull/945#discussion_r2741604336

nielspardon:
> I can try and summarize the discussion, which is effectively that many / most(?) SQL engines don't actually execute these as normal functions. In practice, they get executed _once_ in the query context and then values are reused. #548 tried to capture that distinction with the concept of BATCH (execute once per query) and STREAM (execute once per row), but it fizzled out. This is maybe(?) what `sessionDependent: true` is intended for in the spec, but that's also not a clearly define concept in practice.

At IBM we would like to use Substrait with both batch and streaming engines so we would also be interested in properly representing semantic differences between execution modes. It feels to me like this is a more general topic that should be explored independently of this change. 

Can we introduce these very common functions into the spec while we are still figuring out how to properly represent batch vs streaming semantics in a dedicated issue / discussion?

EpsilonPrime:
The last time we discussed (a year or so ago, check the historical meeting notes) this we noted that it syntactically looks like a function but it's more like a pointer to a literal.  We wanted to see prepared statements working so one could inject the date as needed.

extensions/functions_datetime.yaml
```diff
@@ -997,6 +997,50 @@ scalar_functions:
...
+        deterministic: false
+        sessionDependent: true
+  - name: "current_timezone"
```
yongchul: We discussed this in Feb 11 Substrait meeting. There were two proposals.

1. Add current_timestamp and (maybe) current_timezone as dedicated expression (i.e., add this another field of Expression message) proposed by @jacques-n . This is a very convenient way to access popular/frequent values. current timestamp is just too frequently occurring so could be justified this special treatment.
  +. can't be easier to access the popular value
  ~. we may be opening the door to ever growing list of such thing but I don't think there are many of this kind...

2. Capture current timestamp and current timezone in a standard "execution environment" access functions like `type substrait.getenv(key: string, valueType: type)`. Substrait spec defines standard "key"s and current timestamp and current timezone could be one of those standard keys.
  +.  reusable, a lot of other external scenarios can be modeled in this way.
  ~.  oh well, clunkier than expression 😆 

I'm okay with option 1 for current timestamp (have a mixed feeling with current timezone because this is not too common) because it is just ubiquitous.

I'm going to propose option 2 anyway because we have been considering this internally and beneficial in general to access session specific or environment specific information.

extensions/functions_datetime.yaml
nielspardon: Did we also get some clarity on the semantics of the `integer_parameter` function in the community sync yesterday? It looks like those functions were introduced by @jacques-n in #40 according to git blame.

extensions/functions_datetime.yaml
nielspardon: I guess for option 2 we may also need an extension of the Substrait dialect to define which session variables are available for a given dialect.

nielspardon:
> We wanted to see prepared statements working so one could inject the date as needed.

I did see the discussion in the prior issues Victor linked. I do understand that for the "batch" execution mode case, where the evaluation of `CURRENT_TIMESTAMP` would be at the beginning of the `Plan` execution, one could work around not having a function or variable called `CURRENT_TIMESTAMP` within the Substrait spec by using the `DynamicParameter` feature which was added in #780 (about a year ago). I guess this is what you are referring to?

extensions/functions_datetime.yaml
nielspardon: I also noticed that different systems call these variables differently, e.g. Calcite handles timezone as a "session parameter", Db2 calls them "special registers", Google Big Query/PostgreSQL just list them under datetime functions. If we wanted to add them as a new Expression element what would we call them? Any preferences? Something like`ContextualParameter`, `SessionParameter`, `xxxVariable`, `xxxFunction`?

nielspardon: I updated the PR with an initial proposal for new `Expression` elements in f3fcfb9c6342295e1fd9d63af4e8676bb6eb32fd (I called them `SessionVariable` for now)

nielspardon: I added modifications for the Substrait dialect in 02e023d30a7c263289529918f70da1c0a4c5750f

extensions/functions_datetime.yaml
nielspardon: I created the following issue for updating the documentation of the undocumented return type expression functions: https://github.com/substrait-io/substrait/issues/959

proto/substrait/algebra.proto
```diff
@@ -1688,6 +1689,39 @@ message Expression {
       Rel right = 4;
     }
   }
+
+  // Expression that represents common session dependent variables
+  message SessionVariable {
+    oneof session_variable_type {
+      CurrentTimestamp current_timestamp = 1;
```
yongchul: I'm wondering whether we can just define the type here directly rather than via nested messages.

```protobuf
Type.PrecisionTimestampTZ current_timestamp = 1;
Type.String current_timezone = 2;
Type.Date current_date = 3;
```

Any thoughts?

proto/substrait/algebra.proto
```diff
@@ -1688,6 +1689,39 @@ message Expression {
       Rel right = 4;
     }
   }
+
+  // Expression that represents common session dependent variables
+  message SessionVariable {
```
yongchul: As discussed in the Substrait meeting, we would like to push higher bar for the PR. This one does warrant documentation update. Could you add documentation in the expressions?

@vbarua is there any test coverage that we could add for this?

nielspardon: I'm happy to add documentation. Just want to first get agreement on the general approach so I don't have to redo the documentation

extensions/functions_datetime.yaml
yongchul: Overall, I like your change -- introducing it as a session variable, standardized way to capture anything tied to a session. We can promote one to `oneof` if the concept or value is generic enough. This also opens up the generic way to access to custom environment or context variables. something like

```protobuf

oneof session_variable_Type {
  GenericSessionVariable session_variable = 10;
}

message GenericSessionVariable {
  Type type = 1;
  string name = 2;
}
```

where the generic session variable accesses generic key value store of whatever available in session context for the receiver. Now that I see this again, there is certainly some overlap with this and dynamic parameter. The key difference is that this is based on name, while dynamic parameter is id based...

Anyway, this will be done in a separate PR and +1 to the direction.

Thank you!

proto/substrait/algebra.proto
nielspardon: I gave it a try to see what happens when I implement code on top of this in substrait-java and this works. Updated the proto definitions.

proto/substrait/algebra.proto
nielspardon: I added some proposal for documentation. Let me know if there's anything I can improve.

extensions/functions_datetime.yaml
nielspardon: I introduced an `ExtensionSessionVariable` in the latest commit.

proto/substrait/algebra.proto
```diff
@@ -1688,6 +1689,38 @@ message Expression {
...
+      ExtensionSessionVariable extension_session_variable = 4;
+    }
+
+    // extension session variables
+    message ExtensionSessionVariable {
+      // name of the extension session variable, required
+      string name = 1;
+
+      // type of the extension session variable, required
+      Type type = 2;
```
yongchul: Thank you for adding this (I didn't mean to do this in this PR)!

site/docs/expressions/specialized_record_expressions.md
```diff
@@ -69,10 +69,21 @@ Multi Value:
...
 
+| Session Variable | Description | Return Type |
+| --------------------- | ----------- | ----------- |
+| current_date | a non-deterministic, session dependent variable containing the current date | date |
+| current_timestamp | a non-deterministic, session dependent variable containing the current timestamp in current_timezone | PRECISION_TIMESTAMP_TZ<P> |
+| current_timezone | a deterministic, session dependent variable containing the current session timezone as a string defined by IANA timezone database (https://www.iana.org/time-zones). | string |
 
+### Extension Session Variables
 
+Extension session variables in Substrait are defined by the following properties:
```
yongchul: Some wordings around the names and how those values are populated is defined by the consuming system?

site/docs/expressions/specialized_record_expressions.md
```diff
@@ -69,10 +69,21 @@ Multi Value:
...
 
 
+## Session Variables
 
+Session variables are a special class of expressions whose behavior depends on the current session context. Substrait defines a set of common session variables and allows to define extension session variables.
 
+| Session Variable | Description | Return Type |
+| --------------------- | ----------- | ----------- |
+| current_date | a non-deterministic, session dependent variable containing the current date | date |
+| current_timestamp | a non-deterministic, session dependent variable containing the current timestamp in current_timezone | PRECISION_TIMESTAMP_TZ<P> |
```
yongchul: do we have a function defined to cast PRECISION_TIMESTAMP_TZ to PRECISION_TIMESTAMP? or do we want one returning PRECISION_TIMESTAMP as well?

nielspardon: I would say this one does cover such a conversion:
https://github.com/substrait-io/substrait/blob/6655d98986535f671399495589e569b9138c3aae/extensions/functions_datetime.yaml#L682-L703

and the reverse:
https://github.com/substrait-io/substrait/blob/6655d98986535f671399495589e569b9138c3aae/extensions/functions_datetime.yaml#L652-L680

site/docs/expressions/specialized_record_expressions.md
nielspardon: That's what I would assume. Nullability would be covered by the type. Session dependency is clear by the session variable category. Only determinism might be unclear.

Otherwise this is like a user defined type or extension relation for any cases where Substrait users don't want to contribute to the standard and where the semantics are then agreed upon between the producing and consuming systems.

I had the feeling names must be unique in some way. Just like we are having the URN namespacing for the extension YAMLs.

nielspardon: One thing that we may add to the dialect is an indication of what the maximum supported time/timestamp precision is for a dialect. I think this is currently missing information.

site/docs/expressions/specialized_record_expressions.md
yongchul: perfect! Thank you for due dilligence!

site/docs/expressions/specialized_record_expressions.md
yongchul: How about following?

System specific session variables not covered in Substrait can be accessed using Extension Session Variables. The evaluation of such values are system-dependent.

note: contact Substrait community to propose standard session variables.

To use extension session variables, use following properties:

> One thing that we may add to the dialect is an indication of what the maximum supported time/timestamp precision is for a dialect. I think this is currently missing information.

This is a good point. We can add `session_variables` section and add that information to be explicit. But intuitively, we can infer from the supported timestamp precision (i.e., if the system declares supporting precision 6, the maximum precision supported by current timestamp follows that).

extensions/functions_datetime.yaml
nielspardon: marking this thread as resolved since we moved away from declaring these as functions

extensions/functions_datetime.yaml
nielspardon: marking this thread as resolved since we moved away from declaring these as functions

extensions/functions_datetime.yaml
nielspardon: marking this thread as resolved since we moved away from declaring these as functions

site/docs/expressions/specialized_record_expressions.md
nielspardon: updated the docs with your proposal

nielspardon: I opened https://github.com/substrait-io/substrait/pull/961 for adding the maximum precision to the dialect

site/docs/expressions/specialized_record_expressions.md
```diff
@@ -69,10 +69,25 @@ Multi Value:
 For single value expressions, these are a compact equivalent of `expression = value1 OR expression = value2 OR .. OR expression = valueN`. When using an expression of this type, two things are required; the types of the test expression and all value expressions that are related must be of the same type. Additionally, a function signature for equality must be available for the expression type used.
 
 
+## Session Variables
 
+Session variables are a special class of expressions whose behavior depends on the current session context. Substrait defines a set of common session variables and allows to define extension session variables.
```
benbellick:
```suggestion
Session variables are a special class of expressions whose behavior depends on the current session context. Substrait defines a set of common session variables, but plan producers can also define extension session variables.
```

proto/substrait/algebra.proto
```diff
@@ -1688,6 +1689,38 @@ message Expression {
...
+      // nullability must be NULLABILITY_REQUIRED
+      Type.Date current_date = 3;
+
+      // allow to extend session variables
+      ExtensionSessionVariable extension_session_variable = 4;
+    }
+
+    // extension session variables
+    message ExtensionSessionVariable {
+      // name of the extension session variable, required
```
benbellick: Lets add a note here saying it must be unique?

text/dialect_schema.yaml
```diff
@@ -910,3 +912,27 @@ definitions:
...
+              - CURRENT_TIMEZONE
+              - CURRENT_DATE
+            - $ref: "#/definitions/extension_session_variable"
+    required: [expression]
+  extension_session_variable:
+    type: object
+    additionalProperties: false
+    properties:
+      name:
+        type: string
```
benbellick: Should the dialect also require that you specify the type of the `extension_session_variable`?

proto/substrait/algebra.proto
```diff
@@ -1688,6 +1689,38 @@ message Expression {
       Rel right = 4;
     }
   }
+
+  // Expression that represents common session dependent variables
```
```suggestion
  // Expression that represents session dependent variables
```
This technically includes uncommon ones because of the inclusion of the extension mechanism.

text/dialect_schema.yaml
nielspardon: good question, I wasn't sure on that one

proto/substrait/algebra.proto
nielspardon: makes sense

nielspardon:
> Also, [have we given any thought to how this interacts with the notion of `session_dependent` for scalar functions?](https://substrait.io/expressions/scalar_functions/#scalar-functions) Seems like they may have some overlap.

As you may have seen from the history I initially started with modeling those as session dependent scalar functions though there was feedback that these are not really functions but maybe more variables.

Any more specific concerns on the interaction with session dependent scalar functions from your side? Or just the observation that we may have both session dependent variables and session dependent functions and that it might be a bit unclear on what exactly classifies a session dependent element as a variable vs a function?

text/dialect_schema.yaml
nielspardon: do you have a preference?

proto/substrait/algebra.proto
nielspardon: added the `must be unique` that I also had in the markdown docs

text/dialect_schema.yaml
nielspardon: I guess the dialect would function as an announcement of which extension session variables are supported by a dialect so it probably should include the type

nielspardon: I added a `type` property and descriptions

benbellick:
@nielspardon yes exactly that. I must have missed the previous conversation, but some guidance on when to use one versus the other would be good I think.

nielspardon:
> @nielspardon yes exactly that. I must have missed the previous conversation, but some guidance on when to use one versus the other would be good I think.

the pivot happened due to this feedback: https://github.com/substrait-io/substrait/pull/945#discussion_r2794890929

I'm leaving it to the PMC to help with guidance

proto/substrait/algebra.proto
```diff
@@ -1688,6 +1689,38 @@ message Expression {
       Rel right = 4;
     }
   }
+
+  // Expression that represents session dependent variables
+  message SessionVariable {
+    oneof session_variable_type {
+      // a non-deterministic, session dependent variable containing the current
```
yongchul: I hate github comment system. Why you can't create a thread if a comment is on the PR? 👎 

Anyway, starting a thread on the guidance of extension session variable vs. session dependent property of a function. Thanks @benbellick bring this up.

Now I rethink about it, session dependent property is better in terms of interoperability because it is more explicit in terms of the values that may vary per system.

ExtensionSessionVariable can be an escape hatch but it is not really interoperable... Yes, we could make it more explicit by extending dialect but given that session dependent property already offers the mechanism, I don't see a compelling motivation to pursue this path until we have a good justification.

So I actually want to drop ExtensionSessionVariable from this PR. Sorry @nielspardon but I never intended to make it part of this PR anyway -- you are too eager and generous to pick it up! 😆 

Perhaps we need to setup a guideline on zero argument function though, especially those can return parameterized types.

@benbellick @nielspardon what do you think?

nielspardon: > Perhaps we need to setup a guideline on zero argument function though, especially those can return parameterized types.
> 
> @benbellick @nielspardon what do you think?

Based on my quick Google search for a definition of a `scalar function` brought up multiple definitions that essentially say that a scalar function accepts one or more input arguments. What we modeled as session variables are in some systems modeled as zero-argument functions and thus wouldn't match the definition of a scalar function.

nielspardon: > So I actually want to drop ExtensionSessionVariable from this PR. Sorry @nielspardon but I never intended to make it part of this PR anyway -- you are too eager and generous to pick it up! 😆

done. removed.

yongchul (approved):
LGTM

proto/substrait/algebra.proto
```diff
@@ -1688,6 +1689,26 @@ message Expression {
...
+      // session timezone as a string defined by IANA timezone database
+      // (https://www.iana.org/time-zones).
+      // nullability must be NULLABILITY_REQUIRED
+      Type.String current_timezone = 2;
+
+      // a non-deterministic, session dependent variable containing the current date
+      // nullability must be NULLABILITY_REQUIRED
+      Type.Date current_date = 3;
+    }
+  }
```
benbellick, line 1715: I'm not sure I understand the meaning of deterministic and non-deterministic above. Does non-deterministic just mean it could differ between different plans? I'd almost opt to just drop the usage of `deterministic` and `non-deterministic`. What do you think?

yongchul: Oh, my take is "session variables remain **constants** within a query execution session". Here, the scope of session may vary per system: the narrowest would be just execution of this plan (maybe over a micro batch of rows), or it could span multiple invocations of the plan within a batch of query plans, or a semi-permanent constants like default timezone set in the database.

So... I think we can drop deterministic, nondeterministic from here. If we want to standardize meaning of session... I don't know lol. If we absolutely want to do it, we can add another enum in SessionVariable specifying how to evaluate.

nielspardon: For me a session variable is deterministic when it returns the same value when you resolve it multiple times. When you ask for the current timestamp multiple times it obviously might differ in terms of seconds, subseconds, etc. When you evaluate the current date multiple times then you might be just crossing over into the next day between invocations. Timezone is usually rather static in comparison and should always return the same value unless you consider someone changing the session timezone in between invocation making it non-deterministic or when timezones switch into daylight savings time.

We can also drop it. When I originally modeled it as scalar functions I had to think about which flags to set.

yongchul: > For me a session variable is deterministic when it returns the same value when you resolve it multiple times.

Yes but in many systems current timestamp is usually evaluated once at the time of execution. If you have multiple instances of current timestamps in a plan, they share the same value in that execution -- the granularity could be micro batch but I don't know whether you do truly want to get a different timestamp for every single row... If this is your motivating scenario, I believe we need another concept how you want to evaluate this expression and this can be incorporated as another field in the message.

Anyway, **session** variable is probably a little too wide scope in that sense because there can be many queries executing in a session. What we (or I) wanted to capture is that those captured once (or snapshotted) and remaining as a constant for that particular execution of the plan (e.g., evaluated once at the beginning of execution).

Perhaps, runtime variable, runtime context variable, execution context variable, something along that line captures the intent better than session.

To me, `ExecutionContextVariable` seems capturing the intent the best.

WDYT?

nielspardon: I'm fine with removing the comment about determinism if that helps. 

As I was indicating earlier in the discussion I think whether the evaluation happens per plan or per record is a completely different dimension and we should find a way to express that independently of this PR. In streaming engines you may evaluate such variables like current date / current timestamp per record for e.g. streaming windowing aggregates. Maybe that should be modeled in an `ExecutionContext` data structure based on the latest name for this class of variables.

Fine with renaming the variables to `ExecutionContextVariable`.

yongchul: > whether the evaluation happens per plan or per record is a completely different dimension and we should find a way to express that independently of this PR

Agreed. Let's keep discussing about it separately.

proto/substrait/algebra.proto
```diff
@@ -1688,6 +1689,26 @@ message Expression {
       Rel right = 4;
     }
   }
+
+  // Expression that represents execution context dependent dependent variables
```
vbarua:
```suggestion
  // Expression that represents execution context dependent variables
```

vbarua (commented):
Sorry I'm late to the party, had a busy couple of weeks. Summarizing the state of this PR to make sure I understand what we're doing.

We are attempting to capture the behaviour of functions like `current_date`, `current_time` and `current_timestamp`. These functions are a somewhat special, because in some engines they return values based on the session or transaction and may not be executed as functions on every row. For example, `current_timestamp` may return the time at the start of a transaction, and not literally the current timestamp at evaluation time.

To handle them, we define a new Expression `ExecutionContextVariable` which contains pre-defined variables like:
* current_timestamp
* current_timezone
* current_date

A system executing the Substrait plan, on seeing one of these, would need to provide in the correct data based on its definition of `current`. Definitions of `current` could include:
* `current` is the start of a session/transaction.
* `current` is the time of first variable eval.
* `current` is the time at which the row is evaluated.
* `current` is defined as the time the plan is produced.

We are intentionally not defining `current`, and are leaving that as an execution concern. This does seem fairly reasonable overall, though it is the first case of use leaving something fully up to execution and not captured in the plan.

nielspardon:
> though it is the first case of use leaving something fully up to execution and not captured in the plan.

We are not leaving it undefined indefinitely. My point was I would like to introduce something like an `ExecutionContextConfig` at the plan level though this is something which will most likely require even more discussion and is probably best handled in a separate PR.

vbarua:
> We are not leaving it undefined indefinitely. 

Oh, okay. I think I may have missed something obvious in the discussion then. If the ultimate intent is to define the meaning / behaviour of this more concretely, then what is the utility of including this as is in an underspecified state? How would users know how to use this?

> I would like to introduce something like an ExecutionContextConfig at the plan level .

It sounds like you do have something in mind for this already.

> this is something which will most likely require even more discussion and is probably best handled in a separate PR

You're right that it would require more discussion, and again apologies it's highly likely I missed something while I was absent last month, but if we don't introduce the intended behaviour I'm not sure how we can expect users to implement this feature.

nielspardon:
> It sounds like you do have something in mind for this already.

What I'm thinking of is to define an `expression evaluation time` setting which can be either:
1. per statement list aka Substrait `Plan`
2. per record

One can probably identify many different other points in time. The most common ones are probably the two above.

1. would cover your typical batch/SQL engine where you submit one or multiple SQL statements at once and the execution terminates once it processed all records of the batch
2. would cover streaming engines which continuously process record by record and don't terminate

proto/buf.yaml
```diff
@@ -1,8 +1,9 @@
 version: v1
 name: buf.build/substrait/substrait
 lint:
+  allow_comment_ignores: true
```
nielspardon: explicitly allow comment lint comment ignores so we can have `PER_PLAN` evaluation as default
https://buf.build/docs/configuration/v1/buf-yaml/#allow_comment_ignores

proto/buf.yaml
```diff
@@ -1,8 +1,9 @@
...
   use:
-    - DEFAULT
+    - STANDARD
```
nielspardon, line 5: changed to `STANDARD` since we were getting this deprecation warning:

```
WARN    Category DEFAULT referenced in your buf.yaml is deprecated. It has been replaced by category STANDARD.

        The concept of a default rule has been introduced. A default rule is a rule that will be run
        if no rules are explicitly configured in your buf.yaml. Run buf config ls-lint-rules or
        buf config ls-breaking-rules to see which rules are defaults. With this introduction, having a category
        also named DEFAULT is confusing, as while it happens that all the rules in the DEFAULT category
        are also default rules, the name has become overloaded.

        As with all buf changes, this change is backwards-compatible: DEFAULT will continue to work.
        We recommend replacing DEFAULT in your buf.yaml, but no action is immediately necessary.
```

nielspardon:
I added a new `ExecutionBehavior` to the `Plan` message which allows to specify a `VariableEvaluationMode` which by default is set to `PER_PLAN` and can bet set to `PER_RECORD` for streaming systems.

proto/substrait/plan.proto
```diff
@@ -100,11 +105,29 @@ message Version {
...
+  VariableEvaluationMode variable_eval_mode = 1;
+
+  // VariableEvaluationMode specifies the frequency and scope of execution
+  // context variable evaluation.
+  enum VariableEvaluationMode {
+    // Variables are evaluated once per Substrait plan execution (default).
+    // Skipping the ENUM_ZERO_VALUE_SUFFIX linting rule so we can have
+    // PER_PLAN as default.
+    // buf:lint:ignore ENUM_ZERO_VALUE_SUFFIX
+    VARIABLE_EVALUATION_MODE_PER_PLAN = 0;
```
vbarua: We should just use here.
```proto
VARIABLE_EVALUATION_MODE_UNSPECIFIED = 0;
```

It's the recommended standard practice https://protobuf.dev/programming-guides/proto3/#enum-default, and the lint exists to enforce it.

nielspardon: I do understand this is a recommendation but does that make sense for our use case? Do we want the default to be unspecified?

vbarua: I don't think it's a good idea to stray from the recommendations in the official protobuf docs for this, or in general.

> It is also recommended that this first, default value have no semantic meaning other than “this value was unspecified”.

It is useful to be able to distinguish between and unset field, and an actual value at the protobuf level. It's also the convention we've followed for all of the other enums in the repository.

nielspardon: I removed the ignore lint config from buf.yaml, added the unspecified value, removed the declaration of `PER_PLAN` as default from documentation and added comment to `Plan` proto making the field required instead.

proto/substrait/plan.proto
```diff
@@ -62,6 +64,9 @@ message Plan {
   // parameterized types being declared have many parameters (e.g., struct with
   // many fields, nested parameterized types, string as a type parameter).
   repeated TypeAlias type_aliases = 9;
+
+  // Execution behavior settings for this plan, required.
+  ExecutionBehavior execution_behavior = 10;
```
vbarua, line 69: Does this need to be required always, or can be it be required only if an ExecutionContextVariable is present in the plan?

nielspardon: I have the feeling that it should be required always since I can't imagine anyone creating a query plan without having an execution behavior in mind and this is useful information. Changing the execution behavior based on an analysis of the plan content could be an optimization a Substrait plan optimizer could do.

Also I feel that making this conditional on the presence or absence of ExecutionContextVariables makes the spec more complicated to use.

proto/substrait/plan.proto
```diff
@@ -100,11 +105,27 @@ message Version {
...
   uint32 parameter_anchor = 1;
 
   // The literal value assigned to the parameter at runtime.
   // The type of the literal needs to match the type of the corresponding
   // DynamicParameter expression in the plan.
   Expression.Literal value = 2;
 }
+
+// ExecutionBehavior specifies the execution behavior for plan evaluation.
+message ExecutionBehavior {
```
jacques-n, line 117: who has a use case for this and the per record concept specifically? i can manufacture one but we should hold off on introducing things because someone might want them in the future.

proto/substrait/algebra.proto
```diff
@@ -1688,6 +1689,30 @@ message Expression {
...
+  // These are modeled as expressions rather than functions because they:
+  // - Take no input arguments
+  // - Depend on execution context rather than input data
+  // - Require evaluation mode control (see ExecutionBehavior in plan.proto)
+  message ExecutionContextVariable {
+    oneof execution_context_variable_type {
+      // an execution context dependent variable containing the current
+      // timestamp in current_timezone
+      // nullability must be NULLABILITY_REQUIRED
+      Type.PrecisionTimestampTZ current_timestamp = 1;
```
jacques-n, line 1703: we should fully specific precision for these things i think...

proto/substrait/plan.proto
benbellick: I think the context for this is here: https://github.com/substrait-io/substrait/pull/945#issuecomment-4010091682

However, I agree that it is hard to imagine a context in which e.g. timezone would by dynamically determined on a per row basis. 

I have started investigating if streaming engines actually do this per-row execution for current time, and [Datafusion](https://github.com/apache/datafusion/blob/ccaf8022da368052d8a1d910c28ccc8af8422e03/datafusion/functions/src/datetime/now.rs#L37) (which is one we care about at DD) does not.

nielspardon: We at IBM have this use case since we want to use Substrait with streaming engines like Apache Flink which evaluate per record.

proto/substrait/algebra.proto
nielspardon: my understanding is that the precision is encoded in the type or am I missing something?

https://github.com/substrait-io/substrait/blob/7cceb8369cd533134431979835124de5634a4a47/proto/substrait/type.proto#L205-L210

jacques-n: You're right. I misunderstood the patch upon first review. 

So this plan is unresolved. (E.g. it couldn't be split and sent to multiple nodes and get a correct value as it doesn't carry any way to express what the timestamp actually is).

As a note, I would have been much more inclined to just declare type and use pure sentinels. This creates strange composites that I'm not sure are actually a win.

proto/substrait/plan.proto
jacques-n: Is it really per record? I would assume it is per batch in many cases.

nielspardon: For a true continuously running streaming pipeline the evaluation of expressions and such variables would truly happen per record. Like when you subscribe to a Kafka topic and continuously process any incoming messages then you don't usually operate in batches since you can't control the partitioning of the incoming data into batches upfront and you usually want to keep the latency of processing individual messages as small as possible.

There is actually a section in the Apache Flink documentation explaining the behavior of `CURRENT_TIMESTAMP` in streaming execution mode:

https://nightlies.apache.org/flink/flink-docs-release-2.2/docs/dev/table/concepts/determinism/#3-determinism-in-streaming-processing

Afaik, you would only start batching if you had certain operations like streaming window aggregates where you aggregate on time intervals like calculating an average every 5 minutes

PR #949, opened 2026-01-28
docs: clarify nullability marker rules for MIRROR and DECLARED_OUTPUT

benbellick:
Closes #947

- Clarify that argument and return types must not include nullability markers when using `MIRROR` nullability handling
- Clarify that argument types must not include nullability markers when using `DECLARED_OUTPUT` nullability handling
- Add description field to `nullabilityHandling` in schema
- Fix `is_true`, `is_not_true`, `is_false`, `is_not_false` to remove erroneous nullability markers from arguments

PR #950, opened 2026-01-28
docs: clarify type equality and nullability binding for `any[\d]`

### LINKED ISSUE #941: Clarify type equality definition

### LINKED ISSUE #943: Clarify whether `any[\d]` type parameters can bind arguments with differing outermost nullability
The documentation states that for `any[\d]` types, "all any types with same numeric suffix _must_ be of the same type":
https://github.com/substrait-io/substrait/blob/95f8f350328f63bf44be63f28eb67680983edbb1/site/docs/extensions/index.md?plain=1#L125

However, the existing test case in `tests/cases/comparison/coalesce.test` includes:

https://github.com/substrait-io/substrait/blob/95f8f350328f63bf44be63f28eb67680983edbb1/tests/cases/comparison/coalesce.test#L6

Here `null::i8` is `i8?` (nullable, though it really should be marked as such and isn't) and `2::i8` is `i8` (non-nullable). Are `i8?` and `i8` different types (see #941)? If they are different, they should not bind. It is also possible that they are considered different types, but that the `any[\d]` has special binding logic.

### Resolution depends on #941

**If nullability is part of type equality (per #941), then:**
1. we must clarify that `any[\d]` type parameter binding has special logic that ignores outermost nullability in the case of `MIRROR` [nullability handling](https://substrait.io/expressions/scalar_functions/#nullability-handling). 
2. The test case is incorrect and `coalesce(i8?, i8)` should be rejected. (In this case, we may want to clarify that an explicit cast is necessary.)

**If nullability is NOT part of type equality:**
- The test case is correct as is. But docs should clarify that nullability is not considered when determining type equality.

Related: #941

benbellick:
Closes #941
Closes #943

- Add explicit type equality definition requiring all four components to match
- Clarify that MIRROR nullability ignores outermost nullability when binding `any[\d]` type parameters  
- Add concrete examples demonstrating MIRROR and DECLARED_OUTPUT nullability modes

### DIFF
--- a/site/docs/expressions/scalar_functions.md
+++ b/site/docs/expressions/scalar_functions.md
@@ -74,6 +74,39 @@ A producer may specify multiple values for an option.  If the producer does so t
 | DECLARED_OUTPUT | Input arguments are accepted of any mix of nullability. The nullability of the output function is whatever the return type expression states. Example use might be the function `is_null()` where the output is always `boolean` independent of the nullability of the input. |
 | DISCRETE        | DISCRETE nullability extends DECLARED_OUTPUT. The output nullability must still match the return type expression's nullability. Additionally, the input and arguments all define concrete nullabilities and can only be bound to the types that have those nullabilities. For example, if a type input is declared as `i64?` and one has an `i64` literal, the `i64` literal must be cast to `i64?` to allow the operation to bind. |
 
+#### Examples
+
+##### MIRROR (default)
+
+The [`add`](https://github.com/substrait-io/substrait/blob/main/extensions/functions_arithmetic.yaml#:~:text=name%3A%20%22add%22) and [`coalesce`](https://github.com/substrait-io/substrait/blob/main/extensions/functions_comparison.yaml#:~:text=name%3A%20%22coalesce%22) functions use MIRROR nullability. Under MIRROR, outermost nullability is ignored when binding type parameters. If any input is nullable, the output is nullable.
+
+```
+# add uses MIRROR (the default)
+add(i32, i32) → i32         # both non-nullable → non-nullable output
+add(i32?, i32) → i32?       # one nullable → nullable output
+add(i32?, i32?) → i32?      # both nullable → nullable output
+
+# coalesce binds any1 ignoring outermost nullability
+coalesce(i8?, i8) → i8?     # valid: i8? and i8 bind to same any1
+coalesce(list?<i32>, list<i32>) → list?<i32>  # valid: outermost nullability ignored
+
+# inner nullability still matters for type equality
+coalesce(list<i32>, list<i32?>) → ERROR  # invalid: list<i32> ≠ list<i32?>
+```
+
+##### DECLARED_OUTPUT
+
+The [`is_null`](https://github.com/substrait-io/substrait/blob/main/extensions/functions_comparison.yaml#:~:text=name%3A%20%22is_null%22) and [`is_not_distinct_from`](https://github.com/substrait-io/substrait/blob/main/extensions/functions_comparison.yaml#:~:text=name%3A%20%22is_not_distinct_from%22) functions use DECLARED_OUTPUT. The output nullability is determined solely by the return type, regardless of input nullability.
+
+```
+# is_null always returns non-nullable boolean
+is_null(i32) → boolean      # non-nullable input → non-nullable output
+is_null(i32?) → boolean     # nullable input → still non-nullable output
+
+# is_not_distinct_from compares values including nulls
+is_not_distinct_from(i32?, i32?) → boolean  # handles null comparisons, returns non-nullable
+```
+
 
 
 ### Parameterized Types
--- a/site/docs/types/type_system.md
+++ b/site/docs/types/type_system.md
@@ -35,7 +35,7 @@ The **base type** of a type is the type with all four components (type class, va
 
 For example:
 - The base type of both `i32` and `i32?` is `i32` (when considering the type class alone)
-- The base type of both `list<i32>` and `list<i32>?` is `list<i32>` (when considering the outermost nullability)
+- The base type of both `list<i32>` and `list?<i32>` is `list<i32>` (when considering the outermost nullability)
 - Two types share the same base type if they differ only in their nullability marker
 
 The notion of base type is useful in contexts where nullability should be ignored in type comparisons, such as when binding type parameters with MIRROR nullability handling.
--- a/site/docs/types/type_system.md
+++ b/site/docs/types/type_system.md
@@ -23,4 +23,8 @@ Two types are considered equal if and only if all four components match: type cl
 For example, `i32` and `i32?` are different types because they differ in nullability. Similarly, `list<i32>` and `list<i32?>` are different types because their parameters differ.
 
 !!! note
-    Some contexts explicitly ignore nullability when comparing types. For example, functions with `MIRROR` nullability handling ignore outermost nullability when binding type parameters. See [Nullability Handling](../expressions/scalar_functions.md#nullability-handling) for details.
+    Some contexts explicitly ignore nullability when comparing types:
+
+    - Functions with [MIRROR nullability](../expressions/scalar_functions.md#mirror-default) ignore outermost nullability when binding `any[\d]` type parameters
+    - Functions with [DECLARED_OUTPUT nullability](../expressions/scalar_functions.md#declared_output) accept any mix of nullability in inputs; output nullability is determined solely by return type
+    - [Variadic type parameters](../expressions/scalar_functions.md#type-parameter-resolution-in-variadic-functions) can be marked as consistent or inconsistent for type binding across variadic arguments
--- a/site/docs/types/type_system.md
+++ b/site/docs/types/type_system.md
@@ -28,3 +28,14 @@ For example, `i32` and `i32?` are different types because they differ in nullabi
     - Functions with [MIRROR nullability](../expressions/scalar_functions.md#mirror-default) ignore outermost nullability when binding `any[\d]` type parameters
     - Functions with [DECLARED_OUTPUT nullability](../expressions/scalar_functions.md#declared_output) accept any mix of nullability in inputs; output nullability is determined solely by return type
     - [Variadic type parameters](../expressions/scalar_functions.md#type-parameter-resolution-in-variadic-functions) can be marked as consistent or inconsistent for type binding across variadic arguments
+
+## Base Type
+
+The **base type** of a type is the type with all four components (type class, variation, and parameters) except nullability. In other words, it is the type modulo nullability.
+
+For example:
+- The base type of both `i32` and `i32?` is `i32` (when considering the type class alone)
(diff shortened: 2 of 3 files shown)


site/docs/expressions/scalar_functions.md
```diff
@@ -74,6 +74,39 @@ A producer may specify multiple values for an option.  If the producer does so t
...
 | DISCRETE        | DISCRETE nullability extends DECLARED_OUTPUT. The output nullability must still match the return type expression's nullability. Additionally, the input and arguments all define concrete nullabilities and can only be bound to the types that have those nullabilities. For example, if a type input is declared as `i64?` and one has an `i64` literal, the `i64` literal must be cast to `i64?` to allow the operation to bind. |
 
+#### Examples
+
+##### MIRROR (default)
+
+The [`add`](https://github.com/substrait-io/substrait/blob/main/extensions/functions_arithmetic.yaml#:~:text=name%3A%20%22add%22) and [`coalesce`](https://github.com/substrait-io/substrait/blob/main/extensions/functions_comparison.yaml#:~:text=name%3A%20%22coalesce%22) functions use MIRROR nullability. Under MIRROR, outermost nullability is ignored when binding type parameters. If any input is nullable, the output is nullable.
```
benbellick, line 81: I got this style of link from [this part of the docs](https://github.com/substrait-io/substrait/blob/f43b87625eff1cbabdb77bd23cf09e79085621a2/site/docs/tutorial/sql_to_substrait.md?plain=1#L310). Click [here](https://github.com/substrait-io/substrait/blob/main/extensions/functions_arithmetic.yaml#:~:text=name%3A%20%22add%22) to see where it takes you!

site/docs/expressions/scalar_functions.md
```diff
@@ -74,6 +74,39 @@ A producer may specify multiple values for an option.  If the producer does so t
 | DECLARED_OUTPUT | Input arguments are accepted of any mix of nullability. The nullability of the output function is whatever the return type expression states. Example use might be the function `is_null()` where the output is always `boolean` independent of the nullability of the input. |
 | DISCRETE        | DISCRETE nullability extends DECLARED_OUTPUT. The output nullability must still match the return type expression's nullability. Additionally, the input and arguments all define concrete nullabilities and can only be bound to the types that have those nullabilities. For example, if a type input is declared as `i64?` and one has an `i64` literal, the `i64` literal must be cast to `i64?` to allow the operation to bind. |
 
+#### Examples
```
benbellick, line 77: I didn't include an example of DISCRETE because there are no examples of DISCRETE functions in the standard extension set. That being said, I think it is the most obvious one. Happy to add in a made of example if that would be preferred.

site/docs/types/type_system.md
```diff
@@ -15,3 +15,12 @@ Refer to [Type Parsing](type_parsing.md) for a description of the syntax used to
...
     Substrait employs a strict type system without any coercion rules. All changes in types must be made explicit via [cast expressions](../expressions/specialized_record_expressions.md).
+
+## Type Equality
+
+Two types are considered equal if and only if all four components match: type class, nullability, variation, and parameters (compared recursively for compound types).
+
+For example, `i32` and `i32?` are different types because they differ in nullability. Similarly, `list<i32>` and `list<i32?>` are different types because their parameters differ.
+
+!!! note
+    Some contexts explicitly ignore nullability when comparing types. For example, functions with `MIRROR` nullability handling ignore outermost nullability when binding type parameters. See [Nullability Handling](../expressions/scalar_functions.md#nullability-handling) for details.
```
yongchul: Do we want to call out these in a list? mirror and declared output and variadic arguments?

yongchul (approved):

benbellick:
@yongchul I realize that both #929 and #949 make use of a notion of "base type". What do you think about adding a formal definition of base type in this PR in `site/docs/types/type_system.md` and merging this PR in first? That way, the other two PRs can reference the definition of base type here rather than defining it themselves.

yongchul:
> @yongchul I realize that both #929 and #949 make use of a notion of "base type". What do you think about adding a formal definition of base type in this PR in `site/docs/types/type_system.md` and merging this PR in first? That way, the other two PRs can reference the definition of base type here rather than defining it themselves.

Sounds good to me. `base type` is fine since we don't have type inheritance and I believe we have already using it to describe type variation and it is also the type modulo nullability. So I see clarity by calling it out. I can't think of better name than `base type` though... If others have better name, please chime in.

site/docs/types/type_system.md
```diff
@@ -15,3 +15,27 @@ Refer to [Type Parsing](type_parsing.md) for a description of the syntax used to
...
+For example, `i32` and `i32?` are different types because they differ in nullability. Similarly, `list<i32>` and `list<i32?>` are different types because their parameters differ.
+
+!!! note
+    Some contexts explicitly ignore nullability when comparing types:
+
+    - Functions with [MIRROR nullability](../expressions/scalar_functions.md#mirror-default) ignore outermost nullability when binding `any[\d]` type parameters
+    - Functions with [DECLARED_OUTPUT nullability](../expressions/scalar_functions.md#declared_output) accept any mix of nullability in inputs; output nullability is determined solely by return type
+    - [Variadic type parameters](../expressions/scalar_functions.md#type-parameter-resolution-in-variadic-functions) can be marked as consistent or inconsistent for type binding across variadic arguments
+
+## Base Type
```
benbellick, line 32: Defining this so that https://github.com/substrait-io/substrait/pull/929 and https://github.com/substrait-io/substrait/pull/949 can link to this section when referencing comparing types by their "base types".

site/docs/expressions/scalar_functions.md
vbarua: I think it's reasonable to exclude it for now.

site/docs/expressions/scalar_functions.md
vbarua, line 81: `coalesce` currently uses `MIRROR`, but it probably really shouldn't (see https://github.com/substrait-io/substrait/issues/951) so this example will become out-of-date soon. There's actually not that many functions that consume composite types that we have defined, and nullability handling for composite types is specifically where we've gotten in trouble IMO.

In functions_list.yaml right now, there are 2 functions `transform` and `filter` with MIRROR nullability. These are extra spicy because the `any1` param is both in a `list` and in a `func`. Having examples of nullability handling for those might be worthwhile. I think it would also be reasonable to _add_ a couple of common list function as part of this work to further motivate our examples here.

site/docs/types/type_system.md
```diff
@@ -15,3 +15,27 @@ Refer to [Type Parsing](type_parsing.md) for a description of the syntax used to
...
+
+The **base type** of a type is the type with all four components (type class, variation, and parameters) except nullability. In other words, it is the type modulo nullability.
+
+For example:
+- The base type of both `i32` and `i32?` is `i32` (when considering the type class alone)
+- The base type of both `list<i32>` and `list<i32>?` is `list<i32>` (when considering the outermost nullability)
```
vbarua: I assume you meant:
```suggestion
- The base type of both `list<i32>` and `list?<i32>` is `list<i32>` (when considering the outermost nullability)
```

site/docs/types/type_system.md
```diff
@@ -15,3 +15,27 @@ Refer to [Type Parsing](type_parsing.md) for a description of the syntax used to
...
+- Two types share the same base type if they differ only in their nullability marker
+
+The notion of base type is useful in contexts where nullability should be ignored in type comparisons, such as when binding type parameters with MIRROR nullability handling.
```
vbarua, line 41: Do we want `list<i32>` and `list<i32?>` to have the same base type?

benbellick: I would prefer them to be different. 

The utility of basetype is to say when arguments to functions are compatible. Do you want e.g. `coalesce(list<i32>, list<i32?>)` to be a valid invocation? 
- If yes, then we should consider them the same base type (but I think this will open up a complex can of worms) 
- If not, then I think we should consider them different.

site/docs/expressions/scalar_functions.md
benbellick: I answered in the issue, I'm not convinced we **need** to alter `coalesce`'s nullability handling (even though it would be more correct).

I'm adding an example of `transform`. Should we wait for #951 is resolved before removing the coalesce examples?

benbellick: Also, I think it is important to have an example of the form `some_func(any1...)` because it shows how nullability interacts with bindings to type parameters when the type parameter is itself the outermost thing passed into a function.

site/docs/types/type_system.md
yongchul: If we start to ignore container element nullability, it will go downhill quickly as nested types will need to recurse and align nullability. Maybe it is doable for just top-level container element (i.e., list<T> and list<T?>)... We can start with more restricted form then extend if it turns out to be an absolute pain.

Surveying other languages is something to consider as well before we go down deep... I am curious how SQL standard defines this... 😄

benbellick:
Upon further thought... I am unsure if I agree with the approach of this PR. I ultimately wanted to simplify the type comparison logic, but I now realize I may not even understand it fully myself. 

The complexity is specifically having to do with how the dropping of nullability in function arguments interacts with assignment to type parameters.

For example, consider a function declared with MIRROR as follows:
`fn(any1, any1, list<any1>)`

Obviously the following are correct invocations of the function:
`fn(i32, i32, list<i32>)`
`fn(i32?, i32?, list<i32?>)`

But what about these? 
`fn(i32, i32?, list<i32>)`
`fn(i32, i32?, list<i32?>)`

I _think_ the sensible interpretation is to say that `any1` binds to the non-nullable version here, so that the first is correct and the second is incorrect. But if that is the case, is the following valid?
`fn(i32?, i32?, list<i32>)`

-----
Anyways, I am going to switch this to a draft for now and think it over a little bit.

benbellick:
For the record, clickhouse very much considers `list<i32>` to be different from `list<i32?>`. To demonstrate this, I tested the following queries on [here](https://sql.clickhouse.com/):
```sql
-- this query works totally fine. 
SELECT arraySum([1, 2, 3]::Array(Int32)) AS sum;

-- this query is a type error
SELECT arraySum([1, 2, NULL]::Array(Nullable(Int32))) AS nullable_sum;

-- this query works and returns null
SELECT arraySum(null) AS sum;
```
So in this case, we would model (in substrait) clickhouse's `arraySum` function as taking `list<i32>` as input with nullability `MIRROR` and returning `i32`. Thus, it can accept both `list<i32>` and `list?<i32>` but not `list<i32?>`.

benbellick:
Closing this PR and moving the discussion into the associated issue.

PR #952, opened 2026-01-29
feat!: remove capabilities.proto

nielspardon:
The Substrait capabilities protobuf data structure has been replaced by the Substrait dialect in #816.

BREAKING CHANGE: removes capabilities.proto and corresponding documentation


yongchul:
I did a quick Google search `substrait Capabilities github substrait_versions function_keys` and the result does not have any meaningful references to this...

PR #953, opened 2026-01-29
feat(extensions): add unsigned integer extension types (u8, u16, u32, u64)

### LINKED ISSUE #944: Handling of unsigned numeric types
DataFusion encodes unsigned integers using `type_variation_reference` on signed types, which seemed like a [hack](https://github.com/apache/datafusion/blob/dededf7c78d587afc662c8a06fb949a0bf4d457f/datafusion/substrait/src/logical_plan/consumer/types.rs):

```
Type::I8 { type_variation_reference: 1 }  // means UInt8
```

Are unsigned integers actually a valid type variation of signed integers, or are they semantically distinct types that should be `UserDefined` or added to the spec?

## What the Spec Says

The [type variations spec](https://substrait.io/types/type_variations/) says:

> "All variations of a type are expected to have the same semantics when operated on by functions or other expressions."

The spec also defines a `Function Behavior` property ([PR #25](https://github.com/substrait-io/substrait/pull/25)):

> **INHERITS**: functions supporting the base type implicitly support this variation
> **SEPARATE**: functions must be explicitly defined for this variation

Reading through it seems like `SEPARATE` exists for variations where some operations aren't available, not where operations return different results. From the discussion, an example was: "a physical implementation function that only works for dictionary encoded strings... Both values are logically strings but only a subset of operations may be available for one physical representation."

So I don't think unsigned fits this model. The issue isn't missing operations, it's that the same operations return different results:

```
  // Same bit pattern: 0xFF (11111111)
  -1 < 0   // signed interpretation: true
  255 < 0  // unsigned interpretation: false
```

If variation `SEPARATE` is used, that would mean every arithmetic and comparison operation (<, >, +, -, *, /) has different behavior than its signed counterpart. This feels less like restricting operations and more like redefining all of them.

**Do we agree this is a problem?** This was https://github.com/substrait-io/substrait/discussions/2 when DuckDB needed unsigned integers and resorted to upcasting to signed. Now DataFusion has a different workaround (type variations), and downstream users like https://github.com/lancedb/lance/issues/2972 can't push down filters on unsigned columns because there's no standard approach.

Some other ideas that came from [substrait-io/substrait/discussions/2](https://github.com/substrait-io/substrait/discussions/2):
- A "2 out of N" reference systems criteria
- A "Two Serious Consumers" criteria
- A three class level idea for type inclusion: 3rd party extensions, 1st party extensions, and 1st party native

## Industry Support

Systems with unsigned integers:

| System     | Notes                                                                                                              |
  |------------|--------------------------------------------------------------------------------------------------------------------|
  | Arrow      | UINT8, UINT16, UINT32, UINT64 (https://arrow.apache.org/docs/cpp/api/datatype.html)                                |
  | ClickHouse | UInt8, UInt16, UInt32, UInt64, UInt128, UInt256 (https://clickhouse.com/docs/en/sql-reference/data-types/int-uint) |
  | DuckDB     | UTINYINT, USMALLINT, UINTEGER, UBIGINT, UHUGEINT (https://duckdb.org/docs/stable/sql/data_types/numeric.html)      |
  | Polars     | UInt8, UInt16, UInt32, UInt64, UInt128 (https://docs.pola.rs/user-guide/concepts/data-types-and-structures/)       |
  | MySQL      | UNSIGNED attribute on integer types (https://dev.mysql.com/doc/refman/9.6/en/numeric-type-attributes.html)         |
  | NumPy      | uint8, uint16, uint32, uint64 (https://numpy.org/doc/stable/user/basics.types.html)                                |
  | pandas     | UInt8Dtype, UInt16Dtype, UInt32Dtype, UInt64Dtype (https://pandas.pydata.org/docs/reference/arrays.html)           |
  | Parquet    | INT with isSigned=false parameter (https://parquet.apache.org/docs/file-format/types/logicaltypes/)                |

Systems without unsigned integers:

| System     | Notes                                                                                                                 |
  |------------|-----------------------------------------------------------------------------------------------------------------------|
  | Spark      | Signed only: ByteType, ShortType, IntegerType, LongType (https://spark.apache.org/docs/latest/sql-ref-datatypes.html) |
  | Trino      | Signed only: TINYINT, SMALLINT, INTEGER, BIGINT (https://trino.io/docs/current/language/types.html)                   |
  | Iceberg    | Signed only: int, long (https://iceberg.apache.org/docs/latest/schemas/)                                              |
  | PostgreSQL | Signed only: smallint, integer, bigint (https://www.postgresql.org/docs/current/datatype-numeric.html)                |
  | Snowflake  | All integer types map to signed NUMBER(38,0) (https://docs.snowflake.com/en/sql-reference/data-types-numeric)         |
  | BigQuery   | Signed only: INT64 (https://cloud.google.com/bigquery/docs/reference/standard-sql/data-types)                         |

Would including unsigned be harmful to the systems that don't support them?

I think using type variations for unsigned seems incorrect since signed and unsigned have fundamentally different semantics, but I'm happy to discuss other ideas. If there's an argument that unsigned is too harmful to the type system to accommodate DuckDB/DataFusion, maybe we can use Arrow's [`extension_types.yaml`](https://github.com/apache/arrow/blob/main/format/substrait/extension_types.yaml)

kadinrabo:
Description
---
  Adds unsigned integer types (u8, u16, u32, u64) as first-class extension types with arithmetic function support and test coverage.

  - Self-contained `unsigned_integers.yaml` with type definitions (string structure encoding) and arithmetic function overloads (add, subtract, multiply, divide, modulus, sum, min, max)
  - `functions_arithmetic.yaml` is untouched
  - Test cases in `tests/cases/arithmetic_unsigned/`, following the arithmetic_decimal convention
  - Generic udtArg grammar rule for parsing UDT literals in test cases
  - Test framework updated to scan all extension YAML files for function definitions

Closes https://github.com/substrait-io/substrait/issues/944 and follows up community agreement from [Substrait Meeting Notes on 28 Jan 2026](https://docs.google.com/document/d/1Afg93ojsVWdwo3rBO2Dtng8__RYqAy_zJMrnqhTQapc/edit?usp=sharing) that type variations are not appropriate for unsigned integers due to differing semantics.

- - -
This change is [Reviewable](https://reviewable.io/reviews/substrait-io/substrait/953)


grammar/FuncTestCaseParser.g4
benbellick, line 1: Since you are just adding these types as extension types, rather than as built in types, [you should be able to represent them as UDTs](https://github.com/substrait-io/substrait/blob/c0438dd114ff7de5f965c802c9012e677881a0b4/grammar/FuncTestCaseParser.g4#L248C5-L248C16) instead of introducing new elements to the grammar.

benbellick: Although... I'm not sure if there is a native way already to represent UDT literals, only UDT types. For that, you may have to extend the grammar. It's hard to think of a generic way to handle this for UDTs except by using the structure representation.

For example, `point`s have a structure-based representation:
https://github.com/substrait-io/substrait/blob/c0438dd114ff7de5f965c802c9012e677881a0b4/extensions/extension_types.yaml#L4-L7

and thus could maybe be represented like so in tests:
```
{latitude: 40, longitude: -74}::u!point
```

Just an idea 🤷

extensions/functions_arithmetic.yaml
```diff
@@ -1,6 +1,8 @@
 %YAML 1.2
 ---
 urn: extension:io.substrait:functions_arithmetic
+dependencies:
+  unsigned: extension:io.substrait:extension_types_numeric
```
vbarua: We will likely need to update some of the libraries to handle properly handle dependency references like this in extension files

extensions/functions_arithmetic.yaml
```diff
@@ -40,6 +42,42 @@ scalar_functions:
           overflow:
             values: [ SILENT, SATURATE, ERROR ]
         return: i64
+      - args:
+          - name: x
+            value: u!u8
```
vbarua: I think you're the first user of user-defined types with [dependency aliases](https://substrait.io/extensions/#referencing-user-defined-types).

Technically, `u!u8` would refer to a user-defined type `u8` declared within _this_ extension. The correct name would be `unsigned.u!:u8`, which uses the dependency alias you used above.

extensions/extension_types_numeric.yaml
```diff
@@ -0,0 +1,18 @@
+%YAML 1.2
+---
+urn: "extension:io.substrait:extension_types_numeric"
```
vbarua: Were you thinking that we would include other numeric type extensions in this file, hence a generic name like `extension_types_numeric`?

I would advocate for keeping the scope of these files narrow, and having many of them, instead of one big file. For example, we could use `types_unsigned_integer` for usigned integers specifically. I'm curious to hear what other folks think about narrow scopes vs large scopes.

vbarua: Were you thinking that we would include other numeric type extensions in this file, hence a generic name like `extension_types_numeric`?

I would advocate for keeping the scope of these files narrow, and having many of them, instead of one big file. So like `types_unsigned_integers`

extensions/extension_types_numeric.yaml
```diff
@@ -0,0 +1,18 @@
...
+
+# Unsigned integer literals are encoded as user defined literals.
+types:
+  - name: u8
+    description: Unsigned 8-bit integer
+    structure: {}
```
vbarua: [structure](https://substrait.io/types/type_classes/#communicating-user-defined-types) is a mechanism intended to allow for communication of type literals between systems that don't have a shared notion of it's representation. We should declare an actual encoding for this, and then re-use it for tests.

I can think of two alternative encodings we could include here: 
```yaml
    structure: {
      # value is a encoded as u8 string (i.e. '0', '42', '255')
      value: str
    }
```
or
```yaml
    structure: {
      # value is a encoded as an i8 integer that should be converted to u8 (i.e. 0, 42, -1)
      value: i8
    }
```

It would then be logical/reasonable to piggyback off the structure here to encode literals in the test case.
We can add language to the function test spec along the lines of "User-defined types with a structure containing one field `f` can be written as `<f>:u!<typename>`".

So for example given the two encodings above that would like either:
```
string encoding: subtract('255'::u!u8, '1::u!u8') = '254'::u!u8 
i8 encoding:     subtract(-1::u!u8,    1::'u!u8') = -2::u!u8 
```

vbarua (commented):
I like the inclusion of the new types in their own files + having the new functions functions_arithmetic. This will likely require some library improvements to make sure they can all handle the references to types in other files correctly, but that is already part of the spec.

Testing-wise, once we agree on an encoding for user-defined type literal, I think it would be good to add some checks for overflow behaviours

extensions/extension_types_numeric.yaml
yongchul: Now that I see this URN... can we use shorter names like `extension:io.substrait:types_numeric`? Why do we have to spell out extension twice?

I am inclined to use more specific names but I'm also worried about the growing size of preamble of the plan message as the number of these things grow... If we don't anticipate too many numeric types in the near future (say 5 to 10 years?), we are probably okay to use larger name scope?

Or, we could start with this finer grained stuff then introduce something like extension bundles as needed.

extensions/extension_types_numeric.yaml
yongchul: either signed value or just keep byte array if we want to be explicit. Either way, we need to document how to interprete the value (i.e., 2's complement or zigzag encoding like protobuf).

extensions/extension_types_numeric.yaml
yongchul: Actually, I'm wondering whether this extension type should also carry the functions related to this type -- effectively it is a self-contained module. It is odd to see extension type cripping into core functions.

extensions/functions_arithmetic.yaml
yongchul: Should this file be untouched with extension types?

tests/coverage/extensions.py
```diff
@@ -57,6 +57,11 @@ def build_type_to_short_type():
     any_type = substrait_type_str(FuncTestCaseLexer.Any)
     for i in range(1, 3):
         to_short_type[f"{any_type}{i}"] = f"{any_type}{i}"
+    # Map user-defined unsigned integer types (identity mapping)
+    to_short_type["u!u8"] = "u!u8"
+    to_short_type["u!u16"] = "u!u16"
+    to_short_type["u!u32"] = "u!u32"
+    to_short_type["u!u64"] = "u!u64"
```
yongchul: This better not be hardcoded but loaded from the extension type yaml files directly. It is fine not to do that in this PR but if we go down this path, we should introduce infrastructure to properly support these in the testing.

yongchul: perhaps, we should create an issue to track this.

extensions/extension_types_numeric.yaml
benbellick: Also even further, why separate types from functions? I think it is a bit strange in general that we have this distinction everywhere else. But we have functions in here, too. What if we just called it `extension:io.substrait:numeric`?

benbellick: Or `unsigned_integers` as @vbarua suggested is fine with me, but without the `types`.

extensions/extension_types_numeric.yaml
benbellick: Using signed value is tricky though because we technically wouldn't be able to use arbitrary unsigned values (e.g. we cannot represent $2^{8} - 1$ via i8 but we can via ui8).

I realize it isn't the most elegant, but I would honestly just opt for using the string representation. It is the simplest solution which allows all possible values, and I think we should optimize the structure representation for simplicity rather than speed, considering that they are only used to represent literals in plans, which the parsing of will almost certainly be a minimal time penalty compared to executing the actual query. 


I am willing to do one of the other representations as long as it is appropriately specified, but I do think it is really important that whatever representation we choose, we choose one which is able to represent all possible values. 

Also, if we ever "graduate" this to an offical type, we can just use the native protobuf unsigned ints :)

yongchul: @benbellick the only irk of string is that it will add up quickly and seriously bloat the overall payload if your query has a lot of unsigned integers. 😄  Regarding using signed integer, you can encode 2^8 - 1 if you specify the encoding (e.g., 2's complement  so -1). That's why I mentioned we should be clear about the encoding.

If this were the system that I am working on, I would not suggest to do string (we already suffered from the bloat). Since we don't have support for unsigned types, I practically don't care but still it would be nice to see this as some form of binary. 😆

+0.5 using string as representation.

grammar/FuncTestCaseParser.g4
kadinrabo: Sounds good, I added a generic `udtArg` rule for parsing literals

extensions/functions_arithmetic.yaml
kadinrabo: Yes that would make sense this looks like the first time we're doing this in `extensions/`

extensions/functions_arithmetic.yaml
kadinrabo: Ah, yeah I'll use `unsigned` to match.

extensions/extension_types_numeric.yaml
kadinrabo: Self contained makes sense to me, and I think that was the general consensus in the last sync too. We might just have to by hyper aware of drifting from the core extensions. For example `functions_arithmetic_decimal.yaml` is already missing `on_division_by_zero` and `on_domain_error` on its divide function. It's probably not necessary to address here, but maybe we should consider separating/generating these core extensions and other extensions separately from some kind of shared template.

re @benbellick I'm not sure why types are historically separate from functions but I think a more narrow scope might actually be better as @vbarua suggested.

How do we like `unsigned_integers.yaml` and `extension:io.substrait:unsigned_integers`?

kadinrabo: Agreed, going with narrow scope `unsigned_integers.yaml` with types/functions self-contained.

extensions/extension_types_numeric.yaml
kadinrabo: Yeah if we spec'd the encoding well, it seems we could pretty much do anything we want. Another thing though, the readability of the test cases would be pretty different:

`subtract(-1::u!u8, 1::u!u8) = -2::u!u8` vs `subtract('255'::u!u8, '1'::u!u8) = '254'::u!u8`

Both options seem valid honestly, but string is conveniently super unambiguous. So I think it might make more sense to start simpler with string and see more of the pros/cons of a compact encoding in the spec/libraries, unless there's prior examples of how we've handled this in the past

tests/cases/arithmetic_unsigned/add.test
```diff
@@ -0,0 +1,8 @@
+### SUBSTRAIT_SCALAR_TEST: v1.0
```
kadinrabo, line 1: Since the test framework only supports a single include per test file, and unsigned functions live under a different extension URI, they need their own files. (tried in https://github.com/substrait-io/substrait/pull/953/changes/a14d6d781a548e9bf05b25cb9ab4584ccff4c362)

extensions/extension_types_numeric.yaml
benbellick: I personally think `unsigned_integers` sounds great :)

tests/coverage/extensions.py
kadinrabo: Ended up removing those hardcoded entries entirely 👍

extensions/unsigned_integers.yaml
```diff
@@ -0,0 +1,364 @@
+%YAML 1.2
+---
+urn: "extension:io.substrait:unsigned_integers"
+
+types:
```
benbellick, line 5: nit: I think it would be nice to offer a very short explanation of the structure representation in the description of all of these types.

extensions/unsigned_integers.yaml
```diff
@@ -0,0 +1,364 @@
...
+  - name: u32
+    description: Unsigned 32-bit integer (0 to 4294967295)
+    structure:
+      value: str
+  - name: u64
+    description: Unsigned 64-bit integer (0 to 18446744073709551615)
+    structure:
+      value: str
+
+scalar_functions:
```
benbellick, line 31: These are a lot of functions to introduce in one PR. It may make it easier to get the PR in if you introduce an even smaller set of functions just to get the ball rolling. 

Also just wanted to say that we technically could use the standard `functions_arithmetic.yaml` and import the types from here to that file using the `ext` field. However, I think that this is actually better, so good to keep it in a separate file.

extensions/unsigned_integers.yaml
```diff
@@ -0,0 +1,364 @@
...
+    name: "add"
+    description: "Add two unsigned integer values."
+    impls:
+      - args:
+          - name: x
+            value: u!u8
+          - name: y
+            value: u!u8
+        options:
+          overflow:
```
benbellick, line 42: We can add a description field here explaining what the `overflow` field is. For example [here](https://github.com/substrait-io/substrait/blob/bf679fa94cec4d596113fba0bdbcbaa07f0ea7e3/extensions/functions_list.yaml#L71-L76). I realize you are doing what is already done for the functions in `functions_arithmetic.yaml`, but we may as well get it in with good docs the first go around 🤷

extensions/unsigned_integers.yaml
```diff
@@ -0,0 +1,364 @@
...
+          - name: y
+            value: u!u64
+        options:
+          overflow:
+            values: [ SILENT, SATURATE, ERROR ]
+        return: u!u64
+  -
+    name: "divide"
+    description: >
+      Divide x by y. Partial values are truncated (rounded towards 0).
```
benbellick: what happens if y is zero?

extensions/unsigned_integers.yaml
```diff
@@ -0,0 +1,364 @@
...
+            values: [ TRUNCATE, FLOOR ]
+          overflow:
+            values: [ SILENT, SATURATE, ERROR ]
+          on_domain_error:
+            values: [ "NULL", ERROR ]
+        return: u!u64
+
+aggregate_functions:
+  - name: "sum"
+    description: Sum a set of unsigned integer values. The sum of zero elements yields null.
```
benbellick, line 206: The decision to have the sum of zero elements return null may have come from the other instances of sum. Almost every DB I tested agrees with this (PostgreSQL, DuckDB, Spark, Trino, SQLite). [I believe the SQL standard says this as well](https://modern-sql.com/concept/null#aggregates). 

However, Clickhouse returns 0. 

```shell
docker run clickhouse/clickhouse-server:latest clickhouse-local --query "SELECT SUM(x) FROM (SELECT 1 AS x WHERE false) t"
```

Do we want to account for that in this implementation? Maybe with another option? Or do we just say that because this is a one-off, clickhouse should have its own function?

tests/cases/arithmetic_unsigned/add.test
benbellick, line 1: Is it possible to have null cases added to these tests? Might be overkill but can't hurt.

benbellick (commented):
Thanks for working on this 🙂 

Left a few more comments

extensions/unsigned_integers.yaml
yongchul: can we have this as an optional behavior?

extensions/unsigned_integers.yaml
vbarua: That would be better to handle as a separate change, because we'll want to update all functions care about overflow at once IMO. I think we should file an issue for it and pick it off later.

extensions/unsigned_integers.yaml
vbarua: We should include the divide description from
https://github.com/substrait-io/substrait/blob/2705258137b7272830cf00680e6f3a36b8d3ed4b/extensions/functions_arithmetic.yaml#L177-L186 here.

extensions/unsigned_integers.yaml
vbarua: For this PR, I think it's reasonable to retain the behaviour from https://github.com/substrait-io/substrait/blob/2705258137b7272830cf00680e6f3a36b8d3ed4b/extensions/functions_arithmetic.yaml#L1046-L1049 for consistency.

Whether that _should_ be the default behaviour, or it should be tweakable via option, is a separate question that would warrant its own issue and PR.

tests/cases/arithmetic_unsigned/sum.test
```diff
@@ -0,0 +1,8 @@
+### SUBSTRAIT_AGGREGATE_TEST: v1.0
+### SUBSTRAIT_INCLUDE: '/extensions/unsigned_integers.yaml'
+
+# basic: Basic unsigned integer examples
+sum(('0', '1', '2', '20')::u!u8) = '23'::u!u64
+sum(('100', '200', '50')::u!u16) = '350'::u!u64
+sum(('2000000', '3217908', '629000', '100000', '0', '987654')::u!u32) = '6934562'::u!u64
+sum(('1000000000000', '2000000000000', '3000000000000')::u!u64) = '6000000000000'::u!u64
```
vbarua: Do you think that it would be reasonable to include some of the overlow and null handling checks from

https://github.com/substrait-io/substrait/blob/2705258137b7272830cf00680e6f3a36b8d3ed4b/tests/cases/arithmetic/sum.test#L10-L22

tests/cases/arithmetic_unsigned/add.test
vbarua: > Since the test framework only supports a single include per test file,

I think that's a limitation worth lifting, though not in this PR. Even if it was feasible to support multiple includes, I would argue for keepings the tests under their own directory because it's easy (mentally) to map from the extension to the test directory.

vbarua (commented):
This looks good overall to me.

The one thing I would like to see is that for the test files, let's map over not just the basic tests but also the overflow, null handling, etc tests that are applicable to usigned integers. I think the floating exception tests are the only ones that don't make apply.

extensions/unsigned_integers.yaml
kadinrabo: Totally fair I just opted to include the subset of `functions_arithmetic.yaml` functions that seemed semantically applicable to unsigned

extensions/unsigned_integers.yaml
kadinrabo: I can go with something like this, thoughts?
```
Divide x by y. Partial values are truncated (i.e. rounded towards 0).
The `on_division_by_zero` option governs behavior in cases where y is 0.                                                                                                                                                       
If either x or y are out of range, behavior will be governed by `on_domain_error`.
```

tests/cases/arithmetic_unsigned/sum.test
kadinrabo: Yes, seems reasonable to match signed arithmetic test coverage. I'll add some overflow/null tests in there as well as division_by_zero tests for divide and some modulus tests.

tests/cases/arithmetic_unsigned/add.test
kadinrabo: I'd prefer to keep parity with signed tests here, which don't have null cases for add/subtract/multiply either. Open to adding them if you feel strongly, but might be better as a follow up

extensions/unsigned_integers.yaml
```diff
@@ -0,0 +1,374 @@
...
+      - args:
+          - name: x
+            value: u!u8
+          - name: y
+            value: u!u8
+        options:
+          division_type:
+            values: [ TRUNCATE, FLOOR ]
+          overflow:
+            values: [ SILENT, SATURATE, ERROR ]
```
vbarua: Can... modulus even overflow?

extensions/unsigned_integers.yaml
```diff
@@ -0,0 +1,374 @@
...
+      Calculate the remainder when dividing dividend (x) by divisor (y).
+    impls:
+      - args:
+          - name: x
+            value: u!u8
+          - name: y
+            value: u!u8
+        options:
+          division_type:
+            values: [ TRUNCATE, FLOOR ]
```
vbarua: Is `division_type` an option we need here. The signed modulus operator defines 2 tests for them:
https://github.com/substrait-io/substrait/blob/e4ce3f8710d91cb95cce010c6b59eb1d618bb2a6/tests/cases/arithmetic/modulus.test#L18-L20. I'm not sure this is applicable for unsigned ints.

tests/cases/arithmetic_unsigned/add.test
vbarua: I'm okay with not doing this for now.

extensions/unsigned_integers.yaml
```diff
@@ -0,0 +1,374 @@
...
+      If either x or y are out of range, behavior will be governed by `on_domain_error`.
+    impls:
+      - args:
+          - name: x
+            value: u!u8
+          - name: y
+            value: u!u8
+        options:
+          overflow:
+            values: [ SILENT, SATURATE, ERROR ]
```
vbarua: With signed integers, the overflow cases involve a sign change.
```
# overflow: Examples demonstrating overflow behavior
divide(-9223372036854775808::i64, -1::i64) [overflow:ERROR] = <!ERROR>
divide(-128::i8, -1::i8) [overflow:SATURATE] = 127::i8
```

With unsigned integers, this case isn't applicable, and in general overflow shouldn't be possible so we can drop this option fully for all of the impls.

extensions/unsigned_integers.yaml
```diff
@@ -0,0 +1,374 @@
...
+          on_domain_error:
+            values: [ "NULL", ERROR ]
+          on_division_by_zero:
+            values: [ "NULL", ERROR ]
+        return: u!u64
+  -
+    name: "modulus"
```
vbarua: Let's pull `modulus` out of this PR. I think its definition in `functions_arithmetic.yml` could use a once over 😓

vbarua (commented):
Left one comment on the definition of `divide`, and I think we should pull `modulus` out of this entirely because the existing definition could use a once over, but otherwise looks good to me.

The core set of functions you've added, along with the tests files, set a good example for how to add both future functions for unsigned integers, and also for adding new types.

vbarua (approved):
Changes look good to me, thanks for working on this Kadin.

Looking at this has definitely made me notice some stuff in the existing arithmetic extension that could use improvements as well 🧹

extensions/unsigned_integers.yaml
kadinrabo: Filed https://github.com/substrait-io/substrait/issues/996

extensions/unsigned_integers.yaml
kadinrabo: Filed https://github.com/substrait-io/substrait/issues/997

tests/cases/arithmetic_unsigned/add.test
kadinrabo: Filed https://github.com/substrait-io/substrait/issues/999

tests/coverage/visitor.py
```diff
@@ -428,6 +430,14 @@ def visitLambdaArg(self, ctx: FuncTestCaseParser.LambdaArgContext):
         lambda_type = ctx.funcType().getText()
         return CaseLiteral(value="lambda", type=lambda_type)
 
+    def visitUdtArg(self, ctx: FuncTestCaseParser.UdtArgContext):
+        value, _ = self.visitLiteral(ctx.literal())
+        # Type is "u!" + identifier, e.g., "u!u8"
+        type_str = "u!" + ctx.Identifier().getText().lower()
```
yongchul: does this handle YAML dependency reference or that is out of scope in the test?

yongchul (commented):
+0 using string for representation (because we go at length in defining representation in core types (e.g., decimal)) but overall looks good to me.

yongchul (approved):

tests/coverage/visitor.py
vbarua: By yaml dependency reference, are you talking about this feature for _extensions_?
https://github.com/substrait-io/substrait/blob/00bc3c24078443a23b3b4c92228fae8e2aba3778/text/simple_extensions_schema.yaml#L10-L19

I don't think the test cases support dependency references, yet
https://github.com/substrait-io/substrait/blob/00bc3c24078443a23b3b4c92228fae8e2aba3778/tests/README.md#L92-L101

vbarua:
@yongchul what does

> +0 using string for representation ...

from you mean here? Is that an approval with no vote?

jacques-n:
> +0 using string for representation (because we go at length in defining representation in core types (e.g., decimal)) but overall looks good to me.

I'm more like a -0.1 for the string representation. why not use the corresponding same sized signed types OR fixed binary types (we have those, right?)? (asking for a friend)

vbarua:
> I'm more like a -0.1 for the string representation. why not use the corresponding same sized signed types OR fixed binary types (we have those, right?)? (asking for a friend)

There's a bit of tension here between a good representation for the test cases, and a good representation for actual systems. 

For tests, I find it hard to read them when I have to convert the i8 into u8 values in my head. Like this is what a test for overflow would look like:
```
add(-1::u!u8, 1:u!u8)
```

What I want is to be able to use the string format in the tests, which makes them readable
```
add('255':u!u8, '1':u!u8)
```

and use the signed type encoding for actual systems. Right now we're piggy-backing off of the struct representation defined for the udt to define literals of udt's in the tests. I'm mulling some changes around this, either to allow for multiple struct encodings OR to define test encodings separately.

jacques-n:
> There's a bit of tension here between a good representation for the test cases...

Agree that -1 would suck for test cases. I don't have great ideas on how to resolve the mismatch. One other option I could see is to use the next integer size up for the smaller ones and decimal for the biggest one. then representation isn't ugly and always safe (e.g. a u8 always fits in a u16 so a system that doesn't understand u8 but understands u16 could work with it).

benbellick:
On this topic of test case representation, we do already abuse things a bit by having `decimal` represented via bytes:
https://github.com/substrait-io/substrait/blob/046633f76d83b6b8c29f2a67eb46d899a7889555/proto/substrait/algebra.proto#L1060-L1068
However, in test cases, we just freely use a stringified decimal representation:

https://github.com/substrait-io/substrait/blob/046633f76d83b6b8c29f2a67eb46d899a7889555/tests/cases/arithmetic_decimal/sum_decimal.test#L8

I'm leaning towards @jacques-n's suggestion now. 

How about we do:
- `u8` literals represented as `i16`
- `u16` literals represented as `i32`
- `u32` literals represented as `i64`
- `u64` literals represented as `bytes` **<-- breaks pattern**

This way, the first 3 types have simple representations in both plans and in test cases. 

Unfortunately, `u64` would be a bit of an awkward outsider here, but we could just leave a documentation blurb somewhere specifying that for UDTs with unnatural struct-based representations, we can describe custom representations in text for human readability.

jacques-n:
> On this topic of test case representation, we do already abuse things a bit by having `decimal` represented via bytes:
> 
> https://github.com/substrait-io/substrait/blob/046633f76d83b6b8c29f2a67eb46d899a7889555/proto/substrait/algebra.proto#L1060-L1068
> 
> 
> However, in test cases, we just freely use a stringified decimal representation:
> https://github.com/substrait-io/substrait/blob/046633f76d83b6b8c29f2a67eb46d899a7889555/tests/cases/arithmetic_decimal/sum_decimal.test#L8
> 
> I'm leaning towards @jacques-n's suggestion now.
> 
> How about we do:
> 
> * `u8` literals represented as `i16`
> * `u16` literals represented as `i32`
> * `u32` literals represented as `i64`
> * `u64` literals represented as `bytes` **<-- breaks pattern**
> 
> This way, the first 3 types have simple representations in both plans and in test cases.
> 
> Unfortunately, `u64` would be a bit of an awkward outsider here, but we could just leave a documentation blurb somewhere specifying that for UDTs with unnatural struct-based representations, we can describe custom representations in text for human readability.

you could use DECIMAL(20,0) for u64 I think (if I can count characters correctly).

vbarua:
I'm not a fan of the encode as the next size up scheme. It feels inelegant, breaks down at u64, and it let's people potentially set literals that are too big for the type they are sending.

I've been thinking about a bunch of compilcated things, but maybe we can just do something like:
```yaml
urn: "extension:io.substrait:unsigned_integers"

types:
  - name: u8
    description: >
      Unsigned 8-bit integer (0 to 255).
      Values are encoded as i8 to be reintepreted as u8
    structure:
      value: i8
    # Decimal string to be interpreted as u8 (i.e '0', '42', '255')
    test_encoding: str #
```

We get the nice boring `u8 is just i8 that needs be re-intepreted` for the wire format, and easy to read string format for tests.

`test_encoding` would just use a standard type string like (`i8`, `list<any1>`, etc)

If this seems reasonable, I can expand on it.

benbellick:
IMO, representing `u8` as an `i8` that needs to be reinterpreted is more confusing. Doing this means that in JSON renderings of substrait plans, I would see a number which is not the number I am expecting. At that point I would rather the number be opaque so as not to cause confusion.

vbarua:
It sounds like folks are generally okay with the concept of a string encoding for **tests**.

That leaves the encoding for values. The options thus far are:

## 1:  String
```u8 / u16 / u32 / u64 -> str```

Easy to read in plans, but not necessarily trivial to define an interpretation for. Allows for encoding larger values than are valid for the type. For example '300' for u8.

## 2: Integer Mapping
```
u8 / u16 / u32 / u64
i8 / i16 / i32 / i64
```
Relatively easy to interpret, and impossible to declare values larger than permitted. Potentially difficult to read in plans 
because something like `254:u8` would be encoded as `-2:i8`. 

## 3: Decimal
```
u8       / u16      / u32       / u64
dec<3,0> / dec<5,0> / dec<10,0> / dec<20>
```

Effectively byte encoded in plans so not easy to read. Also allows for encoding larger values than are valid. For example `300` is a valid `dec<3,0>`, but not a valid u8.

## 4: Next Size Up
```
 u8 / u16 / u32 / u64
i16 / i32 / i64 / fbin<16>
```
Two different encodings really, which is itself a bit weird to explain to users.

u8 / u16 / u32 just use the next size up integer. Easy to read, but allows for encoding invalid values. For example, `300` is a valid `i16` but not a valid u32. Commentary below applies to the u64 encoding.

## 5: Bytes
```
 u8 -> fbin<2>
u16 -> fbin<4>
u32 -> fbin<8>
u64 -> fbin<16>
```
Not the easiest to read in plans. Interpretation is mostly clear as long as we declare an endianess, which we've done before for [decimal literals](https://github.com/substrait-io/substrait/blob/dd25f77278820ede469b185399e048215b907b56/proto/substrait/algebra.proto#L1060-L1062).


## 6: Make unsigned integers first class entities

Not discussed in thread, but we could potentially just make unsigned integers part of the core type system. They're common enough 🤷 

## Personal Opinion

For the _value_ encodings used in the user-defined literal message I'm partial to either:
* Signed Integers
* Bytes

Decimal feels like a _slightly_ more constrained String, but we still need to handle invalid values, and the literal representation is just as opaque as bytes. The next size up signed integers also allows invalid values, and falls back to bytes anyways.

cc: @benbellick @yongchul @jacques-n

**Comment by jacques-n** [comment]
I think that what is really coming up here is that we have no vocabulary/specification to define a literal representation of a user defined type.  Anybody done any research on how other systems solve this?

File: tests/cases/arithmetic_unsigned/add.test
Comment (nielspardon) [reply tests/cases/arithmetic_unsigned/add.test:1]: I think we should log an issue for adding null cases if we do not include them in this PR

File: tests/coverage/extensions.py [hunk 18]
```diff
@@ -161,10 +161,16 @@ def add_functions_to_map(func_list, function_map, suffix, extension, urn):
...
+        # self-contained extension type files (e.g. unsigned_integers.yaml)
+        # with function definitions are also discovered. Files without
+        # scalar_functions/aggregate_functions/window_functions sections are
+        # safely skipped by the guards below. unknown.yaml is excluded
+        # because it defines an "unknown" type not in the type mapping.
         extensions = []
         for root, _dirs, files in os.walk(dir_path):
             for file in files:
-                if file.endswith(".yaml") and file.startswith("functions_"):
+                if file.endswith(".yaml") and file != "unknown.yaml":
```
nielspardon: `unknown.yaml` will disappear via #1081 so will need to coordinate

tests/coverage/extensions.py
```diff
@@ -161,10 +161,16 @@ def add_functions_to_map(func_list, function_map, suffix, extension, urn):
     @staticmethod
     def read_substrait_extensions(dir_path: str):
         # read files from extensions directory
+        # Scan all YAML files, not just functions_* prefixed ones, so that
+        # self-contained extension type files (e.g. unsigned_integers.yaml)
+        # with function definitions are also discovered. Files without
+        # scalar_functions/aggregate_functions/window_functions sections are
+        # safely skipped by the guards below. unknown.yaml is excluded
+        # because it defines an "unknown" type not in the type mapping.
```
benbellick:
```suggestion
```
IMO the code here is sufficiently clear without the comment.

extensions/unsigned_integers.yaml
benbellick, line 1: FYI I think in the current state of this PR, we won't auto generate docs for `unsigned_integers` ([example of generated docs for functions_boolean](https://substrait.io/extensions/functions_boolean/)).

site/docs/extensions/generate_function_docs.py
```diff
@@ -21,7 +21,7 @@ def write_markdown(file_obj: dict, file_name: str) -> None:
                 mdFile.new_line(f"{key}: {value}")
 
     for function_classification, _value in file_obj.items():
-        if function_classification == "urn":
+        if function_classification in ("urn", "dependencies"):
```
benbellick, line 24: I think this change is not strictly necessary for this PR by the way. Probably still a reasonable change though.

benbellick (approved):
Some minor comments but LGTM!

michaelmior:
Any chance of this being merged? I have a use case for Substrait that would greatly benefit from an explicit representation of unsigned integers.

tests/cases/arithmetic_unsigned/add.test
benbellick: https://github.com/substrait-io/substrait/issues/1143

extensions/unsigned_integers.yaml
benbellick: https://github.com/substrait-io/substrait/issues/1144

^ made another in the interest of getting this in.

tests/coverage/extensions.py
benbellick: i pushed a fix for it!

benbellick:
@michaelmior thanks for the ping!

@kadinrabo hope you don't mind, I opened issues for all of the things still to be addressed and pushed a small fix for some of the other comments. If there are no other problems, I'll go ahead and merge. Thanks!

vbarua (approved):

michaelmior:
@benbellick No problem! Thanks for all the work on this :)

benbellick:
3PMC approvals, merging! Thanks

PR #955, opened 2026-01-29
docs: forbid masked_reference for lambda parameters

benbellick:
[The markdown documentation already implies that lambda parameters must be referenced with `direct_reference` and `StructField`](https://github.com/substrait-io/substrait/blob/f43b87625eff1cbabdb77bd23cf09e79085621a2/site/docs/expressions/lambda_expressions.md?plain=1#L47), but the proto comments didn't explicitly specify this constraint. This PR clarifies the proto to match the documentation.

Additionally, updated the `RootReference`/`OuterReference` proto comments for consistency.

### DIFF
--- a/proto/substrait/algebra.proto
+++ b/proto/substrait/algebra.proto
@@ -1591,10 +1591,10 @@ message Expression {
     }
 
     // Singleton that expresses this FieldReference is rooted off the root
-    // incoming record type.
+    // incoming record type
     message RootReference {}
 
-    // A root reference for the outer relation's subquery.
+    // A root reference for the outer relation's subquery
     message OuterReference {
       // number of subquery boundaries to traverse up for this field's reference
       //
--- a/proto/substrait/algebra.proto
+++ b/proto/substrait/algebra.proto
@@ -1199,7 +1199,8 @@ message Expression {
     Type.Struct parameters = 1;
 
     // The lambda body expression. Lambda parameters can be referenced using FieldReference
-    // with FieldReference.LambdaParameterReference as root_type.
+    // with FieldReference.LambdaParameterReference as root_type and ReferenceSegment.StructField
+    // as direct_reference to select specific parameters.
     Expression body = 2;
   }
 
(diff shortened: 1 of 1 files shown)


proto/substrait/algebra.proto
```diff
@@ -1575,21 +1576,25 @@ message Expression {
...
+      // ReferenceSegment.StructField.
       OuterReference outer_reference = 5;
+      // Must use direct_reference and the direct_reference
+      // must be of type ReferenceSegment.StructField.
       LambdaParameterReference lambda_parameter_reference = 6;
     }
 
     // Singleton that expresses this FieldReference is rooted off the root
-    // incoming record type
+    // incoming record type.
```
```suggestion
    // incoming record type
```

proto/substrait/algebra.proto
```diff
@@ -1575,21 +1576,25 @@ message Expression {
...
     message RootReference {}
 
-    // A root reference for the outer relation's subquery
+    // A root reference for the outer relation's subquery.
```
```suggestion
    // A root reference for the outer relation's subquery
```

proto/substrait/algebra.proto
```diff
@@ -1575,13 +1576,17 @@ message Expression {
...
     oneof root_type {
       Expression expression = 3;
+      // When used with direct_reference, the direct_reference must be of type
+      // ReferenceSegment.StructField.
       RootReference root_reference = 4;
+      // When used with direct_reference, the direct_reference must be of type
+      // ReferenceSegment.StructField.
       OuterReference outer_reference = 5;
+      // Must use direct_reference and the direct_reference
+      // must be of type ReferenceSegment.StructField.
```
benbellick, line 1589: This is a completely artificial constraint. I am unsure whether or not we want to allow `masked_reference`s with `LambdaParameterReference`s. 

This particular PR was prompted by https://github.com/substrait-io/substrait-go/pull/183 which is adding lambda support to `substrait-go`. 

I am not very familiar with `MaskExpression`s. I could imagine using them for lambda if we wanted to implement something like `(a1,a2,a3,...,a1000) -> <a3,a49,a235>` but that seems like an unrealistic use case. Furthermore, we could always construct that output struct by hand. Since it makes implementation easier, I opted to forbid `MaskReference` here.

proto/substrait/algebra.proto
```diff
@@ -1575,13 +1576,17 @@ message Expression {
       MaskExpression masked_reference = 2;
     }
 
-    // The origin of the data being referenced. When this is a RootReference and
```
benbellick, line 1578: AFAICT, this comment equally applies to `OuterReference`, so I opted to allow each field in the oneof have its own comment.

proto/substrait/algebra.proto
tokoko: Instead of forbidding it in the spec, wouldn't it be better to simply call it out as unimplemented in substrait-go?

yongchul (approved):

proto/substrait/algebra.proto
yongchul: Actually, calling out in substrait-go is a good option although I am fine with this change as I am also not familiar with mask expression and we can relax this later as well.

Does anyone know systems using MaskExpression?

benbellick: @tokoko I am okay with allowing it in the spec provided we can think of a legitimate use case for it. Otherwise, I would prefer to be defensive. If we allow, then we may have to answer questions in the future about what it means :)

benbellick: Also, as linked in the PR description, the markdown docs already suggest that `StructField` is the appropriate way to make this reference.

tokoko: I did some google search and the closest thing I found in the engines was duckdb allowing EXCLUDE inside nested expressions.. my search wasn't exactly thorough though. I guess my point is that whatever reasoning we use for disallowing masked_reference for lambda_parameters specifically could just as easily be used for disallowing (removing) masked_references from substrait entirely.

Having said that, I'm fine with going ahead here if it makes anyone's life easier on the implementation side. We can always come back if it turns out to be an issue.

benbellick:
Closing this issue. Instead, I opened up #956 where we can continue the discussion. I am going to take the position that the use of `masked_reference` will blow up in `substrait-go` for now.

PR #958, opened 2026-02-05
fix: add missing visit methods in TestCaseVisitor

nielspardon:
While working on #945 I realized that the extension test coverage code did not handle some of the newer types correctly.

full disclosure: I used claude code to figure out the required changes for Python test coverage changes but reviewed the changes and they made sense to me.


vbarua (approved):
Changes looks reasonable and match the existing pattern. This change does not change the core specification in any way, it's updating the tooling used to read test case files. I will merge this with just my approval.

PR #960, opened 2026-02-18
feat: clarify interaction between nullability binding and any type parameters

### LINKED ISSUE #943: Clarify whether `any[\d]` type parameters can bind arguments with differing outermost nullability

benbellick:
- Document how nullability mode (MIRROR, DECLARED_OUTPUT, DISCRETE) interacts with `any[\d]` type parameter binding
- Add concrete examples showing outermost vs inner nullability, compound types, cross-structural binding, and DISCRETE matching

Closes #943

- - -
This change is [Reviewable](https://reviewable.io/reviews/substrait-io/substrait/960)


benbellick:
I am aware of the complexity added here. If someone can think of a better way to manage this complexity, that would be much preferred. The fundamental problem is dealing with how `nullabilityHandling` interacts with binding to any type parameters. https://github.com/substrait-io/substrait/pull/949 clarified that with nullability handling `MIRROR` and `DECLARED_OUTPUT`, the formal parameters must not have nullability markers. However, nothing stops us from passing in arguments that contain nullability arguments. 

This PR is a stab at addressing this interaction. 

There is also the case of `DISCRETE` which is accounted for here. The approach is to just say that the outermost nullability must be declared explicitly and must match when invoked.

Slimsammylim (approved):
The explanations and examples here are very clear and make sense to me. My only nit is I prefer if the all the reason boxes are filled, but that's just because I like when explanations are obvious. Either way is good!

benbellick:
@Slimsammylim I added them, good suggestion!

yongchul (approved):

site/docs/expressions/scalar_functions.md
```diff
@@ -74,13 +74,82 @@ A producer may specify multiple values for an option.  If the producer does so t
...
 
 [`add`](https://github.com/substrait-io/substrait/blob/main/extensions/functions_arithmetic.yaml#:~:text=%2D-,name%3A%20%22add%22,-description%3A%20%22Add%20two) is declared as `add(i32, i32) -> i32` with `MIRROR` nullability. `add(i32?, i32)`, `add(i32, i32?)`, and `add(i32?, i32?)` all return `i32?` because at least one argument is nullable, but `add(i32, i32)` returns `i32` because all arguments are non-nullable.
 
 [`is_null`](https://github.com/substrait-io/substrait/blob/main/extensions/functions_comparison.yaml#:~:text=%2D-,name%3A%20%22is_null%22,-description%3A%20Whether) is declared as `is_null(i64) -> boolean` with `DECLARED_OUTPUT` nullability. Both `is_null(i64)` and `is_null(i64?)` return `boolean` because the output type is determined solely by the declared return type regardless of input nullability.
 
+#### `any` Type Binding
+
+##### MIRROR
+
+Given `f(any1, any1) -> any1` with `MIRROR` nullability:
```
benbellick: Per discussion with @jacques-n, I'm going to look into collapsing all of these tables into one by additionally adding a `Definition` column and a `Nullability Handling` column.

vbarua (approved):
These clarifications look good to me. I suspect we'll run continue to run into more weird edge cases as we try and implement this across systems, but this is a great starting point.

PR #961, opened 2026-02-20
feat(dialect): support specifying maximum supported subsecond precision

nielspardon:
While working on #945 I realized it would be useful tom specify the maximum supported subsecond precision in the Substrait dialect since different engines typically supported different levels of precision.

This PR adds a `max_precision` field for the precision time type definitions in the dialect schema.

### DIFF
--- a/text/dialect_schema.yaml
+++ b/text/dialect_schema.yaml
@@ -398,7 +398,7 @@ definitions:
   type_precision_time:
     type: object
     additionalProperties: false
-    required: [type]
+    required: [type, max_precision]
     properties:
       type:
         const: PRECISION_TIME
@@ -417,7 +417,7 @@ definitions:
   type_precision_timestamp:
     type: object
     additionalProperties: false
-    required: [type]
+    required: [type, max_precision]
     properties:
       type:
         const: PRECISION_TIMESTAMP
@@ -436,7 +436,7 @@ definitions:
   type_precision_timestamp_tz:
     type: object
     additionalProperties: false
-    required: [type]
+    required: [type, max_precision]
     properties:
       type:
         const: PRECISION_TIMESTAMP_TZ
@@ -473,7 +473,7 @@ definitions:
   type_interval_compound:
     type: object
     additionalProperties: false
-    required: [type]
+    required: [type, max_precision]
     properties:
       type:
         const: INTERVAL_COMPOUND
@@ -492,7 +492,7 @@ definitions:
   type_interval_day:
     type: object
     additionalProperties: false
-    required: [type]
+    required: [type, max_precision]
     properties:
       type:
         const: INTERVAL_DAY


text/dialect_schema.yaml
```diff
@@ -458,6 +474,10 @@ definitions:
         const: INTERVAL_DAY
       system_metadata:
         $ref: "#/definitions/system_type_metadata"
+      max_precision:
+        type: integer
+        minimum: 0
+        maximum: 12
```
yongchul: documentation says max precision of interval_day is 9.

text/dialect_schema.yaml
```diff
@@ -449,6 +461,10 @@ definitions:
         const: INTERVAL_COMPOUND
       system_metadata:
         $ref: "#/definitions/system_type_metadata"
+      max_precision:
```
yongchul, line 482: Overall +1 to this change.

Some follow up tasks.

1. I see documentation mentions INTERVAL_MONTH but looks like it is not there. Either define it or remove reference from the documentation.
2. I can't comprehend INTERVAL_COMPOUND documentation.
3. If we do this, we should do the same for other parameterized types where it makes sense (e.g., decimal).

text/dialect_schema.yaml
```diff
@@ -458,6 +474,10 @@ definitions:
         const: INTERVAL_DAY
       system_metadata:
         $ref: "#/definitions/system_type_metadata"
+      max_precision:
```
yongchul, line 501: I'm still noob in YAML... is this `max_precision` required or optional?

nielspardon: This is a _YAML schema_ which is built off of [JSON schema](https://json-schema.org) and it describes the schema that a YAML file needs to adhere to. The `max_precision` section here says that a dialect YAML file is allowed to have a field called `max_precision` which is of type `integer` where the `minimum` allowed value is `0` and the `maximum` allowed value is `12` (which should actually be `9` based on your other comment). For reference:

- https://json-schema.org/understanding-json-schema/reference/numeric#integer
- https://json-schema.org/understanding-json-schema/reference/numeric#range

text/dialect_schema.yaml
nielspardon: Thanks, my bad. I intuitively thought that all types with precision time have the same max precision. Fixed it both in the dialect schema and in the test case.

text/dialect_schema.yaml
nielspardon: > 1. I see documentation mentions INTERVAL_MONTH but looks like it is not there. Either define it or remove reference from the documentation.

feels like a typo to me and this should have been mentioning `INTERVAL_YEAR` since this was introduced in https://github.com/substrait-io/substrait/pull/665 where the idea was to combine `IntervalYearToMonth` and `IntervalDayToSecond` into `IntervalCompound`:
https://github.com/substrait-io/substrait/blob/23ac2f30e03a8d28f76e445bcd439064f3d688fd/proto/substrait/algebra.proto#L1017-L1018

https://github.com/substrait-io/substrait/blob/23ac2f30e03a8d28f76e445bcd439064f3d688fd/proto/substrait/algebra.proto#L1112-L1115

nielspardon: > 3\. If we do this, we should do the same for other parameterized types where it makes sense (e.g., decimal).

Agreed, I had a similar thought that we should also do this for e.g. decimal precision or varchar length, etc

Created https://github.com/substrait-io/substrait/issues/965 for this

nielspardon: created https://github.com/substrait-io/substrait/pull/962 to fix the typo

text/dialect_schema.yaml
```diff
@@ -404,6 +404,10 @@ definitions:
         const: PRECISION_TIME
       system_metadata:
         $ref: "#/definitions/system_type_metadata"
+      max_precision:
```
benbellick, line 407: This change overall seems reasonable to me. Perhaps it is overkill, but I'm wondering if it is worth adding a `description` field (like [here](https://github.com/substrait-io/substrait/blob/23ac2f30e03a8d28f76e445bcd439064f3d688fd/text/simple_extensions_schema.yaml#L178)) to all of these fields? 

I know that the meaning might seem really clear from the name alone, but I think it can help make the spec crystal clear to explicitly spell out what that field means in all of these places. 

This is especially true considering we don't have any `md` files for defining dialects, only the yaml schema.

nielspardon: > This is especially true considering we don't have any `md` files for defining dialects, only the yaml schema.

This I think we should definitely change and add a section into the documentation on the website for dialects.

benbellick: I agree with that as well, but also don't think it is a bad thing to have redundant documentation in both places. One reason is that users can use things like lsp support for YAML and see what these fields mean on hover.

nielspardon: created https://github.com/substrait-io/substrait/issues/964 for adding Substrait dialect documentation

nielspardon: I added descriptions in ad9857882df1d85cf58dee980187d55b29c4bfc8

benbellick: Descriptions are awesome! Thanks for adding 🚀

nielspardon: > I agree with that as well, but also don't think it is a bad thing to have redundant documentation in both places. One reason is that users can use things like lsp support for YAML and see what these fields mean on hover.

you were a little too quick to respond or I've been too slow in pushing the new commit :-)

nielspardon: updated https://github.com/substrait-io/substrait/issues/964 to also include updating the YAML schema with description fields

text/dialect_schema.yaml
yongchul: Thank you for explanation but my question is more of "what happens if `max_precision` is missing?" Does YAML parser have a mode to enforce this (e.g., parse error because required field is missing) or we assume something if it is not present (e.g., assume max precision is supported)?

text/dialect_schema.yaml
nielspardon: we can list `max_precision` under `required` above to make it mandatory

text/dialect_schema.yaml
vbarua: > Does YAML parser have a mode to enforce this (e.g., parse error because required field is missing) or we assume something if it is not present (e.g., assume max precision is supported)?

Parsers generated from the schema will fail to parse if the field is missing.

PR #969, opened 2026-02-20
feat: add common list functions to core extensions

### LINKED ISSUE #974: Add `cardinality`, `sort`, `any_match`, and `all_match` to `functions_list.yaml`
These four functions are primitive and display consistent behavior in execution engines. Thus, they are appropriate to upstream into the core repository. 

In the comments, I will outline my analaysis.

benbellick:
Adds 4 commonly-used list functions to `extensions/functions_list.yaml`, based on research of PostgreSQL, DuckDB, Spark SQL, Trino, and DataFusion array operations.

These are: 
- `sort`
- `length`
- `any_match`
- `all_match`

See #974 for detailed cross-engine analysis and verification scripts.

Closes #974

-----


extensions/functions_list.yaml
```diff
@@ -41,3 +41,83 @@ scalar_functions:
...
+
+      When called with a comparator function, the comparator receives two elements
+      and must return a negative integer if the first is less than the second,
+      zero if they are equal, or a positive integer if the first is greater.
+      If the comparator returns null, the behavior is undefined.
+
+      If the input list is null, the result is null.
+
+      Note: Test cases for the comparator overload require a compare function
+      that returns i32, which is not currently available in the extensions.
```
benbellick:
```suggestion
```
This note doesn't belong here.

extensions/functions_list.yaml
```diff
@@ -12,6 +12,8 @@ scalar_functions:
 
       The transformer receives one parameter (the current element) and must return
       the transformed value.
+
+      If the input list is null, the result is null.
```
yongchul, line 16: Isn't this implied by MIRROR?

extensions/functions_list.yaml
```diff
@@ -41,3 +45,75 @@ scalar_functions:
...
+      - args:
+          - name: input
+            value: list<any1>
+        nullability: MIRROR
+        return: i64
+
+  # TODO: Add comparator overload for custom sorting (see #972)
+  - name: "sort"
+    description: >-
+      Sorts the elements of a list. Elements must be orderable.
```
yongchul, line 64: what is orderable? we didn't define it in Substrait right?

extensions/functions_list.yaml
```diff
@@ -41,3 +45,75 @@ scalar_functions:
...
+      Returns false if all elements return false.
+
+      For an empty list, returns false.
+      If the input list or predicate is null, the result is null.
+    impls:
+      - args:
+          - name: input
+            value: list<any1>
+          - name: predicate
+            value: func<any1 -> boolean>
```
yongchul: Just to make sure, we don't care about the nullability here right? especially the return... because it can be null.

tests/cases/list/all_match.test
```diff
@@ -0,0 +1,18 @@
...
+
+# returns_false: At least one element fails the predicate
+all_match([1, 2, 3]::list<i32>, (x -> gt(x, 2::i32))::func<i32 -> bool>) = false::bool
+all_match([1, 2, 3]::list<i32>, (x -> lt(x, 3::i32))::func<i32 -> bool>) = false::bool
+
+# empty: Empty list returns true (vacuously true)
+all_match([]::list<i32>, (x -> gt(x, 0::i32))::func<i32 -> bool>) = true::bool
+
+# three_valued_logic: Null predicate results follow three-valued logic
+all_match([1, Null, 2]::list<i32?>, (x -> gt(x, 0::i32))::func<i32? -> bool>) = Null::bool
```
yongchul: noob question. where are the implementations of these new functions? it somehow magically appeared in this PR?

extensions/functions_list.yaml
benbellick: I wasn't sure about that. Is that guaranteed to be the case? My understanding is that `MIRROR` ensures that the output is _nullable_ (rather than _null_) if one of the inputs is nullable. What about a silly function like this? 

```yaml
- name: "second"
  description: >-
    Returns the second argument, ignoring the first.
  impls:
    - args:
        - name: ignored
          value: any1
        - name: result
          value: any2
      nullability: MIRROR
      return: any2
```

In this case, we would have `second(NULL, 10)` returns `10` as a nullable int. Or would you say that that is incorrect? 

If `MIRROR` implies what you are saying, then I think the docs may need to be updated.

tests/cases/list/all_match.test
benbellick: https://github.com/substrait-io/substrait/issues/974#issuecomment-3946386678

benbellick: I didn't want to clutter up the PR description too much so I put all of the tables I made to analyze these functions in the issue. They have links to the functions in the actual db engines. 

This specific one is here though: 
https://github.com/substrait-io/substrait/pull/969/files#diff-277c58aaedd7a56206398efb068698b5d1c2786b75674beac0f08938727adf17R101-R119

extensions/functions_list.yaml
benbellick: No official definition of orderable. I can leave it out if you'd like. My thinking was it is equally as defined as the notion of orderable for `SortRel` 😅 

https://substrait.io/relations/basics/#orderedness

extensions/functions_list.yaml
benbellick: The way I thought about it was that if a value in the list is null, then the lambda wouldn't even get called and the null gets propagated. But now that actually seems like the wrong decision. Probably the right decision is to have it return `boolean?`. Is that what you mean? I will push a change for that now.

extensions/functions_list.yaml
yongchul: Hmm... okay, after re-reading the doc, MIRROR does not say anything about the behavior. My brain is wired to many SQL functions where if any of arguments are NULL, then the result is NULL so when I saw MIRROR, I interpreted this also implies the behavior.

Okay, given that the documentation says not the behavior, clarification is not bad but we'd better do it consistently across all function definitions though. Shall we create a follow up issue on this?

extensions/functions_list.yaml
yongchul: oh well, that is also underdefined 😆 we will need to define it somehow. not necessary for this PR.

extensions/functions_list.yaml
yongchul: Hmm... pre-filtering NULLs does not sound a great idea to me.

Can we have two options or alternative here?

1. if implementer specify any1 -> boolean, the return value can not be null.
2. if implementer specify any1 -> boolean?, the return value can be null. This is more aligned to AND, OR.

That said, the nullability of output does not depend on the list element type but the return type of predicate. What do you think?

benbellick: I don't think what you are suggesting, having two different impls, is possible. The reason is [this limitation](https://substrait.io/extensions/#uniqueness-constraint):
> A function signature uniquely identifies a function implementation within a single YAML file. As such, every function implementation within a YAML must have a distinct function signature in order for references to the implementation to remain unambiguous. A YAML file in which this is not the case is invalid.

Both of the above would have a signature of `any_match:list_func`, so there is a collision. We could instead have two different functions with different names entirely, but is there any loss of precision by just doing the following? 

```yaml
  - name: "any_match"
    description: >-
      Returns true if any element of the list matches the predicate.

      The predicate function is applied to each element using three-valued logic.
      Returns true if the predicate returns true for at least one element.
      Returns null if no element returns true but at least one returns null.
      Returns false if all elements return false.

      For an empty list, returns false.
      If the input list or predicate is null, the result is null.
    impls:
      - args:
          - name: input
            value: list<any1>
          - name: predicate
            value: func<any1 -> boolean?>
        nullability: MIRROR
        return: boolean
```

^ This is the current state of the file in the PR at the time of writing this comment.

yongchul: duh... signature again... 😭 it is too restricted to have functor type as just func. That means we will have to vary the name per functor type.

The problem is that `MIRROR` is not well-defined here (maybe it is well defined if it is only about the `list` or `predicate` being NULL but still have problems). If it applies to `list` or the element type or both, it will have problem in nullability.

The only safe thing to do is always return `boolean?` with `DECLARED_OUTPUT`.

benbellick: Ah yes I see what you mean now. `any_match(ARRAY[NULL], x -> x > 0)` will return `NULL` but per the nullability handling, it should return a non-null argument. So yes, `DECLARED_OUTPUT` is the only sensible approach. I will fix this now (for `all_match` as well). Thanks!

benbellick: Updated to `DECLARED_OUTPUT`. However, this PR now has me wondering if the previously existing `filter` function is correct, as the lambda there is `func<any1 -> boolean>`. Perhaps it also needs to be `func<any1 -> boolean?>`?

yongchul: Yes, the predicate may return null. So `func<any1 -> boolean?>`

benbellick: I updated it. This technically may be a breaking change, though it is so small that I am inclined to not mark it as such 🤷

yongchul (approved):

Broshen (approved):

extensions/functions_list.yaml
```diff
@@ -33,11 +35,84 @@ scalar_functions:
...
       result. Elements for which the predicate returns false or null are excluded.
+
+      If the input list is null, the result is null.
     impls:
       - args:
           - name: input
             value: list<any1>
           - name: predicate
-            value: func<any1 -> boolean>
+            value: func<any1 -> boolean?>
```
vbarua, line 45: For my own elucidation, this is to indicate that the predicate function may return nulls?

extensions/functions_list.yaml
```diff
@@ -33,11 +35,84 @@ scalar_functions:
...
+    impls:
+      - args:
+          - name: input
+            value: list<any1>
+        options:
+          direction:
+            description: >-
+              Sort direction and null placement, matching SortField.SortDirection
+              in the Sort relation.
+            values: [ASC_NULLS_FIRST, ASC_NULLS_LAST, DESC_NULLS_FIRST, DESC_NULLS_LAST]
```
vbarua, line 76: I _could_ make an argument that this should be declared as an `enum` to force producers to declare the behaviour they want. Options don't have to be set, so it becomes more likely that we end up with sorts calls with _no_ direction set. Saying this though, I realize that we don't have clear guidance on when an enum argument is preferred over an option.

That discussion is probably worth having, but it doesn't need to block this PR. In the worst case, we add a `sort` impl with an enum argument (which would have a different function signature) and deprecate the options-only one.

vbarua (commented):
The definitions and test for `sort` and `cardinality` make sense to me. I need to look at `all_match` and `any_match` with fresh eyes because the trinary logic breaks my brain.

extensions/functions_list.yaml
yongchul: Yes. See https://github.com/substrait-io/substrait/pull/969#discussion_r2854491714 thread

tests/cases/list/all_match.test
```diff
@@ -0,0 +1,18 @@
+### SUBSTRAIT_SCALAR_TEST: v1.0
+### SUBSTRAIT_INCLUDE: '/extensions/functions_list.yaml'
+### SUBSTRAIT_DEPENDENCY: '/extensions/functions_comparison.yaml'
+
```
vbarua, line 4: Can we add a test for a null input?

tests/cases/list/any_match.test
```diff
@@ -0,0 +1,18 @@
+### SUBSTRAIT_SCALAR_TEST: v1.0
+### SUBSTRAIT_INCLUDE: '/extensions/functions_list.yaml'
+### SUBSTRAIT_DEPENDENCY: '/extensions/functions_comparison.yaml'
+
```
vbarua, line 4: Can we add a test for a null input?

vbarua (approved):
Changes LGTM overall.

Did leave one comment for an additional test case, but consider it non-blocking.

tests/cases/list/all_match.test
benbellick: I would love to, but we don't have support for null list literals in these tests yet. Adding it in this PR here: https://github.com/substrait-io/substrait/pull/991 🙂 

How about I add test cases for these functions as part of that PR?

tests/cases/list/any_match.test
benbellick: https://github.com/substrait-io/substrait/pull/969#discussion_r2892191063

extensions/functions_list.yaml
benbellick: Hmm I wasn't even aware of enums to be honest 😅 I don't think there are any functions using them at the moment. Probably would be good to open up an issue about it.

benbellick: #995

tests/cases/list/all_match.test
vbarua: Yep, let's do that. Great first test of that feature 🙂

benbellick: added in that PR!

PR #973, opened 2026-02-21
feat: introduce LateralJoinRel for a correlated subquery evaluation

yongchul:
# Background

SQL or relational query plan may have subqueries. Some subqueries can be easily decorrelated in terms of joins but in other cases it is non-trivial to do such rewriting. Some queries can be de-correlated (i.e., rewritten as a join) but sometimes it is non-trivial to decorrelate complex subqueries.

ApplyRel is one of the way to model a generic correlated subquery execution and implemented in multiple systems (e.g., Google Spanner, Oracle, SQL Server family). 

The semantic of apply is that the subquery is executed per row from the input rather than operating on set of the rows at a time like join. The subquery can reference columns in the input row as outer references (thus each subquery is *correlated* to each input row).

Apply operation is inherently expensive thus typically used in a special context such as index lookup join or in a pinch where a system couldn't decorrelate the operation.

# Proposal

~~Extend JoinRel with `lateral` flag to represent lateral join.~~
`LATERAL` keyword was introduced in SQL 99, as part of table reference. When the table source is wrapped with LATERAL, the subquery *MAY* reference columns and tables preceding the LATERAL table reference, and allows correlated subquery.

~~By this nature of scoping of tables and references, we propose to extend `JoinRel` with lateral flag to denote lateral join, and let right input as the correlated subquery, which can reference the fields from left input.~~

~~In this way, we reuse all the join semantics automatically without introducing a dedicated Rel.~~

~~Same change was made to CrossRel.~~

We decided to introduce a new LateralJoinRel for clarity and to ease handling the new semantic in existing consumers.

The change is backward compatible.

This PR depends on #1031 .

# AI disclaimer

This PR was assisted by Claude Opus 4.6.

- - -
This change is [Reviewable](https://reviewable.io/reviews/substrait-io/substrait/973)


proto/substrait/algebra.proto
```diff
@@ -265,6 +265,20 @@ message JoinRel {
 
   JoinType type = 6;
 
+  // When true, the right input is evaluated once per row of the left input
+  // (lateral join / correlated subquery). The right input may reference fields
+  // from the current left row using FieldReference with OuterReference as the
+  // root_type and steps_out = 1.
```
vbarua: How does this work if there are also proper subqueries in the right input? I don't think that reference to `steps_out = 1` in the RHS will always point to the join.

vbarua (commented):
I'm not super familiar with LATERAL in SQL, but if I was to rephrase the problem we're trying to solve it's that to express LATERAL-style joins in a JOIN like:
```
  JoinRel
  /     \
 LHS    RHS
```
we need to be able to reference LHS outputs in the RHS (or vice-versa).

This type capability has come up before for me in the context of dynamic filtering, where we need to be able to reference fields from other join inputs in the
https://github.com/substrait-io/substrait/blob/dc276549da20d92aa83269219ae267682d55dba4/proto/substrait/algebra.proto#L109 
field to express a (potential) data dependency.

I'm a little wary of re-using Outer References like you've suggested, because as I understand it nothing about the construction were talking prevents from putting actual subqueries in the subtree between the Join relation and where the field is referenced.

To avoid this ambiguity, would it make sense to have a specialized reference type like JoinReference in the same vein as OuterReference and LambdaParameterReference?

yongchul:
> I'm not super familiar with LATERAL in SQL, but if I was to rephrase the problem we're trying to solve it's that to express LATERAL-style joins in a JOIN like:
> 
> ```
>   JoinRel
>   /     \
>  LHS    RHS
> ```
> 
> we need to be able to reference LHS outputs in the RHS (or vice-versa).

No, not vice-versa. The correlation is typically one direction (from left to right) due to scoping.
 
> This type capability has come up before for me in the context of dynamic filtering, where we need to be able to reference fields from other join inputs in the
> 
> https://github.com/substrait-io/substrait/blob/dc276549da20d92aa83269219ae267682d55dba4/proto/substrait/algebra.proto#L109
> 
> field to express a (potential) data dependency.
>
> I'm a little wary of re-using Outer References like you've suggested, because as I understand it nothing about the construction were talking prevents from putting actual subqueries in the subtree between the Join relation and where the field is referenced.

I wish we just ditch the current outer reference honestly and do it differently -- more like global id reference to avoid hair-pulling wiring outer references in the other PR. I don't understand "putting actual subqueries in the subtree join relation"... could you elaborate your concern? Or, it is actually better to do a quick call if you have time.

Anyhow, each subquery would put the "depth" -- only in the context of set predicate today-- and this PR proposing lateral join will put another depth in the nested stack of schema. So I don't see any problem...

> To avoid this ambiguity, would it make sense to have a specialized reference type like JoinReference in the same vein as OuterReference and LambdaParameterReference?

I don't know. OuterReference is OuterReference, you are referencing something outside of your usual column resolution scope. LambdaParameterReference is making sense because there is a clear binding point. OuterReference is generic, and based on the "correlation" boundary -- either the Rel containing set predicate or the left input Rel of lateral join in this PR.

If you prefer more explicit like LambdaParameterReference, we could explore id based declaration, and binding, something like

```
JoinRel {
  repeated CorrelatedFields correlated_left_fields;

  message CorrelatedField
  {
      int anchor = 1;  // plan level unique. or we can give an id to this joinrel, then joinrel id with field reference identify unique reference.
      FieldReference correlated_field = 2;    // correlated field
  }
}

message CorrelatedFieldReference
{
    int correlated_reference = 1;
}
```

Basically, the idea is for outer references, going with more like name or identifier based approach rather than offset based approach. Yes, it is departure from the basic referencing scheme but this is much simpler to reuse expressions and binding the correct columns/fields without going through the hassles of massaging offsets and depth, which depends on context.

I can draft this as well and we can see how it look like.

proto/substrait/algebra.proto
yongchul: If there is a subquery, it will need to use steps_out = 2 or N appropriately depending on its nesting level. Yes, 1 only works when it is not under the nesting.

```
   JoinRel
   /.        \
LHS.    FilterRel
                 |
               AND ( {outerRef {steps_out=1, 0} == fieldRef[1], // outerference to LHS column
                  EXISTS (fieldRef[2], subquery {
                       ... Filter {fieldRef[7] == outerRef {steps_out=2, 2} ) ) // outerference to LHS column. steps out is 2 now
```

The language should be updated. Good catch!

yongchul: Just removed using `steps_out` and preferring the id based approach (#1031) for the lateral join. Let me know what do you think. We can add the description back how to use it with steps_out.

site/examples/proto-textformat/field_reference/outer_reference_lateral_join.textproto
```diff
@@ -0,0 +1,37 @@
...
+#     WHERE orders.customer_id = c.customer_id   -- outer reference
+#   ) o ON true
+#
+# Plan structure:
+#
+#   JoinRel [lateral=true, rel_anchor=5, type=LEFT]
+#     ├── left:  ReadRel(customers)          -- output: (customer_id, name, ...)
+#     └── right: FilterRel
+#                  condition: orders.customer_id = outer_ref(rel_reference=5, field 0)
+#                  └── ReadRel(orders)
```
vbarua, line 23: A bit of tricky-ness is that I'm not sure if the `JoinRel` is the right place for the rel anchor. 

As I understand the various references type that that point to other relations, they reference the **output** of the relation. I think the reference needs to be on the lhs input for this to make sense, like:

```
#   JoinRel [lateral=true, type=LEFT]
#     ├── left:  ReadRel(rel_anchor=5, customers)    -- output: (customer_id, name, ...)
#     └── right: FilterRel
#                  condition: orders.customer_id = outer_ref(rel_reference=5, field 0)
#                  └── ReadRel(orders)
```

Having the anchor on the join itself could get a bit confusing, especially if the join also has any post-join filter or output remappings involved.

yongchul: Good point. Let me try to clarify the reasoning.

* Why not referring output of the left input?

Yes, we could do that but resolving the binding against the "input schema" of the operator feels more ergonomic to me because this is how all the *scalar expressions* are being resolved. It is also clearly states which operator is the outer context. For example, if it were ProjectRel, would expression refer the input of the project or just project? This is also consistent with `steps_out` scheme.

emit and post_join_filter are clearly stating that they are over direct output order, so at least it is not confusing to me.

* But can we make it refer to the output instead?

We can but we will need to look into children to identify the correct one but it is workable. If we go down this path, we need to update outer reference documentation such that it should refer the id of rel whose output is being referenced.

One advantage of this approach is we can express correlated n-ary relations naturally (i.e., think about sequencing multiple relational trees like in a stored procedure).

Or, both scheme may co-exist. There is really no clash other than two rules (i.e., you can either use input schema or output schema based on your nestings).

vbarua: I think I confused myself with regards to the behaviour of OuterReferences

> As I understand the various references type that that point to other relations, they reference the output of the relation.

I'm wrong here, I think.

From the [proto docs](https://github.com/substrait-io/substrait/blob/276296e80436b979bf7e9263f381fbf650d00d6c/proto/substrait/algebra.proto#L1619-L1644)

> // ... The root_type specifies the origin of the data (input record,
  // outer query, lambda parameters, or an arbitrary expression),

For RootReference, that is relative to the input root
```proto
    // Singleton that expresses this FieldReference is rooted off the root
    // incoming record type
    message RootReference {}
```

For OuterReference, that is still a root reference to the input of the referenced rel:
```
    // A root reference for the outer relation's subquery.
    message OuterReference {
```

nielspardon:
It took me a while to wrap my ahead around this. I think I understand the idea now. 

What makes me feel a little uneasy is that the `lateral` flag changes quite a bit of the `JoinRel` and `CrossRel` semantics like restricting the supported join types. This makes me think that having a separate relation would be easier to understand. We would have to re-declare some of the join semantics on the new relation but we would have less conditional semantics baked into `JoinRel` / `CrossRel`. If we had a separate relation which only had the join types that make sense for a lateral join that would be easier to understand than the conditional logic of the current proposal.

I was curious how Calcite deals with lateral joins and they have a separate `Correlate` operator which they also chose since it was easier to distinguish between `Join` and `Correlate` in optimization rules since they are claiming most join optimization rules shouldn't be applied to correlates (if I understand it correctly).

https://github.com/apache/calcite/blob/17bf870d32d1362055e1df0050eec4e7b0f7e03c/core/src/main/java/org/apache/calcite/rel/core/Correlate.java#L54

site/examples/proto-textformat/field_reference/outer_reference_lateral_join.textproto
```diff
@@ -0,0 +1,37 @@
...
+#   FROM customers c
+#   LEFT JOIN LATERAL (
+#     SELECT *
+#     FROM orders
+#     WHERE orders.customer_id = c.customer_id   -- outer reference
+#   ) o ON true
+#
+# Plan structure:
+#
+#   LateralJoinRel [rel_anchor=5, type=LEFT]
```
nielspardon, line 18: nit: should we include an `expression: true::bool` or whatever the syntactically correct equivalent is to match the SQL query more closely?

nielspardon (approved):
LGTM

nielspardon:
I think we would also need to update the Substrait dialect schema to allow implementations to declare their support for the new relation.

proto/substrait/algebra.proto
```diff
@@ -296,6 +296,35 @@ message JoinRel {
...
+// semantically identical to JoinRel, except the right input is evaluated once
+// per row of the left input and may reference fields from the current left row
+// via OuterReference rel_reference.
+//
+// LateralJoinRel must set RelCommon.rel_anchor so the right input can
+// reference fields of the current left row.
+message LateralJoinRel {
+  RelCommon common = 1;
+  Rel left = 2;
+  Rel right = 3;
```
```suggestion
  // The independent input whose rows drive evaluation of the lateral join.
  Rel left = 2;

  // The dependent input evaluated once per left row. It may reference fields
  // from the current left row using OuterReference.rel_reference pointing to
  // this LateralJoinRel's RelCommon.rel_anchor.
  Rel right = 3;
```

What do you think about this? Just to make things super explicit.

proto/substrait/algebra.proto
```diff
@@ -296,6 +296,35 @@ message JoinRel {
...
+  // join-type-specific output formation is complete.
+  // Semantically equivalent to placing a FilterRel directly above this join.
+  Expression post_join_filter = 5;
+
+  // Only INNER and left-oriented join types are valid for lateral joins:
+  // INNER, LEFT, LEFT_SEMI, LEFT_ANTI, LEFT_SINGLE, LEFT_MARK.
+  // RIGHT-oriented join types and OUTER are not valid because the right input
+  // is evaluated in the context of each left row and has no independent
+  // existence.
+  JoinRel.JoinType type = 6;
```
vbarua, line 328: 👿  Hat: One could make the case that this should be it's own enum with only the variants supported by LateralJoinRel. Which is what we've done in the past.

Taking the 👿 off, that has historically made the libraries more complicated because we end up with a bunch of custom enum wrappers. Dialects feel like a better solution and story for validations in going forwards as well.

I like this unification on the JoinRel enum here, and would be interested in updating the other Join relations do this if possible.

site/docs/relations/logical_relations.md
```diff
@@ -274,6 +274,47 @@ The join operation will combine two separate inputs into a single output, based
 %%% proto.algebra.JoinRel %%%
     ```
 
+## Lateral Join Operation
+
+The lateral join operation is represented by `LateralJoinRel`.
+
+`LateralJoinRel` uses the same fields and `JoinRel.JoinType` enum as [Join Operation](#join-operation). For field meanings and join-type behavior, refer to the `JoinRel` documentation above.
+
+In `LateralJoinRel`, the right input is evaluated once per row of the left input and may reference fields of the current left row via `OuterReference.rel_reference`. `LateralJoinRel` must set `RelCommon.rel_anchor`; when the right input references the current left row, `rel_reference` points to that anchor. Right-input outer references may also point to other outer roots when applicable. See [Field References — Outer References](../expressions/field_references.md#outer-references) for details.
```
vbarua:
```suggestion
In `LateralJoinRel`, the right input is evaluated once per row of the left input and may reference fields of the current left row via `OuterReference.rel_reference`. `LateralJoinRel` must set `RelCommon.rel_anchor` when the right input references the current left row, `rel_reference` points to that anchor. Right-input outer references may also point to other outer roots when applicable. See [Field References — Outer References](../expressions/field_references.md#outer-references) for details.
```

The `;` seems extraneous.

site/docs/relations/logical_relations.md
```diff
@@ -274,6 +274,47 @@ The join operation will combine two separate inputs into a single output, based
...
+
+```
+--8<-- "examples/proto-textformat/field_reference/outer_reference_lateral_join.textproto"
+```
+
+=== "LateralJoinRel Message"
+
+        ```proto
+%%% proto.algebra.LateralJoinRel %%%
+        ```
```
vbarua:
```suggestion
    ```proto
%%% proto.algebra.LateralJoinRel %%%
    ```
```

This wasn't rendering correctly. The indentation is somehow load bearing 🙃 

<img width="970" height="208" alt="Image" src="https://github.com/user-attachments/assets/a660fea8-3303-4dfd-b29d-f05d025c73b9" />

vbarua (approved):
Left some very minor suggestions, but otherwise LGTM!

proto/substrait/algebra.proto
yongchul: Either this (use JoinRel.JoinType) or just lift the join type to top level as you did in one of the PR. This is something we should clean up IMO.

site/docs/relations/logical_relations.md
```diff
@@ -274,6 +274,45 @@ The join operation will combine two separate inputs into a single output, based
...
+can be represented as an inner lateral join where `T1` is the left input and the scalar subquery `SELECT MAX(b) FROM T2 WHERE T2.x = T1.a` is the right input.
+
+Here is a concrete `FieldReference` example for a lateral join's outer reference:
+
+```
+--8<-- "examples/proto-textformat/field_reference/outer_reference_lateral_join.textproto"
+```
+
+=== "LateralJoinRel Message"
+
```
benbellick, line 307: This is rendering incorrectly but I think you meant to include

```proto
%%% proto.algebra.LateralJoinRel %%%
```

site/docs/relations/logical_relations.md
```diff
@@ -274,6 +274,45 @@ The join operation will combine two separate inputs into a single output, based
 %%% proto.algebra.JoinRel %%%
     ```
 
+## Lateral Join Operation
+
+The lateral join operation is represented by `LateralJoinRel`.
+
+`LateralJoinRel` uses the same fields and `JoinRel.JoinType` enum as [Join Operation](#join-operation). For field meanings and join-type behavior, refer to the `JoinRel` documentation above.
+
+In `LateralJoinRel`, the right input is evaluated once per row of the left input and may reference fields of the current left row via `OuterReference.rel_reference`. `LateralJoinRel` must set `RelCommon.rel_anchor` when the right input references the current left row, `rel_reference` points to that anchor. Right-input outer references may also point to other outer roots when applicable. See [Field References — Outer References](../expressions/field_references.md#outer-references) for details.
```
benbellick:
```suggestion
In `LateralJoinRel`, the right input is evaluated once per row of the left input and may reference fields of the current left row via `OuterReference.rel_reference`. When the right input references the current left row, `LateralJoinRel` must set `RelCommon.rel_anchor`, and `rel_reference` points to that anchor. Right-input outer references may also point to other outer roots when applicable. See [Field References — Outer References](../expressions/field_references.md#outer-references) for details.
```

benbellick (approved):
Apologies for such a late review. a few minor things but otherwise looks great! Thanks for your patience. 🚀

site/docs/relations/logical_relations.md
yongchul: good catch!

PR #978, opened 2026-02-27
feat(extensions): add fp16, decimal256, and duration extension types

kadinrabo:
Adds three numeric extension types as self-contained modules following the pattern from #953 (unsigned integers). Each file defines types with string structure encoding and arithmetic function overloads.                                                                                                      

  - `float16.yaml`: fp16 with add, subtract, multiply, divide, sum, min, max
  - `decimal256.yaml`: decimal256<precision, scale> with add, subtract, multiply, divide, modulus, abs, sum, min, max, avg
  - `duration.yaml`: duration<unit> with add, subtract, sum, min, max

  These types are already defined in Arrow's `extension_types.yaml` and are currently encoded in DataFusion via `type_variation_reference`, which violates the spec's "same semantics" requirement for type variations.

  Depends on [#953](https://github.com/substrait-io/substrait/pull/953).

- - -
This change is [Reviewable](https://reviewable.io/reviews/substrait-io/substrait/978)


extensions/duration.yaml
nielspardon, line 1: What is the difference between `duration` and the `interval_year` / `interval_day` data types?

Issue #968, opened 2026-02-20
feat: extend test grammar to support nested list types and null list literals


benbellick:
The test case grammar (defined in `grammar/FuncTestCaseParser.g4`) does not currently support:

1. Nested list literals: e.g. `[[1, 2], [3, 4]]::list<list<i32>>`
2. Null list arguments: e.g. `null::list<i32>` as a function input
3. Null list results: e.g. `= null::list<i32>` as an expected return value

This limitation was discovered while adding test cases for new list functions in #967

benbellick:
In #969, we wanted to add a test case for `cardinality(Null::list<i32>) = Null::i64` but the grammar doesn't support null list literals.

PR #987, opened 2026-03-04
feat: add has_overlap function to functions_list.yaml

### LINKED ISSUE #986: feat: Add arrays_overlap function to functions_list.yaml / functions_set.yaml
The `arrays_overlap` function (returns whether two arrays share any common elements) is a common primitive across engines and would be a good addition to `functions_list.yaml` or `functions_set.yaml` (I am not particular about which).

## Cross-Engine Function Names

| Engine | Function |
|--------|----------|
| PostgreSQL | [`&&` operator](https://www.postgresql.org/docs/current/functions-array.html) |
| DuckDB | [`list_has_any`](https://duckdb.org/docs/sql/functions/list.html#list_has_anylist-list) |
| Spark | [`arrays_overlap`](https://spark.apache.org/docs/latest/api/sql/#arrays_overlap) |
| Trino | [`arrays_overlap`](https://trino.io/docs/current/functions/array.html#arrays_overlap) |
| DataFusion | [`array_has_any`](https://datafusion.apache.org/user-guide/sql/scalar_functions.html#array-has-any) |

## Cross-Engine Semantic Comparison

The following table was generated by running equivalent queries across all five engines via Docker.

| Test Case | PostgreSQL | DuckDB | Spark | Trino | DataFusion |
|-----------|-----------|--------|-------|-------|------------|
| `[1,2,3] && [3,4,5]` | true | true | true | true | true |
| `[1,2,3] && [4,5,6]` | false | false | false | false | false |
| `[1,2,3] && []` | false | false | false | false | false |
| `[] && []` | false | false | false | false | false |
| `[1,NULL,3] && [3,4]` | true | true | true | true | true |
| ⚠️ `[1,NULL,3] && [4,5]` | **false** | **false** | **NULL** | **NULL** | **false** |
| ⚠️ `[1,NULL] && [NULL,4]` | **false** | **false** | **NULL** | **NULL** | **true** |
| `NULL && [1,2]` | NULL | NULL | NULL | NULL | NULL |
| `[1,1,2] && [1,3]` | true | true | true | true | true |
| `['a','b','c'] && ['c','d']` | true | true | true | true | true |

All five engines agree on the straightforward cases. The divergence is entirely about **how NULL elements are handled** when there is no definitive non-null overlap (cases 6 & 7). 

## Proposal

Lets add an `arrays_overlap` function to `functions_{list/set}.yaml`. We can handle the nullability stuff with a function option.

benbellick:
Closes #986

Adds a `has_overlap` scalar function to `functions_list.yaml` that determines whether two lists share any common elements.

This was initially placed in `functions_set.yaml`, but per discussion it makes more sense in `functions_list.yaml` since the function operates on `list` types and bundling functions under a `function_<datatype>` namespace is the more natural convention.

## Cross-Engine Semantic Comparison

The following table was generated by running equivalent queries across engines via Docker (see details below for the script used to generate it).

| Test Case | PostgreSQL | DuckDB | Spark | Trino | DataFusion | ClickHouse |
|-----------|-----------|--------|-------|-------|------------|------------|
| `[1,2,3] && [3,4,5]` | true | true | true | true | true | true |
| `[1,2,3] && [4,5,6]` | false | false | false | false | false | false |
| `[1,2,3] && []` | false | false | false | false | false | false |
| `[] && []` | false | false | false | false | false | false |
| `[1,NULL,3] && [3,4]` | true | true | true | true | true | true |
| ⚠️ `[1,NULL,3] && [4,5]` | **false** | **false** | **NULL** | **NULL** | **false** | **false** |
| ⚠️ `[1,NULL] && [NULL,4]` | **false** | **false** | **NULL** | **NULL** | **true** | **true** |
| `NULL && [1,2]` | NULL | NULL | NULL | NULL | NULL | NULL |
| `[1,1,2] && [1,3]` | true | true | true | true | true | true |

The divergence is entirely about how NULL elements are handled when there is no definitive non-null overlap:

| Behavior | Engines |
|----------|---------|
| **THREE_VALUED** | Spark, Trino |
| **IGNORE_NULLS** | PostgreSQL, DuckDB |
| **NULL_EQUALS_NULL** | DataFusion, ClickHouse |

This is modeled via a `null_handling` option with those three values.

Note: test cases for null list arguments (e.g. `has_overlap(null::list<i32>, ...)`) will be added once #968 lands.

<details>
<summary>Cross-engine verification script</summary>

```bash
#!/bin/bash

# Start Trino in background
docker run --rm -d --name trino-test trinodb/trino:latest > /dev/null 2>&1
echo "Waiting for Trino to start..."
sleep 25

run_pg() {
    result=$(docker run --rm -e POSTGRES_PASSWORD=test postgres:latest su postgres -c "initdb -D /tmp/pgdata >/dev/null 2>&1 && pg_ctl start -D /tmp/pgdata -l /tmp/pg.log -o '-k /tmp' >/dev/null 2>&1 && sleep 2 && psql -h /tmp -t -A -c \"$1\"" 2>/dev/null)
    if [ -z "$result" ]; then echo "NULL"; else echo "$result"; fi
}

run_duckdb() {
    result=$(docker run --rm duckdb/duckdb:latest duckdb -noheader -list -c "$1" 2>/dev/null)
    if [ -z "$result" ]; then echo "NULL"; else echo "$result"; fi
}

run_spark() {
    result=$(docker run --rm apache/spark:latest /opt/spark/bin/spark-sql -e "$1" 2>&1 | grep -v "WARN\|INFO\|Time taken\|Spark Web\|Spark master\|ObjectStore\|^Using\|^Setting\|^24\|^Ivy\|^Preparing\|^Resolving\|^Downloading\|^::" | grep -v "^$" | tail -1)
    if [ -z "$result" ]; then echo "NULL"; else echo "$result"; fi
}

run_trino() {
    result=$(docker exec trino-test trino --execute "$1" 2>&1 | grep -v "WARNING\|jline" | tr -d '"')
    if [ -z "$result" ]; then echo "NULL"; else echo "$result"; fi
}

run_datafusion() {
    result=$(datafusion-cli -c "$1" 2>&1 | grep -E "^\|" | tail -1 | sed 's/|//g' | tr -d ' ')
    if [ -z "$result" ]; then echo "NULL"; else echo "$result"; fi
}

run_clickhouse() {
    result=$(docker run --rm clickhouse/clickhouse-server:latest clickhouse-local --query "$1" 2>/dev/null)
    if [ -z "$result" ]; then echo "NULL"; else echo "$result"; fi
}

run_test() {
    local label="$1" pg="$2" duck="$3" spark="$4" trino="$5" df="$6" ch="$7"
    echo ""
    echo "$label"
    echo "  PostgreSQL: $(run_pg "$pg")"
    echo "  DuckDB:     $(run_duckdb "$duck")"
    echo "  Spark:      $(run_spark "$spark")"
    echo "  Trino:      $(run_trino "$trino")"
    echo "  DataFusion: $(run_datafusion "$df")"
    echo "  ClickHouse: $(run_clickhouse "$ch")"
}

run_test "[1,2,3] && [3,4,5]" \
    "SELECT ARRAY[1,2,3] && ARRAY[3,4,5]" \
    "SELECT list_has_any([1,2,3], [3,4,5])" \
    "SELECT arrays_overlap(array(1,2,3), array(3,4,5))" \
    "SELECT arrays_overlap(ARRAY[1,2,3], ARRAY[3,4,5])" \
    "SELECT array_has_any(make_array(1,2,3), make_array(3,4,5))" \
    "SELECT hasAny([1,2,3], [3,4,5])"

run_test "[1,2,3] && [4,5,6]" \
    "SELECT ARRAY[1,2,3] && ARRAY[4,5,6]" \
    "SELECT list_has_any([1,2,3], [4,5,6])" \
    "SELECT arrays_overlap(array(1,2,3), array(4,5,6))" \
    "SELECT arrays_overlap(ARRAY[1,2,3], ARRAY[4,5,6])" \
    "SELECT array_has_any(make_array(1,2,3), make_array(4,5,6))" \
    "SELECT hasAny([1,2,3], [4,5,6])"

run_test "[1,2,3] && []" \
    "SELECT ARRAY[1,2,3] && ARRAY[]::int[]" \
    "SELECT list_has_any([1,2,3], []::INT[])" \
    "SELECT arrays_overlap(array(1,2,3), array())" \
    "SELECT arrays_overlap(ARRAY[1,2,3], ARRAY[])" \
    "SELECT array_has_any(make_array(1,2,3), make_array())" \
    "SELECT hasAny([1,2,3], [])"

run_test "[] && []" \
    "SELECT ARRAY[]::int[] && ARRAY[]::int[]" \
    "SELECT list_has_any([]::INT[], []::INT[])" \
    "SELECT arrays_overlap(array(), array())" \
    "SELECT arrays_overlap(ARRAY[], ARRAY[])" \
    "SELECT array_has_any(make_array(), make_array())" \
    "SELECT hasAny([], [])"

run_test "[1,NULL,3] && [3,4]" \
    "SELECT ARRAY[1,NULL,3] && ARRAY[3,4]" \
    "SELECT list_has_any([1,NULL,3], [3,4])" \
    "SELECT arrays_overlap(array(1,NULL,3), array(3,4))" \
    "SELECT arrays_overlap(ARRAY[1,NULL,3], ARRAY[3,4])" \
    "SELECT array_has_any(make_array(1,NULL,3), make_array(3,4))" \
    "SELECT hasAny([1,NULL,3], [3,4])"

run_test "[1,NULL,3] && [4,5]" \
    "SELECT ARRAY[1,NULL,3] && ARRAY[4,5]" \
    "SELECT list_has_any([1,NULL,3], [4,5])" \
    "SELECT arrays_overlap(array(1,NULL,3), array(4,5))" \
    "SELECT arrays_overlap(ARRAY[1,NULL,3], ARRAY[4,5])" \
    "SELECT array_has_any(make_array(1,NULL,3), make_array(4,5))" \
    "SELECT hasAny([1,NULL,3], [4,5])"

run_test "[1,NULL] && [NULL,4]" \
    "SELECT ARRAY[1,NULL] && ARRAY[NULL,4]" \
    "SELECT list_has_any([1,NULL], [NULL,4])" \
    "SELECT arrays_overlap(array(1,NULL), array(NULL,4))" \
    "SELECT arrays_overlap(ARRAY[1,NULL], ARRAY[NULL,4])" \
    "SELECT array_has_any(make_array(1,NULL), make_array(NULL,4))" \
    "SELECT hasAny([1,NULL], [NULL,4])"

run_test "NULL && [1,2]" \
    "SELECT NULL::int[] && ARRAY[1,2]" \
    "SELECT list_has_any(NULL::INT[], [1,2])" \
    "SELECT arrays_overlap(NULL, array(1,2))" \
    "SELECT arrays_overlap(NULL, ARRAY[1,2])" \
    "SELECT array_has_any(NULL::INT[], make_array(1,2))" \
    "SELECT hasAny([1,2]::Nullable(Array(Int32)), NULL)"

run_test "[1,1,2] && [1,3]" \
    "SELECT ARRAY[1,1,2] && ARRAY[1,3]" \
    "SELECT list_has_any([1,1,2], [1,3])" \
    "SELECT arrays_overlap(array(1,1,2), array(1,3))" \
    "SELECT arrays_overlap(ARRAY[1,1,2], ARRAY[1,3])" \
    "SELECT array_has_any(make_array(1,1,2), make_array(1,3))" \
    "SELECT hasAny([1,1,2], [1,3])"

docker stop trino-test > /dev/null 2>&1
```

</details>

-----

- - -
This change is [Reviewable](https://reviewable.io/reviews/substrait-io/substrait/987)


extensions/functions_set.yaml
```diff
@@ -26,3 +26,40 @@ scalar_functions:
...
+          - name: right
+            value: list<any1>
+        options:
+          null_handling:
+            description: >
+              Controls how null elements affect the result when no non-null
+              elements overlap.
+
+              - THREE_VALUED: Returns NULL when null elements are present in
+                both lists but no non-null overlap is found.
```
yongchul: same comment as @vbarua had in the other PR... if option is not specified, I guess it is assumed to be engine dependent?

I keep forgetting the dialect... does the dialect allow specify supported options or enum arguments?

extensions/functions_set.yaml
yongchul, line 6: it's odd that this function is actually in the `set` 😄

extensions/functions_set.yaml
```diff
@@ -26,3 +26,40 @@ scalar_functions:
             values: [ NAN_IS_NAN, NAN_IS_NOT_NAN ]
         nullability: DECLARED_OUTPUT
         return: i64?
+  -
+    name: "has_overlap"
+    description: >
+      Determines whether two lists share any common elements.
```
yongchul: I'm fine having `list` functions reside in `set` as long as the functions work as "set" semantic. But how many functions are there as a set?

* intersect, set and multi-set
* difference, set and multi-set
* union, set and multi-set
* distinct? (multi-set to set)
* membership test - returns bool

After written down, looks like it has enough content 😄

benbellick: Yeah I was also unsure about where to place this one. I'm happy to put it in list, I don't feel particularly strongly. My thinking was that this operation is kind-of "set" like in that we are taking an intersection and we don't care about duplicates. But I can see a case for still considering it a list operation. 🤷 

Let me know if you'd like me to move it and I'd be happy to do it.

extensions/functions_set.yaml
benbellick: Yes 🤦 if it returned a `boolean` I could see it. But returning the index? That isn't  very set like 😅

extensions/functions_set.yaml
yongchul: Now that I read again those set-like operations, they are too vague (in the end, it is list. what would be the "unioned" list would look like? what about diff?) and not sure whether they have actual usages.  Also, it kinda makes sense to bundle functions specific to a data type under `function_<datatype>` namespace. So I am inclined to put it in the `list` namespace.

benbellick: That's totally fine. It seems reasonable that we would want to introduce a set UDT in the future anyways, so I'm happy to switch it over. Thanks!

extensions/functions_set.yaml
benbellick: I'm still not sure what the right thing to do here is 😟

extensions/functions_set.yaml
yongchul: Do we also want to move this to list? Sorry, this comment was not published. It is technically a breaking change but wondering how many would have taken dependencies...

yongchul (approved):

benbellick:
@yongchul I agree it may be a good idea to just get rid of `functions_set`. In the interest of keeping this PR from becoming a breaking change, let's leave it for future work if that is alright with you :)

yongchul:
> @yongchul I agree it may be a good idea to just get rid of `functions_set`. In the interest of keeping this PR from becoming a breaking change, let's leave it for future work if that is alright with you :)

I'm okay with keeping the file but it would be nice if we can do some prep work (i.e., duplicating the function and have description in the file for deprecation warning). It's fine to do that in a separate PR.

extensions/functions_set.yaml
yongchul: we can now deprecate this and move it to list function! 😆 please consider it in different PR.

extensions/functions_list.yaml
```diff
@@ -130,3 +130,40 @@ scalar_functions:
...
+
+      If either input list is empty, returns `false`.
+    impls:
+      - args:
+          - name: left
+            value: list<any1>
+          - name: right
+            value: list<any1>
+        options:
+          null_handling:
```
yongchul, line 153: my 2 cents are `enum` for clarity and interoperability. One might argue that this is an option that should follow system default but

a. the alternatives are rather well-understood
b. it may not be consistent (i.e., system may not implement null handling consistently)
c. it works better with the dialect I believe...
d. null handling is... kind of the essential to the semantic of the overlap (null is not an exception in real data unfortunately).

vbarua:
+1 for using an enum to handle the nullability differences.

Deprecating `functions_set`, if that's something we want to do, should be its own PR.

extensions/functions_list.yaml
vbarua: Agree with this. Null handling is kind of essential here, and if we allow producers to not set this, or consumers to ignore it, were going to wind up with some very interesting execution differences across systems.

Issue #988, opened 2026-03-04
fix: enforce nullable types for null literals in test cases


benbellick:
Null literals in test cases currently use non-nullable types, e.g. `null::i32`, `null::bool`, `null::dec<38, 0>`. This is semantically incorrect. A null value must have a nullable type (`null::i32?`, `null::bool?`, `null::dec?<38, 0>`). 

This matters because null-semantics are something we _want_ to test with these test files. Even when the value is a literal, we want to know if its type is nullable or non-nullable. 

We can either amend the grammar to enforce this, or amend the visitor to actually check it.

vbarua:
A third option would be to clarify in the _test_ spec that `null` literals are always nullable. Effectively:
```
null:i8  -> Implicitly Nullable I8 Type
null:i8? -> Explicitly Nullable I8 Type
5:i8     -> Explicitly Non-Null I8 Type
5:i8?    -> Explicitly Nullable I8 Type
```

At the written spec level, we can clarify the null literals _must_ always be of nullable type, which is what we've indicated in the protobufs
https://github.com/substrait-io/substrait/blob/e2dca291fb9061a38bbd03574f11a85b56f96c28/proto/substrait/algebra.proto#L1042-L1047

benbellick:
@vbarua I am fine with that approach if that is what the community wants. I personally think explicit is always better, and a one time large PR feels like a relatively minor cost to pay for more explicit tests. It also means that nulls aren't a sort of "special" case.

benbellick:
This was closed in #989

PR #989, opened 2026-03-04
fix: enforce nullable types for null literals in test cases

benbellick:
Test cases now respect the nullability rules defined by each function's `nullability` handling in the extension YAML. MIRROR functions require the output to be marked nullable (`?`) iff any argument is nullable, and DECLARED_OUTPUT functions require the output nullability to match the declared return type.

The Python test infrastructure now enforces these rules automatically via `test_substrait_nullability_consistency`. This drove the ~105 test case line changes to add missing `?` markers on return types. The visitor was also fixed to stop silently dropping nullable markers for several types, and to distinguish outer nullability (`list?<i32>`) from inner element nullability (`list<i32?>`).

DISCRETE nullability handling only validates output type for now — argument nullability matching is not yet implemented. This is fine since no extensions currently use DISCRETE.

-----

### DIFF
--- a/tests/coverage/extensions.py
+++ b/tests/coverage/extensions.py
@@ -121,14 +121,14 @@ class Extension:
                 "nullability", "MIRROR"
             )  # MIRROR is the spec default
             return_type_raw = str(impl["return"])
-            return_nullable = type_str_is_outer_nullable(return_type_raw)
+            is_return_nullable = type_str_is_outer_nullable(return_type_raw)
             overloads.append(
                 FunctionOverload(
                     args,
                     Extension.get_short_type(impl["return"]),
                     "variadic" in impl,
                     nullability=nullability,
-                    return_nullable=return_nullable,
+                    is_return_nullable=is_return_nullable,
                 )
             )
         return overloads
@@ -227,7 +227,7 @@ class FunctionVariant:
         variadic,
         func_type,
         nullability="MIRROR",
-        return_nullable=False,
+        is_return_nullable=False,
     ):
         self.name = name
         self.uri = uri
@@ -237,7 +237,7 @@ class FunctionVariant:
         self.variadic = variadic
         self.func_type = func_type
         self.nullability = nullability
-        self.return_nullable = return_nullable
+        self.is_return_nullable = is_return_nullable
         self.test_count = 0
 
     def __str__(self):
@@ -249,16 +249,21 @@ class FunctionVariant:
 
 class FunctionOverload:
     def __init__(
-        self, args, return_type, variadic, nullability="MIRROR", return_nullable=False
+        self,
+        args,
+        return_type,
+        variadic,
+        nullability="MIRROR",
+        is_return_nullable=False,
     ):
         self.args = args
         self.return_type = return_type
         self.variadic = variadic
         self.nullability = nullability
-        self.return_nullable = return_nullable
+        self.is_return_nullable = is_return_nullable
 
     def __str__(self):
-        return f"FunctionOverload(args={self.args}, result={self.return_type}, variadic={self.variadic}, nullability={self.nullability})"
+        return f"FunctionOverload(args={self.args}, result={self.return_type}, variadic={self.variadic}, nullability={self.nullability}, is_return_nullable={self.is_return_nullable})"
 
 
 # define function type enum
@@ -298,7 +303,7 @@ class FunctionRegistry:
                     overload.variadic,
                     func_type,
                     nullability=overload.nullability,
-                    return_nullable=overload.return_nullable,
+                    is_return_nullable=overload.is_return_nullable,
                 )
                 fun_arr.append(function)
             self.registry[f_name] = fun_arr
--- a/tests/coverage/coverage.py
+++ b/tests/coverage/coverage.py
@@ -198,12 +198,12 @@ def _validate_declared_output(test_case, variant, loc, sig):
     type in the extension YAML.
     """
     result_nullable = _result_is_nullable(test_case)
-    if variant.return_nullable and not result_nullable:
+    if variant.is_return_nullable and not result_nullable:
         return (
             f"{loc}: DECLARED_OUTPUT violation: {sig} — "
             f"output should be nullable to match declared return type"
         )
-    if not variant.return_nullable and result_nullable:
+    if not variant.is_return_nullable and result_nullable:
         return (
             f"{loc}: DECLARED_OUTPUT violation: {sig} — "
             f"output should not be nullable; declared return type is not nullable"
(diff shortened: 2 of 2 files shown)


tests/coverage/extensions.py
```diff
@@ -281,7 +281,7 @@ def is_type_any(func_arg_type):
 
     @staticmethod
     def is_same_type(func_arg_type, arg_type):
-        arg_type_base = arg_type.split("<")[0]
+        arg_type_base = arg_type.split("<")[0].rstrip("?")
```
yongchul, line 317: I'm not familiar with this part of repo... but this looks like comparing only the base type. `is_same_type` is the proper name for this function? also, shouldn't it consider DISCRETE nullability https://substrait.io/expressions/scalar_functions/#nullability-handling?

benbellick: I'm not so familiar with this part of it either. At one point in history I was looking into rewriting quite a bit of it to use a more typed representation, but ended up focusing on other things. 

Yes, I think it is only comparing the base type. Though that was the behavior before this PR 🤷. I agree it should consider DISCRETE nullability. Is it possible to leave those sort of validations for another PR? 

I think there is a lot of room for improvement in these test cases, but wanted to keep changes manageable.

yongchul: I don't mean to hold back this PR, just want to understand whether I am missing something. 😄 

But overall, it appears that this needs to go under some careful review and clean up. Yes, we should do it incrementally -- otherwise, we are going nowhere. 😆

tests/cases/boolean/and.test
```diff
@@ -7,8 +7,8 @@ and(true::bool, false::bool) = false::bool
...
 
 # null_input: Examples with null as input
-and(true::bool, null::bool) = null::bool
-and(null::bool, true::bool) = null::bool
-and(false::bool, null::bool) = false::bool
-and(null::bool, false::bool) = false::bool
-and(null::bool, null::bool) = null::bool
+and(true::bool, null::bool?) = null::bool?
+and(null::bool?, true::bool) = null::bool?
+and(false::bool, null::bool?) = false::bool
```
benbellick: Actually maybe this output should be nullable, too?

vbarua:
Added a potential 3rd option for this in the original issue https://github.com/substrait-io/substrait/issues/988#issuecomment-4012834395

It's a tiny bit obnoxious to have to add `?` to every usage of `null`. At the spec* level, we already say that null literals have to be of nullable type.

\* specifically the proto

tests/test_extensions.py
```diff
@@ -39,6 +39,25 @@ def test_substrait_extension_coverage():
         print(json.dumps(asdict(actual_baseline), indent=2))
 
 
+def test_substrait_nullability_consistency():
```
benbellick, line 42: This is really the core driver of this PR. I realize there are a lot of cases changed, but all of the changes were _driven_ by making this test pass. This test is just checking that the test cases have nullability which is consistent with the nullability handling of the functions as they are defined in the YAML files.

tests/coverage/nodes.py
```diff
@@ -3,6 +3,26 @@
 from typing import List
 
 
+def type_str_is_outer_nullable(type_str):
```
benbellick, line 6: In the interest of keeping this diff small, I used this hack. I would love to have a *better* solution, but I think that is worth its own dedicated PR.

vbarua:
> It's a tiny bit obnoxious to have to add ? to every usage of null. At the spec* level, we already say that null literals have to be of nullable type.

This was a weakly held opinion, and I have been easily convinced the explicit is better than implicit in this case.

tests/coverage/visitor.py
```diff
@@ -326,92 +331,157 @@ def visitFloatLiteral(self, ctx: FuncTestCaseParser.FloatLiteralContext):
 
     def visitNullArg(self, ctx: FuncTestCaseParser.NullArgContext):
         datatype = ctx.dataType().getText()
-        return CaseLiteral(value=None, type=datatype)
+        if not type_str_is_outer_nullable(datatype):
+            raise ParseError(
+                f"Null literal must have a nullable type (add '?'), "
+                f"got: null::{datatype}"
```
vbarua, line 337: Nice, this will make it easy for other folks to fix this issue when adding new tests.

tests/coverage/visitor.py
```diff
@@ -213,7 +214,11 @@ def visitTableData(self, ctx: FuncTestCaseParser.TableDataContext):
     def visitDataColumn(self, ctx: FuncTestCaseParser.DataColumnContext):
         column = self.visitColumnValues(ctx.columnValues())
         column_type = ctx.dataType().getText()
-        return CaseLiteral(value=column, type=column_type)
+        return CaseLiteral(
+            value=column,
+            type=column_type,
+            nullable=type_str_is_outer_nullable(column_type),
+        )
```
vbarua, line 221: You were asking if you could do something better*. You can technically just reach down through the nodes and look for the QMark() directly.

```python
        column = self.visitColumnValues(ctx.columnValues())
        column_type = ctx.dataType()
        is_nullable = column_type.children[0].children[0].QMark() == '?'
```

\* not sure if this is better tbh, but it was fun to figure out

vbarua: Talking with @/benbellick about this, in some ways this is more fragile than his approach of producing the type string and parsing that. I'm happy to go with his approach, and while there's always room for improvement, I don't think it needs to happen in this PR.

tests/coverage/extensions.py
```diff
@@ -208,14 +217,27 @@ class FunctionType:
...
+        self,
+        name,
+        uri,
+        description,
+        args,
+        return_type,
+        variadic,
+        func_type,
+        nullability="MIRROR",
+        return_nullable=False,
```
vbarua: minor: it might be clearer readability wise to call this `is_return_nullable` or `can_return_nullable`.

tests/coverage/extensions.py
```diff
@@ -226,13 +248,17 @@ def increment_test_count(self, count=1):
...
+    ):
         self.args = args
         self.return_type = return_type
         self.variadic = variadic
+        self.nullability = nullability
+        self.return_nullable = return_nullable
 
     def __str__(self):
-        return f"FunctionOverload(args={self.args}, result={self.return_type}, variadic={self.variadic})"
+        return f"FunctionOverload(args={self.args}, result={self.return_type}, variadic={self.variadic}, nullability={self.nullability})"
```
vbarua: Do we not want to stringify the `return_nullable` value?

tests/coverage/extensions.py
vbarua: +1 to this being something we can improve, but that which  we should improve in the future.

tests/coverage/nodes.py
```diff
@@ -3,6 +3,26 @@
...
+    ``isnull`` token rather than re-parsing the string, but ``dataType``
+    is an intermediate rule and the ``isnull`` token lives on concrete
+    type contexts nested several levels deep.  Getting it would require a
+    non-trivial refactor of the visitor to propagate nullability up
+    through ``dataType``.
+    """
+    bracket_pos = type_str.find("<")
+    if bracket_pos == -1:
+        return type_str.endswith("?")
+    return "?" in type_str[:bracket_pos]
```
vbarua, line 23: This is reasonable enough, and the type string shape can be considered stable.

PR #991, opened 2026-03-04
feat: extend test grammar to support null list types and nested list literals

### LINKED ISSUE #968: feat: extend test grammar to support nested list types and null list literals

benbellick:
Closes #968

Extends the test case grammar (`FuncTestCaseParser.g4`) to support:

- Null list arguments and results
- Nested list literals

It also adds some new test cases to demonstrate that these work. 

-----

### DIFF
--- a/tests/cases/list/all_match.test
+++ b/tests/cases/list/all_match.test
@@ -13,6 +13,12 @@ all_match([1, 2, 3]::list<i32>, (x -> lt(x, 3::i32))::func<i32 -> bool?>) = fals
 # empty: Empty list returns true (vacuously true)
 all_match([]::list<i32>, (x -> gt(x, 0::i32))::func<i32 -> bool?>) = true::bool
 
+# null_list: Null list input returns null
+all_match(null::list<i32>, (x -> gt(x, 0::i32))::func<i32 -> bool?>) = null::bool
+
+# null_predicate: Null predicate returns null
+all_match([1, 2, 3]::list<i32>, null::func<i32 -> bool?>) = null::bool
+
 # three_valued_logic: Null predicate results follow three-valued logic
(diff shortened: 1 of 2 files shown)


tests/cases/list/filter.test
```diff
@@ -10,3 +10,6 @@ filter([]::list<i32>, (x -> gt(x, 0::i32))::func<i32 -> bool>) = []::list<i32>
 # nullable_elements: Handling nullable elements
 filter([1, null, 3]::list<i32?>, (n -> gt(n, 0::i32))::func<i32 -> bool>) = [1, 3]::list<i32?>
 filter([1, null, 3]::list<i32?>, (e -> is_null(e))::func<i32 -> bool>) = [null]::list<i32?>
+
+# null_list: Null list input
+filter(null::list<i32>, (x -> gt(x, 0::i32))::func<i32 -> bool>) = null::list<i32>
```
benbellick, line 15: I personally think this should be `null::list?<i32>` but leaving it this way to be consistent with other test cases. 

Left for another issue here: https://github.com/substrait-io/substrait/issues/988

yongchul: And you are fixing this in #989? Agree that the `list?<i32>` should be the type for null.

benbellick: Yes, that is exactly the plan :)

PR #992, opened 2026-03-04
ci: send notification to substrait-packaging on release

vbarua:


vbarua:
Wires in `notify.sh` to fire after a successful release, which will trigger the spec_release.yml workflow in substrait-packaging. See associated change there https://github.com/substrait-io/substrait-packaging/pull/10.

`notify.sh` is best-effort, because if it fails we can always trigger substrait-packaging manually. I expect that I'll need to tweak the auth a little bit to make this work correctly.

These changes were generated with assistance from Claude.

.releaserc.json
```diff
@@ -23,7 +23,8 @@
       {
         "verifyConditionsCmd": "ci/release/verify.sh",
         "prepareCmd": "ci/release/prepare.sh",
-        "publishCmd": "ci/release/publish.sh ${nextRelease.version}"
+        "publishCmd": "ci/release/publish.sh ${nextRelease.version}",
+        "successCmd": "ci/release/notify.sh ${nextRelease.version}"
```
vbarua, line 27: See available options [here](https://www.npmjs.com/package/@semantic-release/exec#configuration).

yongchul (approved):

vbarua:
This is not spec related, so I will merge with one approval.

PR #971, opened 2026-02-20
feat!: remove deprecated usage of URIs for referencing simple extensions

benbellick:
Closes #970 

BREAKING CHANGE: dropping URI usage from protobufs in favor of URNs

Issue #959, opened 2026-02-13
docs: document return type expression functions


nielspardon:
PR https://github.com/substrait-io/substrait/pull/40 introduced return type expressions for functions defined in Substrait extensions which are documented here: https://substrait.io/expressions/scalar_functions/#return-type-expressions

The challenge is that while a lot of the added functions are trivial to understand there are some functions which are lacking documentation and are not currently used.

I think specifically the following two should be documented:
- `type_parameter(string name) => type`
- `integer_parameter(string name) => integer`

@jacques-n since you are the original author of #40 would you be willing to help adding a sentence each for those two functions?

This came up in https://github.com/substrait-io/substrait/pull/945#discussion_r2760339735

PR #994, opened 2026-03-05
feat!: remove deprecated time, timestamp and timestamp_tz types

### LINKED ISSUE #980: drop support for deprecated time types
For a while, the following deprecations have been ongoing:
- `Timestamp` -> `PrecisionTimeStamp`
- `Time` -> `PrecisionTime`
- `TimestampTZ` -> `PrecisionTimestampTz`

https://github.com/substrait-io/substrait/blob/2705258137b7272830cf00680e6f3a36b8d3ed4b/proto/substrait/type.proto#L28-L46

Lets get rid of them an make it easier on library implementors.

nielspardon:
BREAKING CHANGE: removes the deprecated `time`, `timestamp` and `timestamp_tz` types from:

- proto files
- dialect schema
- extension yamls
- ANTLR grammar
- test cases
- coverage python code
- documentation

closes #980

- - -
This change is [Reviewable](https://reviewable.io/reviews/substrait-io/substrait/994)

### DIFF
--- a/tests/baseline.json
+++ b/tests/baseline.json
@@ -6,7 +6,7 @@
     "num_aggregate_functions": 28,
     "num_scalar_functions": 170,
     "num_window_functions": 11,
-    "num_function_overloads": 8466
+    "num_function_overloads": 498
   },
   "coverage": {
     "total_test_count": 1177,


extensions/functions_datetime.yaml
```diff
@@ -278,25 +202,17 @@ scalar_functions:
...
-          - name: x
-            value: timestamp
-          - name: y
-            value: interval_day<P>
-        return: timestamp
+          - name: precision
+            value: i8
+        return: |-
+          precision = integer_parameter(precision)
+          precision_timestamp<precision>
```
nielspardon: introducing a new `precision` argument here, to make the return type precision predictable

extensions/functions_datetime.yaml
```diff
@@ -675,9 +503,15 @@ scalar_functions:
...
+            description: >-
+              Timezone string from IANA tzdb. Returned precision_timestamp_tz will have time set to
+              00:00:00 with 0 subseconds in the given precision (0 <= precision <= 12).
             value: string
-        return: timestamp_tz
+          - name: precision
+            value: i8
+        return:  |-
+          precision = integer_parameter(precision)
+          precision_timestamp_tz<precision>
```
nielspardon, line 510: introducing a new `precision` argument here, to make the return type precision predictable

extensions/functions_datetime.yaml
```diff
@@ -704,15 +531,19 @@ scalar_functions:
...
           - name: time_string
             value: string
           - name: format
             value: string
-        return: time
+          - name: precision
+            value: i8
+        return: |-
+          precision = integer_parameter(precision)
+          precision_time<precision>
```
nielspardon, line 542: introducing a new `precision` argument here, to make the return type precision predictable

extensions/functions_datetime.yaml
```diff
@@ -745,13 +576,21 @@ scalar_functions:
           - name: timezone
             description: Timezone string from IANA tzdb.
             value: string
-        return: timestamp_tz
+          - name: precision
+            value: i8
+        return: |-
+          precision = integer_parameter(precision)
+          precision_timestamp_tz<precision>
```
nielspardon, line 579: introducing a new `precision` argument here, to make the return type precision predictable

extensions/functions_datetime.yaml
```diff
@@ -745,13 +576,21 @@ scalar_functions:
...
           - name: timestamp_string
             value: string
           - name: format
             value: string
-        return: timestamp_tz
+          - name: precision
+            value: i8
+        return: |-
+          precision = integer_parameter(precision)
+          precision_timestamp_tz<precision>
```
nielspardon, line 589: introducing a new `precision` argument here, to make the return type precision predictable

tools/proto_prefix.py
```diff
@@ -34,7 +34,7 @@ def tokenize(data):
         ident=re.compile(r"[a-zA-Z_][a-zA-Z_0-9.]*"),
         string=re.compile(r'"(?:[^"\\]|\\.)*"'),
         number=re.compile(r"[0-9]+"),
-        symbol=re.compile(r"[=;{}\[\]]"),
+        symbol=re.compile(r"[=;,{}\[\]]"),
```
nielspardon, line 37: since we now have the `reserved 14, 17, 27;` statements in the proto files we need to add comma `,` to the symbol regex

yongchul (approved):
Overall, okay with the change but I want us to discuss and document deprecation and release policy... for instance, @vbarua proposed to hold off a breaking change and make an explicit release with the breaking change (https://github.com/substrait-io/substrait/pull/971#issuecomment-4001228856) -- which to me seems a good idea.

Will create a discussion item for this.

benbellick:
I haven't investigated myself yet but has this migration been managed within the reasonably maintained implementation libraries? (For us, this means `substrait-{go, java, py, rs}`)

nielspardon:
> I haven't investigated myself yet but has this migration been managed within the reasonably maintained implementation libraries? (For us, this means `substrait-{go, java, py, rs}`)

Here the problem is that there is no clean migration path right now since some datetime function variants simply do not exist for the new types in the current default extensions. We would need to at least introduce the new function variants. 

Even then I would argue the time and timestamp types have been deprecated for a long time and it should be ok to force a migration to the new types by removing the old types.

yongchul:
My previous comment still holds so I'm okay to do this breaking change. But since we have deprecation mechanism, that is another option and delay this removal a few releases (i.e., make this PR as *official* deprecation then remove).

nielspardon:
> My previous comment still holds so I'm okay to do this breaking change. But since we have deprecation mechanism, that is another option and delay this removal a few releases (i.e., make this PR as _official_ deprecation then remove).

I guess you mean what you introduced in #1014 with deprecation mechanism which is for extensions. This PR mainly removes the data types from the protobuf files which have been deprecated for a while now and as consequence of that removes / updates functions using the data types. The protobuf deprecation already led to deprecation warnings in the code generated from the protobufs (at least in Java) ever since they were marked deprecated. Marking the extensions deprecated does not lead to any warnings right now and is mainly informational. Marking the extensions using long deprecated types as deprecated first would satisfy the new process but feels a bit redundant for such a longstanding deprecation especially since there are no warning mechanisms in place anyway.

benbellick:
Yeah I'm +1 with @nielspardon about just doing this breaking change. I just need to validate that Datafusion consumer actually accepts the new value. I will do that right now!

benbellick:
Seems like we are good on the DF side of things: 



### Type Handling
- [`PrecisionTimestamp`](https://github.com/apache/datafusion/blob/81423085b45111fa0768742810b73a77d2af06dd/datafusion/substrait/src/logical_plan/consumer/types.rs#L131-L133) - converts to Arrow `Timestamp` with precision-derived `TimeUnit`
- [`PrecisionTimestampTz`](https://github.com/apache/datafusion/blob/81423085b45111fa0768742810b73a77d2af06dd/datafusion/substrait/src/logical_plan/consumer/types.rs#L135-L138) - converts to Arrow `Timestamp` with UTC timezone
- [`PrecisionTime`](https://github.com/apache/datafusion/blob/81423085b45111fa0768742810b73a77d2af06dd/datafusion/substrait/src/logical_plan/consumer/types.rs#L140-L148) - converts to Arrow `Time32`/`Time64` based on precision

### Literal Handling
- [`PrecisionTimestamp` literals](https://github.com/apache/datafusion/blob/81423085b45111fa0768742810b73a77d2af06dd/datafusion/substrait/src/logical_plan/consumer/expr/literal.rs#L127-L137) - consumes literal values with precision 0/3/6/9 → Second/Milli/Micro/Nano
- [`PrecisionTimestampTz` literals](https://github.com/apache/datafusion/blob/81423085b45111fa0768742810b73a77d2af06dd/datafusion/substrait/src/logical_plan/consumer/expr/literal.rs#L138-L160) - consumes timezone-aware literal values
- [`PrecisionTime` literals](https://github.com/apache/datafusion/blob/81423085b45111fa0768742810b73a77d2af06dd/datafusion/substrait/src/logical_plan/consumer/expr/literal.rs#L162-L200) - consumes time literal values as `Time32`/`Time64` scalars

tests/baseline.json
```diff
@@ -6,11 +6,11 @@
     "num_aggregate_functions": 29,
     "num_scalar_functions": 170,
     "num_window_functions": 11,
-    "num_function_overloads": 533
```
benbellick, line 9: Why did this one go up so much?

tests/cases/datetime/add_datetime.test
```diff
@@ -1,15 +1,15 @@
...
-add('2020-12-31'::date, 'P5D'::iday) = '2021-01-05T00:00:00'::ts
-add('2020-12-31'::date, 'P5Y'::iyear) = '2025-12-31T00:00:00'::ts
-add('2020-12-31'::date, 'P5M'::iyear) = '2021-05-31T00:00:00'::ts
+# date_to_timestamp: examples using the date types and resulting in a precision_timestamp
+add('2020-12-31'::date, 'P5D'::iday<6>) = '2021-01-05T00:00:00'::pts<6>
+add('2020-12-31'::date, 'P5Y'::iyear, 6::i8) = '2025-12-31T00:00:00'::pts<6>
+add('2020-12-31'::date, 'P5M'::iyear, 9::i8) = '2021-05-31T00:00:00'::pts<9>
 
 # null_input: examples with null args or return
-add(null::date, 'P5D'::iday) = null::ts
```
benbellick: I expect these to fail now that #989 is in. Probably will have to fix the conflicts here.

extensions/functions_datetime.yaml
benbellick: I have no idea what the level of support is for `interger_parameter` actually is in the downstream libraries. It doesn't seem very well specified, considering that the only line about it is:
> * `integer_parameter(string name) => integer`

I propose we tackle this in a separate PR where we can have dedicated discussion about it.

tests/cases/datetime/add_datetime.test
nielspardon: yes, needed a rebase

tests/baseline.json
nielspardon: I've been asking Claude for an explanation and it gave a very detailed one:


> When the deprecated time, timestamp, and timestamp_tz types were removed, they were replaced with parameterized types like precision_time<P> and precision_timestamp<P>.
> 
> The massive increase in overloads (533 → 4,482) is due to parametric type expansion. Previously, functions like extract(HOUR, x: timestamp) had a single overload. Now with precision_timestamp<P>, the type system generates separate overloads for each supported precision value (microsecond, millisecond, etc.), causing a combinatorial explosion in the number of overload signatures.
> 
> For example:
> 
> Old: 1 overload for extract(HOUR, x: timestamp)
> New: Multiple overloads for extract(HOUR, x: precision_timestamp<microsecond>), extract(HOUR, x: precision_timestamp<millisecond>), etc.
> This pattern repeated across many datetime functions explains the ~8.4× increase. The later jump from 4,482 → 8,466 in commit 63f3fd5 likely comes from additional function additions or test coverage adjustments.

tests/cases/datetime/add_datetime.test
nielspardon: tests were failing but I updated them to handle the nullability correctly.

tests/baseline.json
benbellick: Hmm... if that is the case, what do you think about only deleting the old impls in this PR, and we can add new ones if we need in a subsequent one?

nielspardon: I'm only adding new impls when there would be no equivalent impls after removing the old ones. This PR retain parity in terms of functionality with before the change aka if there was an impl with time, timestamp, timestamp_tz before and there wouldn't be one after removing all impls using those then I'm adding the equivalent precision impl. If we do these in separate PRs then we risk having a Substrait release that has less functionality than before.

nielspardon: I have no idea what's going on here. The explanation that Claude Code gave does not match the current code on `main`. Currently, `num_function_overloads` is essentially the total number of `impls` across all extension YAMLs. I rebased this PR branch on latest main and now `num_function_overloads` decreases now from `533` to `498` which sounds more reasonable than the previous `8,466`.

extensions/functions_datetime.yaml
nielspardon: btw I had already created an issue for improving the documentation of these return type functions and had asked @jacques-n if he wants to contribute since he originally introduced these functions (unfortunately did not get any reaction from him) https://github.com/substrait-io/substrait/issues/959

PR #1005, opened 2026-03-12
feat(docs): clarify distinction between enumeration arguments and options

### LINKED ISSUE #995: Document when to use enum arguments vs function options
The spec provides two mechanisms for functions to accept a choice from a set of values, enum arguments and function options, but doesn't document when to use which. Both are used in core extensions but the extensions docs only show enum arguments in signature examples without explaining the distinction.

For example, `extract` uses an enum argument for the component:

https://github.com/substrait-io/substrait/blob/e4ce3f8710d91cb95cce010c6b59eb1d618bb2a6/extensions/functions_datetime.yaml#L71-L76

While `add` uses a function option for overflow behavior:

https://github.com/substrait-io/substrait/blob/e4ce3f8710d91cb95cce010c6b59eb1d618bb2a6/extensions/functions_arithmetic.yaml#L13-L17

They also look very similar in YAML, with both using the keyword `options`.

The extensions docs should clarify the difference between these two mechanisms and provide guidance on when to use each.

Related: #254

benbellick:
The spec has two distinct mechanisms for passing fixed-set string values to functions: enumeration arguments and options. The existing docs described options as "similar to a required enumeration" which muddied the distinction. This PR clarifies that enumeration arguments are always required and represent choices core to the function's semantics, while options are always optional and represent engine-level behavioral preferences.

Also removes the misleading `add(optional enumeration, ...)` example from the extensions index (the `add` function has options, not enum arguments) and drops the "Required Enumeration" / "optional enumeration" terminology in favor of just "Enumeration".

These changes are being proposed in light of the community sync conversation yesterday (Mar 11, 2026). 

Closes #995

-----

- - -
This change is [Reviewable](https://reviewable.io/reviews/substrait-io/substrait/1005)


site/docs/expressions/scalar_functions.md
```diff
@@ -22,7 +22,7 @@ There are three main types of arguments: value arguments, type arguments, and en
...
-* Enumeration: arguments that support a fixed set of declared values as constant arguments. These arguments must be specified as part of an expression. While these could also have been implemented as constant string value arguments, they are formally included to improve validation/contextual help/etc. for frontend processors and IDEs. An example might be `extract([DAY|YEAR|MONTH], <date value>)`. In this example, a producer must specify a type of date part to extract. Note, the value of a required enumeration cannot be used in type derivation.
+* Enumeration: arguments that require the caller to specify exactly one value from a fixed set of declared string values. They represent choices that are integral to the function's core semantics. An example is `extract([DAY|YEAR|MONTH], <date value>)`, where the caller must specify which date part to extract. Note, the value of an enumeration argument cannot be used in type derivation.
```
benbellick, line 26: I thought it made sense to drop the line about "while these could also have been implement as constant string..." as I thought that seemed like more of an implementation detail. People are generally familiar with the concept of an enum IMO.

site/docs/expressions/scalar_functions.md
```diff
@@ -41,7 +41,7 @@ There are three main types of arguments: value arguments, type arguments, and en
 | Name        | A human-readable name for this argument to help clarify use.        | Optional, defaults to a name based on position (e.g. `arg0`) |
 | Description | Additional description of this argument.                            | Optional                                                   |
 
-#### Required Enumeration Properties
```
benbellick, line 44: Consistent with the above now.

site/docs/extensions/index.md
```diff
@@ -122,11 +122,13 @@ A function signature uniquely identifies a function implementation within a sing
 
 #### Examples
 
+Function-level [options](../expressions/scalar_functions.md#options) are not part of the function signature and do not appear here. Only enumeration arguments (which are positional and required) contribute to the signature as `req`.
+
 | Function Signature                                | Function Name       |
 | ------------------------------------------------- | ------------------- |
-| `add(optional enumeration, i8, i8) => i8`         | `add:i8_i8`         |
```
benbellick, line 124: `add` doesn't take an enum, it takes an option. It is confusing to refer to this as an "optional enumeration"

site/docs/expressions/scalar_functions.md
nielspardon, line 48: The use of `Options` for the enum elements here is a bit peculiar given that the term`Options` in the next section is defined as a distinct concept. I do understand this is baked into our current extension YAML schema. Maybe something we also want to change to avoid confusion between the `options` of an enum argument vs the `options` of a function.

benbellick: I also don't love that part of the YAML schema. I would support changing it, but considering it would be breaking (and might warrant a migration 😅) I'm inclined to leave it for a future PR.

benbellick: It might be worth adding a warning section just acknowledging this confusion in the docs though, until it is resolved at least.

benbellick: I added it in the most recent commit to try and make that clear for the time being. 
https://github.com/substrait-io/substrait/pull/1005/files#diff-4779922140497bc0eeb3d0b59c4884aab559c44d307a4681e395821b5212a5fdR81-R84

site/docs/expressions/scalar_functions.md
```diff
@@ -51,19 +51,37 @@ There are three main types of arguments: value arguments, type arguments, and en
...
+| --- | --- | --- |
+| Required? | Yes, must be specified by the producer | No, may be omitted |
+| Semantics | Core to the function's operation (e.g., which date component to extract) | Behavioral preference for corner cases or engine-specific behavior (e.g., overflow handling) |
+| Position | Positional, part of the argument list | Named, separate from arguments |
+| In function signature? | Yes (as `req`) | No |
+| If omitted | Invalid plan | Consumer chooses its own behavior |
+
+!!! warning "YAML keyword overlap"
+
+    In the YAML extension format, both enumeration arguments and function-level options use the keyword `options`. An enumeration argument appears inside `args` as `options: [VAL1, VAL2, ...]`, while a function-level option appears under a top-level `options` key with named sub-keys and a `values` list. Take care not to confuse the two when reading or writing extension YAML files.
```
yongchul, line 83: I am confused. 😆 perhaps a short snippet with comment would work?

yongchul (approved):

site/docs/expressions/scalar_functions.md
benbellick: Sorry getting back to you late here. I can include an example. Another option would be to just drop this section entirely and open up a ticket to make this more clear in the future. As an example here: 
```yaml
  -
    name: "add"
    description: "Add two values."
    impls:
      - args:
          - name: x
            value: i8
          - name: y
            value: i8
        options: # this is an option
          overflow:
        return: i8
```

```yaml
  -
    name: extract
    # ...
      - args:
          - name: component
            options: [ YEAR, ISO_YEAR, US_YEAR, HOUR, MINUTE, SECOND,
                       MILLISECOND, MICROSECOND, SUBSECOND, UNIX_TIME ] # and this is an enum
            description: The part of the value to extract.
          - name: x
            value: timestamp
```

I was just trying to call out that both use the YAML key name `options` which could be confusing 🤔

site/docs/expressions/scalar_functions.md
yongchul: I actually do not want to use string but somehow declare the enum in the YAML and reference... The key idea remains the same though.

site/docs/expressions/scalar_functions.md
```diff
@@ -51,19 +51,51 @@ There are three main types of arguments: value arguments, type arguments, and en
...
+### Enumeration Arguments vs Options
+
+Both enumeration arguments and options accept values from a fixed set of strings, but they serve different purposes. As a general rule, prefer enumeration arguments for anything the caller must decide for the function to be meaningful. Options are best suited for engine-level behavioral properties (like overflow or rounding mode) where a reasonable default exists and the caller may not care which behavior the engine picks.
+
+| | Enumeration Argument | Option |
+| --- | --- | --- |
+| Required? | Yes, must be specified by the producer | No, may be omitted |
+| Semantics | Core to the function's operation (e.g., which date component to extract) | Behavioral preference for corner cases or engine-specific behavior (e.g., overflow handling) |
+| Position | Positional, part of the argument list | Named, separate from arguments |
+| In function signature? | Yes (as `req`) | No |
```
yongchul, line 78: perhaps we may revisit this `req` at some point...

yongchul (approved):

yongchul:
+1. I see a couple of follow ups but not a blocker for this PR.

PR #1007, opened 2026-03-13
fix(extensions)!: correct return type for `add:date_iyear` operation

nielspardon:
BREAKING CHANGE: changes the return type of the `add:date_iyear` function from `timestamp` to `date`

Fix the return type of the add function when adding an interval_year to a date. Since interval_year is an interval of years and months the operation should return a date type, not a timestamp type.

This corrects the function signature in functions_datetime.yaml where date + interval_year was incorrectly specified to return timestamp instead of date.

The function signature for `subtract:date_iyear` already correctly returns a `date`.
https://github.com/substrait-io/substrait/blob/150d86d10ce10384685513715c78d63a9a6c1e37/extensions/functions_datetime.yaml#L431-L436

### DIFF
--- a/extensions/functions_datetime.yaml
+++ b/extensions/functions_datetime.yaml
@@ -278,7 +278,7 @@ scalar_functions:
             value: date
           - name: y
             value: interval_year
-        return: timestamp
+        return: date
       - args:
           - name: x
             value: timestamp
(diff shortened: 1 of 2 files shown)


extensions/functions_datetime.yaml
```diff
@@ -278,7 +278,7 @@ scalar_functions:
             value: date
           - name: y
             value: interval_year
-        return: timestamp
+        return: date
```
yongchul, line 281: Good catch! I am wondering how many of these mistakes are out there. I'll try to run AI and see whether it is capable of finding these. 😄

yongchul (approved):

vbarua (approved):
LGTM. This is a breaking change as it modifies a return type, but also it shouldn't have worked if you exercised it anyways. Good catch.

PR #1008, opened 2026-03-13
ci: fix commitlint tinyexec module resolution failure

benbellick:
npx fetches commitlint into an isolated cache that fails to resolve the
tinyexec transitive dependency. Install @commitlint/cli explicitly
alongside the config and use --no-install to avoid the broken npx
resolution.


vbarua:
@/mbwhite helped chase this down: https://github.com/conventional-changelog/commitlint/issues/4657

If you re-run the failing conventional commit jobs, they should Just Work (TM) now.

PR #1009, opened 2026-03-17
feat: add TopNRel physical operator with WITH TIES support

jcamachor:
Add TopNRel message to algebra.proto as a physical relation that combines sort and fetch into a single operator. This fills the gap between the existing Top-N documentation in physical_relations.md and the protobuf definitions.

The operator supports:
- Sort fields (reusing SortField)
- Offset and count as expressions
- WITH TIES semantics via fetch mode enum

Update physical_relations.md to reflect the new properties including WITH TIES documentation.

- - -
This change is [Reviewable](https://reviewable.io/reviews/substrait-io/substrait/1009)


jcamachor:
@yongchul , @vbarua , @jacques-n , any thoughts on this proposal? The documentation for TopNRel already existed though it had not been added to the proto spec.

proto/substrait/algebra.proto
```diff
@@ -557,6 +557,30 @@ message ExpandRel {
...
+  // records to skip. Null is treated as 0. Recommended type is int64.
+  Expression offset = 4;
+
+  // Expression evaluating to a non-negative integer specifying the number of
+  // records to return. Null signals ALL. Recommended type is int64.
+  Expression count = 5;
+
+  // If true, additional rows tied with the last row (per the sort fields) are
+  // also included.
+  bool with_ties = 6;
```
yongchul: This is fine with me but some may prefer `enum`. 😆

site/docs/relations/physical_relations.md
```diff
@@ -136,7 +136,9 @@ A receiving operation that will merge multiple streams in an arbitrary order.
 
 ## Top-N Operation
 
-The top-N operator reorders a dataset based on one or more identified sort fields as well as a sorting function. Rather than sort the entire dataset, the top-N will only maintain the total number of records required to ensure a limited output. A top-n is a combination of a logical sort and logical fetch operations.
+The top-N operator reorders a dataset based on one or more identified sort fields as well as a sorting function. Rather than sort the entire dataset, the top-N will only maintain the total number of records required to ensure a limited output. A top-N is a combination of a logical sort and logical fetch operations.
```
yongchul, line 139: Thanks for proposal. I have a COI with this proposal 😆 but this is physical, related but different. I am fine with this proposal (looks *very* familiar) but will still revive and push for the logical counterpart.

proto/substrait/algebra.proto
jcamachor: I've created an enum and kept it outside of TopN so it can be easily re-used in the logical counterpart.

site/docs/relations/physical_relations.md
vbarua: > (looks very familiar) 

This is similar-ish to what you were thinking about in https://github.com/substrait-io/substrait/pull/797 right?

> will still revive and push for the logical counterpart.

I think this can model all of what you wanted to handle in that PR, or am I incorrect. Instead of having another logical node alongside this, I might even suggest just treating this as a logical node.

At this point I'd be pretty okay with having:
* LimitRel
* SortRel
* TopNRel / SortLimitRel

even if the last one overlaps with the other two.

jcamachor: Thanks, @vbarua, @yongchul. Whether we keep TopNRel as physical and augment LimitRel, or reclassify TopNRel as logical, doesn't seem like a blocker for this PR. But let me know if you'd like me to make any changes. I've just merged latest main into the PR.

vbarua: Let's assume that we'll only have one of either TopNRel or SortLimitRel, and we'll make it a logical relation. Which name do we prefer?

vbarua: I asked this, and it helped me realized that TopN is the better name because it's more specific and helps explain why you can't just FetchRel over a SortRel.

proto/substrait/algebra.proto
```diff
@@ -557,6 +557,39 @@ message ExpandRel {
...
+  // Expression evaluating to a non-negative integer specifying the number of
+  // records to skip. Null is treated as 0. Recommended type is int64.
+  Expression offset = 4;
+
+  // Expression evaluating to a non-negative integer specifying the number of
+  // records to return. Null signals ALL. Recommended type is int64.
+  Expression count = 5;
+
+  // Determines how to handle rows tied with the last returned row.
+  LimitMode mode = 6;
```
vbarua: What do you think about using FetchMode for this? It's both consistent with our usage of FetchRel instead of LimitRel, and also more consistent with the SQL spec language.

proto/substrait/algebra.proto
```diff
@@ -557,6 +557,39 @@ message ExpandRel {
...
+
+  substrait.extensions.AdvancedExtension advanced_extension = 7;
+}
+
+// Determines how a limit operation handles rows tied with the last returned row.
+enum LimitMode {
+  // This default value implies LIMIT_MODE_ROWS_ONLY.
```
vbarua: I would push for not having a default and forcing producers to populate this correctly. It helps make plans more readable in isolation, because you don't have to remember what the default is when it's not set.

site/docs/relations/physical_relations.md
```diff
@@ -147,12 +149,13 @@ The top-N operator reorders a dataset based on one or more identified sort field
...
+| Property          | Description                                                                                                                                                                                                | Required                           |
+| ----------------- | ---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- | ---------------------------------- |
+| Input             | The relational input.                                                                                                                                                                                      | Required                           |
+| Sort Fields       | List of one or more fields to sort by. Uses the same properties as the [orderedness](basics.md#orderedness) property.                                                                                      | One sort field required            |
```
vbarua:
```suggestion
| Sort Fields       | List of one or more fields to sort by. Uses the same properties as the [orderedness](basics.md#orderedness) property.                                                                                      | At least one field is required     |
```

proto/substrait/algebra.proto
jcamachor: Sounds good, done!

proto/substrait/algebra.proto
jcamachor: Done.

site/docs/relations/physical_relations.md
jcamachor: Thanks, done!

PR #1010, opened 2026-03-17
feat(tests): add enum argument support to FuncTestCase grammar

nielspardon:
In the FuncTestCase grammar enum arguments where not handled as enum arguments but as string literals (only example so far the datetime `extract` function). 

- This PR changes the grammar to explicitly support enum arguments with a new `enum` data type in the grammar.
- The Python code is regenerated from the grammar.
- The `extract` test cases are updated to use the new syntax.


nielspardon:
I rebased and resolved the conflicts. Github offers me to merge the PR myself with the given approvals. Is it ok if I do this as a committer or should a PMC member push the button here?

PR #1011, opened 2026-03-17
fix(extensions): add signatures with distribution enum arg for std_dev and variance

nielspardon:
Given the [recent clarification around function options vs enum arguments](https://github.com/substrait-io/substrait/pull/1005) I do believe the `std_dev` and `variance` function definitions should change the `distribution` function to be an enum argument.

The `distribution` parameter changes the semantic behavior of these functions which is why SQL systems typically model the different `distribution` values as distinct functions e.g. `STDDEV_SAMP` and `STDDEV_POP`. Providing the `distribution` parameter thus is not optional but required.

This PR deprecates the existing function signatures for `std_dev` and `variance` and adds new signatures that use a `distribution` enum argument instead of a function option.

This PR depends on #1010 whose changes are included in this PR since it uses the new enum argument syntax in the test cases I generated using AI for the two functions.

The main changes for this PR are in the second commit 8ec63ae102b8b3ffb1d896fa47c30c8fc757bed6.

### DIFF
--- a/extensions/functions_arithmetic.yaml
+++ b/extensions/functions_arithmetic.yaml
@@ -1403,10 +1403,10 @@ aggregate_functions:
         nullability: DECLARED_OUTPUT
         return: fp32?
       - args:
-          - name: x
-            value: fp32
           - name: distribution
             options: [ SAMPLE, POPULATION]
+          - name: x
+            value: fp32
         options:
           rounding:
             values: [ TIE_TO_EVEN, TIE_AWAY_FROM_ZERO, TRUNCATE, CEILING, FLOOR ]
@@ -1423,10 +1423,10 @@ aggregate_functions:
         nullability: DECLARED_OUTPUT
         return: fp64?
       - args:
-          - name: x
-            value: fp64
           - name: distribution
             options: [ SAMPLE, POPULATION]
+          - name: x
+            value: fp64
         options:
           rounding:
             values: [ TIE_TO_EVEN, TIE_AWAY_FROM_ZERO, TRUNCATE, CEILING, FLOOR ]
@@ -1446,10 +1446,10 @@ aggregate_functions:
         nullability: DECLARED_OUTPUT
         return: fp32?
       - args:
-          - name: x
-            value: fp32
           - name: distribution
             options: [ SAMPLE, POPULATION]
+          - name: x
+            value: fp32
         options:
           rounding:
             values: [ TIE_TO_EVEN, TIE_AWAY_FROM_ZERO, TRUNCATE, CEILING, FLOOR ]
@@ -1466,10 +1466,10 @@ aggregate_functions:
         nullability: DECLARED_OUTPUT
         return: fp64?
       - args:
-          - name: x
-            value: fp64
           - name: distribution
             options: [ SAMPLE, POPULATION]
+          - name: x
+            value: fp64
         options:
           rounding:
             values: [ TIE_TO_EVEN, TIE_AWAY_FROM_ZERO, TRUNCATE, CEILING, FLOOR ]
--- a/tests/cases/arithmetic/std_dev.test
+++ b/tests/cases/arithmetic/std_dev.test
@@ -2,52 +2,52 @@
 ### SUBSTRAIT_INCLUDE: '/extensions/functions_arithmetic.yaml'
 
 # basic: Basic examples without any special cases
-((1.0, 2.0, 3.0, 4.0, 5.0)) std_dev(col0::fp32, SAMPLE::enum) = 1.5811388::fp32?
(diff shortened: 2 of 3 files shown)


benbellick:
I'm in support of this change, but think we should wait until #1005 lands to make sure we all agree on the meaning of options / enums.

extensions/functions_arithmetic.yaml
```diff
@@ -1395,21 +1395,21 @@ aggregate_functions:
       - args:
           - name: x
             value: fp32
+          - name: distribution
+            options: [ SAMPLE, POPULATION]
```
yongchul, line 1407: The change makes sense to me however wondering whether we have to make a breaking change. The proposal can have the new functions as separate entry and the old ones can be annotated as deprecated (we need a mechanism to do this) or at least some comments.

nielspardon: I added a `deprecated` flag for types, type variations, functions and function signatures into `simple_extensions_schema.yaml`, added back and deprecated the old signatures using options

nielspardon: ah, I see you opened this PR: https://github.com/substrait-io/substrait/pull/1014

nielspardon:
Rebased, resolved conflicts, used the new deprecation syntax.

nielspardon:
I merged #1010 and rebased this PR

nielspardon:
updated the deprecation version to `0.87.0` in b364972e5e94ef7fdafe717c467350f9718db800

nielspardon:
removed the deprecation of the old function signatures per conversation on Slack

extensions/functions_arithmetic.yaml
```diff
@@ -1404,12 +1404,32 @@ aggregate_functions:
         return: fp32?
       - args:
           - name: x
-            value: fp64
+            value: fp32
+          - name: distribution
+            options: [ SAMPLE, POPULATION]
```
vbarua, line 1407: minor suggestion
```
      - args:
          - name: distribution
            options: [ SAMPLE, POPULATION]
          - name: x
            value: fp32
```

I would use `distribution` as the first argument for readability. The enum args will always be a fixed size roughly, but `x` can be an arbitrarily deep expression.

In pseudo-invocations
```
std_dev(POPULATION, <really big expression>)
```
as opposed to
```
std_dev(<really big expression>, POPULATION)
```

extensions/functions_arithmetic.yaml
```diff
@@ -1404,12 +1404,32 @@ aggregate_functions:
...
             values: [ SAMPLE, POPULATION]
+          rounding:
+            values: [ TIE_TO_EVEN, TIE_AWAY_FROM_ZERO, TRUNCATE, CEILING, FLOOR ]
+        nullability: DECLARED_OUTPUT
+        return: fp64?
+      - args:
+          - name: x
+            value: fp64
+          - name: distribution
+            options: [ SAMPLE, POPULATION]
```
vbarua, line 1427: Same as above.

extensions/functions_arithmetic.yaml
```diff
@@ -1419,20 +1439,40 @@ aggregate_functions:
...
+          rounding:
+            values: [ TIE_TO_EVEN, TIE_AWAY_FROM_ZERO, TRUNCATE, CEILING, FLOOR ]
         nullability: DECLARED_OUTPUT
         return: fp32?
       - args:
           - name: x
-            value: fp64
+            value: fp32
+          - name: distribution
+            options: [ SAMPLE, POPULATION]
```
vbarua, line 1450: Same as above

extensions/functions_arithmetic.yaml
```diff
@@ -1419,20 +1439,40 @@ aggregate_functions:
...
             values: [ SAMPLE, POPULATION]
+          rounding:
+            values: [ TIE_TO_EVEN, TIE_AWAY_FROM_ZERO, TRUNCATE, CEILING, FLOOR ]
+        nullability: DECLARED_OUTPUT
+        return: fp64?
+      - args:
+          - name: x
+            value: fp64
+          - name: distribution
+            options: [ SAMPLE, POPULATION]
```
vbarua, line 1470: Same as above

extensions/functions_arithmetic.yaml
nielspardon: done

extensions/functions_arithmetic.yaml
nielspardon: done

extensions/functions_arithmetic.yaml
nielspardon: done

extensions/functions_arithmetic.yaml
nielspardon: done

PR #1014, opened 2026-03-19
feat(extensions): support deprecation info in extensions

yongchul:
There are a few recent PRs making breaking changes to extensions but we can do this more gracefully.

This PR proposes adding deprecation information to all kinds of extensions: `type`, `type variations`, `scalar`, `aggregate`, and `window` functions so that we can present deprecation information to the consumers. This also makes possible tooling leverage this information to produce warnings.

# AI Disclaimer
* Claude Opus 4.6
* Example YAML was generated by AI
* Documentation was edited by AI

### DIFF
--- a/site/examples/extensions/deprecation_example.yaml
+++ b/site/examples/extensions/deprecation_example.yaml
@@ -1,7 +1,7 @@
 # Example showing deprecation of extensions
 urn: extension:org.example:deprecation_examples
 
-# A deprecated type using Substrait version
+# A deprecated type
 types:
   - name: old_type
     deprecated:
@@ -11,7 +11,7 @@ types:
       x: i32
       y: i32
 
-# A deprecated scalar function using generic version
+# A deprecated scalar function
 scalar_functions:
   - name: "old_func"
     deprecated:
@@ -23,7 +23,7 @@ scalar_functions:
             value: i32
         return: i32
 
-  # A function with a deprecated impl using generic version
+  # A function with a deprecated impl
   - name: "evolving_func"
     impls:
       - args:
@@ -38,7 +38,7 @@ scalar_functions:
           reason: "Use the fp32 overload instead"
         return: i32
 
-  # A function with a deprecated impl using generic version
+  # A function with a deprecated impl with metadata
   - name: "deprecated_func_with_metadata"
     deprecated:
       since: "2.0.0"


text/simple_extensions_schema.yaml
```diff
@@ -205,13 +209,23 @@ $defs:
...
   ordered:
     type: boolean
+  deprecationStatus:
+    type: object
+    additionalProperties: false
+    properties:
+      deprecatedSince: # Substrait version since which the function is deprecated (for documentation only)
+        type: string
+      description: # description of the deprecation (for documentation only)
+        type: string
```
nielspardon: I guess you could shorten the `deprecatedSince` to just `since` given that you did use `description` and not `deprecationDescription`:
```suggestion
      since: # Substrait version since which the function is deprecated (for documentation only)
        type: string
      description: # description of the deprecation (for documentation only)
        type: string
```

vbarua: Thinking about how the whole messages look like in context, what about something like:
```
deprecated:
  version: '0.82.4'
  reason: "this is why this was deprecated"
```

text/simple_extensions_schema.yaml
```diff
@@ -205,13 +209,23 @@ $defs:
     type: number
   ordered:
     type: boolean
+  deprecationStatus:
+    type: object
+    additionalProperties: false
+    properties:
+      deprecatedSince: # Substrait version since which the function is deprecated (for documentation only)
```
benbellick: Instead of using a comment, can you use a description field? E.g.
https://github.com/substrait-io/substrait/blob/62147cf26b4020ee8433f506809f410cc40d5806/text/simple_extensions_schema.yaml#L178-L190

This way, the documentation is available via YAML LSPs. Same for the next item, description (which unfortunately will then have `description.description` 😅)

text/simple_extensions_schema.yaml
```diff
@@ -205,13 +209,23 @@ $defs:
     type: number
   ordered:
     type: boolean
+  deprecationStatus:
+    type: object
+    additionalProperties: false
```
benbellick, line 220: Jumping ahead of the curve here, is it possible we will want to allow arbitrary extension points here via a `metadata` field (or some other name)? So that non-core extension authors can extend this `deprecationStatus` field.

text/simple_extensions_schema.yaml
yongchul: Thanks! Like the suggestion!

text/simple_extensions_schema.yaml
yongchul: done

text/simple_extensions_schema.yaml
yongchul: done

text/simple_extensions_schema.yaml
```diff
@@ -205,13 +209,33 @@ $defs:
...
+          Substrait version since which the extension is deprecated.
+          Should follow semantic versioning format (e.g. "1.2.3").
+        type: string
+      reason:
+        description: >-
+          Description of the deprecation (for documentation only).
+        type: string
+      metadata:
+        description: >-
+          Arbitrary data created by extension author (for documentation only).
```
benbellick:
```suggestion
          Arbitrary data created by extension author.
```
Do we want to enforce that? I think of these extension points as kind of free-for-alls 😆

text/simple_extensions_schema.yaml
```diff
@@ -205,13 +209,33 @@ $defs:
...
+  deprecationStatus:
+    type: object
+    additionalProperties: false
+    required: [since]
+    properties:
+      since:
+        description: >-
+          Substrait version since which the extension is deprecated.
+          Should follow semantic versioning format (e.g. "1.2.3").
+        type: string
```
benbellick: I am neutral here, but we could add a regex in the properties, as I am attempting to do [here](https://github.com/substrait-io/substrait/pull/881/files#diff-4080c304939982e966bde3c7802c81077680453c16820559fc91b1afe17e17e0R9-R10).

text/simple_extensions_schema.yaml
```diff
@@ -205,13 +209,33 @@ $defs:
     type: number
   ordered:
     type: boolean
+  deprecationStatus:
+    type: object
+    additionalProperties: false
+    required: [since]
+    properties:
+      since:
+        description: >-
```
benbellick, line 224: Ah I just realized something... which is that internally we would like to use this, but that means that we will need to deprecate based off of something different than substrait version. We don't have versions internally, but at the very least, it doesn't make sense to always mark an internal deprecation as being tied to a particular substrait version. 

Does it make sense to make the since field a "oneof" which is either:
- `substraitVersion`
- `other`

?

yongchul: Added `substraitVersion`, `sementicVersion`, and `genericVersion`.

site/docs/extensions/index.md
```diff
@@ -161,6 +161,25 @@ Example:
 
 Consumers of extension files are not required to understand or validate metadata fields.
 
+### Deprecation of Extensions
+
+Any extension entry can be deprecated at some point and producing a plan using deprecated extensions is **highly discouraged** for forward compatibility. The `deprecated` field can be added to types, type variations, functions (scalar, aggregate, and window), and individual function implementations.
```
benbellick:
```suggestion
An extension entry can be deprecated. Producing a plan using deprecated extensions is discouraged for forward compatibility. The `deprecated` field can be added to types, type variations, functions (scalar, aggregate, and window), and individual function implementations.
```
Just suggesting a slightly less severe tone, feel free to ignore

site/docs/extensions/index.md
```diff
@@ -161,6 +161,25 @@ Example:
...
+
+Consumers of extension files are not required to understand or validate deprecation fields.
+
+The `deprecated` field requires a `since` object that specifies the version at which the entry was deprecated. Three versioning schemes are supported:
+
+* `substraitVersion` — a [semantic version](https://semver.org) string denoting since which Substrait release the extension has been deprecated. Substrait core extensions use this scheme.
+* `semanticVersion` — a [semantic version](https://semver.org) string for extensions that maintain their own versioning independent of Substrait releases.
+* `genericVersion` — a freeform versioning scheme with `name` (the versioning scheme name) and `version` (the version string) fields. The interpretation is up to the extension authors.
```
benbellick: I say let's keep things simple for now and just have only `substraitVersion` and `genericVersion`, though I don't feel super strong about this. It just feels to me like adding `semanticVersion` is just adding a special case of `genericVersion`. 

If others would prefer having both then that is fine with me.

nielspardon: I'm also in favor of keeping things simple. Couldn't we just add an optional `version` field for each YAML file and if it's not set it is implicitly the substrait version?

yongchul: @nielspardon if we put file-level default, then we need to make `reason` required if I understood correctly how this YAML thing works? I think it is okay to just have it per entry level and being explicit as we don't have too many deprecated stuff...

 Also, @benbellick  and I discussed offline briefly about introducing per YAML file version and deprecation can be with respect to the YAML file version. We may explore this idea with future URN changes. So please let me know what do you think.

@benbellick I just put `semanticVersion` there because it is such a popular scheme. But in the end, we still need to know which version you need to follow... without anchoring point, there is no meaning to compare verisons.  Will remove the semanticVersion.

site/examples/extensions/deprecation_example.yaml
```diff
@@ -0,0 +1,43 @@
...
+    structure:
+      x: i32
+      y: i32
+
+# A deprecated scalar function using semantic version
+scalar_functions:
+  - name: "old_func"
+    deprecated:
+      since:
+        semanticVersion: "1.2.0"
```
benbellick: Should this one say `substraitVersion` or be dropped from the examples?

text/simple_extensions_schema.yaml
```diff
@@ -205,13 +209,66 @@ $defs:
     type: number
   ordered:
     type: boolean
+  semanticVersionString:
+    type: string
+    pattern: "^[0-9]+\\.[0-9]+\\.[0-9]+$"
```
benbellick: 😢 [semver recommended regex is quite complex](https://semver.org/#is-there-a-suggested-regular-expression-regex-to-check-a-semver-string)

We could just include it and put a comment to that link?

site/examples/extensions/deprecation_example.yaml
yongchul: reworked.

site/docs/extensions/index.md
nielspardon: > without anchoring point, there is no meaning to compare verisons.

Yes, that was also my main thought with introducing a version per YAML file since then you can deprecate relative to the YAML file version. My understanding is that for the default YAML files that come with the Substrait spec we use the same version as the Substrait spec by convention. 3rd party extension YAMLs would need to explicitly define a version per file.

nielspardon: if we assume for a second that this relatively short extension file was a 3rd party one:

```
%YAML 1.2
---
urn: extension:org.example:functions_aggregate_approx
version: 1.2.3
aggregate_functions:
  - name: "approx_count_distinct"
    description: >-
      Calculates the approximate number of rows that contain distinct values of the expression argument using
      HyperLogLog. This function provides an alternative to the COUNT (DISTINCT expression) function, which
      returns the exact number of rows that contain distinct values of an expression. APPROX_COUNT_DISTINCT
      processes large amounts of data significantly faster than COUNT, with negligible deviation from the exact
      result.
    deprecated:
      since: 1.2.0
      reason: deprecated in favor of exact count distinct
    impls:
      - args:
          - name: x
            value: any
        nullability: DECLARED_OUTPUT
        decomposable: MANY
        intermediate: binary
        return: i64

```

then it is clear that `since` is relative to the top-level `version` number

nielspardon: there is no need for `substraitVersion` vs `genericVersion` vs `semanticVersion`

benbellick: I think introducing versioning for YAML files opens up a whole can of worms. We may want to tackle that eventually, but it is worth its own dedicated PR and discussion. 

At the current moment, there is nothing stopping extension authors outside of the core extension set from managing their own _internal_ version schema via one of the `metadata` endpoints.

nielspardon: for default Substrait extensions you would just not populate the version field since it would be the same as the Substrait version of the `Plan` message:

```
%YAML 1.2
---
urn: extension:io.substrait:functions_aggregate_approx
aggregate_functions:
  - name: "approx_count_distinct"
    description: >-
      Calculates the approximate number of rows that contain distinct values of the expression argument using
      HyperLogLog. This function provides an alternative to the COUNT (DISTINCT expression) function, which
      returns the exact number of rows that contain distinct values of an expression. APPROX_COUNT_DISTINCT
      processes large amounts of data significantly faster than COUNT, with negligible deviation from the exact
      result.
    deprecated:
      since: 1.2.0
      reason: deprecated in favor of exact count distinct
    impls:
      - args:
          - name: x
            value: any
        nullability: DECLARED_OUTPUT
        decomposable: MANY
        intermediate: binary
        return: i64
```

nielspardon: > I think introducing versioning for YAML files opens up a whole can of worms. We may want to tackle that eventually, but it is worth its own dedicated PR and discussion.
> 
> At the current moment, there is nothing stopping extension authors outside of the core extension set from managing their own _internal_ version schema via one of the `metadata` endpoints.

So what are you actually proposing here? adding the multitude of versioning fields as you initially proposed and then clean up later? I'm a bit confused.

text/simple_extensions_schema.yaml
yongchul: Pondered this, and the expression is too complex and we don't need that complexity (capturing tags and release) as a spec IMO. So I'll stick to the simpler regex for now. We can always increase the complexity later. 😄

site/docs/extensions/index.md
benbellick: Honestly, I'm okay with dropping my proposal. You can technically put any extra deprecation info you want into the existing metadata fields. 

I was just trying to allow a way for unofficial extensions to mark themselves as deprecated in an "official" way without (so their deprecation has nothing to do with substrait versioning). My suggestion was to just add an extension point where people can put whatever string they want.

nielspardon: btw. another simple solution that avoids the whole versioning debate would be to just add a boolean `deprecated` field as I did here not being aware that YongChul started this PR: https://github.com/substrait-io/substrait/pull/1011/changes/44c2c5b47539135920f8272a6bed1fbc7b4b16c7

yongchul: @nielspardon thanks for the example. Now I got what you mentioned and I think I like the idea. In fact, adding `boolean` field was indeed my first approach but then I realized that it is not particularly useful without which release it has been deprecated for official documentation. So here we are. 😄 

The main points for this is following.
* versioning scheme
* granularity of versioning

For Substrait, it appears that we all agree that we use Substrait release [granularity] version [semantic version].
 
For third party, we could leave the version field free form (i.e., just string and not enforcing anything) for maximal flexibility but we lose systematic validation and perhaps tracking as well.
Or, we can enforce the version must be following semantic versioning scheme.
Or, we can hand wave that if it look like semantic.

That's where my attempt to genericSemanticVersion to handle things in systematic way...

Here is probably the middle ground, just to unblock the official extensions.

```yaml
deprecated:
  since: "X.Y.Z" # this can be extended with oneOf to support other convention but by default it is semantic version string.
```

and that's it. If third party happens to want to use it, it can as long as they use semantic version string (and of course use the metadata field as well).

We can continue debate whether we want file level version field or not in other PR or issue.

How does that sound?

text/simple_extensions_schema.yaml
```diff
@@ -205,13 +209,38 @@ $defs:
...
+      We intentiaonally allow only core components (major.minor.patch) of the semantic version string.
+  deprecationStatus:
+    type: object
+    additionalProperties: false
+    required: [since]
+    properties:
+      since:
+        description: >-
+          Version since when the item is deprecated.
+        $ref: "#/$defs/coreSemanticVersionString"
```
yongchul, line 226: @benbellick @nielspardon thanks for reviewing and discussion!

PTAL this simplified version -- almost going back to the initial proposal.

* `since` is now only semantic version string.
* If we need to add more explicit versioning scheme, we can make the `since` as oneof in a backward compatible way to incorporate other scheme.
* Documented explicitly what this means for core substrait extensions and intentionally underspecified  third-party extensions.
* In a follow up PR or issue, we can further discuss standardizing the deprecation in third party extensions. We can actually punt this until someone in pain steps up. ;-)

yongchul: I believe this is sufficient to unblock the breaking changes in core extensions in more graceful way.

site/examples/extensions/deprecation_example.yaml
```diff
@@ -0,0 +1,39 @@
+# Example showing deprecation of extensions using different versioning schemes
+urn: extension:io.substrait:deprecation_examples
```
nielspardon: what about using `org.example` here instead of `io.substrait` to make it even more obvious that this is an example?
```suggestion
urn: extension:org.example:deprecation_examples
```

text/simple_extensions_schema.yaml
nielspardon: thanks, makes sense to me and also thanks for being mindful of unblocking other PRs quickly depending on this. hopefully there are enough other PMCs members around so we can get the minimum approvals required in a reasonable time

text/simple_extensions_schema.yaml
```diff
@@ -205,13 +209,38 @@ $defs:
     type: number
   ordered:
     type: boolean
+  coreSemanticVersionString:
+    type: string
+    pattern: "^(0|[1-9]\\d*)\\.(0|[1-9]\\d*)\\.(0|[1-9]\\d*)$"
```
nielspardon, line 214: just curious, wouldn't this regex be a little simpler and achieve the same:

```suggestion
    pattern: "^(\\d+)\\.(\\d+)\\.(\\d+)$"
```

nielspardon: nevermind, the current regex prevents matching multiple zeros like: `00.00.00` does not match, only single zeros match `0.0.0`

text/simple_extensions_schema.yaml
```diff
@@ -205,13 +209,38 @@ $defs:
...
+          Version since when the item is deprecated.
+        $ref: "#/$defs/coreSemanticVersionString"
+      reason:
+        description: >-
+          Description of the deprecation (for documentation only).
+        type: string
+      metadata:
+        description: >-
+          Arbitrary data created by extension author.
+        type: object
```
nielspardon, line 234: do you also want to add an example showing the use of the `metadata` field for completeness?

site/examples/extensions/deprecation_example.yaml
```diff
@@ -0,0 +1,52 @@
+# Example showing deprecation of extensions
+urn: extension:org.example:deprecation_examples
+
+# A deprecated type using Substrait version
```
vbarua: minor: We use both `using Substrait version` and `using generic version`? Is there a distinction between the two, or is this an artifact from an earlier version?

yongchul: forgot to update that. thanks!

PR #1015, opened 2026-03-19
fix(extensions)!: random extraneous argument for repeat varchar

mbwhite:
The extension definition for `repeat()` is correct for type `str` but for the `varchar` there was a random `value` entry.


nielspardon:
who's doing the honors and merges this PR?

PR #1019, opened 2026-03-23
feat(extensions)!: deprecate std_dev and variance using function options

nielspardon:
BREAKING CHANGE: deprecates the function signatures for std_dev and variance using function options in favor of the versions using enum arguments

Follow-up to and depending on #1011 main change in commit d20dde393612d6501e517316ea886399b94fe2a4

- - -
This change is [Reviewable](https://reviewable.io/reviews/substrait-io/substrait/1019)


nielspardon:
rebased on latest `main` which fixed conflicts and update the deprecation version to `0.88.0` since this PR missed the `0.87.0` release

jacques-n (approved):

Issue #946, opened 2026-01-28
pixi-fy substrait repo tooling


nielspardon:
While working on https://github.com/substrait-io/substrait/pull/945 I really missed the convenience of having [pixi](https://pixi.prefix.dev/latest/) take care of setting up a consistent and reproducible dev environment [like we recently got in substrait-python](https://github.com/substrait-io/substrait-python/pull/137).

I would like the setup of ANTLR, Python dependencies and dev tools like black, yamllint, etc to be included.

PR #1021, opened 2026-03-24
build: use pixi to manage build environment

nielspardon:
fixes https://github.com/substrait-io/substrait/issues/946

Migrates the build logic from using a fragmented approach requiring many manual setup steps to a centralized approach which uses [pixi](https://pixi.prefix.dev) to manage the full build environment.

Only prerequisite now is to have pixi installed, then pixi will handle:
- installing the right versions of Python, ANTLR, buf, Python dependencies, etc
- setting up Python virtual environments
- running commands with the right virtual environments and in the right working directories

We are already using pixi successfully in [substrait-packaging](https://github.com/substrait-io/substrait-packaging) and [substrait-python](https://github.com/substrait-io/substrait-python).

- - -
This change is [Reviewable](https://reviewable.io/reviews/substrait-io/substrait/1021)

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -22,7 +22,7 @@ You can also use other Python tooling like `uv` with the PyPI dependencies decla
 Pixi manages two types of dependencies:
 
 - **non-PyPI**: Includes all non-PyPI dependencies (Python itself, buf, ANTLR, Node.js, etc.) as `[tool.pixi.dependencies]` in `pyproject.toml`
-- **PyPI**: Includes all PyPI development dependencies (Black, Flake8, pytest, check-jsonschema, yamllint, etc.) and documentation dependencies (mkdocs and plugins) as a regular pyproject.toml `dev` dependency group in `pyproject.toml` which can be used with other Python package managers like `uv`
+- **PyPI**: Includes all PyPI development dependencies (Black, Flake8, pytest, check-jsonschema, yamllint, etc.) and documentation dependencies (mkdocs and plugins) as a regular pyproject.toml `dev` dependency group in `pyproject.toml` which can be used with other Python package managers like `uv`.
 
 ## Common Development Tasks
 


.github/workflows/pr.yml
```diff
@@ -8,106 +8,117 @@ jobs:
...
+          cache: true
       - name: Generate Static Site
-        run: mkdocs build
-        working-directory: ./site
+        run: pixi run mkdocs build
   editorconfig-checker:
     name: Check editorconfig
     runs-on: ubuntu-latest
     steps:
-      - uses: editorconfig-checker/action-editorconfig-checker@v2
```
nielspardon, line 24: previously we just configured the editorconfig-checker action but did not actually run it since we were missing a `run: editorconfig-checker`.

pixi now actually runs the editorconfig-checker which uncovered some violations which I'm also fixing in this PR

.github/workflows/pr.yml
```diff
@@ -8,106 +8,117 @@ jobs:
...
+          cache: true
       - name: Generate protobuf bindings
-        run: buf generate
+        run: pixi run generate-protobuf
       - name: Black
-        run: python3 -m black --diff --check .
+        run: pixi run black
       - name: Flake8
-        run: python3 -m flake8 .
+        run: pixi run flake8
```
nielspardon, line 104: I created https://github.com/substrait-io/substrait/issues/1022 for migrating these to `ruff` in a follow-up

extensions/functions_datetime.yaml
nielspardon, line 1: for this file editorconfig complained that we were not using multiples of 2 for indentation

grammar/prepend_license.sh
```diff
@@ -2,4 +2,4 @@
 
 for f in $1/*.py; do
     echo '# SPDX-License-Identifier: Apache-2.0' | cat - $f > temp && mv temp $f
-done
\ No newline at end of file
+done
```
nielspardon, line 5: editorconfig expects newline at end of file

.licenserc.yaml
```diff
@@ -2,7 +2,7 @@ header:
...
     spdx-id: Apache-2.0
 
-  paths: 
-    - 'proto/substrait/**'  
+  paths:
+    - 'proto/substrait/**'
 
-  comment: never
\ No newline at end of file
+  comment: never
```
nielspardon, line 8: editorconfig expects newline at end of file

.python-version
nielspardon, line 1: I think we had this file for users using `pyenv`. Since pixi manages the right Python version we do no longer need this.

requirements.txt
nielspardon, line 1: moved to `pyproject.toml`

site/requirements.txt
nielspardon, line 1: moved to `pyproject.toml`

.editorconfig
```diff
@@ -14,5 +14,30 @@ insert_final_newline = unset
...
+# Makefiles require tabs for indentation
+[Makefile]
+indent_style = tab
+indent_size = 4
+
+# Go files use tabs for indentation (gofmt standard)
+[*.go]
+indent_style = tab
+indent_size = 4
+
```
nielspardon, line 45: since editorconfig-checker wasn't actually running in CI it did not flag these issues with the generated ANTLR code, Makefile and Go code

pyproject.toml
```diff
@@ -16,4 +16,65 @@ force-exclude = '''
...
+generate-protobuf = { cmd = "buf generate", description = "Generate protobuf Python code" }
+yamllint = { cmd = "yamllint .", description = "Run yamllint" }
+check-jsonschema = { cmd = "check-jsonschema", description = "Validate YAML extensions" }
+editorconfig-checker = { cmd = "ec", description = "Check editorconfig"}
+mkdocs = { cmd = "mkdocs", cwd = "site", default-environment = "docs", description = "Build Website" }
+black = { cmd = "black --extend-exclude '.pixi' --diff --check .", default-environment = "dev", description = "Black" }
+flake8 = { cmd = "flake8 .", default-environment = "dev", description = "Flake8" }
+test = { cmd = "pytest .", default-environment = "dev", description = "Run tests including test_substrait_extension_coverage" }
+dry-run = { cmd = "./ci/release/dry_run.sh", default-environment = "dev", description = "dry-run release" }
+release = { cmd = "./ci/release/run.sh", default-environment = "dev", description = "release" }
```
nielspardon, line 56: we can consolidate these tasks into a simpler set of tasks in a follow-up, e.g. `lint`, `format`

benbellick:
If this goes in, are users required to install `pixi` to contribute to this repo? Or is it an optional way to manage setting up your dev environment?

nielspardon:
> If this goes in, are users required to install `pixi` to contribute to this repo? Or is it an optional way to manage setting up your dev environment?

Users would be required to install `pixi` which in my opinion is not a huge barrier given all the benefits it offers. You do not need to configure anything else other than having `pixi` installed. As I said we already use it in other substrait repos.

benbellick:
I'm a very mild -0.2 on this one. I see the benefit, but the current setup isn't that complex IMO. 

It's basically just "install antlr + buf + use your favorite Python tool to handle `requirements.txt`". 

I'd be more open if it were optional, like in `substrait-python`. But requiring it feels like added complexity without proportional benefit.

That being said, if others feel strongly about it I'm not going to hold anyone back. Have you or others actually had trouble getting your dev environment setup for this project?

nielspardon:
> It's basically just "install antlr + buf + use your favorite Python tool to handle `requirements.txt`".

It's not just that. It's also handling `editorconfig-checker`, `yamllint`, `check-jsonschema` which are not referenced in any requirements.txt.

> Have you or others actually had trouble getting your dev environment setup for this project?

For me it's a major annoyance that I have to manually set up all of these tools and ensure they are consistent with the required versions. If you remember we had similar issues in substrait-python where contributors were using outdated / older versions of ANTLR or protoc to generate the Python code. That has been resolved now with centralized configuration in substrait-packaging but is not solved for this repository.

> I'd be more open if it were optional, like in substrait-python. But requiring it feels like added complexity without proportional benefit.

I checked once more and it is optional here as well since we are just using basic `pyproject.toml` dependency groups so you can also use `uv` or `pip` if you wish. I'm pushing one more commit that should make this even easier by consolidating the `dev` and `docs` dependency groups and adding some additional `pyproject.toml` metadata to make it more complete.

CHANGELOG.md
```diff
@@ -633,7 +633,7 @@ SQLite returns null for some inputs such as negative infinity
 
 ### ⚠ BREAKING CHANGES
 
-* Renamed modulus to modulo.
```
benbellick, line 644: I think the `CHANGELOG.md` file is autogenerated. Are we sure it won't break anything if we change it by hand? Probably not but just double checking.

benbellick:
Fair enough. With it being optional, I'm slowing moving from a -0.2 -> +0.2 🙂 

I'll try and take a look today

CHANGELOG.md
nielspardon: I also wouldn't think so. alternative would be to add an exclude to editorconfig for CHANGELOG.md. also fine with me

nielspardon:
> Fair enough. With it being optional, I'm slowing moving from a -0.2 -> +0.2 🙂

hopefully, I can get you up to +0.5 so it rounds up to +1 vote in favor 😀

pyproject.toml
```diff
@@ -12,8 +21,68 @@ force-exclude = '''
...
+antlr = ">=4.13.2,<5"
+buf = ">=1.62.1,<2"
+uv = ">=0.9.25,<1"
+check-jsonschema = ">=0.37,<1"
+yamllint = ">=1.38,<2"
+python = ">=3.13,<3.14"
+nodejs = ">=24,<25"
+
+[tool.pixi.pypi-dependencies]
+editorconfig-checker = ">=3.6.1,<4"
```
tokoko: any reason why this is not in dependency-groups?

nielspardon: no good reason. only reason I had is that it was not in the original `requirements.txt` files and I manually added that. you're right I can move it into the dev group with the rest of the dependencies.

CHANGELOG.md
nielspardon: Thinking about it I tend to exclude the CHANGELOG.md file since I'm not sure whether otherwise the automatic release might trip up when adding trailing spaces in a future release.

pyproject.toml
nielspardon: @tokoko anything else you see can be optimized? Would appreciate an approval if you think this good with your pixi experience.

pyproject.toml
```diff
@@ -12,8 +21,66 @@ force-exclude = '''
...
+
+[tool.pixi.workspace]
+name = "substrait"
+channels = ["conda-forge"]
+platforms = ["linux-64", "osx-64", "osx-arm64", "linux-aarch64"]
+
+[tool.pixi.dependencies]
+antlr = ">=4.13.2,<5"
+buf = ">=1.62.1,<2"
+uv = ">=0.9.25,<1"
```
tokoko: do we still need to keep uv around as well? this is not a python project after all, nothing is published, we also don't use (?) uv in ci, so having two lockfiles feels unnecessary.

even w/o a uv.lock, people are still free to use uv if they want of course. I'm hoping the deletion won't downgrade Ben's vote to +0.1 😄

pyproject.toml
tokoko: looks good, a great improvement imho. left one new comment.

pyproject.toml
nielspardon: fair enough, I can drop it

benbellick: Don't we still use python to run the tests in here though?

nielspardon: yes, that does not require `uv` to be a dependency for pixi

nielspardon: For your benefit: the `[tool.pixi.dependencies]` are dependencies pixi installs which then are available in the environments of the tasks defined in `[tool.pixi.tasks]`. When we run `pixi run test` it will first install Python 3.13 since we have `python = ">=3.13,<3.14"` as a pixi dependency, then install the Python dev dependency group which installs `pytest` and then it will simply run `pytest` using the installed Python 3.13.

If we wanted to, we could change the pixi task definition to use `uv` to run pytest by changing the `test` task to `uv run pytest` then we would need `uv` as a pixi dependency so that `uv` is available when we run `pixi run test`.

.editorconfig
```diff
@@ -7,12 +7,41 @@ insert_final_newline = true
 indent_style = space
 trim_trailing_whitespace = true
 
+# Markdown files - preserve formatting
+[*.md]
+trim_trailing_whitespace = false
```
vbarua, line 12: Did you disable this because you don't like it, or because there's a lot of whitespace that gets trimmed? If it's the latter, it would good to enable in its own PR following this work.

.editorconfig
vbarua: I'm not familiar with editorconfig. Does it just run `go fmt` under the hood, or is it doing it's own formatting on top?

CONTRIBUTING.md
```diff
@@ -2,13 +2,101 @@
...
+```
+
+### Release Management
+
+```bash
+# Perform a dry-run release (testing only)
+pixi run dry-run
+
+# Execute the release process (maintainers only)
+pixi run release
```
vbarua: I would remove this last example, as (ideally) releases should only run via github actions.

CONTRIBUTING.md
```diff
@@ -2,13 +2,101 @@
...
+```
+
+### Environment Management
+
+Pixi manages two separate environments:
+
+- **dev**: Includes all development dependencies (Black, Flake8, pytest, etc.)
+- **docs**: Includes documentation dependencies (mkdocs and plugins)
```
vbarua: Is there a reason to keep them separate? It seems like an extra thing to track.

Looking at pyproject.toml, it doesn't even seem like there is a docs environment?

tokoko: this seems out of date

.editorconfig
nielspardon: you may want to read this thread: https://github.com/substrait-io/substrait/pull/1021#pullrequestreview-4007203884

.editorconfig
nielspardon: To my knowledge editorconfig is just a configuration file format which is interpreted by editors and IDEs supporting it. https://editorconfig.org

It was already present in the repository but the code was not actively checked using the file in CI due to an incomplete Github Actions configuration. 

We can decide whether we want to keep it and use it (which seemed to be the original intent of the authors of the editorconfig configuration file and the Github Actions setup) or we drop it.

If we keep it then we need to update the config file since it is no longer in sync with the contents of the repo aka there were no `*.go` files in the repo when the config file was added.

CONTRIBUTING.md
nielspardon: Thanks for catching this. I did clean up the documentation to reflect the current state better. Also cleaned up the dependencies by moving all PyPI dependencies into the `dev` dependency group and only keep non-PyPI dependencies as pixi dependencies.

CONTRIBUTING.md
nielspardon: I removed this one from the documentation

.editorconfig
nielspardon: There was one whitespace that had to be trimmed. I'm more worried that somehow the semantic release process might introduce whitespaces in the CHANGELOG which then would break the release process. We could limit the markdown file pattern to only `CHANGELOG.md` so we still catch any other md files.

.editorconfig
vbarua: > It was already present in the repository but the code was not actively checked using the file in CI due to an incomplete Github Actions configuration.

We can keep it as is, as we can always improve it later if we need to.

PR #1026, opened 2026-03-27
docs: supported libraries + breaking change policy

vbarua:
This change is [Reviewable](https://reviewable.io/reviews/substrait-io/substrait/1026)

### DIFF
--- a/site/docs/community/active_libraries.md
+++ b/site/docs/community/active_libraries.md
@@ -13,7 +13,7 @@ The following language libraries have active users, with at least one party acti
 
 # Inactive Libraries
 
-The following libraries in the ecosystem do not currently have active users:
+The following libraries in the ecosystem are currently looking for active contributors:
 
 - [substrait-cpp](https://github.com/substrait-io/substrait-cpp)
 - [substrait-csharp](https://github.com/substrait-io/substrait-csharp)
--- a/site/docs/community/active_libraries.md
+++ b/site/docs/community/active_libraries.md
@@ -1,10 +1,8 @@
 ---
-title: Library Support Policy
+title: Active Libraries
 ---
 
-# Library Support Policy
-
-## Active Libraries
+# Active Libraries
 
 The following language libraries have active users, with at least one party actively working to maintain and improve each of them:
 
@@ -13,13 +11,13 @@ The following language libraries have active users, with at least one party acti
 - [substrait-python](https://github.com/substrait-io/substrait-python)
 - [substrait-rs](https://github.com/substrait-io/substrait-rs)
 
-## Inactive Libraries
+# Inactive Libraries
 
 The following libraries in the ecosystem do not currently have active users:
 
 - [substrait-cpp](https://github.com/substrait-io/substrait-cpp)
-- [substrait-ruby](https://github.com/substrait-io/substrait-ruby)
 - [substrait-csharp](https://github.com/substrait-io/substrait-csharp)
 - [substrait-js](https://github.com/substrait-io/substrait-js)
+- [substrait-ruby](https://github.com/substrait-io/substrait-ruby)
 
 The Substrait project is always open to contributions and would be happy to work with anyone seeking to improve them.
--- a/site/docs/community/library_support.md
+++ b/site/docs/community/library_support.md
@@ -1,0 +1,25 @@
+---
+title: Library Support Policy
+---
+
+# Library Support Policy
+
+## Active Libraries
+
+The following language libraries have active users, with at least one party actively working to maintain and improve each of them:
+
+- [substrait-go](https://github.com/substrait-io/substrait-go)
+- [substrait-java](https://github.com/substrait-io/substrait-java)
+- [substrait-python](https://github.com/substrait-io/substrait-python)
+- [substrait-rs](https://github.com/substrait-io/substrait-rs)
+
+## Inactive Libraries
+
+The following libraries in the ecosystem do not currently have active users:
+
+- [substrait-cpp](https://github.com/substrait-io/substrait-cpp)
+- [substrait-ruby](https://github.com/substrait-io/substrait-ruby)
+- [substrait-csharp](https://github.com/substrait-io/substrait-csharp)
+- [substrait-js](https://github.com/substrait-io/substrait-js)
+
+The Substrait project is always open to contributions and would be happy to work with anyone seeking to improve them.
--- a/site/docs/spec/breaking_change_policy.md
+++ b/site/docs/spec/breaking_change_policy.md
@@ -1,0 +1,41 @@
+---
+title: Breaking Change Policy
+---
+
+# Substrait Breaking Change Policy
+
+Breaking changes proposed to the specification should include an explicit migration strategy when they are proposed.
+
+The migration strategy must be implemented in all [active libraries](../community/library_support.md) before the breaking change is implemented.
+
+For the purposes of this policy, deprecations are treated as breaking changes, because they are usually the first step in implementing a breaking change.
+
+## Context
+
+There exists in reality a rich set of query languages, query engines and data formats with an equally rich set of semantics. In the course of developing Substrait, we are constantly learning fun and arcane tidbits about all of these systems as we attempt to craft a model to encapsulate them all.
+
+Our model is not perfect. Over time, we have introduced features into Substrait that have not stood the test of practice. In many cases, our work integrating Substrait with new systems has revealed the limitations of our model. We of course take the opportunity to improve our model when this happens, but it does mean that we accumulate cruft in the specification over time — cruft that we would like to remove.
+
+Substrait is still a pre-1.0 project, so while we allow ourselves to make breaking changes, we want to do so in a controlled manner that allows users to deal with the breakages easily. For this reason, we strive to treat breaking changes as migrations and provide users with clear paths to upgrade their systems.
+
+When making any breaking changes we should assume that users are operating Substrait in a multi-system, multi-language environment where parallel and synchronous deployments are impossible. Breaking changes should be structured so that users can update systems independently and asynchronously.
+
+## Breaking Change Case Studies
+
+### Field Replacement
+
+Changes requiring the introduction of a new field to replace an old one.
+
+#### Potential Strategy
+
+1. Mark old field as deprecated.
+2. Update the specification to include both fields.
+3. Update producers to emit both the new and old fields.
+4. Update consumers to read both new and old fields.
+5. Remove old field.
+
+#### Example: URI to URI Migration
+
+- Initial Deprecation [PR](https://github.com/substrait-io/substrait/pull/859)
+- substrait-java Migration [PR](https://github.com/substrait-io/substrait-java/pull/522)
+- substrait-go Migration [PR](https://github.com/substrait-io/substrait-go/pull/166)
--- a/site/docs/spec/breaking_change_policy.md
+++ b/site/docs/spec/breaking_change_policy.md
@@ -6,7 +6,7 @@ title: Breaking Change Policy
 
 Breaking changes proposed to the specification should include an explicit migration strategy when they are proposed.
 
-The migration strategy must be implemented in all [active libraries](../community/library_support.md) before the breaking change is implemented.
+The migration strategy must be implemented in all [active libraries](../community/active_libraries.md) before the breaking change is implemented.
 
 For the purposes of this policy, deprecations are treated as breaking changes, because they are usually the first step in implementing a breaking change.
 
--- a/site/docs/spec/breaking_change_policy.md
+++ b/site/docs/spec/breaking_change_policy.md
@@ -10,7 +10,7 @@ The migration strategy must be implemented in all [active libraries](../communit
 
 ## Breaking Change Cookbooks
 
-#### URI to URN Migration
+### URI to URN Migration
 
 In 2025, the Substrait community noticed that the values used for `uris` were not consistent within the ecosystem.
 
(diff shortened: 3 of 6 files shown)


site/docs/spec/breaking_change_policy.md
```diff
@@ -0,0 +1,41 @@
...
+### Field Replacement
+
+Changes requiring the introduction of a new field to replace an old one.
+
+#### Potential Strategy
+
+1. Mark old field as deprecated.
+2. Update the specification to include both fields.
+3. Update producers to emit both the new and old fields.
+4. Update consumers to read both new and old fields.
```
yongchul: this should be reversed. consumer should be ready first, then producer can make change.

site/docs/spec/breaking_change_policy.md
```diff
@@ -0,0 +1,41 @@
...
+
+When making any breaking changes we should assume that users are operating Substrait in a multi-system, multi-language environment where parallel and synchronous deployments are impossible. Breaking changes should be structured so that users can update systems independently and asynchronously.
+
+## Breaking Change Case Studies
+
+### Field Replacement
+
+Changes requiring the introduction of a new field to replace an old one.
+
+#### Potential Strategy
```
yongchul: It would be also nice to spell out this in terms of active library support and release cadence in terms of library and substrait core.

| Step | Substrait core | Substrait library support |
| ---- | --------------| ----------------|
| Mark old field as deprecated, spec, proto is marked deprecated | X | Y |
| Consumers read both new and old fields | X | Y + 1 | 
| Producers produces new fields, remove old fields | X | Y + graceperiod releases |
| Remove deprecated field from spec and proto | X + 1 | Y + graceperiod releases |

yongchul: depending on scenario, we could make producers produce both old and new... if a system integrate heterogenous substrait systems. probably that's the only way to make it work. It would be nice that we can have some sort of configuration in the library about the policy (whether to produce new or produce new and old).

nielspardon: I was also thinking about a grace period the other day and intuitively would have opted for a time based grace period. Like we remove deprecated features from the spec in a release 3 months after the release containing the deprecation. That is a bit more predictable than number of releases given that while we do have automatic weekly releases we also sometimes do intermittent releases.

- deprecation release T
- removal: deprecation release T + 3 months

nielspardon: Also if we have a fixed time window for removal of deprecated features then we could simplify the voting process for the removal:
- introduction of deprecation -> breaking change = 2 PMC votes
- removal of deprecated feature after grace period -> non-breaking change = 1 PMC vote? 1 committer vote?

site/docs/spec/breaking_change_policy.md
```diff
@@ -0,0 +1,41 @@
...
+title: Breaking Change Policy
+---
+
+# Substrait Breaking Change Policy
+
+Breaking changes proposed to the specification should include an explicit migration strategy when they are proposed.
+
+The migration strategy must be implemented in all [active libraries](../community/library_support.md) before the breaking change is implemented.
+
+For the purposes of this policy, deprecations are treated as breaking changes, because they are usually the first step in implementing a breaking change.
```
jacques-n: I find this set of sentences confusing. It feels like it is mixing up marking something as deprecated versus removing it. marking something as deprecated (what I would call a deprecation) should not be considered a breaking change.

What are you trying to say in this paragraph? That we follow a policy of semantic versioning?

site/docs/community/library_support.md
```diff
@@ -0,0 +1,25 @@
+---
+title: Library Support Policy
```
jacques-n: I'm not sure "support" is the right word here. this isn't a support policy. it is really about activity level.

site/docs/spec/breaking_change_policy.md
```diff
@@ -0,0 +1,41 @@
...
+
+## Context
+
+There exists in reality a rich set of query languages, query engines and data formats with an equally rich set of semantics. In the course of developing Substrait, we are constantly learning fun and arcane tidbits about all of these systems as we attempt to craft a model to encapsulate them all.
```
jacques-n: I would remove this section. I'm not sure who the audience is or what information it expresses.

site/docs/spec/breaking_change_policy.md
```diff
@@ -0,0 +1,41 @@
...
+
+Our model is not perfect. Over time, we have introduced features into Substrait that have not stood the test of practice. In many cases, our work integrating Substrait with new systems has revealed the limitations of our model. We of course take the opportunity to improve our model when this happens, but it does mean that we accumulate cruft in the specification over time — cruft that we would like to remove.
+
+Substrait is still a pre-1.0 project, so while we allow ourselves to make breaking changes, we want to do so in a controlled manner that allows users to deal with the breakages easily. For this reason, we strive to treat breaking changes as migrations and provide users with clear paths to upgrade their systems.
+
+When making any breaking changes we should assume that users are operating Substrait in a multi-system, multi-language environment where parallel and synchronous deployments are impossible. Breaking changes should be structured so that users can update systems independently and asynchronously.
+
+## Breaking Change Case Studies
```
jacques-n: Are these really cookbooks for specific kinds of breaking changes?

It seems like this is the content we should really focus on: how do we do breaking changes for different kinds of specific examples. Maybe we take several from history and describe those?

site/docs/spec/breaking_change_policy.md
benbellick: I think this potentially confusing language comes from the last sync we had (or maybe the one before). @nielspardon raised the point that the introducing of the new field isn't a breaking change, but the dropping of the old is. 

At the time, it seemed to us like the governance structure required a different number of votes between breaking changes and non-breaking changes. 

However, reading it back, it actually seems clear: 
https://github.com/substrait-io/substrait/blob/7cceb8369cd533134431979835124de5634a4a47/site/docs/governance.md?plain=1#L60-L70

I would call these non-breaking changes "Specification Modifications" and thus it _is_ required to get 2 PMC votes. 

What about just amending the language in the table in `goverance.md` if the current language seems unclear to either @vbarua or @nielspardon?

nielspardon: We had the exact same discussion at the community sync that marking something deprecated is not technically a breaking change but from the voting process perspective should be treated like a breaking change aka marking something deprecated should already require 2 PMC votes and not only the actual removal of previously deprecated functionality.

site/docs/spec/breaking_change_policy.md
vbarua: I think time-based would be easier to reason about.

> we remove deprecated features from the spec in a release 3 months after the release containing the deprecation.

Would it be 3-months after the deprecation, or 3-months after all the active libraries have been updated with the migration strategy?

site/docs/community/library_support.md
vbarua: That makes sense. Updated this to be more about activity.

site/docs/spec/breaking_change_policy.md
vbarua: > What about just amending the language in the table in goverance.md

That actually works way better Ben, I'll do that and remove the wording here.

site/docs/spec/breaking_change_policy.md
vbarua: I've reframed this example to be less prescriptive.

site/docs/spec/breaking_change_policy.md
vbarua: Honestly, I was in a really conciliatory headspace when I wrote this. In hindsight, I don't we need to explain ourselves this much. Removed.

site/docs/spec/breaking_change_policy.md
vbarua: I like the cookbook idea. It's less prescriptive and more grounded in what we've actually done. Updated.

site/docs/spec/breaking_change_policy.md
nielspardon: good question, I'm wondering whether it would be sufficient to declare this only for the specification release. If the active libraries are truly actively maintained they would be picking up the deprecation and the removal not too far from the spec removal release anyway.

nielspardon: we could include a diagram showing an example timeline

```mermaid
gantt
    title Substrait Deprecation Policy
    dateFormat  YYYY-MM-DD
    axisFormat %m
    tickInterval 1month

    section Specification
    Feature A              :a2, 2020-01-01, 4M
    Feature A Deprecation  : milestone, m1, 2020-02-01, 0d
    Feature A Removal      : milestone, m2, 2020-05-01, 0d
    Feature B              :a3, 2020-02-01, 5M

    section Libraries
    Feature A              :a2, 2020-01-15, 4M
    Feature A Deprecation  : milestone, m1, 2020-02-15, 0d
    Feature A Removal      : milestone, m2, 2020-05-15, 0d
    Feature B              :a3, 2020-02-15, 5M
```

yongchul (approved):
+1 for this change.

site/docs/spec/breaking_change_policy.md
```diff
@@ -0,0 +1,24 @@
...
+
+# Substrait Breaking Change Policy
+
+Breaking changes proposed to the specification should include an explicit migration strategy when they are proposed. This is based on the assumption that users are operating Substrait in a multi-system, multi-language environment where parallel and synchronous deployments are impossible. Breaking changes should be made so that users can update systems independently and asynchronously.
+
+The migration strategy must be implemented in all [active libraries](../community/active_libraries.md) before the breaking change is implemented.
+
+## Breaking Change Cookbooks
+
+#### URI to URN Migration
```
yongchul: nesting should be `###`?

site/docs/spec/breaking_change_policy.md
vbarua: Would we be okay with introducing a more formal timeline as a followup? I feel like we could have a whole separate discussion about this. There's also a world were what we have now will be sufficient in practice and folks will generally update their systems in a timely fashion. I know that we were the most recent counterexample of the latter, but knowing that other folks want to get rid of stuff will also make us prioritize it differently.

site/docs/community/active_libraries.md
```diff
@@ -0,0 +1,23 @@
...
+The following language libraries have active users, with at least one party actively working to maintain and improve each of them:
+
+- [substrait-go](https://github.com/substrait-io/substrait-go)
+- [substrait-java](https://github.com/substrait-io/substrait-java)
+- [substrait-python](https://github.com/substrait-io/substrait-python)
+- [substrait-rs](https://github.com/substrait-io/substrait-rs)
+
+# Inactive Libraries
+
+The following libraries in the ecosystem do not currently have active users:
```
nielspardon: I'm wondering whether it would make sense to phrase this more actively:

```suggestion
The following libraries in the ecosystem are currently looking for active contributors:
```

site/docs/spec/breaking_change_policy.md
nielspardon: > Would we be okay with introducing a more formal timeline as a followup?

Fine with me.

site/docs/spec/breaking_change_policy.md
vbarua: It should
```suggestion
### URI to URN Migration
```

nielspardon (approved):
LGTM

site/docs/spec/breaking_change_policy.md
```diff
@@ -0,0 +1,24 @@
...
+## Breaking Change Cookbooks
+
+### URI to URN Migration
+
+In 2025, the Substrait community noticed that the values used for `uris` were not consistent within the ecosystem.
+
+To deal with this, a new URN format was [introduced](https://github.com/substrait-io/substrait/issues/856) to replace URIs. To migrate the ecosystem to use URNs, producers were updated to dual write URIs and URNs in plans, and consumers were updated to preferentially consume URNs when available.
+
+This work can be seen in:
+- substrait-java Migration [PR](https://github.com/substrait-io/substrait-java/pull/522)
```
benbellick: nit: 
```suggestion
- `substrait-java` Migration [PR](https://github.com/substrait-io/substrait-java/pull/522)
```

site/docs/spec/breaking_change_policy.md
```diff
@@ -0,0 +1,24 @@
...
+- substrait-go Migration [PR](https://github.com/substrait-io/substrait-go/pull/166)
```
benbellick:
```suggestion
- `substrait-go` Migration [PR](https://github.com/substrait-io/substrait-go/pull/166)
```

site/docs/spec/breaking_change_policy.md
```diff
@@ -0,0 +1,24 @@
...
+- substrait-python Migration [PR](https://github.com/substrait-io/substrait-python/pull/114)
```
benbellick:
```suggestion
- `substrait-python` Migration [PR](https://github.com/substrait-io/substrait-python/pull/114)
```

site/docs/spec/breaking_change_policy.md
```diff
@@ -0,0 +1,24 @@
...
+
+Breaking changes proposed to the specification should include an explicit migration strategy when they are proposed. This is based on the assumption that users are operating Substrait in a multi-system, multi-language environment where parallel and synchronous deployments are impossible. Breaking changes should be made so that users can update systems independently and asynchronously.
+
+The migration strategy must be implemented in all [active libraries](../community/active_libraries.md) before the breaking change is implemented.
+
+## Breaking Change Cookbooks
+
+### URI to URN Migration
+
+In 2025, the Substrait community noticed that the values used for `uris` were not consistent within the ecosystem.
```
benbellick: IMO, might as well include a link to the issue here so that people can see the discussion https://github.com/substrait-io/substrait/issues/828

benbellick (approved):
LGTM! I haven't actually checked out this branch locally and made the docs.

site/docs/spec/breaking_change_policy.md
vbarua:
```suggestion
In [2025](https://github.com/substrait-io/substrait/issues/828), the Substrait community noticed that the values used for `uris` were not consistent within the ecosystem.
```

Issue #576, opened 2023-11-14
Inconsistent  type inference for date add/subtract interval operations


LeonidChistov:
### Discussed in https://github.com/substrait-io/substrait/discussions/574

Following operations are defined in Substrait prototypes:

```
add(date, interval_year): -> timestamp
add(date, interval_day): -> timestamp
subtract(date, interval_year): -> date
subtract(date, interval_day): -> date
```
It looks like `add` and `subtract` behavior need to be consistent with regards to result type.

However, different databases/frameworks have different type inference for these operations:
- Postrgres returns timestamp in all four cases
- Calcite infers return type as `date` in all four cases

Probably it is not so important which type inference approach will be used by Substrait, because type casts from/to timestamp for operands/result can be used to adjust to specific DB semantics.

But there is no reason to let `add` and `subtract` have different type inference.

nielspardon:
This has been partially fixed recently in https://github.com/substrait-io/substrait/pull/1007:
```
add(date, interval_year): -> date
```

nielspardon:
I opened PR https://github.com/substrait-io/substrait/pull/1029 for fixing the `subtract:date_iday` return type.

nielspardon:
Both parts of this are now fixed and released, so the type inference for `add` and `subtract` is consistent:

| signature | return |
|---|---|
| `add(date, interval_year)` | `date` |
| `add(date, interval_day<P>)` | `precision_timestamp<P>` |
| `subtract(date, interval_year)` | `date` |
| `subtract(date, interval_day<P>)` | `precision_timestamp<P>` |

- #1007 corrected `add:date_iyear` to return `date`
- #1029 corrected `subtract:date_iday` to return `precision_timestamp` (released in v0.91.0)

The remaining year/day difference is deliberate and symmetric across both functions: `interval_day` carries sub-day components, so the result has to widen to a timestamp, while `interval_year` does not and the result stays a `date`.

Closing as fixed.

PR #1029, opened 2026-04-01
fix(extensions)!: correct return type for subtract:date_iday operation

nielspardon:
BREAKING CHANGE: changes the return type of the `subtract:date_iday` function from `date` to `precision_timestamp`

Fix the return type of the `subtract` function when subtracting an `interval_day` from a `date`. Since `interval_day` is an interval of days to subseconds the operation should return a `precision_timestamp` type, not a `date` type.

This corrects the function signature in `functions_datetime.yaml` where `date` + `interval_day` was incorrectly specified to return `date` instead of `precision_timestamp`.

This is similar to #1007 and was originally reported in #576.

This also fixes the `subtract:ts_iday` test cases to use a precision type parameter for `iday`.

- - -
This change is [Reviewable](https://reviewable.io/reviews/substrait-io/substrait/1029)

### DIFF
--- a/extensions/functions_datetime.yaml
+++ b/extensions/functions_datetime.yaml
@@ -342,7 +342,7 @@ scalar_functions:
             value: date
           - name: y
             value: interval_day<P>
-        return: date
+        return: precision_timestamp<P>
   -
     name: "lte"
     description: less than or equal to
--- a/tests/cases/datetime/subtract_datetime.test
+++ b/tests/cases/datetime/subtract_datetime.test
@@ -7,9 +7,9 @@ subtract(2016-12-01T13:30:15::pts<6>, P5Y::iyear) = 2011-12-01T13:30:15::pts<6>
 subtract(2016-12-01T13:30:15::pts<6>, PT5H::iday) = 2016-12-01T08:30:15::pts<6>
 
 # date: examples using the date type
-subtract(2020-12-31::date, P5D::iday) = 2020-12-26::date
+subtract(2020-12-31::date, P5D::iday<0>) = 2020-12-26T00:00:00::pts<0>
 subtract(2020-12-31::date, P5Y::iyear) = 2015-12-31::date
 subtract(2020-12-31::date, P5M::iyear) = 2020-07-31::date
 
 # null_input: examples with null args or return
-subtract(null::date?, P5D::iday) = null::date?
+subtract(null::date?, P5D::iday<0>) = null::pts?<0>


extensions/functions_datetime.yaml
```diff
@@ -463,7 +463,7 @@ scalar_functions:
             value: date
           - name: y
             value: interval_day<P>
-        return: date
+        return: precision_timestamp<P>
```
yongchul, line 345: Wait what? The existing change looks correct to my eyes.

nielspardon: If you subtract 1 day, 2 hours, 3 seconds, 4 microseconds from a date what do you expect to be returned? Certainly not a full date anymore but a timestamp and given that interval_day has a precision for subseconds it would be a precision timestamp of the same precision as interval_day.

yongchul: Thank you for pointing that out but that opens even more questions. If that is the case, we should update all other things? For instance, `timestamp` and `timestamp_tz` are at microsecond precision. So if you do math with `interval_day<P>`, it should be the higher precision of `max(6,P)`?

or,

```
if (P <= 6) then (timestamp) else (precision_timestamp<P>) 
```

nielspardon: So I already have [this PR](https://github.com/substrait-io/substrait/pull/994) open for removing the old timestamp, timestamp_tz and time types which should simplify such scenarios somewhat. We can check if the remaining precision_timestamp, precision_timestamp_tz, precision_time functions handle the precision correctly.

nielspardon: I merged #994 and rebased this PR on latest main.

nielspardon: We do indeed have the situation in function signatures where datetime return type precision is derived from directly from the argument precision where e.g. for `add:pts_iday` both arguments need to have the same precision `P`:

https://github.com/substrait-io/substrait/blob/87d73b663bcc153f9dda7dfe574dca36b1e4ed28/extensions/functions_datetime.yaml#L206-L211

We could (in a follow-up) change this for more flexibility to:

```yaml
      - args:
          - name: x
            value: precision_timestamp<P1>
          - name: y
            value: interval_day<P2>
        return: |
          P = max(P1, P2)
          precision_timestamp<P>
```

nielspardon: created a new issue for looking into the return type precision: https://github.com/substrait-io/substrait/issues/1033

yongchul: Sounds good. Thank you for creating the follow up issue!

yongchul (approved):
+1 but I have no idea whether this will cause any problem to anyone or not...

extensions/functions_datetime.yaml
nielspardon: FYI: found another issue today raising the exact same issue #926 

Looks like @vbarua is of the opinion that one should explicitly cast the argument precision to the same precision before the function call in the older issue.

benbellick (approved):
[Since this makes it consistent with add now](https://github.com/substrait-io/substrait/blob/99ba9d1f452416b8d6267993142b427062888078/extensions/functions_datetime.yaml#L218-L223), LGTM!

Issue #965, opened 2026-02-20
Substrait dialect: allow specifying maximum supported scale/precision/length for parameterized data types


nielspardon:
https://github.com/substrait-io/substrait/pull/961 introduced a `max_precision` property in the Substrait dialect for datetime data types. We should allow providing similar properties for maximum supported scale/precision/length for other parameterized data types like `DECIMAL<S, P>`, `VARCHAR<L>`, `FIXEDCHAR<L>`, etc

PR #1030, opened 2026-04-01
feat(dialect): support max length, scale, precision for parameterized types

nielspardon:
BREAKING CHANGE: explicitly defines the maximum length of `FIXEDBINARY` in the documentation which was already existent in the protobuf specification due to the use of `int32` for the `length` argument

- adds `max_length` for `FIXED_BINARY`, `VARCHAR`, `FIXED_CHAR` to the dialect schema
- adds `max_precision` and `max_scale` for `DECIMAL` to the dialect schema
- updated the documentation to mention the maximum length for `FIXEDBINARY` which is given by the [use of `int32` for the `length` field in the proto](https://github.com/substrait-io/substrait/blob/7cceb8369cd533134431979835124de5634a4a47/proto/substrait/type.proto#L178-L179)

follow-up to https://github.com/substrait-io/substrait/pull/961
fixes https://github.com/substrait-io/substrait/issues/965

- - -
This change is [Reviewable](https://reviewable.io/reviews/substrait-io/substrait/1030)


nielspardon:
> This is technically breaking change 😆 because we are imposing an explicit limit now but it is aligning with respect to other types so I'm okay with that. 

I can mark it as breaking to be safe.

> Also, now that I see it again, there are systems supporting larger than 2 GiB data but it is what it is... We will see whether there are enough requests to support such very large objects. +1.

We can have a follow-up where we can increase the limits.

Issue #1024, opened 2026-03-25
Unified outer reference handling with common subexpression


yongchul:
# Problem

The current outer reference in Substrait is offset based. To resolve the outer reference, it needs to walk up the *tree* to find the binding `Rel`. There can be multiple layers of nestings, thus the implementer must be aware what does provide such boundary. Today, the boundary is set by Subquery, and it is the only construct introducing such outer reference boundary.

The main problem is this assumes *tree* -- you have a single parent. If you have a common subexpressions, or a graph, this scheme no longer works any more.

The following motivating example uses a `lateral join` (#973 ), SQL standard feature allowing outer references in a table reference, for the sake of demonstration but the same problem can be demonstrated in scalar subqueries or set predicates.

# Example

Consider following SQL.

```sql
WITH x as (
  select a,b
  from tableA
  where b = @p1
),
y as (
  select a, count(*) as b
  from x
  group by a
  order by a limit 10
),
z as (
  select a, count(y.b) as b
  from x left outer lateral join y on x.a = y.a
  where y.a > b
  group by a
)
select *
from tableB LATERAL JOIN (
    SELECT a, b
    FROM y
    WHERE a > tableB.a
    UNION
    SELECT a, b
    FROM z
    WHERE a > tableB.a
)
ON true
```

A rather straight forward translation to Substrait Rel is following.

```mermaid
graph TD;
  A["FilterRel(b=@p1)"] --- B["ReadRel(tableA)"];
  D["FetchRel(10)"] --- E["AggRel(a, count(*))"] --- A
  F["AggRel(a,count(y.b))"] --- G["FilterRel(y.a > b)"] --- H[LateralJoin2] --- A
  H --- D
  LateralJoin1 --- ReadRel(tableB)
  LateralJoin1 --- Union --- X["FilterRel(a > tableB.a)"] --- D
  Union --- Y["FilterRel(a > tableB.a)"] --- F
```

After simple predicate push down, the plan will look like following. Note that the common correlated predicate `a > tableB.a` is pushed all the way down to `ReadRel(tableA)` (see the node with `!!!`).

```mermaid
graph TD;
  A["!!! FilterRel(b=@p1 AND a > tableB.a) !!!"] --- B["ReadRel(tableA)"];
  D["FetchRel(10)"] --- E["AggRel(a, count(*))"] ---|step=2| A
[...]

PR #1031, opened 2026-04-02
feat!: add id-based outer reference resolution for DAG plans

yongchul:
BREAKING CHANGE: NOT A REAL BREAKING CHANGE. OuterReference.steps_out field is now under oneof -- implicit optional to explicit optional. This is not a real breaking change as Substrait always required non-zero value.

Add optional RelCommon.id field and OuterReference.id_reference to support unambiguous outer reference resolution in plans with common subexpressions (ReferenceRel). The existing steps_out offset-based mechanism remains for backward compatibility with tree-shaped plans.

The `RelCommon.rel_anchor` can be used for other purpose but for now only required when it is used to resolve outer reference.

See #1024 for detailed motivating example and discussion.

# AI disclaimer
This PR was assisted by Claude Opus 4.6.

# NOT A REAL BREAKING CHANGE

`OuterReference.steps_out` field is moved into `oneof`. The field requires non-zero values thus there is no wire format change (the value has been *always* present on the wire for `steps_out` when used).

- - -
This change is [Reviewable](https://reviewable.io/reviews/substrait-io/substrait/1031)


benbellick:
Am I understanding correctly that this problem only arises when a `Plan.relations` entry contains outer references that reach outside the bounds of that particular root?

My interpretation of `Plan.relations` is that each entry is a standalone thing. Having a relation use `steps_out` to reference something outside its own scope feels like a misuse of `Plan.relations` to me. Of course, we can always use a `ReferenceRel` to reference one of the other relations, but their evaluation wouldn't be context dependent. 

I feel like we may need a new construct here rather than tacking on the ability to reference outside of the scope of a particular root. This feels like some kind of "parameterized" relation notion, where we would want to have call sites providing values to pass in dynamically. I'm not sure yet what the right way to represent this would be but wanted to get my thoughts down for now. 

What do you think?

yongchul:
I love github missing thread in the PR comment... 😠 

> Am I understanding correctly that this problem only arises when a `Plan.relations` entry contains outer references that reach outside the bounds of that particular root?

Yes.
 
> My interpretation of `Plan.relations` is that each entry is a standalone thing. Having a relation use `steps_out` to reference something outside its own scope feels like a misuse of `Plan.relations` to me. Of course, we can always use a `ReferenceRel` to reference one of the other relations, but their evaluation wouldn't be context dependent.

Yes, you could put that restriction, you have to have each relation context independent. However, this proposal is actually making it context independent because the specific outer reference is bound to one **consistent** value without ambiguity [modulo there is still restriction where that particular value **can** coming from].
 
> I feel like we may need a new construct here rather than tacking on the ability to reference outside of the scope of a particular root. This feels like some kind of "parameterized" relation notion, where we would want to have call sites providing values to pass in dynamically. I'm not sure yet what the right way to represent this would be but wanted to get my thoughts down for now.

This goes to inline table function (or parameterized view), which is a generic solution (or sledge hammer) and I actually feel like it is an overkill. Yes, I did think initially this is the way to go when I confine myself in existing Substrait outer reference structure, but this quickly went out of control. If your binding actually does not change in the plan, why do you have to go through indirect mappings only to point to the same thing?

To truly use inline table function, either you have a) a fairly sophisticated common subexpression engine to extract the common relational template or b) support for the parameterized views. I'm not aware of (a) but many systems do support (b). In (b), you won't actually have an outer reference if you inline the view definition or if you model it actual function call. Because you already have the common definition, it is straight forward to implement in producer and they are in fact used multiple times with *different* binding points. For (a), if you don't have such detection, it's pretty challenging for producer extract the common pattern, and generate the definition. It's nice to have such fancy decomposition for our system to have, but I don't see us doing this anytime soon...


So, I'm not arguing we don't need inline table function. We will need that and useful for parameterized views. But I don't feel like we *have to* use inline table function for outer reference, because we *can* represent outer reference in more generic way.


I'm biased. 😆  Happy to chat more about this but I strongly prefer anything else except github PR-level comment.

drin:
Could there be a common mechanism with #726? or is this change totally independent?

I don't have context on why an outer reference and a reference rel are both necessary. This seems to be the reason, but I'm not sure what context an outer reference has that can't be included in a reference rel:
> Of course, we can always use a ReferenceRel to reference one of the other relations, but their evaluation wouldn't be context dependent.


(Of course, I would have to reopen #726 or maybe you can combine the change in this one since it's small)

yongchul:
@drin 

> Could there be a common mechanism with #726? or is this change totally independent?

I believe they are independent. If you squint, you could use the `rel_id` of the root rel of `PlanRel` but it is indirect while you seemed to look into more direct reference in #726. Or, `rel_id` **value** is sufficient for `PlanRel` anchor **value** you are looking for but not necessary, and I expect you want more direct reference at PlanRel message level.

> I don't have context on why an outer reference and a reference rel are both necessary. This seems to be the reason, but I'm not sure what context an outer reference has that can't be included in a reference rel:

reference rel is about sharing a relational subtree (e.g., common table expressions, views). outer reference is a concept that you can correlate one relational tree to another (see lateral join #973 for example). They are orthogonal.

> > Of course, we can always use a ReferenceRel to reference one of the other relations, but their evaluation wouldn't be context dependent.
> (Of course, I would have to reopen #726 or maybe you can combine the change in this one since it's small)

I wouldn't combine them because they are different concept.

Please let me know if there is some misunderstanding. Again, I hate PR-level comment to discuss something. If you want to discuss, let's create a comment in a file, then starts a thread.

jacques-n:
@yongchul , have you looked at all at how to change this in Datafusion/Duckdb? I think maybe one/both have implemented subqueries (maybe?) and am curious the complexity to update to new pattern.

yongchul:
> @yongchul , have you looked at all at how to change this in Datafusion/Duckdb? I think maybe one/both have implemented subqueries (maybe?) and am curious the complexity to update to new pattern.

@jacques-n conceptually, if the implementer already maintains schema stack, the new approach can be implemented by look up by the id. So the content of stack changes from `(schema)` to `(id, schema)`, then you can simply go over the stack until you find the schema with matching id.

I schemed datafusion. With my inexperienced rusty eyes, it does seem to follow above structure so fairly easy to extend. I couldn't find handling of `RelCommon` but this could be done easily as well when you push schema to the stack.

I admit that I used a little loose definition of `uniqueness of id` here though. I don't know how hard to check the uniqueness of non-zero `id`s across the plan... maybe this is difficult in datafusion?

I will look into duckdb as well.

jacques-n (approved):
looks good to me.

site/docs/expressions/field_references.md
```diff
@@ -155,3 +155,34 @@ By default, when only a single field is selected from a struct, that struct is r
 
     * Should we support column reordering/positioning using a masked complex expression? (Right now, you can only mask things out.)
 
+### Outer References
+
+Outer references allow expressions inside a subquery to access records from an enclosing relation. The `OuterReference` root type supports two resolution strategies that can be used independently or together:
+
+#### `steps_out` (offset-based)
+
+`steps_out` resolves the reference by counting subquery boundaries upward (`steps_out >= 1`). This works correctly whenever the plan is a **tree** — each relation has exactly one parent, so the path to the binding relation is unique.
```
benbellick:
```suggestion
`steps_out` resolves the reference by counting subquery boundaries upward (`steps_out >= 1`). This works correctly whenever the plan is a **tree**, i.e when each relation has exactly one parent, the path to the binding relation can be uniquely determined via `steps_out`. 
```

site/docs/expressions/field_references.md
```diff
@@ -155,3 +155,34 @@ By default, when only a single field is selected from a struct, that struct is r
...
+
+#### `id_reference` (id-based)
+
+When a plan contains **common subexpressions** via `ReferenceRel`, the same relation can be reached through multiple paths with different depths. In that case, offset-based resolution is ambiguous because `steps_out` depends on *which* path is followed.
```
benbellick:
```suggestion
When a plan contains **common subrelation** via `ReferenceRel`, the same relation can be reached through multiple paths with different depths. In that case, offset-based resolution is ambiguous because `steps_out` depends on *which* path is followed.
```
might make more sense?

benbellick: I think that this section may read a bit better if instead it was 
```markdown
#### `steps_out` (offset-based)
<explanation of WHAT steps_out is with an example>
#### `id_reference` (id-based)
<explanation of WHAT id_reference is with an example>

#### When `id_reference` is required
example of case where it is required. 
```

I think it would read clearer to people without the context of the problem this is solving to just see the concept in isolation. Then in the subsequent section, they see _why_ you need to use one vs the other. WDYT?

site/docs/expressions/field_references.md
```diff
@@ -155,3 +155,34 @@ By default, when only a single field is selected from a struct, that struct is r
...
+
+`id_reference` resolves this by naming the binding relation directly via its plan-wide unique `RelCommon.id`. The `id` on the referenced relation must be set (> 0) and unique across all relations in the plan.
+
+At least one of `steps_out` or `id_reference` must be set. When both are present, `id_reference` is authoritative and `steps_out` is advisory (useful for backward compatibility with consumers that only understand `steps_out`).
+
+For example, consider a plan with two nested scalar subqueries that share a common subexpression `x`. The outer reference to `tableA.a` lives inside `x`, which is reached via paths of different depth:
```
benbellick, line 174: again maybe subrelation is a better term here?

site/docs/expressions/subqueries.md
```diff
@@ -68,6 +68,26 @@ WHERE x < ANY(SELECT y from t2)
 
 
 
+## Outer References in Subqueries
+
+Subqueries may contain *outer references* — field references that reach
```
benbellick:
```suggestion
Subqueries may contain *outer references*, which are field references that reach
```

nit: sorry but LLMs have made me em-dash averse, even though this is a fine use case for them 😅

site/docs/expressions/subqueries.md
```diff
@@ -68,6 +68,26 @@ WHERE x < ANY(SELECT y from t2)
...
+  upward. This works correctly when the plan is a tree (each relation has a
+  single parent).
+
+* **`id_reference`**: Resolves the reference by naming the binding relation
+  via its plan-wide unique `RelCommon.id`. This is recommended when the plan
+  contains shared subexpressions (`ReferenceRel`) because offset-based
+  resolution is ambiguous when multiple paths exist to the same relation.
+
+At least one must be set. When both are present, `id_reference` is
+authoritative. See
```
benbellick, line 87: Why allow both to be set? Could we just consider it an error?

site/docs/expressions/field_references.md
```diff
@@ -155,3 +155,34 @@ By default, when only a single field is selected from a struct, that struct is r
...
+├── ReadRel(tableA)  [id=1]
+└── Subquery.Scalar ── (1)
+    └── SetRel(MINUS_PRIMARY)
+        ├── Subquery.Scalar ── (2)
+        │   └── x: FilterRel(a > outer_ref(tableA.a))  [id=2]
+        │         └── ReadRel(tableB)
+        └── x  (shared via ReferenceRel)
+```
+
+The common subexpression `x` contains an outer reference to `tableA.a`. It is reached through two paths — one via subquery (2), one directly — so `steps_out` would need a different value depending on which path is followed. With `id_reference = 1`, the reference inside `x` unambiguously names `ReadRel(tableA)` regardless of path.
```
benbellick, line 187: again would subrelation make sense here?

site/docs/expressions/field_references.md
benbellick, line 1: What do you think about explicitly banning the usage of `steps_out` to reference relations outside of your scope? In that case, you _must_ use an `id_reference`?

proto/substrait/algebra.proto
```diff
@@ -1590,12 +1596,22 @@ message Expression {
...
-      //
-      // This value must be >= 1
+      // Number of subquery boundaries to traverse up for this field's
+      // reference. This value must be >= 1 when set. A value of 0 indicates
+      // that steps_out is not provided and id_reference must be used instead.
       uint32 steps_out = 1;
+
+      // Plan-wide unique id of the relation whose output this reference is
+      // rooted on. Must match an id defined on a RelCommon in the plan.
+      // Must be > 0 when set. Recommended when the plan contains shared
```
benbellick, line 1611: mentioned elsewhere, but could we just make it required in this case?

site/docs/expressions/field_references.md
benbellick, line 174: [Could we add some textformat examples](https://github.com/substrait-io/substrait/tree/main/site/examples/proto-textformat)?

site/docs/expressions/subqueries.md
benbellick: Going even further, would it be possible to wrap the two fields in a `oneof` so that this invariant is enforced at the proto level?

benbellick (commented):
I generally think this is good now that I properly understand the problem, but just have some things I'd like clarified. 

Thanks!

site/docs/expressions/subqueries.md
yongchul: I thought about using `oneof`... but then went back for `optional` for some reason I can't remember. 😦  since I don't remember the reason, I'll just switch over to oneof. I may recall why... but then we can discuss whether that's valid.

site/docs/expressions/field_references.md
yongchul: I wish but that will make this change effectively a breaking change. We can discuss about this in the meeting.

proto/substrait/algebra.proto
yongchul: I am trying to avoid presenting this as a breaking change. Perhaps it is fine. Let's see how others think.

site/docs/expressions/field_references.md
yongchul: reworded as shared relation.

jacques-n (approved):

site/docs/expressions/subqueries.md
yongchul: It is oneof now.

site/docs/expressions/subqueries.md
```diff
@@ -68,6 +68,26 @@ WHERE x < ANY(SELECT y from t2)
 
 
 
+## Outer References in Subqueries
+
+Subqueries may contain *outer references*, which are field references that reach
+outside the subquery boundary to access records from an enclosing relation.
+The `OuterReference` root type provides two resolution fields:
+
+* **`steps_out`**: Resolves the reference by counting subquery boundaries
```
benbellick:
```suggestion
* `steps_out`: Resolves the reference by counting subquery boundaries
```

site/docs/expressions/subqueries.md
```diff
@@ -68,6 +68,26 @@ WHERE x < ANY(SELECT y from t2)
...
+  upward. This works correctly when the plan is a tree (each relation has a
+  single parent).
+
+* **`id_reference`**: Resolves the reference by naming the binding relation
```
benbellick:
```suggestion
* `id_reference`: Resolves the reference by naming the binding relation
```

Just suggesting the LLM-isms to go 😆

site/docs/expressions/field_references.md
```diff
@@ -155,3 +155,38 @@ By default, when only a single field is selected from a struct, that struct is r
...
+ProjectRel [id=1]
+├── ReadRel(tableA)
+└── Subquery.Scalar ── (1)
+    └── SetRel(MINUS_PRIMARY)
+        ├── Subquery.Scalar ── (2)
+        │   └── ReferenceRel(0)
+        └── ReferenceRel(0)
+```
+
+The shared relation `x` contains an outer reference to `tableA.a`. It is reached through two paths — one via subquery (2), one directly — so `steps_out` would need a different value depending on which path is followed. With `id_reference = 1`, the reference inside `x` unambiguously names `ProjectRel` to resolve the outer reference regardless of path.
```
benbellick: It would be nice to show the example of the "good" case right after, i.e. the first relation but using an id reference.

benbellick (approved):

site/docs/expressions/field_references.md
yongchul: Hmm... Do you mean making the example more explicit using steps out and id? I'll make the change.

yongchul: Revised. PTAL

site/docs/expressions/field_references.md
yongchul: Revised accordingly. PTAL

site/docs/expressions/field_references.md
```diff
@@ -155,3 +155,60 @@ By default, when only a single field is selected from a struct, that struct is r
...
+
+`id_reference` resolves the reference by referring to the binding relation via its plan-wide unique `RelCommon.id`. The `id` on the referenced relation must be set (>= 1) and unique across all relations in the plan.
+
+#### Coexistence rules
+
+Exactly one of `steps_out` or `id_reference` must be set on each `OuterReference`. A single plan may contain outer references using different strategies (e.g., some using `steps_out` and others using `id_reference`), as long as every individual reference is unambiguous. However, if any shared relation (via `ReferenceRel`) contains an unresolved outer reference, that reference **must** use `id_reference`.
+
+#### When to use `id_reference`
+
+`id_reference` must be used instead of `steps_out` when a plan contains **shared relations** via `ReferenceRel` with unresolved outer references in the shared relations. In such plans, the binding relation (i.e., the relation providing the actual value of the outer reference) can be reached through multiple paths with different depths, making offset-based resolution ambiguous because `steps_out` depends on *which* path is followed.
```
benbellick:
```suggestion
`id_reference` must be used instead of `steps_out` when an outer reference appears inside a relation shared via `ReferenceRel` and the shared relation can be reached through multiple paths with different subquery depths, making `steps_out` ambiguous. In this case, the same outer reference could require different `steps_out` values depending on which path is followed.
```

This is just one suggestion. The thing I am hoping to fix here is that the first sentence reads a bit ambiguously to me: 
> `id_reference` must be used instead of `steps_out` when a plan contains **shared relations** via `ReferenceRel` with unresolved outer references in the shared relations.
This sounds like we may require `id_reference` _always_, rather than merely where it is ambiguous.

site/docs/expressions/field_references.md
```diff
@@ -155,3 +155,60 @@ By default, when only a single field is selected from a struct, that struct is r
...
+
+PlanRel.relations[1].root:
+ProjectRel # Correct binding for tableA.a for the outer reference tableA.a in x.
+├── ReadRel(tableA)
+└── Subquery.Scalar # Subquery (1)
+    └── SetRel(MINUS_PRIMARY)
+        ├── ProjectRel
+        |   └── Subquery.Scalar # Subquery (2)
+        │       └── ReferenceRel(0) # Reference 1: correct steps_out = 2
+        └── ReferenceRel(0) # Reference 2: correct steps_out = 1
```
benbellick:
```suggestion
        │       └── ReferenceRel(0) # Here steps_out=1 binds incorrectly, because tableA.a is actually two subquery boundaries out.
        └── ReferenceRel(0) # Here steps_out=1 binds correctly, because tableA.a is one subquery boundary out.
```

site/docs/expressions/field_references.md
```diff
@@ -155,3 +155,60 @@ By default, when only a single field is selected from a struct, that struct is r
...
+```
+
+From the reference 1, the correct `steps_out` is 2 because it needs to go through 2 subqueries to reach the ProjectRel. From the reference 2, the correct `steps_out` is 1 because it only needs to go over 1 subquery. Thus, the outer reference is malformed.
```
benbellick:
```suggestion
The same shared relation `x` contains a single stored `steps_out=1` outer reference, but that value is only correct for one of its uses. The other use would need `steps_out=2`, so offset-based resolution is ambiguous.
```

site/docs/expressions/subqueries.md
```diff
@@ -68,6 +68,26 @@ WHERE x < ANY(SELECT y from t2)
...
+
+* `steps_out`: Resolves the reference by counting subquery boundaries
+  upward. This works correctly when the plan is a tree (each relation has a
+  single parent).
+
+* `id_reference`: Resolves the reference by naming the binding relation
+  via its plan-wide unique `RelCommon.id`. Must be used instead of
+  `steps_out` when the plan contains shared relations (`ReferenceRel`)
+  because offset-based resolution is ambiguous when multiple paths exist to
+  the same relation.
```
benbellick:
```suggestion
  `steps_out` when an outer reference appears inside a relation shared via
  `ReferenceRel` and that shared relation can be reached through multiple
  paths with different subquery depths, making `steps_out` ambiguous.
```

Just echoing the same language of my other comment.

proto/substrait/algebra.proto
```diff
@@ -1616,12 +1622,20 @@ message Expression {
...
+      oneof outer_reference_type {
+        // Number of subquery boundaries to traverse up for this field's
+        // reference. Must be >= 1.
+        uint32 steps_out = 1;
+
+        // Plan-wide unique id of the relation whose input order this reference is
+        // rooted on. Must match an id defined on a RelCommon in the plan.
+        // Must be >= 1. Must be used instead of steps_out when the plan
+        // contains shared relations (ReferenceRel) and offset-based
+        // resolution via steps_out would be ambiguous.
```
```suggestion
        // References the plan-wide unique id of the relation that this field reference
        // is rooted on. Must match an id defined on a RelCommon in the plan.
        // Must be >= 1. Must be used instead of steps_out when this outer reference
        // appears inside a relation shared via ReferenceRel and offset-based resolution
        // via steps_out would be ambiguous.
```

benbellick (approved):
Just some documentation suggestions I have suggested, feel free to take or leave them, they are mostly nits. Otherwise LGTM!

proto/substrait/algebra.proto
```diff
@@ -24,6 +24,12 @@ message RelCommon {
   Hint hint = 3;
   substrait.extensions.AdvancedExtension advanced_extension = 4;
 
+  // Optional plan-wide unique identifier for this relation. Required when
+  // this relation is the binding point for an OuterReference using
+  // id_reference. Must be unique across all rels within a Plan.
+  // Must be >= 1 when set.
+  optional uint32 id = 5;
```
benbellick: Ah one more thing I forgot to add. Do you think it is worth using the terms `relation_anchor` and `relation_reference` or something like that to be consistent with the other patterns of anchor / reference in this spec? 

I am maybe a +0.1 for doing that just to be consistent, unless someone has a compelling reason not to. But not a huge deal and isn't a blocker 🙂

yongchul: good point. I'm opt to `rel_anchor` to be consistent with other anchor -- reference scheme, which are more aligned to the proto structure.

yongchul: Updated. `rel_anchor` and `rel_reference`.

Issue #1042, opened 2026-04-10
Grammar: support enum args in inline aggregate test format


benbellick:
The `singleArgAggregateFuncCall` grammar rule only accepts a single `dataColumn`:

```
functName=identifier OParen dataColumn CParen
```

Functions like `std_dev` and `variance` take an enum arg (SAMPLE/POPULATION) alongside the data column, so they can't use this format. Test authors must use the compact table format instead, which is less readable for single-column data.

Ideally, the grammar would support something like:

```
std_dev(SAMPLE::enum, (1.0, 2.0, 3.0, 4.0, 5.0)::fp32) = 1.5811388::fp32?
```

The main challenge is that a general `(argument | dataColumn)*` arg list is structurally ambiguous with scalar test cases at the ANTLR level, which affects error recovery on intentionally malformed inputs. A more restrictive rule like `(enumArg Comma)* dataColumn` avoids this but doesn't generalize to all possible arg combinations (e.g. multiple inline data columns for `corr`).

PR #1043, opened 2026-04-10
fix(tests): use correct compact table row format in std_dev and variance tests

benbellick:
The compact aggregate test format uses each inner parenthesis group as a row. The `std_dev` and `variance` test files were using a flat list of values in a single group, e.g. `((1.0, 2.0, 3.0))`, which parses as 1 row with 3 columns rather than 3 rows with 1 column.

This changes them to the correct per-row format: `((1.0), (2.0), (3.0))`.

This format is documented here (this is format 2):
https://github.com/substrait-io/substrait/blob/87d73b663bcc153f9dda7dfe574dca36b1e4ed28/tests/README.md?plain=1#L71-L84

This came up in some work on `substrait-go`. 

(I opened #1042 to discuss how the inline-format 1 above could be used in the future. We would have to figure out how to deal with intermixing enum args.)

- - -
This change is [Reviewable](https://reviewable.io/reviews/substrait-io/substrait/1043)

### DIFF
--- a/tests/cases/arithmetic/std_dev.test
+++ b/tests/cases/arithmetic/std_dev.test
@@ -2,52 +2,52 @@
 ### SUBSTRAIT_INCLUDE: extension:io.substrait:functions_arithmetic
 
 # basic: Basic examples without any special cases
-((1.0, 2.0, 3.0, 4.0, 5.0)) std_dev(SAMPLE::enum, col0::fp32) = 1.5811388::fp32?
-((1.0, 2.0, 3.0, 4.0, 5.0)) std_dev(SAMPLE::enum, col0::fp64) = 1.5811388300841898::fp64?
-((1.0, 2.0, 3.0, 4.0, 5.0)) std_dev(POPULATION::enum, col0::fp32) = 1.4142135::fp32?
-((1.0, 2.0, 3.0, 4.0, 5.0)) std_dev(POPULATION::enum, col0::fp64) = 1.4142135623730951::fp64?
+((1.0), (2.0), (3.0), (4.0), (5.0)) std_dev(SAMPLE::enum, col0::fp32) = 1.5811388::fp32?
+((1.0), (2.0), (3.0), (4.0), (5.0)) std_dev(SAMPLE::enum, col0::fp64) = 1.5811388300841898::fp64?
+((1.0), (2.0), (3.0), (4.0), (5.0)) std_dev(POPULATION::enum, col0::fp32) = 1.4142135::fp32?
+((1.0), (2.0), (3.0), (4.0), (5.0)) std_dev(POPULATION::enum, col0::fp64) = 1.4142135623730951::fp64?
 
 # uniform_values: Standard deviation of uniform values
-((5.0, 5.0, 5.0, 5.0)) std_dev(SAMPLE::enum, col0::fp32) = 0.0::fp32?
-((5.0, 5.0, 5.0, 5.0)) std_dev(POPULATION::enum, col0::fp64) = 0.0::fp64?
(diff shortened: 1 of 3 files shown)


tests/cases/arithmetic/std_dev.test
```diff
@@ -2,52 +2,52 @@
 ### SUBSTRAIT_INCLUDE: extension:io.substrait:functions_arithmetic
 
 # basic: Basic examples without any special cases
-((1.0, 2.0, 3.0, 4.0, 5.0)) std_dev(SAMPLE::enum, col0::fp32) = 1.5811388::fp32?
-((1.0, 2.0, 3.0, 4.0, 5.0)) std_dev(SAMPLE::enum, col0::fp64) = 1.5811388300841898::fp64?
-((1.0, 2.0, 3.0, 4.0, 5.0)) std_dev(POPULATION::enum, col0::fp32) = 1.4142135::fp32?
-((1.0, 2.0, 3.0, 4.0, 5.0)) std_dev(POPULATION::enum, col0::fp64) = 1.4142135623730951::fp64?
+((1.0), (2.0), (3.0), (4.0), (5.0)) std_dev(SAMPLE::enum, col0::fp32) = 1.5811388::fp32?
```
yongchul, line 5: what am I missing here? why the values don't change?

benbellick: Before this diff, the table in this test was:
| col0 | col1 | col2 | col3 | col4 |
|------|------|------|------|------|
| 1.0  | 2.0  | 3.0  | 4.0  | 5.0  |

After this change, the table is:
| col0 |
|------|
| 1.0  |
| 2.0  |
| 3.0  |
| 4.0  |
| 5.0  |

That is my understanding at least. Since this is an aggregate function, I assume that the intention here was for the second thing, with one column, rather than the first. 

I just assumed that the computed values were based off of that.

benbellick: I think this was contributed by @nielspardon in https://github.com/substrait-io/substrait/pull/1011. Is my understanding correct?

benbellick: I cannot really find other example of this syntax used in actual tests, other than in: 
https://github.com/substrait-io/substrait/blob/ff70d18c25fe3e078d40bb3cff6074cf299d0b50/tests/coverage/test_coverage.py#L220

yongchul: Yeah, the transpose makes sense. I just can't get what the test was doing previously... was it just running aggregate of scalar values or aggregate rows? or, this is just funky way of behaving differently based on nuanced argument(s)?

nielspardon: Thanks for catching this. My bad, I was relying on AI too much and didn't check the format myself. 

Do you think it make sense to have some sort of check whether there are columns in the table format not referenced in the function invocation? Like in this case since there is only a reference to `col0` but no references to `col1` - `col4` could have helped as a consistency check.

nielspardon: Since you were using actual markdown tables to explain the behavior in this thread should we also add markdown tables to the `tests/README.md`?

nielspardon (approved):
LGTM, thanks for catching this

yongchul (approved):

tests/cases/arithmetic/std_dev.test
benbellick: Will do

benbellick: As for introducing a check, I think it would be a great idea. I added a ticket here: #1047 

Note that it may involve some serious refactors of this python code 😅

PR #1044, opened 2026-04-13
feat: clarify post_join_filter. add residual_expressions to hash join and merge join

yongchul:
# Background

JoinRel, HashJoinRel, and MergeJoinRel have a field named`post_join_filter`, causing confusion. The confusing point is whether `post_join_filter` is part of the join predicate (i.e., two rows are **matched** when it evaluates to `true`) or not (i.e., as the name says, this is **post** join filter, thus `filter` evaluates after the join operation).

Substrait [FAQ](https://substrait.io/faq/#what-is-the-purpose-of-the-post-join-filter-field-on-join-relations) calls out that the `post_join_filter` is part of join predicate but @yongchul was deeply discontent about the explanation and the naming.

After lengthy discussion in the Slack channel, #807 was created. After more back-and-forth, the original intent of `post_join_filter` was indeed [**post** join filter, not part of join predicate](https://github.com/substrait-io/substrait/pull/807#issuecomment-4230380222).

Also, it appears that Calcite has precisely the same name and same semantic, post join filter. Following code point shows placing the filter on top of the join.
 https://github.com/apache/calcite/blob/0a39568b167592ded8db1128b5838982ffe264f3/core/src/main/java/org/apache/calcite/rel/rules/LoptOptimizeJoinRule.java#L553-L559

# What this changes do?

TL;DR; Make `post_join_filter` as `post_join_filter` again.

1. Documents `post_join_filter` is not join condition but semantically filter on top of join.
2. Introduce `residual_expression` to HashJoinRel and MergeJoinRel so that they can express non-equality join predicate -- I guess this was the confusion who tried to shove non-equality join predicate to `HashJoin` and `MergeJoin` somewhere and distorted the meaning without much thought.
3. Drop `equi` from hash join and merge join documentation as with `residual_expression` they are capable of supporting arbitrary join predicate.
4. Fix the FAQ.

# Breaking change?

This is debatable. According to previous FAQ, `post_join_filter` was `residual_expression` introduced in this PR. If we take the FAQ as part of the "spec", then this is a breaking change.

If we don't consider FAQ as a spec, then this is not a breaking change but clarification and adding a new feature -- HashJoinRel, MergeJoinRel now supports arbitrary join predicate.

# AI disclaimer

The PR is assisted by Claud Opus 4.6.

- - -
This change is [Reviewable](https://reviewable.io/reviews/substrait-io/substrait/1044)


site/docs/faq.md
```diff
@@ -6,9 +6,12 @@ title: FAQ
 
 ## What is the purpose of the post-join filter field on Join relations?
 
-The post-join filter on the various Join relations is not always equivalent to an explicit Filter relation AFTER the Join.
+`post_join_filter` is an optional filter on the output records of a join. It is applied after matching and any join-type–specific post-processing (e.g., null-padding unmatched rows for outer joins, emitting unmatched rows for anti joins) are complete. It is semantically equivalent to placing a separate `Filter` relation directly above the join. Because it does not participate in matching, it has no effect on which rows are considered matched or unmatched for the purpose of outer, anti, semi, or mark join semantics. When omitted it defaults to `true`.
```
nielspardon, line 9: super minor remark: I noticed that in the previous changes in the other md files you changed `true` to `True` while here you are using `true`. Not sure if we want to enforce consistent casing of keywords like `True|true` / `False|false`.

site/docs/relations/physical_relations.md
nielspardon: should the residual expression also be mentioned here?

yongchul: Thank you. I just noticed `Equijoin`. This is gotta be just join with `residual_expression`. Will update the documentation.

site/docs/faq.md
yongchul: I'll blame copilot 😆 I'll stick with `true`. I also noticed that other parts of documentation using `True`, probably need to clean up as well.

proto/substrait/algebra.proto
```diff
@@ -855,12 +863,23 @@ message HashJoinRel {
   // hash function for a given comparsion function or to reject the plan if it cannot
   // do so.
   repeated ComparisonJoinKey keys = 8;
+  // An optional boolean filter applied to each output record after join
+  // matching and null-padding (for outer/mark joins) are complete.
+  // Semantically equivalent to placing a FilterRel directly above this join.
   Expression post_join_filter = 6;
```
westonpace, line 869: Is there any known engine that has this feature?

yongchul: Great question. I know one that I can fix (it is under my control lol). I will check data fusion, duckdb, and velox.

@benbellick @vbarua @nielspardon any impact on your sides?

yongchul: * datafusion: [not supported](https://github.com/apache/datafusion/blob/5c653bee5da64003915f6dfeb3da15759b091a8d/datafusion/substrait/src/logical_plan/consumer/rel/join_rel.rs#L32-L34)
* duckdb: does not populate post_join_filter
* gluten: does use post_join_filter but not sure whether it is consistent with this PR.

nielspardon: The engines we are currently dealing with would also handle this as a separate `FilterRel` instead of merging this into the Join operation. We may have some use cases in the future for representing fused operators in Substrait though I would look at this separately.

For me I'm fine with both this improved documentation or removing it.

benbellick: I just checked and we do not use it internally.

Clarifying the documentation is a good idea, though I would prefer removing it if no user of substrait is actually using it.

jacques-n (approved):
lgtm

proto/substrait/algebra.proto
yongchul: @nielspardon @benbellick any final thoughts on the PR? This one I can wait until next week meeting.

proto/substrait/algebra.proto
```diff
@@ -855,12 +863,23 @@ message HashJoinRel {
...
 
   JoinType type = 7;
 
-  // Specifies which side of input to build the hash table for this hash join. Default is `BUILD_INPUT_RIGHT`.
+  // Specifies which side of input to build the hash table for this hash join.
+  // Defaults to `BUILD_INPUT_RIGHT`.
```
benbellick, line 880: Does this mean that `BUILD_INPUT_UNSPECIFIED` => `BUILD_INPUT_RIGHT`? 

I assume that reason for the presence of `UNSPECIFIED` [is to follow the protobuf best practices, but the protobuf docs are pretty explicit that the unspecified value should not have any meaningful semantics](https://protobuf.dev/best-practices/dos-donts/):
> It may be tempting to declare this default as a semantically meaningful value but as a general rule, do not, to aid in the evolution of your protocol as new enum values are added over time.

This doesn't have to be for this PR, but I think we should should force producers to specify what they mean explicitly. Maybe we do that in one larger PR next.

site/docs/relations/logical_relations.md
benbellick, line 1: This file says:
> All field references of [Join Properties](#join-properties) are over this order.

where "this order" is the input order. 

However, this is no longer the case since `post_join_filter` operates over the output order. Could we change that language as well?

proto/substrait/algebra.proto
```diff
@@ -255,12 +255,18 @@ message ProjectRel {
...
+// types, a join condition, and an optional filter on the joined output.
 message JoinRel {
   RelCommon common = 1;
   Rel left = 2;
   Rel right = 3;
+  // A boolean condition evaluated over the left and right inputs that
+  // determines whether a pair of records is a match.
   Expression expression = 4;
+  // An optional boolean filter applied to each output record after join
+  // matching and null-padding (for outer/mark joins) are complete.
```
```suggestion
  // An optional boolean filter applied to each output record after
  // join-type-specific output formation is complete.
```
I'm not too familiar with mark joins, but my understanding is that they append a nullable boolean mark column rather than null-padding the opposite input. 

In that case, it might make more sense to keep this language generic, because otherwise it reads like we are saying mark joins null-pad.

site/docs/relations/physical_relations.md
```diff
@@ -23,7 +23,8 @@ The hash equijoin join operator will build a hash table out of one input (defaul
 | Build Input         | Specifies which input is the `Build`.                                                                                                                                                                                   | Optional, defaults to build `Right`, probe `Left`. |
```
benbellick, line 23: FYI these tables still only discuss left / right keys (deprecated) rather than comparison join keys. 

We can address in another PR if you'd like. Though I would love to just get rid of the deprecated keys instead.

benbellick (commented):
Generally looks good to me, just a few small comments!

proto/substrait/algebra.proto
benbellick: I made an issue to track this: #1069

yongchul: > Does this mean that BUILD_INPUT_UNSPECIFIED => BUILD_INPUT_RIGHT?

Yes, for the backward compatibility. Mandating would be a breaking change and we can do it in another PR as we discussed in the last meeting.

site/docs/relations/logical_relations.md
yongchul: Good catch. It should be over direct output order -- but that is the same as direct input order.

site/docs/relations/physical_relations.md
yongchul: Let me see whether I make the change easily in this PR. otherwise, I'll follow up with another one.

yongchul: Looked into it and looks like better be handled in a separate PR.

yongchul: Created #1070

benbellick (approved):
LGTM!

proto/substrait/algebra.proto
nielspardon: what was the conclusion on this discussion? we update the documentation in this PR and deprecate the post join filter in a follow-up?

yongchul: @nielspardon deprecation will be a separate PR if we wanted to deprecate.

nielspardon: just watched the community sync replay and it seems @benbellick found a use of it in substrait-go

nielspardon (approved):

PR #1046, opened 2026-04-13
ci: use actions/create-github-app-token

nielspardon:
- replaces the [deprecated `tibdex/github-app-token`](https://github.com/tibdex/github-app-token) with [`actions/create-github-app-token`](https://github.com/actions/create-github-app-token)
- requests token for substrait and substrait-packaging repos to enable `ci/release/notify.sh` script

- - -
This change is [Reviewable](https://reviewable.io/reviews/substrait-io/substrait/1046)

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -18,11 +18,11 @@ jobs:
     runs-on: ubuntu-latest
     if: github.repository == 'substrait-io/substrait'
     steps:
-      - uses: tibdex/github-app-token@v2
+      - uses: actions/create-github-app-token@v3
         id: generate-token
         with:
-          app_id: ${{ secrets.APP_ID }}
-          private_key: ${{ secrets.APP_PRIVATE_KEY }}
+          client-id: ${{ secrets.APP_ID }}
+          private-key: ${{ secrets.APP_PRIVATE_KEY }}
 
       - uses: actions/checkout@v6
         with:
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -23,6 +23,12 @@ jobs:
         with:
           client-id: ${{ secrets.APP_ID }}
           private-key: ${{ secrets.APP_PRIVATE_KEY }}
+          owner: ${{ github.repository_owner }}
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yml
```diff
@@ -18,11 +18,17 @@ jobs:
...
-          app_id: ${{ secrets.APP_ID }}
-          private_key: ${{ secrets.APP_PRIVATE_KEY }}
+          client-id: ${{ secrets.APP_ID }}
+          private-key: ${{ secrets.APP_PRIVATE_KEY }}
+          owner: ${{ github.repository_owner }}
+          # request token for both substrait and substrait-packaging repos
+          # for ci/release/notify.sh
+          repositories: |
+            substrait
+            substrait-packaging
```
nielspardon, line 31: @vbarua I think this should help with the 403 we are getting currently:

```
[4:43:35 AM] [semantic-release] [@semantic-release/exec] › ℹ  Call script ci/release/notify.sh 0.88.0
could not create workflow dispatch event: HTTP 403: Resource not accessible by integration (https://api.github.com/repos/substrait-io/substrait-packaging/actions/workflows/226147488/dispatches)
```

https://github.com/substrait-io/substrait/actions/runs/24298853195/job/70948495423#step:5:156

nielspardon: I think we also need to change the permissions for the Github App to include `Actions` `write` access to allow it to dispatch the workflow run in the `substrait-packaging` repo. This requires Github org admin permissions which apparently PMC members do not have.

vbarua (approved):
Thanks for updating this. I can take a look at the permissions issue, but I think I also gave you the correct access now.

.github/workflows/release.yml
vbarua: Based on https://docs.github.com/en/rest/actions/workflows?apiVersion=2026-03-10#create-a-workflow-dispatch-event, I added `Actions: Write` permissions to the Substrait Project Bot. 

Going to merge this, and we can see if that solves the issue in the next release.

PR #1048, opened 2026-04-13
fix(tests): correct decimal sum and nullif results

bvolpato:
## Summary

Correct two inconsistent expected result types in the function test corpus:

- decimal `sum` preserves input scale `S`, so `dec<2,1>` returns `dec?<38,1>`
- `nullif(any1, any1) -> any1?`, so decimal arguments return nullable decimal rather than nullable boolean

The earlier strict type-checker implementation has been removed. A future checker can reuse the generated ANTLR type grammar without adding a second parser.

## Testing

- `pixi run lint`
- `pixi run test` (`137 passed`)

- - -
This change is [Reviewable](https://reviewable.io/reviews/substrait-io/substrait/1048)


tests/coverage/type_checker.py
```diff
@@ -0,0 +1,426 @@
...
+    ]
+)
+
+
+def canon_base(base):
+    base = base.lower()
+    return _SHORT_TO_LONG.get(base, base)
+
+
+def parse_type(s):
```
benbellick, line 60: I will try and find time to review this more thoroughly, but on first glance, it feels unnecessary to hand-write a parser for a type language with a formal grammar and generated parsers. 

There is something brittle about the current way the code is written before this PR which I can't quite recall, but I remember in the past trying to rewrite this code to better take advantage of the antlr grammar and stopping because I ran into some hurdles. Might be worth reconsidering a larger rewrite if it means we can avoid having to maintain this.

bvolpato:
[review-prs-verification]
Head: 2e77365aaac2f1a1b5e7795c1de80454fa285fb3
Ready to merge: yes

Files reviewed:
- tests/cases/arithmetic_decimal/sum_decimal.test
- tests/cases/comparison/nullif.test

Changes:
- narrowed the PR to two verified testcase corrections
- removed the handwritten type parser and strict-checker integration
- updated PR title and description to match final scope
- replied to and resolved the ANTLR reuse review thread

Tests/checks:
- local: `pixi run lint` passed
- local: `pixi run test` passed (`137 passed`)
- GitHub: all checks passed, including style, tests, dry-run release, protobuf, YAML, title, license, and CLA
- both branch commits have valid signatures

tests/coverage/type_checker.py
bvolpato: Agreed. I narrowed this PR to the two verified testcase corrections and removed the handwritten parser plus its integration. Strict parameter validation can follow separately using the generated ANTLR grammar.

nielspardon (approved):
LGTM

yongchul (approved):

PR #1064, opened 2026-04-30
fix(tests): make typed temporal test literals unquoted

benbellick:
Quoted date/time-shaped values are ambiguous in the test grammar. The old lexer matched values like `'2020-05-10'` as [`DateLiteral`](https://github.com/substrait-io/substrait/blob/19716584708f98b06eb3f0d73a1f62e15bf6b169/grammar/FuncTestCaseLexer.g4#L77-L79) before the parser could see that the argument was annotated as [`::str`](https://github.com/substrait-io/substrait/blob/19716584708f98b06eb3f0d73a1f62e15bf6b169/grammar/FuncTestCaseParser.g4#L153-L155).

This resulted in cases like `identity('2020-05-10'::str) = '2020-05-10'::str` failing to parse with `mismatched input 'str' expecting 'DATE'`. Because `'2020-05-10'` is lexed as `DateLiteral` instead of `StringLiteral`, it fails the [`stringArg` rule](https://github.com/substrait-io/substrait/blob/19716584708f98b06eb3f0d73a1f62e15bf6b169/grammar/FuncTestCaseParser.g4#L153-L155):

```antlr
stringArg
    : StringLiteral DoubleColon stringType
    ;
```

To fix this, typed temporal and interval literals are now unquoted, so quoted values are always available as strings. Concretely, tests like:
```
lt('2020-12-30'::date, '2020-12-31'::date) = true::bool
```
instead become:
```
lt(2020-12-30::date, 2020-12-31::date) = true::bool
```

Existing datetime tests have been updated to use the unquoted form.

## Alternatives

I first tried adding a shared `quotedLiteral` parser rule and allowing `stringArg` to accept date/time/interval tokens. That was smaller and backwards compatible, but it left the grammar treating string arguments as date/time tokens in some cases. Making typed temporal literals unquoted removes the ambiguity instead.

## Breaking Change

I am unsure if this should be treated as breaking. It breaks the scalar function test format, but not the Substrait format itself.

---
Note: This PR was developed with AI assistance. All changes have been reviewed, and I take full responsibility for this contribution.

- - -
This change is [Reviewable](https://reviewable.io/reviews/substrait-io/substrait/1064)


grammar/FuncTestCaseLexer.g4
```diff
@@ -62,20 +62,20 @@ fragment FourDigits: [0-9][0-9][0-9][0-9];
 fragment TwoDigits: [0-9][0-9];
 
 TimestampTzLiteral
-    : '\'' FourDigits '-' TwoDigits '-' TwoDigits 'T' TwoDigits ':' TwoDigits ':' TwoDigits ( '.' [0-9]+ )?
-        [+-] TwoDigits ':' TwoDigits '\''
+    : FourDigits '-' TwoDigits '-' TwoDigits 'T' TwoDigits ':' TwoDigits ':' TwoDigits ( '.' [0-9]+ )?
```
yongchul, line 65: The previous quoted string is weird... It shouldn't exist from the beginning IMO -- it should have been just string literal that is casted to something else like timestamp or something. The actual validation could be done in the casting, not in the lexer lexer (or call lexer with this rule let's say).

So this change makes sense to me.

yongchul (approved):

grammar/FuncTestCaseLexer.g4
benbellick: I agree. This is both easier to parse and a nicer format. Thanks for the review! 

------
Do you think this counts as a breaking change? We are technically altering the test format... however, I'm not sure if it is worth invoking a breaking change when there is no spec change.

yongchul: I don't think it's a breaking change...

benbellick:
Per today's sync, this is not considered a breaking change. I will merge later today if there aren't any last minute thoughts.

PR #1075, opened 2026-05-11
ci: switch from dependabot to renovate

nielspardon:
We are currently using dependabot which has a very limited feature set, for example it can not handle managing a `pixi.lock` file as demonstrated in https://github.com/substrait-io/substrait/actions/runs/25411815587/job/74535026421?pr=1072

Renovate is also free to use and a much more powerful dependency auto update tool compared to dependabot. It can handle `pixi.lock` files among other things like e.g. it could also be used to update Git submodules to tags instead of just the latest commit (https://docs.renovatebot.com/modules/manager/git-submodules/#updating-to-specific-tag-values).

It gives more flexibility and helps to address the issue with dependabot not handling the `pixi.lock` file properly.

## Why this is config-only now

Renovate only learned to refresh `pixi.lock` after bumping dependencies declared in `pyproject.toml` in https://github.com/renovatebot/renovate/pull/44004, released in Renovate 44.0.0. Because `pixi lock` can execute arbitrary code from conda package hooks, that same change gated the behaviour — for the `pixi` manager as well, as an explicit breaking change — behind the `allowedUnsafeExecutions` option. That option is `globalOnly`, so it can never be set from a repository's `renovate.json`, which is why earlier revisions of this PR carried a self-hosted `.github/workflows/renovate.yml` and needed a `substrait-io`-owned GitHub App.

Mend has since added `pixi` to `allowedUnsafeExecutions` on the hosted Renovate App — in place since 2026-08-04, confirmed by @jamietanna below — so the hosted App refreshes `pixi.lock` on its own and the self-hosted workflow is gone from this PR. That also settles the `pixi lock` code-execution exposure @benbellick raised: it is covered by Mend's hardening instead of running on our own runners with an App token in the environment. All three pixi repositories are therefore config-only, and this PR is now identical in approach to substrait-io/substrait-python#244 and substrait-io/substrait-packaging#52.

That Renovate release was verified end to end on a fork before the hosted App gained the option: https://github.com/nielspardon/substrait/pull/28 bumps `protobuf` in `pyproject.toml` and regenerates `pixi.lock` across all four platforms in the same commit, with CI green.

## Changes

- Adds a `renovate.json` config file configuring semantic commits as the commit style for Renovate and the commit type as `build` and scope as `deps` to match the dependabot style following: https://docs.renovatebot.com/configuration-options/#semanticcommitscope
  - `:semanticCommitTypeAll(build)` is needed for the `build` type to actually take effect. `config:recommended` pulls in `:semanticPrefixFixDepsChoreOthers`, which contributes a `matchPackageNames: ["*"]` rule setting the type to `chore`, and package rules override top-level config.
  - Groups `pyproject.toml` dependency bumps into one PR. The Python dev dependencies and the conda dependencies all share a single `pixi.lock`, so ungrouped PRs would each rewrite that file and conflict with one another as soon as the first one merges. Major bumps still get their own PR.
  - Disables updates to `python`. Renovate treats both `requires-python` and the `[tool.pixi.dependencies]` pin as dependencies, but changing the supported Python version is a project decision rather than a routine bump. Leaving it enabled also puts a `pep440: failed to calculate newValue` warning on the Dependency Dashboard, because the range `>=3.13,<3.14` cannot be rewritten against a Python pre-release.
  - Enables the `pre-commit` manager, which is off by default. Our `.pre-commit-config.yaml` has drifted well behind `pyproject.toml` (for example `ruff` `v0.8.4` there against `>=0.15.22` here) and dependabot cannot see that file at all.
  - Holds new releases for 14 days before proposing them (`minimumReleaseAge`), so that a compromised or broken release has time to be caught, following the cooldown advice in https://pixi.prefix.dev/latest/security/#2-exclude-newer. This covers the conda dependencies in `[tool.pixi.dependencies]` as well: the `conda` datasource has always returned release timestamps (from `upload_time` on the Anaconda.org API and `createdAt` on prefix.dev), it just never declared support for them, which is corrected in https://github.com/renovatebot/renovate/pull/45622 — shipped in Renovate 44.57.0 and now running on the hosted App. `internalChecksFilter` stays at its default of `strict`, so held versions are filtered out of PR creation and surface on the Dependency Dashboard instead. Two limits worth knowing: a dependency on a datasource that genuinely has no release timestamps would sit pending rather than be proposed, because `minimumReleaseAgeBehaviour` defaults to `timestamp-required`; and Renovate does not enforce this for `lockFileMaintenance` updates, since the package manager performs those changes itself. Putting a cooldown on the transitive pins that `pixi lock` chooses therefore needs pixi's own `exclude-newer`, which is a follow-up rather than part of this PR.
- Removes the `dependabot.yml` configuration file to disable dependabot.
- Adds a commitlint ignore pattern for Renovate PR descriptions. The dependabot patterns stay for now because dependabot PRs opened before this change are still open.
- Replaces #1074

## What has to happen for Renovate to start

The hosted Renovate App was installed on this repository when it opened #1074, and merging a `renovate.json` onto `main` is what makes it run: a closed onboarding PR only holds Renovate back while the repository has no config file. Nothing else is needed — no GitHub App to create and no repository secrets. If no Dependency Dashboard issue shows up after this merges, the App installation is the thing to check.

## What changes about the PRs we receive

Renovate covers ground that our dependabot configuration never did, so expect a batch of new PRs on the first run:

- The conda dependencies in `[tool.pixi.dependencies]`, which dependabot's `pip` ecosystem never looked at. They are all within their declared ranges right now, so they only show up via lock file maintenance until one of them moves out of range.
- `go.mod`, which had no dependabot ecosystem entry at all. Note that we have no Go job in CI, so those PRs arrive without coverage. Add `"gomod": { "enabled": false }` if we would rather not receive them.
- The `pre-commit` hooks described above. Renovate attaches a note to those PRs saying the `pre-commit` manager is not supported by the pre-commit maintainers.
- The `pixi-version` pinned in our workflows, which Renovate picks up from the `prefix-dev/setup-pixi` inputs in addition to the action itself.

Dependencies declared as ranges only get a PR once the new version falls outside the range, so a quiet dependency like `ruff>=0.15.22,<1` is expected rather than a sign that something is broken. `lockFileMaintenance` covers the rest by refreshing the transitive pins in `pixi.lock` weekly.

- - -
This change is [Reviewable](https://reviewable.io/reviews/substrait-io/substrait/1075)


benbellick:
Thanks @nielspardon. Can we include a description in the PR body for why we should switch from dependabot to renovate? I _think_ it has something to do with how dependabot interacts with pixi but it would be great to have it in writing somewhere so we can reference it in the future if something comes up. Thank you!

nielspardon:
updated the description. let me know if the description looks good for you.

.github/workflows/pr_title.yml
```diff
@@ -16,8 +16,11 @@ jobs:
           echo 'module.exports = {
             // Workaround for https://github.com/dependabot/dependabot-core/issues/5923
```
benbellick: Is this comment still relevant if we no longer use dependabot?

renovate.json
benbellick, line 1: Is there a way to validate that this works before merging?

nielspardon: I've been testing this on my private fork and I found that the current pixi support for Renovate is only covering the direct Pixi dependencies in `[tool.pixi.dependencies]` e.g. NodeJS which when it does update the version also updates the pixi.lock file: https://github.com/nielspardon/substrait/pull/5/changes

When it updates one of the dependencies in the dev dependency group in `[dependency-groups]` it does not update those using the Pixi manager and Renovate thus also not updating the `pixi.lock` file currently: https://github.com/nielspardon/substrait/pull/7/changes

nielspardon: I logged a feature request with Renovate to help us fix this gap: https://github.com/renovatebot/renovate/discussions/43260

nielspardon: FYI: Renovate also supports `postUpgradeTasks` if you are hosting it yourself. When using the hosted one with the Github App it does not allow custom commands for security reasons.

benbellick: Thanks for investigating. Hopefully we can figure out something that will work for our setup.

nielspardon: I opened a PR to fix renovate for pixi lock file updates: https://github.com/renovatebot/renovate/pull/44004

tokoko: do we need dependency groups? in our case, what's the difference between dependencies and a dev dependency group? isn't the whole thing a single dev environment anyway?

nielspardon: maybe we should discuss this in a separate issue. feels unrelated to switching from dependabot to renovate.

tokoko: yeah, just a possible shortcut :laughing: doesn't really matter though if the issue is being addressed in renovate anyway

nielspardon: The issue with Renovate is not caused by the dependencies being in a dependency group. We would have the same issue if they were regular Python dependencies in `[project.dependencies]` in `pyproject.toml`. The issue is that the `pixi` support in Renovate today would only update the `pixi.lock` file when suggesting a dependency update in `pixi.toml` or in `[tool.pixi.dependencies]` in `pyproject.toml`. This is better than dependabot which has no pixi support at all but wouldn't help since we are using `pyproject.toml` in this repo and no `pixi.toml`.

The PR I opened in Renovate which got one approval already makes sure the `pixi.lock` file gets updated when Python dependencies in `pyproject.toml` (outside of `[tool.pixi.dependencies]`) get bumped including dependency groups.

Whether we should declare the dependencies as project dependencies or in a dev dependency group is orthogonal in my opinion.

jamietanna: Also, when the Renovate "configure Renovate" / onboarding PR is raised, y'all will see what's pending from it - you can even put these config changes into that branch (requires write access to repo) and it'll update the PR description to show changes it'll make :)

nielspardon:
I ran this end to end on my fork so we can see the behaviour before deciding. The setup there is identical to what this PR proposes, only the GitHub App is mine rather than a `substrait-io` one.

**The result we were after:** https://github.com/nielspardon/substrait/pull/28 bumps `protobuf` from `6.33.1` to `6.33.6` in `[dependency-groups]` of `pyproject.toml` **and** regenerates `pixi.lock` in the same commit, across all four platforms. All 13 CI jobs pass, which also confirms the lock file that Renovate's pixi 0.75.0 produced is still readable by the `pixi-version: v0.73.0` we pin in CI. The commit is authored by the App and signed (`verified=true`) because Renovate uses GitHub's commit API when running as an App.

From the run log, the part that the hosted App cannot do:

```
install-tool pixi 0.75.0
pixi lock --no-progress --color=never --quiet
  stderr: "✔ Updated lock file\nEnvironment: dev\n  ~ (pypi) protobuf  6.33.1  ->  6.33.6\n"
  updatedArtifacts: ["pixi.lock"]
  2 file(s) to commit
```

**Dependency Dashboard:** https://github.com/nielspardon/substrait/issues/27 lists everything Renovate detected, including the rate-limited updates queued behind the default limit of two PRs per hour, and gives us checkboxes to pull any of them forward. Dependabot has no equivalent.

A few things the fork run surfaced that are worth knowing before review:

- Renovate detects `prefix-dev/pixi v0.73.0` from the `pixi-version` input in our workflows, so the pixi version CI uses gets kept up to date too.
- The conda dependencies in `[tool.pixi.dependencies]` are all inside their declared ranges right now, so they produce no PR of their own. `lockFileMaintenance` is what refreshes their pins in `pixi.lock`, and it is queued on the dashboard.
- Leaving `python` updatable put a `pep440: failed to calculate newValue` warning in the dashboard's Repository Problems, because `>=3.13,<3.14` cannot be rewritten against the `3.15.0b4` pre-release. The config in this PR disables `python` updates, which removes the warning and also keeps the supported Python version a deliberate decision rather than a routine bump.
- The `github-actions` manager needs the App to have `Workflows: read+write`, otherwise Renovate silently cannot open PRs for anything under `.github/workflows`, which is a job dependabot currently does for us.

@benbellick to answer your earlier question about the commitlint comment: the dependabot ignore patterns still need to stay, because #1061, #1092, #1125, #1126 and #1135 were opened by dependabot and would start failing the PR Title Check without them. I have scoped the `dependabot-core` link to just those two patterns and added a separate one for Renovate. We can drop the dependabot patterns once those PRs are closed out.

One correction to my earlier comments on this PR: I said the gap was that Renovate only updated `pixi.lock` for `[tool.pixi.dependencies]`. That is fixed now, but as of Renovate 44.0.0 the gate applies to the `pixi` manager as well, so the hosted App now refreshes `pixi.lock` for nothing at all. Self-hosting is what makes it work, not merely what makes it work better.

jamietanna:
> That is fixed now, but as of Renovate 44.0.0 the gate applies to the pixi manager as well, so the hosted App now refreshes pixi.lock for nothing at all. Self-hosting is what makes it work, not merely what makes it work better.

When I get an internal PR review, we'll be setting `allowedUnsafeExecutions=[..., "pixi"]` so you'll be able to use `pixi lock` on the Mend-hosted apps, no need to self-host (unless you want to!)

nielspardon:
> > That is fixed now, but as of Renovate 44.0.0 the gate applies to the pixi manager as well, so the hosted App now refreshes pixi.lock for nothing at all. Self-hosting is what makes it work, not merely what makes it work better.
> 
> When I get an internal PR review, we'll be setting `allowedUnsafeExecutions=[..., "pixi"]` so you'll be able to use `pixi lock` on the Mend-hosted apps, no need to self-host (unless you want to!)

That's great to hear. Thanks.

.github/workflows/renovate.yml
```diff
@@ -0,0 +1,67 @@
...
+        with:
+          token: ${{ steps.generate-token.outputs.token }}
+          renovate-version: ${{ env.RENOVATE_VERSION }}
+        env:
+          LOG_LEVEL: ${{ inputs.logLevel || 'info' }}
+          # `pixi lock` can execute arbitrary code from conda package hooks, so
+          # Renovate requires it to be allow-listed. This option is self-hosted
+          # only, which is why the hosted Renovate App cannot keep pixi.lock in
+          # sync and why this workflow exists.
+          RENOVATE_ALLOWED_UNSAFE_EXECUTIONS: pixi
```
benbellick: This makes me a bit nervous. Could a malicious package execute code during `pixi lock` and access the GitHub App token?

jamietanna: We (Renovate maintainers) take steps to avoid sensitive env vars from being accessible to commands, but if the command being executed can ie read files on disks or check memory on the runner, it may be able to leverage secrets

This is largely why it's behind a scary sounding flag, because it can be risky, so it's a choice y'all have to make

(see also: https://docs.renovatebot.com/environment-variable-handling/)

jamietanna: Mend-hosted Renovate handles this with additional hardening not commonly done by others (also because we're hosting it as a service for the general public)

nielspardon: There's a pretty good write-up by the pixi maintainers on the security exposure when using pixi: https://pixi.prefix.dev/latest/security/

Section 4 explains risks from Conda post-link scripts and environment activation scripts. This does not only apply to the execution of `pixi lock` while running Renovate but also on our PR and release builds.

The Github App permissions recommended for Renovate are documented here: https://docs.renovatebot.com/modules/platform/github/#running-as-a-github-app

I just checked and I still need to set up the Github App for the `substrait-io` org and install it for this repo before we merge.

nielspardon: @jamietanna did you get pixi enabled on the hosted service in the end? I'm not sure where I could check.

.github/workflows/renovate.yml
```diff
@@ -0,0 +1,67 @@
...
+concurrency:
+  group: renovate
+  cancel-in-progress: false
+
+env:
+  # Renovate >= 44.0.0 is required: it is the first release that regenerates
+  # pixi.lock after bumping dependencies declared in pyproject.toml.
+  # See https://github.com/renovatebot/renovate/pull/44004
+  # renovate: datasource=docker depName=ghcr.io/renovatebot/renovate
+  RENOVATE_VERSION: 44.3.3
```
nielspardon: if we go with the self-hosted approach we should also bump the version before merging since there have been a couple CVEs fixed since I created this PR

.github/workflows/renovate.yml
```diff
@@ -0,0 +1,67 @@
...
+
+# Two Renovate runs against the same repository would fight over branches.
+concurrency:
+  group: renovate
+  cancel-in-progress: false
+
+env:
+  # Renovate >= 44.0.0 is required: it is the first release that regenerates
+  # pixi.lock after bumping dependencies declared in pyproject.toml.
+  # See https://github.com/renovatebot/renovate/pull/44004
```
jamietanna: No longer relevant?

```suggestion
```

.github/workflows/renovate.yml
jamietanna: I'd upgrade to a maximum of:

```suggestion
  RENOVATE_VERSION: 44.14.7
```

(As it includes https://github.com/renovatebot/renovate/discussions/45495)

Then, you can make sure that Renovate bumps itself - it'll give you a confirmation everything's wired up correctly

.github/workflows/renovate.yml
```diff
@@ -0,0 +1,67 @@
...
+    timeout-minutes: 60
+    # As in release.yml, don't run the schedule on forks - but still let a fork
+    # owner dispatch a run manually to test their own setup.
+    if: github.event_name == 'workflow_dispatch' || github.repository == 'substrait-io/substrait'
+    steps:
+      - uses: actions/create-github-app-token@v3
+        id: generate-token
+        with:
+          client-id: ${{ secrets.RENOVATE_APP_CLIENT_ID }}
+          private-key: ${{ secrets.RENOVATE_APP_PRIVATE_KEY }}
```
jamietanna: Make sure this is only accessible to the `main` branch

.github/workflows/renovate.yml
jamietanna: Yep, it's been there since 2026-08-04. You can verify by checking any repo's job logs, and looking for the `allowedUnsafeExecutions` listed at the top of the job

.github/workflows/renovate.yml
nielspardon: dropped the self-host GH actions workflow

.github/workflows/renovate.yml
nielspardon: dropped the self-host GH actions workflow

.github/workflows/renovate.yml
nielspardon: dropped the self-host GH actions workflow

.github/workflows/renovate.yml
nielspardon: I dropped the self-host GH actions workflow

renovate.json
```diff
@@ -0,0 +1,31 @@
...
+    {
+      "description": "Changing the supported Python version is a project decision, not a routine bump. Covers both requires-python and the pixi python pin",
+      "matchPackageNames": ["python"],
+      "enabled": false
+    },
+    {
+      "description": "Hold new releases for 14 days so a compromised or broken release has time to be caught. The conda datasource provides no release timestamps, and minimumReleaseAgeBehaviour defaults to timestamp-required, so including it would suppress conda updates indefinitely rather than delay them. A cooldown for conda packages and for the transitive pins in pixi.lock has to come from pixi's own exclude-newer instead",
+      "matchDatasources": ["!conda"],
+      "minimumReleaseAge": "14 days"
+    }
```
nielspardon: This is one security measure on our side to delay package update PRs for new packages by default. This currently does not work for conda packages through Renovate. This seems to be more a Renovate vs a conda limitation. I'm looking into that in parallel.

nielspardon: Renovate follow-up (already merged): https://github.com/renovatebot/renovate/pull/45622

nielspardon: The follow-up is now also available on the Renovate Github app.

renovate.json
```diff
@@ -0,0 +1,27 @@
...
+  "$schema": "https://docs.renovatebot.com/renovate-schema.json",
+  "extends": [
+    "config:recommended",
+    ":semanticCommitTypeAll(build)",
+    ":maintainLockFilesWeekly"
+  ],
+  "semanticCommits": "enabled",
+  "semanticCommitType": "build",
+  "semanticCommitScope": "deps",
+  "minimumReleaseAge": "14 days",
```
andrew-coleman, line 14: `rangeStrategy` is unset, so it resolves to `replace`, which only rewrites a constraint when the new version falls outside it. Every dep here is a range, and dependabot is actively widening them (`16c4319` bumped `mkdocs-material >=9.7.6 → >=9.7.7`). Without this those lower bounds freeze until a major release — and since nothing then arrives as a manifest PR, `minimumReleaseAge` stops covering the path versions actually travel (`:maintainLockFilesWeekly`, which it doesn't apply to).

```suggestion
  "rangeStrategy": "widen",
  "minimumReleaseAge": "14 days",
```

renovate.json
```diff
@@ -0,0 +1,27 @@
...
+  "pre-commit": {
+    "enabled": true
+  },
```
andrew-coleman, line 17: `config:recommended` newly enables `gomod`; the deleted config declared only `github-actions` and `pip`. No workflow runs Go, and a bump to the `go 1.22.0` directive raises the minimum toolchain for every consumer of the published module.

```suggestion
  "pre-commit": {
    "enabled": true
  },
  "gomod": {
    "enabled": false
  },
```

renovate.json
```diff
@@ -0,0 +1,27 @@
+{
+  "$schema": "https://docs.renovatebot.com/renovate-schema.json",
+  "extends": [
+    "config:recommended",
+    ":semanticCommitTypeAll(build)",
+    ":maintainLockFilesWeekly"
+  ],
+  "semanticCommits": "enabled",
+  "semanticCommitType": "build",
+  "semanticCommitScope": "deps",
```
andrew-coleman, line 10: `fetchChangeLogs` defaults to `"pr"`, and per `pr_title.yml` the description becomes the merge commit message, which `.releaserc.mjs:89` reads with `releaseRules: [{ breaking: true, release: "minor" }]` — an upstream changelog quoting `BREAKING CHANGE:` would cut a spec release.

`labels` is for `stale.yml`: it closes PRs after 60+7 days with no `exempt-pr-labels`, and Renovate's default `recreateWhen: "auto"` suppresses an update whose branch+title match a closed PR — both stable for `all pixi deps`. Dependabot got `dependencies` automatically; Renovate sets none. (Needs `exempt-pr-labels: dependencies` on the action too.)

```suggestion
  "semanticCommitScope": "deps",
  "fetchChangeLogs": "off",
  "labels": ["dependencies"],
```

renovate.json
```diff
@@ -0,0 +1,27 @@
...
+  "minimumReleaseAge": "14 days",
+  "pre-commit": {
+    "enabled": true
+  },
+  "packageRules": [
+    {
+      "description": "Group pyproject.toml dependency bumps so pixi.lock is regenerated once",
+      "matchManagers": ["pep621", "pixi"],
+      "groupName": "all pixi deps"
```
andrew-coleman: `.pre-commit-config.yaml` pins `ruff v0.8.4` against pyproject's `ruff>=0.15.22,<1`; enabling the manager without grouping gives rev bumps their own PR, and no workflow runs pre-commit.

```suggestion
      "description": "Group dependency bumps so pixi.lock is regenerated once, and keep the pre-commit ruff hook in step with the pyproject range",
      "matchManagers": ["pep621", "pixi", "pre-commit"],
      "groupName": "all pixi deps"
```

renovate.json
nielspardon: Right on the diagnosis, but `widen` won't do it. Renovate's pep440 `handleWidenStrategy` opens with `if (satisfies(newVersion, currentValue)) return [currentValue]` — the same early return as `handleReplaceStrategy` — so a satisfied `>=9.7.6` stays put. `handleBumpStrategy` is the one that does `if (range.operator === '>=') return range.operator + newVersion`, which is exactly the rewrite in `16c4319`, and the docs say the same thing: "If instead you'd prefer to be updated to `^1.2.0` in cases like this, then configure `rangeStrategy` to `bump`."

Set `"rangeStrategy": "bump"` instead. Your `auto` → `replace` reading checks out: neither `pep621` nor `pixi` exports a `getRangeStrategy`, so `getRangeStrategy` in `lib/modules/manager/index.ts` falls through to the `default to 'replace' for auto` branch. The conda dependencies are covered too — the `conda` versioning module delegates `getNewValue` to pep440.

Two things keep `bump` from being noisy here: the `all pixi deps` group already funnels every manifest update into a single PR, and `pyproject.toml` is `Private :: Do Not Upload`, so a raised floor never reaches a consumer.

renovate.json
nielspardon: Added. Confirmed there is no Go step anywhere in `.github/workflows/`, so those PRs would arrive with no coverage at all — better off as a deliberate `"gomod": { "enabled": false }` than as a note in the description saying we could add one.

renovate.json
nielspardon: Took both. The `fetchChangeLogs` risk holds on the repository settings too: `squash_merge_commit_message` is `PR_BODY`, so fetched release notes land verbatim in the message `commit-analyzer` reads, and the commitlint `ignores` pattern this PR adds means nothing would flag it on the way in.

The labels rationale is inverted though. `recreateWhen: "auto"` recreates these rather than suppressing them: `lib/workers/repository/updates/generate.ts` sets `recreateClosed = recreateWhen !== 'never'` for lock file maintenance (line 246) and again for grouped upgrades (325, 342), and the docs list package groups and lock file maintenance as the cases where Renovate overrides the option precisely because branch names and PR titles have to be reused. `all pixi deps` is therefore an immortal PR — the failure mode is the opposite one, where the stale action closes it at 60+7 days and Renovate reopens it on the next run, forever. Still worth fixing, so the label is in along with `exempt-pr-labels: dependencies` on the action.

renovate.json
nielspardon: Done, with your wording. One thing to flag: `matchManagers` takes the whole file, so `buf-lint`, `yamllint` and `commitlint` rev bumps join `all pixi deps` as well. That looks like the right trade for a config no workflow runs, but say so if you would rather scope the grouping to the hooks that actually shadow `pyproject.toml`.

commitlint.config.js
```diff
@@ -1,9 +1,13 @@
...
   extends: ["@commitlint/config-conventional"],
-  // Workaround for https://github.com/dependabot/dependabot-core/issues/5923
+  // Bot PR descriptions are not conventional-commit bodies, so don't lint them.
   ignores: [
+    // Dependabot, see https://github.com/dependabot/dependabot-core/issues/5923.
+    // Kept while dependabot PRs opened before the switch to renovate are still open.
     (message) => /^Bumps \[.+]\(.+\) from .+ to .+\.$/m.test(message),
     (message) => /^Updates the requirements on .+ to permit the latest version\.$/m.test(message),
+    // Renovate
+    (message) => /^This PR contains the following updates:$/m.test(message),
```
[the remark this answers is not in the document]
nielspardon: We will fix that separately in a follow-up. We had a similar finding in https://github.com/substrait-io/substrait-validator/pull/589 and we should change this across the various repos independently of the Renovate switch.

andrew-coleman (approved):

nielspardon:
Merging with 1 committer approval as: CI Modifications (no format/specification changes)

PR #1078, opened 2026-05-11
ci: poll for tag visibility before notifying substrait-packaging

vbarua:
- Adds a polling loop in `notify.sh` that checks the GitHub API for the newly pushed tag before triggering the downstream `spec_released.yml` workflow in `substrait-packaging`
- Polls up to 4 times at 30s intervals (2 min max); breaks early once the tag is confirmed visible
- Proceeds with the dispatch even if the tag isn't visible after all retries, with a warning

- - -
This change is [Reviewable](https://reviewable.io/reviews/substrait-io/substrait/1078)

### DIFF
--- a/ci/release/notify.sh
+++ b/ci/release/notify.sh
@@ -7,6 +7,21 @@
 
 VERSION="${1}"
 
+# Wait for the tag to be visible via the GitHub API before notifying downstream.
+# The tag is pushed by semantic-release moments before this script runs, and
+# GitHub's ref replication can lag enough for the receiving workflow to miss it.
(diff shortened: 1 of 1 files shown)


vbarua:
This changes is intended to help with the error seen in https://github.com/substrait-io/substrait-packaging/actions/runs/25701463429/job/75462320467.

The release PR in substrait-packaging triggered before the tag was visible, causing it to fail even though the branch had been deleted. Adding a small arbitrary delay to work around this issue.

nielspardon (approved):

PR #1085, opened 2026-06-02
feat(protos): remove deprecated args fields

vbarua:
BREAKING CHANGE: removed args field from ScalarFunction
BREAKING CHANGE: removed args field from AggregateFunction
BREAKING CHANGE: removed args field from WindowFunction

- - -
This change is [Reviewable](https://reviewable.io/reviews/substrait-io/substrait/1085)

### DIFF
--- a/proto/substrait/algebra.proto
+++ b/proto/substrait/algebra.proto
@@ -1315,8 +1315,8 @@ message Expression {
     // using the declaration in the extension.
     Type output_type = 3;
 
-    // Deprecated; use arguments instead.
-    repeated Expression args = 2 [deprecated = true];
+    reserved 2;
+    reserved "args";
   }
 
   // A window function call.
@@ -1395,8 +1395,8 @@ message Expression {
     // Optional; defaults to the end of the partition.
     Bound upper_bound = 4;
 
-    // Deprecated; use arguments instead.
-    repeated Expression args = 8 [deprecated = true];
+    reserved 8;
+    reserved "args";
 
     enum BoundsType {
       BOUNDS_TYPE_UNSPECIFIED = 0;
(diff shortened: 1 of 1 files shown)


vbarua:
These fields have been deprecated for years. `arguments` is the field that should (and is) used today. `args` is not used in the active libraries. In those with language specific representations:
* substrait-go: only sets the [arguments](https://github.com/substrait-io/substrait-go/blob/16e339a2ed34b4aeaf6d4bcd03f72b501dfcbfc9/expr/functions.go#L395-L411) field when producing plans from POGOs.
* substrait-java: only sets the [arguments](https://github.com/substrait-io/substrait-java/blob/7b5722a08e6494d85d15bc28eb7d08e7563cfe39/core/src/main/java/io/substrait/expression/proto/ExpressionProtoConverter.java#L510-L513) when producing plans from POJOs.

nielspardon (approved):
LGTM

proto/substrait/algebra.proto
```diff
@@ -1315,8 +1315,8 @@ message Expression {
     // using the declaration in the extension.
     Type output_type = 3;
 
-    // Deprecated; use arguments instead.
-    repeated Expression args = 2 [deprecated = true];
+    reserved 2;
+    reserved "args";
```
benbellick: huh, didn't know we could reserve field names as well! Nice

benbellick: Minor nit, what do you think about moving the reserved block to the beginning of the message? That is where I usually see them and thus is where I would expect to find them. 

Same for the others.

benbellick (approved):
LGTM, just the minor comment.

proto/substrait/algebra.proto
vbarua: Good idea. Done.

yongchul (approved):

Issue #1057, opened 2026-04-23
feat: add user-defined type literal syntax to test case grammar


benbellick:
**Problem**

The test case grammar has no way to write a non-null literal of a user-defined type.

**Example**

`functions_geometry.yaml` declares `u!geometry` with functions like `point(fp64, fp64) -> u!geometry` and `make_line(u!geometry, u!geometry) -> u!geometry`. Neither the inputs nor the expected output for `make_line` can be written as a test case today.

**Background**

This was discussed on PR #953 and again at a [recent community sync](https://www.youtube.com/watch?v=bNzNYyiDbO4). We need to unify on how UDT literals should be represented and then make the grammar reflect it. The `structure` field wasn't designed with tests in mind, but maybe it should be stretched to cover them.

**Options discussed**

A few approaches came up, and we agreed it would help to prototype them before picking one:

1. **First-class UDT literal syntax.** Add a `userDefinedArg` grammar rule so a test can write something like `<literal>::u!foo` directly. The open question is what form `<literal>` takes:
   - **Recycle the `structure` field.** The literal form is derived from the type's declared `structure`, e.g. a point with `structure: {x: fp64, y: fp64}` gets written as a struct literal. Cheap to implement and reuses existing literal syntax, but stretches `structure` beyond its original purpose and doesn't help atomic UDTs whose structure is a handwave (e.g. unsigned ints encoded as signed ints).
   - **Per-type custom encoding.** Each UDT declares its own literal form as part of the type definition (custom string formats, decimal integers for `u8`, etc.). More expressive and lets each type pick something natural, but needs a new spec for how a UDT declares its literal grammar and semantics, which starts to look a lot like reinventing the function framework.
2. **No UDT literals at all.** Write test inputs as built-in literals (`string`, `binary`, `fp64`, etc.) and wrap them in extension functions that return the UDT, e.g. `make_line(point(1.0, 2.0), point(3.0, 4.0))`. Composes with machinery we already have and is close to zero work on the grammar side. The cost is giving up UDTs as a standalone literal form.

**Proposal**

Start with 1a. It's simple to specify since we already have struct literal syntax, and it unblocks the grammar gap for any UDT with a usable `structure`. Once that's in, we can open separate discussions about forms of syntactic sugar (e.g. more readable decimal literals, custom string formats for geometry) for cases where the raw structure form is painful to read.

---

[...]

PR #1088, opened 2026-06-02
feat: allow for function composition in test cases

vbarua:
Allows for functions to invoked as arguments in test. This enables tests like:

```
add(add(119::i8, 1::i8), 5::i8) = 125::i8
add(a, b) = add(b, a).
```

- - -
This change is [Reviewable](https://reviewable.io/reviews/substrait-io/substrait/1088)


vbarua:
See https://github.com/substrait-io/substrait/issues/1057#issuecomment-4482882960 for more context with this change.

Function compositions in tests cases allows us to utilize UDT constructor functions in tests cases, making them easier to write and read.

nielspardon:
I'm generally in favor of this change especially in the context of #1057. 

What I'm a bit worried about is that there is a risk that we create tight coupling between functions due to them being used together in test cases. This could lead to situations where hypothetically speaking we have an engine that only has support for the `add` function but does not have support for the `multiply` function and then the testcases that use a combination of `add` and `multiply` fail. This by itself does not have to be an issue since we may have the same situation when only a subset of function signatures are supported due to e.g. missing type support then we also may have a subset of the testcases in a testcase file fail.

I've been seeing the testcases in the spec repo as a foundation for a potential Substrait spec compliance validation and reporting in the future. When I think about this reporting then it would be great to not just have an overall percentage of this Substrait producer / consumer correctly supports 63% of the Substrait v0.90.0 testcases but to list distinct categories of which features are correctly implemented. Maybe this could be solved by requiring the definition of a Substrait dialect as the input for compliance validation which can be used to filter testcases so the compliance testing measures how much of the declared Substrait support in the dialect is correctly implemented.

Not something that needs to be solved in this PR.

yongchul (approved):

nielspardon (approved):
LGTM, as I wrote in my earlier comment my concerns can be addressed separately from this PR

PR #1096, opened 2026-06-04
ci: open release-tracking issues in downstream repos

nielspardon:
## What

After the release process triggers the `substrait-packaging` build, `ci/release/notify.sh` now also opens an **"Update to Substrait vX.Y.Z"** issue in each downstream implementation repo to notify them of the new spec release and track its adoption:

- `substrait-io/substrait-java`
- `substrait-io/substrait-python`
- `substrait-io/substrait-go`
- `substrait-io/substrait-rs`

This automates what was previously done by hand (e.g. substrait-io/substrait-java#809). The issue body reuses the release notes from the just-published GitHub release, matching the manual format.

## Details

- **Notes source:** read from the published GitHub release (created by `@semantic-release/github` during `publish`, before this `success` hook runs).
- **Retry:** the notes fetch is retried briefly (3×10s) to absorb read-replica lag, avoiding issues with a blank notes section.
- **Graceful degradation:** if notes still can't be fetched, the issue is opened with a minimal body linking to the release rather than skipping the notification entirely — a notifier should fail loudly, not silently.
- **Duplicate guard:** a title-based check avoids reopening issues if the release workflow is re-run for the same version.
- **Token scope:** the GitHub App token in `release.yml` is expanded to cover the four downstream repos so `gh` can create issues in them.

## Prerequisite

The GitHub App backing the release workflow (`APP_CLIENT_ID`/`APP_PRIVATE_KEY`) must be installed on the four downstream repos with **Issues: write** before this takes effect. If not, issue creation fails non-fatally (warning only) and the release itself still succeeds.

- - -
This change is [Reviewable](https://reviewable.io/reviews/substrait-io/substrait/1096)

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -24,11 +24,16 @@ jobs:
           client-id: ${{ secrets.APP_CLIENT_ID }}
           private-key: ${{ secrets.APP_PRIVATE_KEY }}
           owner: ${{ github.repository_owner }}
-          # request token for both substrait and substrait-packaging repos
-          # for ci/release/notify.sh
+          # request a token scoped to every repo ci/release/notify.sh touches:
+          # substrait (tag lookup), substrait-packaging (build trigger), and the
+          # downstream implementation repos (release notification issues)
           repositories: |
             substrait
             substrait-packaging
+            substrait-java
+            substrait-python
+            substrait-go
+            substrait-rs
 
       - uses: actions/checkout@v6
         with:
--- a/ci/release/notify.sh
+++ b/ci/release/notify.sh
@@ -26,3 +26,61 @@ gh workflow run spec_released.yml \
   --repo substrait-io/substrait-packaging \
   --field substrait_version="v${VERSION}" \
   || echo "Warning: failed to trigger release workflow in substrait-packaging"
+
+# Open an adoption-tracking issue in each downstream implementation repo so
+# they know a new spec version is available and can track upgrading to it.
+DOWNSTREAM_REPOS=(
+  "substrait-io/substrait-java"
+  "substrait-io/substrait-python"
+  "substrait-io/substrait-go"
+  "substrait-io/substrait-rs"
+)
+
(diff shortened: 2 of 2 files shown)


nielspardon:
@benbellick I think we discussed this a couple of times. I already installed the [Substrait Project Bot](https://github.com/organizations/substrait-io/settings/apps/substrait-project-bot) in the 4 downstreams repos and gave it issue write permissions.

yongchul (approved):
Is there any way to test this other than leap of faith?

benbellick:
This will be awesome! 🚀 

Same question as @yongchul. Is it possible to test or do we just have to go for it? And if it doesn't work, is there any risk?

nielspardon:
The question is probably how much of it we wanted to test.

- `notify.sh` already runs after everything else in the release process has succeeded and it shouldn't have an impact on the release process itself other than issues not being created or potentially duplicate issues being created in case someone somehow reruns the release workflow (which is very unlikely and the script should detect).
- We would need to test that the actual bot user has the right permissions in the target repos to create issues. That would require us to run the `notify.sh` script in a GH actions workflow other than the release workflow e.g. manually with a previous release. It might be useful to have a manual workflow trigger that allows us to backfill issue creation for older releases.
- A lot of the setup requires the actual repos and actual GH action execution environment. Main risk is that we end up with no issues or not properly formatted issues in 4 of our repositories where we can easily clean things up.

I would argue there's no siginificant risk and we could do the small leap of faith.

vbarua (approved):

vbarua:
Happy to test this live. As you point out Niels, the worst case is we have 4 badly formatted issues we need to delete.

nielspardon:
This has worked as expected for the latest release:
- https://github.com/substrait-io/substrait-go/issues/267
- https://github.com/substrait-io/substrait-java/issues/863
- https://github.com/substrait-io/substrait-python/issues/176
- https://github.com/substrait-io/substrait-rs/issues/501

PR #1098, opened 2026-06-05
feat(tests): support user-defined type literals in test cases

vbarua:
## Release Notes
Allow for user-defined type (UDT) literals to be used in tests cases.

Given user-defined types like
```yaml
types:
  - name: point
    structure:
      longitude: i32
      latitude: i32
  - name: line
    structure:
      start: u!point
      end: u!point
```
UDT test literals can be written as
```
(4, 2)::u!point 
((4, 2), (1, 1))::u!line
```

## Change Summary
- Adds `userDefinedArg` grammar rule so UDT literals can be written as `(v1, v2)::u!typename` in test cases, using the structure encoding defined in the UDT's extension YAML
- Extracts `userDefinedType` as a named grammar rule (previously inlined in `scalarType`)
- Updates the visitor to parse and return `CaseLiteral` for UDT args
- Documents the new syntax in `tests/README.md`

Part of the roadmap in #1057 (Step 2).

Note that this will require additional work in the test case handlers as well, because when parsing `(4, 2)::u!point ` it will need to resolved the struct encoding of `u!point` from the extension in order to be able to decode the `(4, 2)` component correctly.

🤖 Generated with [Claude Code](https://claude.com/claude-code) but reviewed by @vbarua 

- - -
This change is [Reviewable](https://reviewable.io/reviews/substrait-io/substrait/1098)


benbellick (approved):
Thinking out loud here, but I wonder if it would be valuable to somehow mark a struct-based UDT when it is being used in this context? 
- `u!(1,2)::u!line`, 
- `!(1,2)::u!line`, or
- `(1,2)::u!line` (the current proposal)

I _think_ it is fine to just do the last one, since it should be clear from context that structs used for UDTs mean the struct-based rep. Just sharing another option :)

yongchul (approved):

nielspardon (approved):

PR #1099, opened 2026-06-08
ci: strip git trailers from changelog and default to dry-run release

### LINKED ISSUE #980: drop support for deprecated time types

nielspardon:
## Problem

In [v0.88.0](https://github.com/substrait-io/substrait/releases/tag/v0.88.0), a `Signed-off-by` line leaked into the changelog and GitHub release notes under **BREAKING CHANGES**.

Root cause: the conventional-commits parser only ends a `BREAKING CHANGE` note at a recognized reference (`closes #…`) or another note keyword — **not** at a git trailer. So when a `Signed-off-by:` trailer immediately follows the breaking-change description (with no intervening `closes`/`fixes` line), it gets absorbed into the note text. That is exactly why #1019 leaked but #994 (which had a `closes #980` line before its sign-off) did not.

Because `@semantic-release/changelog` and `@semantic-release/github` both render `nextRelease.notes`, the single fix point is `@semantic-release/release-notes-generator`.

## Changes

- **Convert `.releaserc.json` → `.releaserc.mjs`.** JSON can't carry functions, and the fix needs a `writerOpts.transform`. The new ESM config is a faithful translation of the old one plus the transform.
- **Strip trailing git trailers from note text.** The transform wraps the preset's own transform and only rewrites `commit.notes[].text`, removing trailing trailer lines (`Signed-off-by`, `Co-authored-by`, etc.). Grouping, templates, and sorting from the preset are preserved (the generator merges `{ ...preset.writer, ...writerOpts }`).
- **Default to a dry run as a fail-safe.** The publishing plugins (`@semantic-release/github`, `@semantic-release/git`) and the publish/notify exec commands are only loaded when `RELEASE_DRY_RUN=false`. This matters because semantic-release still runs `prepare`/`publish`/`success` in dry-run mode (it only skips tag and push), so those side-effecting plugins must be omitted entirely rather than merely guarded by `--dry-run`.
  - `ci/release/run.sh` opts in to a real release via `RELEASE_DRY_RUN=false`.
  - `ci/release/dry_run.sh` now reads the config file (instead of overriding plugins on the CLI), so its preview reflects the real changelog. It inherits the safe default.

## Verification

Tested against the actual `conventional-changelog-conventionalcommits@8.0.0` + `@semantic-release/release-notes-generator@14` packages:

- Reproduced the v0.88.0 leak and confirmed the parser mechanism.
- Loaded the real `.releaserc.mjs`, pulled its `transform`, and re-rendered the actual `v0.87.0..v0.88.0` commits: the `Signed-off-by` is gone, while the breaking-change text, the `closes #980` reference, and all other entries are intact.
- Confirmed the config resolves correctly for `RELEASE_DRY_RUN` unset / `true` / `false` (safe reduced plugin set by default; full publishing set only on `false`).
- `bash -n` and `node --check` pass on all three files.

> [!NOTE]
> Tradeoff of the dry-run default: if CI ever fails to set `RELEASE_DRY_RUN=false`, a release silently becomes a no-op (green build, no new tag). That is visible and cheap to recover, versus an accidental release which is not.

- - -
This change is [Reviewable](https://reviewable.io/reviews/substrait-io/substrait/1099)


benbellick:
Hmm it looks like the "signed off by" issue is not present in other releases. What was actually different about that merge that caused this issue? Are we likely to see it again? 

EDIT: [Ah so this is the culprit commit](https://github.com/substrait-io/substrait/commit/fc6ab3cc3bb189f47e525db7f4b0c8a94b7b69f9)

nielspardon:
> EDIT: [Ah so this is the culprit commit](https://github.com/substrait-io/substrait/commit/fc6ab3cc3bb189f47e525db7f4b0c8a94b7b69f9)

Yes, when you have a `BREAKING CHANGE: ...` followed by anything not part of the breaking change message it gets included into the changelog since semantic release thinks it's part of the breaking change message.

.releaserc.mjs
```diff
@@ -0,0 +1,151 @@
...
+  "Helped-by",
+  "Cc",
+];
+const TRAILER = new RegExp(`^(?:${TRAILER_KEYS.join("|")}):\\s`, "i");
+
+// The conventional-commits parser ends a BREAKING CHANGE note only at a
+// recognized reference (closes #..., fixes #...) or another note keyword, not
+// at a git trailer -- so a trailing `Signed-off-by:` gets absorbed into the
+// note text. Strip such trailing trailer lines.
+const stripTrailers = (text) => {
```
yongchul, line 61: wow... wish the commit message is actually structured, rather than relying on adhoc parsing 😮‍💨

yongchul (approved):

PR #1100, opened 2026-06-08
ci: enforce conventional commits via committed commitlint config

nielspardon:
## Problem

The local `commitlint` pre-commit hook had no config to read — there was no `commitlint.config.js` / `.commitlintrc` anywhere in the repo. As a result the hook rejected **every** commit message with the `empty-rules` error and never applied the Conventional Commits rules at all (breaking-change notation like `feat!:` included). The rules only existed inside the PR-title CI workflow as an inline heredoc, so local and CI could drift.

## Changes

- **Add `commitlint.config.js`** that extends `@commitlint/config-conventional` and folds in the dependabot `ignores` and line-length overrides that previously lived only in CI. One shared source of truth.
- **Wire `@commitlint/config-conventional` into the pre-commit hook** via `additional_dependencies`, so the shareable config resolves inside the hook's isolated node environment.
- **Point the PR-title workflow at the committed config**: add a checkout of the (trusted) base ref and drop the inline `.commitlintrc.js` heredoc and `--extends` flag.

The PR title/body still flow into commitlint as data via `env`, so the `pull_request_target` security posture is unchanged — config comes from the trusted base branch, not the PR head.

## Verification

Ran commitlint locally against the new config:
- `feat!: ...` (+ body) → passes
- `feat(protos)!: ...` → passes
- dependabot `Bumps [...] from x to y.` → ignored (passes)
- malformed title (no type) → fails with `type-empty` / `subject-empty`

Also ran the `commitlint` pre-commit hook end-to-end: valid messages pass, malformed messages now fail with rule violations instead of `empty-rules`.

- - -
This change is [Reviewable](https://reviewable.io/reviews/substrait-io/substrait/1100)

### DIFF
--- a/.github/workflows/pr_title.yml
+++ b/.github/workflows/pr_title.yml
@@ -8,24 +8,14 @@ jobs:
     name: PR title / description conforms to semantic-release
     runs-on: ubuntu-latest
     steps:
+      # Check out the base ref (trusted, not the PR head) so the committed
+      # commitlint.config.js cannot be weakened by the PR being validated.
+      - uses: actions/checkout@v5
       - uses: actions/setup-node@v6
         with:
-          node-version: "20"
+          node-version: "24"
       - run: npm install @commitlint/config-conventional
-      - run: >
-          echo 'module.exports = {
-            // Workaround for https://github.com/dependabot/dependabot-core/issues/5923
-            "ignores": [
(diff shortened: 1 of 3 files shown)


vbarua (approved):
This should help developers catch this locally more easily 🙂

Issue #902, opened 2025-11-25
Add support for GROUPS window frame units/bounds in Substrait


kosiew:
# Add support for `GROUPS` window frame units/bounds in Substrait

## Summary

Some engines (e.g. Apache DataFusion) support window frames with `GROUPS` units (e.g. `ROWS`, `RANGE`, `GROUPS`). When trying to round-trip such plans through Substrait, there doesn’t appear to be a fully supported representation for `GROUPS` in the current Substrait proto, which makes it unclear how to encode these frames without producing invalid or non-portable Substrait plans.

This issue is to track adding first-class support for `GROUPS` window frame bounds/units in the Substrait spec and protobuf definitions.

## Background / Context

In DataFusion we’re adding support for Substrait window functions that map to internal `WindowFrameUnits`, including a `Groups` variant. On the consumer side we’ve worked around this by introducing an internal extension enum to handle the numeric value that the Substrait spec reserves for `GROUPS`.

However, when producing Substrait, if we simply emit the numeric value used for `GROUPS`, we risk generating plans that are not valid / not understood by tools that rely on the existing `BoundsType` protobuf enum, which currently only exposes:

* `ROWS`
* `RANGE`
* `UNSPECIFIED`

In other words, an engine that supports `GROUPS` has to choose between:

1. Emitting a `bounds_type` value that isn’t part of the protobuf enum (likely treated as invalid or “unrecognized” by consumers), or
2. Refusing to export plans that use `GROUPS` window frames at all.

Neither outcome is ideal if `GROUPS` is intended to be a first-class concept in Substrait.

## Current Behaviour

* The Substrait specification references `GROUPS` window frame units/bounds (and assigns it a numeric value).
* The generated protobuf enum for `BoundsType` (and associated tooling) does not yet expose a `GROUPS` variant.

From the engine side (example: DataFusion):

* We can parse/interpret `GROUPS` by mapping the raw integer to an internal extended enum.
* When trying to **produce** Substrait, we don’t have a safe, forward-compatible way to encode `GROUPS` without stepping outside the current proto definition.

## Desired Behaviour

It would be helpful to:

1. Add a `GROUPS` variant to the relevant Substrait protobuf enum(s) for window frame units/bounds (e.g. `BoundsType`), matching the numeric value reserved in the spec.
2. Clarify in the specification how `GROUPS` should behave, ideally with examples that cover:

   * `GROUPS BETWEEN UNBOUNDED PRECEDING AND CURRENT ROW`
[...]

PR #1105, opened 2026-06-15
feat(protos)!: support expression in window aggregate bounds

### LINKED ISSUE #617: Window bound offset field is too restrictive for RANGE mode
The current definitions of [Preceding and Following Window Bounds](https://github.com/substrait-io/substrait/blob/e564cf0b3b769491c19016d92d195b5f62361775/proto/substrait/algebra.proto#L1063-L1079) both declare the offset as
``` proto
int64 offset = 1;
```
with a note
> A strictly positive integer specifying the number of records that the window extends back from the current record. Required. 

This is true for ROWS and GROUPS mode (the latter of which is not yet supported in Substrait). However RANGE mode allows more than just integer offsets. As described in the Postgres docs:

> In RANGE mode, these options require that the ORDER BY clause specify exactly one column. The offset specifies the maximum difference between the value of that column in the current row and its value in preceding or following rows of the frame. The data type of the offset expression varies depending on the data type of the ordering column. For numeric ordering columns it is typically of the same type as the ordering column, but for datetime ordering columns it is an interval.

The offset field should be reworked to allow fully describing bounds using RANGE mode, which is effectively all numeric types + intervals

Sources:
* https://www.postgresql.org/docs/current/sql-expressions.html#SYNTAX-WINDOW-FUNCTIONS
* https://dev.mysql.com/doc/refman/8.3/en/window-functions-frames.html
* https://trino.io/blog/2021/03/10/introducing-new-window-features.html#-full-support-for-frame-type-range

### LINKED ISSUE #930: RANGE bounds spec is overly restrictive for UNBOUNDED/CURRENT ROW

yongchul:
# Motivation

Like #748 in FetchRel, `offset` can be expressions in window spec.

# Proposal

* Extend the offset to expression, the same way we did for FetchRel.
* deprecate offset from preceding and following message.
* Clarify window frame bound documentation especially around window frame, ordering, and current rows.

The proto and documentation changes closes #617 and #930 

BREAKING CHANGE: deprecates the `offset` field of the `Preceding` and `Following` window bounds in favor of the new `offset_expr` field, which accepts an arbitrary expression instead of an integer literal. Consumers should add `offset_expr` support before producers switch to it. The PR also requires explicit window frame bounds type. Producers should set the bounds type and Consumers should reject `BOUNDS_TYPE_UNSPECIFIED` in such cases.
 

- - -
This change is [Reviewable](https://reviewable.io/reviews/substrait-io/substrait/1105)


nielspardon:
It looks like for `FetchRel` the decision was to deprecate the old int64 offset field in favor of the expression field. In this PR you seem to propose we continue to support both?

yongchul:
> It looks like for FetchRel the decision was to deprecate the old int64 offset field in favor of the expression field. In this PR you seem to propose we continue to support both?

I'm trying to separate the `deprecation` as a separate PR as discussed sometime ago. But for this one, maybe we can do that in one PR... let's see how others think.

proto/substrait/algebra.proto
```diff
@@ -1402,17 +1402,33 @@ message Expression {
       // Defines that the bound extends this far back from the current record.
       message Preceding {
         // A strictly positive integer specifying the number of records that
-        // the window extends back from the current record. Required. Use
-        // CurrentRow for offset zero and Following for negative offsets.
-        int64 offset = 1;
+        // the window extends back from the current record. Required.
+        // Use CurrentRow for offset zero and Following for negative offsets.
+        oneof offset_mode {
```
yongchul: @vbarua here is a requested SQL example for the scenario. adapted from SQL server documentation.
https://learn.microsoft.com/en-us/sql/t-sql/functions/lag-transact-sql?view=sql-server-ver17#a-compare-values-between-years

```sql
USE AdventureWorks2022;  
GO  
SELECT BusinessEntityID, YEAR(QuotaDate) AS SalesYear, SalesQuota AS CurrentQuota,   
       LAG(SalesQuota, LookBack,0) OVER (ORDER BY YEAR(QuotaDate)) AS PreviousQuota  
FROM Sales.SalesPersonQuotaHistory  
WHERE BusinessEntityID = 275 AND YEAR(QuotaDate) IN ('2005','2006');
```

The modification I made is `LookBack` in the `LAG()`, means you differentiate the offset per business entity id (let's say per rank of business entity id, for the senior rank look into further history). You can change `LookBack` to arbitrary scalar expression except another analytical function.

`LookBack` creates a new windowing spec including expression.

proto/substrait/algebra.proto
```diff
@@ -1402,17 +1402,33 @@ message Expression {
...
+        // Use CurrentRow for offset zero and Preceding for negative offsets.
+        oneof offset_mode {
+          int64 offset = 1;
+
+          // Expression evaluated into a strictly positive integer specifying
+          // the offset from the current record. Evaluating to null, zero, or
+          // a negative integer should result in an error.
+          // Recommended type for offset is int64.
+          Expression offset_expr = 2;
+        }
```
benbellick: I wonder if we should instead deprecate `offset` and only keep `offset_expr`? Is there value in keeping both?

yongchul: This is a recurring scenario and I keep forgetting what we do. We can discuss this probably again (and this time, I will follow up with a practice document) on Wednesday.

Options
1. introduce a new field and deprecate old in one release 
2. introduce transient `oneof` then deprecate in another release

nielspardon: I would do 1. introduce and deprecate then in a later follow-up remove old. Especially since the old field semantics can be expressed with the new field. Otherwise the spec supports both equally for a while and it is unclear what the direction is.

yongchul: deprecated the old field and updated the PR. PTAL

nielspardon:
Thanks for this — it's a clean application of the `FetchRel` `offset_expr` pattern. I've tagged it `apache-gluten` because it covers a modification Apache Gluten currently carries in its Substrait fork (an `Expression`-valued window frame bound), so landing it upstream helps that fork converge on the spec.

One semantics question for consideration: the `offset_expr` docs say it "should evaluate to a strictly positive integer and should use type `int64`." That's exactly right for `BOUNDS_TYPE_ROWS` (a row count). But for `BOUNDS_TYPE_RANGE`, the bound is a value delta on the ordering column — per the `BoundsType` comment, rows where the ordering value is within `current_value ± bound` — which is frequently **not** an integer. For example `RANGE BETWEEN INTERVAL '7' DAY PRECEDING` over a temporal ordering column, or a decimal offset over a decimal column.

Would it make sense to broaden the RANGE case so that, for `BOUNDS_TYPE_RANGE`, `offset_expr` may evaluate to a value **type-compatible with the ordering column** (one that can be added to / subtracted from it), while keeping the strictly-positive-integer requirement for `BOUNDS_TYPE_ROWS`? That fits into `offset_expr` without any shape change — just a doc/semantics clarification — and it happens to be the main case Gluten's fork needs (its bound `ref` is used for exactly these interval/typed RANGE boundaries).

Concretely, the `offset_expr` doc could read something like:

> Expression specifying this bound's offset from the current record.
>
> - For `BOUNDS_TYPE_ROWS`: evaluates to a strictly positive integer number of records (recommended type `int64`).
> - For `BOUNDS_TYPE_RANGE`: evaluates to a strictly positive offset that is subtracted from (`Preceding`) or added to (`Following`) the ordering column's value to form the boundary. Its type must support that arithmetic with the ordering column — e.g. an interval for a temporal ordering column, or a numeric value for a numeric one.
>
> In both cases, use `CurrentRow` for a zero offset and the opposite bound direction for negative offsets. Evaluating to null, or to a negative value, should result in an error.

proto/substrait/algebra.proto
```diff
@@ -1402,17 +1402,31 @@ message Expression {
...
-        int64 offset = 1;
+        // the window extends ahead of the current record. Required.
+        // Use CurrentRow for offset zero and Preceding for negative offsets.
+        // Deprecated: use `offset_expr` instead.
+        int64 offset = 1 [deprecated = true];
+
+        // Expression evaluated into a strictly positive integer specifying
+        // the offset from the current record. Evaluating to null, zero, or
+        // a negative integer should result in an error.
+        // Recommended type for offset is int64.
```
```suggestion
        // Expression must have type int64.
```
Are there cases where we would ever want something other than i64?

text/dialect_schema.yaml
```diff
@@ -1100,6 +1100,13 @@ definitions:
       metadata:
         type: object
         description: Arbitrary data created by the dialect author.
+      supports_offset_expr:
+        type: boolean
+        default: false
+        description: >
+          Whether this dialect supports expression-based offsets for window
+          function bounds. If false, only literal offsets are supported for
+          Preceding and Following bounds. Defaults to false.
```
benbellick:
```suggestion
```
I'm not sure it makes sense to include this as part of a dialect. If the plan is to deprecate the old field, then accepting the new field is not optional, but is instead a matter of what version of substrait you are using.

dialects/tests/expressions_test.yaml
```diff
@@ -21,3 +21,5 @@ supported_expressions:
       - CURRENT_DATE
       - CURRENT_TIMESTAMP
       - CURRENT_TIMEZONE
+  - expression: WINDOW_FUNCTION
+    supports_offset_expr: false
```
benbellick:
```suggestion
```
Related to the below comment about removing it from the dialect.

proto/substrait/algebra.proto
```diff
@@ -1402,17 +1402,31 @@ message Expression {
...
-        int64 offset = 1;
+        // the window extends back from the current record. Required.
+        // Use CurrentRow for offset zero and Following for negative offsets.
+        // Deprecated: use `offset_expr` instead.
+        int64 offset = 1 [deprecated = true];
+
+        // Expression evaluated into a strictly positive integer specifying
+        // the offset from the current record. Evaluating to null, zero, or
+        // a negative integer should result in an error.
+        // Recommended type for offset is int64.
```
```suggestion
        // Expression must have type int64.
```
Same here as below: are there cases where we would ever want something other than i64?

benbellick (approved):
A few small comments but otherwise looks good to me!

proto/substrait/algebra.proto
nielspardon: Yes — for `BOUNDS_TYPE_ROWS` the offset is a row count and `int64` is correct (that's the case @yongchul's `LAG(…, LookBack, …)` example exercises). But `BOUNDS_TYPE_RANGE` bounds are a value delta on the ordering column, so the offset type follows that column. The common non-`int64` case is a time-based range:

```sql
-- 7-day trailing window over a DATE/TIMESTAMP ordering column
SELECT order_date,
       SUM(amount) OVER (
         ORDER BY order_date
         RANGE BETWEEN INTERVAL '7' DAY PRECEDING AND CURRENT ROW
       ) AS rolling_7d
FROM orders;
```

Here the `PRECEDING` offset is `INTERVAL '7' DAY` (frame = `order_date - 7 days … order_date`), not an integer; a decimal ordering column likewise takes a decimal offset. So `int64` fits ROWS, but RANGE wants a type compatible with the ordering column — I sketched proposed wording covering both in the top-level comment.

text/dialect_schema.yaml
nielspardon: I had the the same thought. The whole topic of which type of expressions are supported in which place feels like we either want to solve holistically across relations or keep open.

proto/substrait/algebra.proto
yongchul: Ah, thanks! Yes, indeed the range should be delta -- or something that can be subtracted from the value. I'll update the comments.

yongchul: Revised the comments and documentation around window frame. PTAL.

proto/substrait/algebra.proto
```diff
@@ -1387,32 +1387,57 @@ message Expression {
...
+        // A strictly positive integer specifying the offset of this bound.
+        // Use CurrentRow for offset zero and Preceding for negative offsets.
+        // Deprecated: use `offset_expr` instead.
+        int64 offset = 1 [deprecated = true];
+
+        // Expression specifying the distance of this bound from the current
+        // row. Its interpretation depends on `WindowFunction.bounds_type`:
+        //
+        // * BOUNDS_TYPE_ROWS: evaluates to a strictly positive integer number
+        //   of rows. The recommended type is int64.
```
```suggestion
        //   of rows. The type must be int64.
```
I don't think there is any benefit to leaving the door open there 🤷

proto/substrait/algebra.proto
```diff
@@ -1387,32 +1387,57 @@ message Expression {
...
+        // * BOUNDS_TYPE_RANGE: evaluates to a strictly positive distance
+        //   compatible with the single ordering expression and applies that
+        //   distance toward values later in the declared ordering.
```
benbellick: I think we should try an explicitly say that the type _must_ match the ordering column.

benbellick: Ah I realize there may be something more profound to figure out here, which is that the types don't need to match. 
e.g.

```sql
  SELECT
    event_time,
    COUNT(*) OVER (
      ORDER BY event_time
      RANGE BETWEEN INTERVAL '1' HOUR PRECEDING AND CURRENT ROW
    )
  FROM events;
```

This is doing arithmetic between a timestamp and an interval, like the example @nielspardon provided. 

I think we need to figure out a way to clarify what is meant by: 
> a strictly positive distance compatible with the single ordering expression

What does it mean to be compatible?

yongchul: Compatible type D for offset_expr with respect to ordering `O` means that there are well-defined functions `SUCCESSOR<O>(T, D) -> T` for given order `O`, value type `T`, and offset_expr type `D`. Oh, well, this is for single ordering column...

proto/substrait/algebra.proto
```diff
@@ -1387,32 +1387,57 @@ message Expression {
 
     enum BoundsType {
       BOUNDS_TYPE_UNSPECIFIED = 0;
```
nielspardon: The new text defines `offset_expr` entirely by switching on `bounds_type`, which leaves `offset_expr` with no defined interpretation when `bounds_type` is unset. Not asking to give the zero value semantics — it should stay a pure sentinel per [protobuf's enum default guidance](https://protobuf.dev/programming-guides/proto3/#enum-default). A producer-obligation line in the style of `FetchMode` covers it:

```suggestion
      // Unspecified. Producers must set one of the defined bounds types when
      // either bound is Preceding or Following.
      BOUNDS_TYPE_UNSPECIFIED = 0;
```

Note `Bound` predates `bounds_type` (#136 vs #399), so this is forward-looking guidance only — existing plans carrying `Preceding`/`Following` with no `bounds_type` shouldn't become invalid retroactively. Happy to drop this one if we'd rather not touch pre-existing wording here.

proto/substrait/algebra.proto
```diff
@@ -1387,32 +1387,57 @@ message Expression {
...
-      // The lower and upper bound specify how many rows before and after the current row
-      // the window should extend.
+      // Bounds are physical row offsets in the declared ordering.
       BOUNDS_TYPE_ROWS = 1;
```
nielspardon: `sorts` is optional and SQL permits `ROWS BETWEEN 1 PRECEDING AND CURRENT ROW` with no `ORDER BY`. The old wording ("rows before and after the current row") didn't presume an ordering; this one does.

```suggestion
      // Bounds are physical row offsets in the declared ordering. If no
      // ordering is declared, the row order is implementation-defined.
      BOUNDS_TYPE_ROWS = 1;
```

proto/substrait/algebra.proto
```diff
@@ -1387,32 +1387,57 @@ message Expression {
...
+      // Defines a bound earlier than the current record in the declared ordering.
       message Preceding {
-        // A strictly positive integer specifying the number of records that
-        // the window extends back from the current record. Required. Use
-        // CurrentRow for offset zero and Following for negative offsets.
-        int64 offset = 1;
+        // A strictly positive integer specifying the offset of this bound.
+        // Use CurrentRow for offset zero and Following for negative offsets.
+        // Deprecated: use `offset_expr` instead.
+        int64 offset = 1 [deprecated = true];
```
nielspardon: "Required." disappeared in the rewrite and nothing replaced it, so three cases are currently undefined: both fields set, neither set, and `offset = 0` with no `offset_expr` (proto3 gives the scalar no presence, so it's indistinguishable from unset).

My first instinct was "setting both is invalid", but that would forbid exactly the dual-write step the [breaking change policy](https://substrait.io/spec/breaking_change_policy/) relies on — it requires changes that let users "update systems independently and asynchronously", as in the URI→URN migration. So both-set should be legal with a defined winner:

```suggestion
        // A strictly positive integer specifying the offset of this bound.
        // Use CurrentRow for offset zero and Following for negative offsets.
        //
        // At least one of `offset` and `offset_expr` must be set; setting
        // neither is invalid. Consumers must use `offset_expr` when it is set
        // and ignore `offset`. Producers may set both during migration, in
        // which case `offset` must be the int64-literal equivalent of
        // `offset_expr` -- possible only for BOUNDS_TYPE_ROWS with a literal
        // offset -- so that consumers which do not yet read `offset_expr` can
        // still execute the plan.
        //
        // Deprecated: use `offset_expr` instead.
        int64 offset = 1 [deprecated = true];
```

Dual-write is only *possible* where the old field can express the same thing (ROWS with a literal offset). A RANGE interval bound or a per-row offset has no `offset` equivalent — acceptable, since an old consumer couldn't have executed those anyway.

Either way this is purely a docs/validation matter: field 2 as an `Expression` encodes identically with or without a `oneof`, so nothing about the wire format changes.

proto/substrait/algebra.proto
```diff
@@ -1387,32 +1387,57 @@ message Expression {
...
+
+        // Expression specifying the distance of this bound from the current
+        // row. Its interpretation depends on `WindowFunction.bounds_type`:
+        //
+        // * BOUNDS_TYPE_ROWS: evaluates to a strictly positive integer number
+        //   of rows. The recommended type is int64.
```
nielspardon: Mirror of @benbellick's suggestion on the `Following` copy at L1433, same reasoning — a row count has no legitimate alternative width, and making a producer `CAST` an `i32` is cheaper than every consumer having to accept any integer type.

```suggestion
        //   of rows. The type must be int64.
```

proto/substrait/algebra.proto
```diff
@@ -1387,32 +1387,57 @@ message Expression {
...
+        // * BOUNDS_TYPE_RANGE: evaluates to a strictly positive distance
+        //   compatible with the single ordering expression and applies that
+        //   distance toward values earlier in the declared ordering.
```
nielspardon: Grounding "compatible" in machinery the spec already has, so a validator can check it and the timestamp/interval case @benbellick raised is covered — the types don't have to match, they just have to be addable:

```suggestion
        // * BOUNDS_TYPE_RANGE: evaluates to a strictly positive distance
        //   applied toward values earlier in the declared ordering. Its type D
        //   must be compatible with the type T of the single ordering
        //   expression, i.e. `add(T, D) -> T` and `subtract(T, D) -> T` must
        //   be defined (e.g. timestamp/interval_day, decimal/decimal,
        //   i64/i64).
```

The per-direction formula (`subtract` toward earlier values, `add` toward later) is in the docs suggestion so it lives in one place rather than four.

proto/substrait/algebra.proto
```diff
@@ -1387,32 +1387,57 @@ message Expression {
...
+        //
+        // A null, zero, or negative result is invalid. Use CurrentRow for zero
+        // and Following for a negative distance.
```
nielspardon: With a literal `offset`, "strictly positive" was statically checkable; with an arbitrary expression it isn't, so this converts previously-expressible plans into runtime errors. Zero is also legal SQL — `<window frame preceding>` takes an *unsigned* value specification, so `ROWS BETWEEN 0 PRECEDING` and `RANGE BETWEEN INTERVAL '0' DAY PRECEDING` are both valid, and both equal `CurrentRow` in their respective modes. A producer translating `ROWS BETWEEN LookBack PRECEDING` can't rewrite to `CurrentRow` when `LookBack` may be 0 at runtime; it has to error or emit a `switch`.

```suggestion
        // A null or negative result is invalid; a zero result is equivalent
        // to CurrentRow. Use Following for a negative distance.
```

Separately: `FetchRel.offset_expr` — the precedent this PR follows — treats null as 0 and errors only on negatives. Making null invalid here is defensible (a frame offset has no natural "no bound" reading, unlike `count`), just worth being a deliberate divergence.

nielspardon (commented):
Clean application of the `FetchRel` `offset_expr` pattern, and the field lands at position 2 as an `Expression` — byte-identical to what Apache Gluten's fork already carries as `ref`, so this lets that fork converge with a rename. The descending-sort rewrite is a real fix too: the old `current_value ± bound` formula was simply wrong for `DESC`, and the docs table listed a nonexistent "Bound Trailing(int64)". My comments are all about semantics a consumer can't currently derive; the wire shape looks right to me.

One item that can't be an inline suggestion because it sits outside the diff — `ConsistentPartitionWindowRel` (`algebra.proto` L415-419): `sorts` there is rel-level and shared across all `window_functions`, so a single RANGE-bounded function constrains the entire relation to one sort field and can't coexist with a function needing a two-key ordering. Worth a sentence under "Consistent Partition Window Operation" in `physical_relations.md` so consumers don't have to deduce it.

### Announcing the deprecation

Using the breaking-change mechanism for the deprecation makes sense, and it's cheap: `.releaserc.mjs` maps `breaking: true` to a *minor* release, so on 0.x a `!` bumps the same as a plain `feat:` (cf. 0.98.0, which carried three `feat(protos)!:` removals), and `pr_breaking.yml` only demands the footer when `buf breaking` actually fails — so a buf-clean deprecation can carry one voluntarily with CI green. The payoff is the `### ⚠ BREAKING CHANGES` block at the top of the release notes, and it lines up with governance already giving "Format/Specification Deprecations" its own 2-PMC row.

Since the footer *is* the migration notice, it should be addressed at consumers — they have to move first so producers can follow:

```
feat(protos)!: support expression in window aggregate bounds

BREAKING CHANGE: `Bound.Preceding.offset` and `Bound.Following.offset` are
deprecated in favor of `offset_expr`. Consumers should accept `offset_expr`
and prefer it over `offset` when both are set, so that producers can migrate
to `offset_expr` before the deprecated fields are removed in a future
release.
```

That gives the three-step order the policy asks for: consumers read `offset_expr` (preferring it over `offset`) → producers switch to writing it, dual-writing `offset` where it is expressible → the deprecated fields are removed, with its own "have been removed" footer at that point (cf. #1130). Concretely, step 1 means `offset_expr` window-bound support in substrait-go, substrait-java, substrait-python and substrait-rs before this counts as migrated, per the breaking-change policy's "implemented in all active libraries" clause — probably worth tracking issues against those four rather than leaving it implicit here.

Two smaller notes: the branch is 26 commits behind `main`, and one of those (#1130) removed `FetchRel`'s deprecated `offset`/`count` — the tail end of exactly the lifecycle this PR starts. And there's no CI surface for any of this: `tests/` covers function cases, types and the text grammar, with no plan-level window fixtures, so whatever we settle on here is enforced by prose alone.

proto/substrait/algebra.proto
```diff
@@ -1387,32 +1387,57 @@ message Expression {
...
+        Expression offset_expr = 2;
```
nielspardon: The biggest remaining gap for me: the comment never says *when* this is evaluated or what it may reference. The `LAG(SalesQuota, LookBack, 0)` example implies per-row evaluation with field references into the input, and Gluten's fork describes its equivalent field as "the reference to pre-project range frame boundary" — but engines genuinely disagree here. PostgreSQL forbids variables, aggregates and window functions in a frame offset; Spark requires a foldable boundary; DuckDB permits per-row column references. Without a normative statement, a producer emitting a per-row offset and a consumer assuming constant-foldable will silently disagree instead of erroring.

```suggestion
        //
        // Field references resolve against the input schema of the relation
        // containing this window function; the expression is evaluated once
        // per input record. It must not contain window or aggregate functions.
        Expression offset_expr = 2;
```

If per-row is *not* the intent, "must be constant-foldable" works instead — but that would rule out the LAG example that motivated the PR.

proto/substrait/algebra.proto
```diff
@@ -1387,32 +1387,57 @@ message Expression {
...
+      // Defines a bound later than the current record in the declared ordering.
       message Following {
-        // A strictly positive integer specifying the number of records that
-        // the window extends ahead of the current record. Required. Use
-        // CurrentRow for offset zero and Preceding for negative offsets.
-        int64 offset = 1;
+        // A strictly positive integer specifying the offset of this bound.
+        // Use CurrentRow for offset zero and Preceding for negative offsets.
+        // Deprecated: use `offset_expr` instead.
+        int64 offset = 1 [deprecated = true];
```
nielspardon: Mirror of the `Preceding` change above.

```suggestion
        // A strictly positive integer specifying the offset of this bound.
        // Use CurrentRow for offset zero and Preceding for negative offsets.
        //
        // At least one of `offset` and `offset_expr` must be set; setting
        // neither is invalid. Consumers must use `offset_expr` when it is set
        // and ignore `offset`. Producers may set both during migration, in
        // which case `offset` must be the int64-literal equivalent of
        // `offset_expr` -- possible only for BOUNDS_TYPE_ROWS with a literal
        // offset -- so that consumers which do not yet read `offset_expr` can
        // still execute the plan.
        //
        // Deprecated: use `offset_expr` instead.
        int64 offset = 1 [deprecated = true];
```

proto/substrait/algebra.proto
nielspardon: Mirror for `Following`, direction reversed — rationale in the thread with @benbellick above.

```suggestion
        // * BOUNDS_TYPE_RANGE: evaluates to a strictly positive distance
        //   applied toward values later in the declared ordering. Its type D
        //   must be compatible with the type T of the single ordering
        //   expression, i.e. `add(T, D) -> T` and `subtract(T, D) -> T` must
        //   be defined (e.g. timestamp/interval_day, decimal/decimal,
        //   i64/i64).
```

proto/substrait/algebra.proto
```diff
@@ -1387,32 +1387,57 @@ message Expression {
...
+        // row. Its interpretation depends on `WindowFunction.bounds_type`:
+        //
+        // * BOUNDS_TYPE_ROWS: evaluates to a strictly positive integer number
+        //   of rows. The recommended type is int64.
+        // * BOUNDS_TYPE_RANGE: evaluates to a strictly positive distance
+        //   compatible with the single ordering expression and applies that
+        //   distance toward values later in the declared ordering.
+        //
+        // A null, zero, or negative result is invalid. Use CurrentRow for zero
+        // and Preceding for a negative distance.
```
nielspardon: Mirror of the `Preceding` change above.

```suggestion
        // A null or negative result is invalid; a zero result is equivalent
        // to CurrentRow. Use Preceding for a negative distance.
```

proto/substrait/algebra.proto
```diff
@@ -1387,32 +1387,57 @@ message Expression {
...
+        Expression offset_expr = 2;
```
nielspardon: Mirror of the `Preceding` change above.

```suggestion
        //
        // Field references resolve against the input schema of the relation
        // containing this window function; the expression is evaluated once
        // per input record. It must not contain window or aggregate functions.
        Expression offset_expr = 2;
```

site/docs/expressions/window_functions.md
```diff
@@ -18,8 +18,22 @@ When binding a window function, the binding must include the following additiona
...
+`Preceding` and `Following` define offsets relative to the current record and with respect to the declared ordering. `Preceding` moves toward records or values earlier in that ordering, while `Following` moves toward records or values later in that ordering. They therefore apply in opposite value directions for ascending and descending sorts:
+
+| Sort direction | `Preceding` boundary              | `Following` boundary            |
+| -------------- | --------------------------------- | ------------------------------- |
+| Ascending      | Toward lower ordering values      | Toward higher ordering values   |
+| Descending     | Toward higher ordering values     | Toward lower ordering values    |
+
+The interpretation of an offset depends on `BoundsType`:
+
+* `BOUNDS_TYPE_ROWS` defines physical row offsets in the declared ordering. `offset_expr` must evaluate to a strictly positive integer number of rows; `int64` is the recommended type.
```
nielspardon: Matching the `int64` mandate and the no-declared-ordering case:

```suggestion
* `BOUNDS_TYPE_ROWS` defines physical row offsets in the declared ordering. `offset_expr` must evaluate to a strictly positive number of rows and must have type `int64`. If no ordering is declared, the row order is implementation-defined.
```

site/docs/expressions/window_functions.md
```diff
@@ -18,8 +18,22 @@ When binding a window function, the binding must include the following additiona
...
+* `BOUNDS_TYPE_RANGE` defines value offsets from the current row's ordering value and requires exactly one ordering expression. `offset_expr` must evaluate to a strictly positive distance compatible with that ordering expression. A range boundary is inclusive.
```
nielspardon: Making "compatible" concrete, and carrying the per-direction formula that the old `current_value ± bound` wording got wrong for descending sorts:

```suggestion
* `BOUNDS_TYPE_RANGE` defines value offsets from the current row's ordering value and requires exactly one ordering expression. `offset_expr` must evaluate to a strictly positive distance whose type `D` is compatible with the ordering expression's type `T` — that is, `add(T, D) -> T` and `subtract(T, D) -> T` must be defined (for example `timestamp`/`interval_day`, `decimal`/`decimal`, `i64`/`i64`). For a `Preceding` bound the boundary value is `subtract(current, offset)` under an ascending ordering and `add(current, offset)` under a descending ordering; `Following` is the mirror. A range boundary is inclusive.
```

site/docs/expressions/window_functions.md
```diff
@@ -18,8 +18,22 @@ When binding a window function, the binding must include the following additiona
...
+
+`offset_expr` is the recommended way to specify an offset. A literal integer `offset` is still accepted for compatibility but is deprecated. Evaluating `offset_expr` to null, zero, or a negative value should result in an error. Use `CurrentRow` for zero and the opposite bound direction for a negative distance.
```
nielspardon: Folding in the removal timeline, the at-least-one rule with the dual-write allowance, the evaluation context, and zero-as-`CurrentRow`:

```suggestion
`offset_expr` is the recommended way to specify an offset. A literal integer `offset` is still accepted for compatibility but is deprecated and will be removed in a future release. At least one of the two must be set; consumers must use `offset_expr` when it is set and ignore `offset`. During migration a producer may set both, with `offset` carrying the int64-literal equivalent — possible only for `BOUNDS_TYPE_ROWS` with a literal offset — so that consumers which do not yet read `offset_expr` can still execute the plan.

`offset_expr` is evaluated once per input record and may reference the input's fields, but must not contain window or aggregate functions. Evaluating it to null or a negative value should result in an error; evaluating it to zero is equivalent to `CurrentRow`.
```

proto/substrait/algebra.proto
nielspardon: Agree, and I'd take this as written — a row count has no legitimate alternative width, and making a producer `CAST` an `i32` is cheaper than every consumer having to accept any integer type. The `Preceding` copy needs the same, so I've left a matching suggestion on L1412. I'd keep "recommended" only in the RANGE bullets, where the type legitimately follows the ordering column.

proto/substrait/algebra.proto
nielspardon: `SUCCESSOR<O>(T, D) -> T` captures it, but it introduces a concept the spec doesn't have elsewhere. We can say the same thing with machinery we already have, which makes it validator-checkable and answers the timestamp/interval case directly — the types don't have to match, they just have to be addable:

> Let `T` be the type of the single ordering expression and `D` the type of `offset_expr`. `D` is compatible with `T` if `add(T, D) -> T` and `subtract(T, D) -> T` are defined. The boundary is `subtract(current, offset)` toward values earlier in the declared ordering and `add(current, offset)` toward values later.

I've left suggestions on both blocks (L1413-1415 and L1434-1436) plus the docs bullet, which carries the per-direction formula so it isn't repeated four times. One adjacent gap while we're in here: nullable ordering values under a RANGE-with-offset frame are undefined — Postgres treats nulls as their own peer group — so a sentence on how that interacts with `NULLS FIRST/LAST` would help.

proto/substrait/algebra.proto
yongchul: I don't like implementation-defined because there is really no order 😆  If you squint, no-order is actually declared order. Let me think if there is a better way to phrase that.

proto/substrait/algebra.proto
yongchul: That's the main reason behind the dialect change in the beginning -- support only `constant-foldable` expressions as engine capabilities are varying.

proto/substrait/algebra.proto
```diff
@@ -1413,33 +1413,93 @@ message Expression {
...
+        // A strictly positive integer specifying the offset of this bound.
+        // Use CurrentRow for offset zero and Following for negative offsets.
+        //
+        // At least one of `offset` and `offset_expr` must be set; setting
+        // neither is invalid. Consumers must use `offset_expr` when it is set
+        // and ignore `offset`. Producers may set both during migration, in
+        // which case `offset` must be the int64-literal equivalent of
+        // `offset_expr` -- possible only for BOUNDS_TYPE_ROWS with a literal
+        // offset -- so that consumers which do not yet read `offset_expr` can
+        // still execute the plan.
```
nielspardon: My earlier suggestion was wrong on the dual-write condition, sorry -- pre-PR `BOUNDS_TYPE_RANGE` was defined against this same `int64 offset` ("the value of the ordering column is greater than or equal to (current_value - lower bound)"), so a RANGE bound with an integer distance was expressible via `offset` too. The restriction belongs on the distance, not the bounds type.

Also folding in a presence clarifier: proto3 gives `offset` no explicit presence, so "must be set" isn't checkable as written -- a consumer can't tell `offset = 0` from unset. Since zero was never a legal offset (the first paragraph already says to use `CurrentRow` instead), defining set-ness as non-zero closes the hole for free, and it also settles the case where `offset_expr` evaluates to zero: that bound simply can't be dual-written and should be emitted as `CurrentRow`.

```suggestion
        // At least one of `offset` and `offset_expr` must be set; setting
        // neither is invalid. Because `offset` has no explicit presence, it is
        // considered set when it is non-zero; zero was never a valid offset.
        // Consumers must use `offset_expr` when it is set and ignore `offset`.
        // Producers may set both during migration, in which case `offset` must
        // be the int64-literal equivalent of `offset_expr` -- possible only
        // when the distance is a positive int64 literal -- so that consumers
        // which do not yet read `offset_expr` can still execute the plan.
```

proto/substrait/algebra.proto
```diff
@@ -1413,33 +1413,93 @@ message Expression {
...
+        // A strictly positive integer specifying the offset of this bound.
+        // Use CurrentRow for offset zero and Preceding for negative offsets.
+        //
+        // At least one of `offset` and `offset_expr` must be set; setting
+        // neither is invalid. Consumers must use `offset_expr` when it is set
+        // and ignore `offset`. Producers may set both during migration, in
+        // which case `offset` must be the int64-literal equivalent of
+        // `offset_expr` -- possible only for BOUNDS_TYPE_ROWS with a literal
+        // offset -- so that consumers which do not yet read `offset_expr` can
+        // still execute the plan.
```
nielspardon: Same two fixes as on `Preceding` -- this paragraph is byte-identical, so the mirror is too.

```suggestion
        // At least one of `offset` and `offset_expr` must be set; setting
        // neither is invalid. Because `offset` has no explicit presence, it is
        // considered set when it is non-zero; zero was never a valid offset.
        // Consumers must use `offset_expr` when it is set and ignore `offset`.
        // Producers may set both during migration, in which case `offset` must
        // be the int64-literal equivalent of `offset_expr` -- possible only
        // when the distance is a positive int64 literal -- so that consumers
        // which do not yet read `offset_expr` can still execute the plan.
```

site/docs/expressions/window_functions.md
```diff
@@ -18,8 +18,24 @@ When binding a window function, the binding must include the following additiona
...
+* `BOUNDS_TYPE_RANGE` defines value offsets from the current row's ordering value and requires exactly one ordering expression. `offset_expr` must evaluate to a strictly positive distance whose type `D` is compatible with the ordering expression's type `T` — that is, `add(T, D) -> T` and `subtract(T, D) -> T` must be defined (for example `timestamp`/`interval_day`, `decimal`/`decimal`, `i64`/`i64`). For a `Preceding` bound the boundary value is `subtract(current, offset)` under an ascending ordering and `add(current, offset)` under a descending ordering; `Following` is the mirror. A range boundary is inclusive.
+
+`offset_expr` is the recommended way to specify an offset. A literal integer `offset` is still accepted for compatibility but is deprecated and will be removed in a future release. At least one of the two must be set; consumers must use `offset_expr` when it is set and ignore `offset`. During migration a producer may set both, with `offset` carrying the int64-literal equivalent — possible only for `BOUNDS_TYPE_ROWS` with a literal offset — so that consumers which do not yet read `offset_expr` can still execute the plan.
```
nielspardon: Matching the proto change: the dual-write condition is about the distance being a positive int64 literal, not about `BOUNDS_TYPE_ROWS` (RANGE with an integer distance was expressible via `offset` before this PR). Plus the same presence clarifier, so the prose and the proto comment state the same rule.

```suggestion
`offset_expr` is the recommended way to specify an offset. A literal integer `offset` is still accepted for compatibility but is deprecated and will be removed in a future release. At least one of the two must be set; because `offset` has no explicit presence it is considered set when it is non-zero, and zero was never a valid offset. Consumers must use `offset_expr` when it is set and ignore `offset`. During migration a producer may set both, with `offset` carrying the int64-literal equivalent — possible only when the distance is a positive int64 literal — so that consumers which do not yet read `offset_expr` can still execute the plan.
```

site/docs/expressions/window_functions.md
```diff
@@ -18,8 +18,24 @@ When binding a window function, the binding must include the following additiona
...
+| -------------- | --------------------------------- | ------------------------------- |
+| Ascending      | Toward lower ordering values      | Toward higher ordering values   |
+| Descending     | Toward higher ordering values     | Toward lower ordering values    |
+
+The interpretation of an offset depends on `BoundsType`:
+
+* `BOUNDS_TYPE_ROWS` defines physical row offsets in the declared ordering. `offset_expr` must evaluate to a strictly positive number of rows and must have type `int64`. If no ordering is declared, the row order is implementation-defined.
+* `BOUNDS_TYPE_RANGE` defines value offsets from the current row's ordering value and requires exactly one ordering expression. `offset_expr` must evaluate to a strictly positive distance whose type `D` is compatible with the ordering expression's type `T` — that is, `add(T, D) -> T` and `subtract(T, D) -> T` must be defined (for example `timestamp`/`interval_day`, `decimal`/`decimal`, `i64`/`i64`). For a `Preceding` bound the boundary value is `subtract(current, offset)` under an ascending ordering and `add(current, offset)` under a descending ordering; `Following` is the mirror. A range boundary is inclusive.
```
nielspardon: Nit: the two formulas say `offset`, which now reads as the deprecated int64 field. The subject of the sentence is `offset_expr`.

```suggestion
* `BOUNDS_TYPE_RANGE` defines value offsets from the current row's ordering value and requires exactly one ordering expression. `offset_expr` must evaluate to a strictly positive distance whose type `D` is compatible with the ordering expression's type `T` — that is, `add(T, D) -> T` and `subtract(T, D) -> T` must be defined (for example `timestamp`/`interval_day`, `decimal`/`decimal`, `i64`/`i64`). For a `Preceding` bound the boundary value is `subtract(current, offset_expr)` under an ascending ordering and `add(current, offset_expr)` under a descending ordering; `Following` is the mirror. A range boundary is inclusive.
```

nielspardon (commented):
Thanks for applying the earlier round -- everything landed, and the line-ending normalization plus the merge with `main` cleared those two nits.

Four follow-ups, all on the deprecation/migration wording. Two of them correct a mistake in my own earlier suggestion: I claimed dual-writing `offset` alongside `offset_expr` is "possible only for `BOUNDS_TYPE_ROWS` with a literal offset", but `BOUNDS_TYPE_RANGE` was defined against that same `int64 offset` before this PR, so a RANGE bound with an integer distance could be expressed with it as well. The condition should be on the distance being a positive int64 literal. The other two are a proto3 presence clarifier for "at least one must be set" and a field-name fix in the docs formulas.

nielspardon (approved):
LGTM, with the comment fixes

I guess we want to mark this as a breaking change `feat(protos)!:` and add a breaking change note to announce the deprecation, e.g.:

```
BREAKING CHANGE: deprecates the `offset` field of the `Preceding` and
`Following` window bounds in favor of the new `offset_expr` field, which
accepts an arbitrary expression instead of an integer literal. Consumers
should add `offset_expr` support before producers switch to it.
```

nielspardon:
FYI: I configured an exception for `Copilot` in the CLA assistant config just like we have already for dependabot

pyproject.toml
nielspardon, line 1: I think because Copilot showed up as a co-author the GH actions workflows did not run and needed an approval. Just approved the run and there's now failures since `pixi.lock` has not been updated after the `pyproject.toml` changes. A simple `pixi lock` run should update it.

The `pyproject.toml` changes also seem unrelated to the spec change. They might still be valuable to include in the repo. Then it would make sense to cover them in a separate PR.

proto/substrait/algebra.proto
```diff
@@ -1390,59 +1396,132 @@ message Expression {
...
+    // less than the lower bound (i.e., empty frame due to contradiction),
+    // then the behavior follows the semantics of the window function over
+    // the empty frame (e.g., SUM yields NULL, COUNT(*) yields 0).
     // Optional; defaults to the end of the partition.
     Bound upper_bound = 4;
 
     enum BoundsType {
+      // Unspecified. Producers must set one of the defined bounds types
+      // when either bound is CurrentRow, Preceding or Following.
       BOUNDS_TYPE_UNSPECIFIED = 0;
```
nielspardon: I'd raised a compatibility concern here and then went and checked the implementations, which showed I had it backwards — recording that so the next reviewer doesn't repeat it. Requiring this is right, and there is no coherent existing behavior for a fallback to preserve. Across consumers, an unset `bounds_type` gets four different treatments:

| Behavior | Implementation |
|---|---|
| Reject | substrait-java — `ExpressionRexConverter.java:705-712`, `"bounds type on window function must be specified"` |
| Always RANGE | duckdb-substrait-extension — `from_substrait.cpp:1639`, `is_rows = bounds_type == BOUNDS_TYPE_ROWS`, so `CurrentRow` falls through to `CURRENT_ROW_RANGE` |
| Conditional on `sorts` | DataFusion — `consumer/expr/window_function.rs:74-83`, `if order_by.is_empty() { Rows } else { Range }` |
| Accept, never interpret | substrait-go, substrait-rs, substrait-validator |

The pre-PR text said `CurrentRow` "extends to or from the current record" unconditionally, and no implementation implements *that* rule — DataFusion lands on record-based only when `sorts` is empty. So a documented record-based fallback would invent behavior nobody has, and would force both DuckDB and DataFusion to change results.

It also leaves a live divergence that this rule removes: `CurrentRow` with no `ORDER BY` and unset `bounds_type` gives ROWS on DataFusion (frame ends at the current record) and RANGE on DuckDB (no ordering, so every record is a peer, so the frame is the whole partition). Same plan, different results. With `ORDER BY` present the two agree on RANGE.

The one thing I'd add is the consumer obligation, since silently guessing is the actual failure mode this should eliminate — a producer requirement alone doesn't tell a consumer not to guess.

```suggestion
      // Unspecified. Producers must set one of the defined bounds types when
      // either bound is CurrentRow, Preceding or Following. Consumers must
      // reject such plans rather than assuming a bounds type, since the two
      // types disagree whenever the ordering has ties, including when no
      // ordering is declared and every record is a peer.
      BOUNDS_TYPE_UNSPECIFIED = 0;
```

Worth knowing the cost: substrait-go currently defaults its builder to UNSPECIFIED and has tests locking that in as valid (`expr/window_function_test.go:97`, and one named "UNSPECIFIED with multiple sort fields should succeed"). So this does invalidate substrait-go's default output and needs a follow-up there, alongside DuckDB and DataFusion — which is a good reason for the `BREAKING CHANGE:` footer.

*Edited to fold in the DataFusion survey I'd flagged as missing in the review body, and to correct an overstatement: I originally wrote that no implementation implements the record-based reading, which was too absolute. The suggestion now also covers the no-ordering case, which is where DuckDB and DataFusion actually diverge.*

proto/substrait/algebra.proto
```diff
@@ -1390,59 +1396,132 @@ message Expression {
...
+        // containing this window function; the expression must behave
+        // as if evaluated once per input record.
+        // It must not contain window or aggregate functions.
+        Expression offset_expr = 2;
       }
 
-      // Defines that the bound extends to or from the current record.
+      // Defines that the bound extends to or from the current record with ROWS bound,
+      // or peer records (i.e., the set of records with equal sort key values) with RANGE bound.
       message CurrentRow {}
```
nielspardon: "records with equal sort key values" doesn't say whether two nulls are equal, and the new null rule on `offset_expr` depends on the answer. Given "when the current row's ordering value is null, a RANGE offset bound is equivalent to CurrentRow" — if nulls are peers that frame is *every* null row; if not, it's the single current row. `SUM` over a null-ordered partition differs either way. SQL treats nulls as not-distinct for peer grouping, so I've made that explicit.

```suggestion
      // Defines that the bound extends to or from the current record under
      // BOUNDS_TYPE_ROWS, or to or from the current peer group under
      // BOUNDS_TYPE_RANGE. A peer group is the set of records with equal sort
      // key values, where two null sort key values are considered equal.
      message CurrentRow {}
```

proto/substrait/algebra.proto
```diff
@@ -1390,59 +1396,132 @@ message Expression {
...
+      // Bounds are value offsets from the current row's ordering value.
+      // This bounds type requires exactly one ordering expression if Preceding or Following is used.
+      // Preceding moves toward values earlier in the declared ordering in 'sorts';
+      // Following moves toward values later in the declared ordering in 'sorts'.
+      // Bounds are inclusive.
       BOUNDS_TYPE_RANGE = 2;
     }
 
     // Defines one of the two boundaries for the window of a window function.
     message Bound {
```
nielspardon: Since fixing the `ConsistentPartitionWindowRel` underspecification is part of what you're after here: `WindowRelFunction` reuses this `Bound` and this `BoundsType`, but L443-447 carry no comments at all — `lower_bound`, `upper_bound` and `bounds_type = 12` are all bare, where `WindowFunction.bounds_type` at least has a one-liner at L1406. Every obligation added in this PR is invisible from the relation side.

There's a real ambiguity to nail down while you're here too: the RANGE restrictions are phrased against `'sorts'`, and *both* messages have a field of that name — `WindowFunction.sorts` (L1393) and `ConsistentPartitionWindowRel.sorts`. Which one governs is left to the reader.

Proposed text for L443-447:

```proto
    // Defines the lower bound of the window frame. Optional; defaults to the
    // start of the partition. See `Expression.WindowFunction.Bound`.
    Expression.WindowFunction.Bound lower_bound = 5;

    // Defines the upper bound of the window frame. Optional; defaults to the
    // end of the partition. See `Expression.WindowFunction.Bound`.
    Expression.WindowFunction.Bound upper_bound = 4;

    // Defines the bounds type: ROWS, RANGE. The producer obligations
    // documented on `Expression.WindowFunction.BoundsType` apply here as well;
    // the ordering restrictions for BOUNDS_TYPE_RANGE apply to `sorts` on the
    // enclosing ConsistentPartitionWindowRel.
    Expression.WindowFunction.BoundsType bounds_type = 12;
```

proto/substrait/algebra.proto
```diff
@@ -1390,59 +1396,132 @@ message Expression {
...
+        //   applied toward values earlier in the declared ordering in 'sorts'.
+        //   Its type D must be compatible with the type T of the single ordering
+        //   expression, i.e. `add(T, D) -> T` and `subtract(T, D) -> T` must
+        //   be defined (e.g. timestamp/interval_day, decimal/decimal,
+        //   i64/i64).
+        //
+        // A null offset expression result is invalid. When the current row's
+        // ordering value is null, a RANGE offset bound is equivalent to CurrentRow.
+        // A negative result is invalid; a zero result is equivalent
+        // to CurrentRow. Use Following for a bound after the current row.
```
nielspardon: Relaxing `offset_expr` from strictly-positive to non-negative is right, but it opens a small hole in the migration story. `offset` is documented as strictly positive ("use CurrentRow for offset zero"), and `offset = 0` is indistinguishable from unset anyway — so a producer emitting `offset_expr = 0` cannot mirror it into `offset`, and the plan is silently unreadable by consumers that only understand `offset`. Worth saying the zero case should use `CurrentRow`.

```suggestion
        // A negative result is invalid; a zero result is equivalent
        // to CurrentRow. Because `offset` cannot represent zero, producers
        // should emit CurrentRow rather than a zero `offset_expr` so that the
        // bound remains readable by consumers that do not yet support
        // `offset_expr`. Use Following for a bound after the current row.
```

The identical edit applies to `Following` at L1513-1514, with the last sentence as "Use Preceding for a bound before the current row."

nielspardon (commented):
Following up on our offline conversation — thanks for picking up the peer semantics and the RANGE narrowing. Worth noting substrait-go already implements "RANGE requires exactly one sort field" (`expr/functions.go:634`), so that rule is codifying existing practice.

I went through the implementations to check whether requiring `bounds_type` breaks existing plans, and it doesn't — full detail in my comment on L1425. Short version: an unset `bounds_type` gets four different treatments today. substrait-java rejects it, DuckDB silently assumes RANGE, DataFusion infers from whether `sorts` is empty (`if order_by.is_empty() { Rows } else { Range }`), and substrait-go/-rs/validator accept it without ever interpreting it. Nothing implements the unconditional record-based reading the old comment described.

That leaves a live divergence which this rule removes: `CurrentRow` with no `ORDER BY` and unset `bounds_type` gives ROWS on DataFusion (frame ends at the current record) and RANGE on DuckDB (no ordering, so every record is a peer, so the frame is the whole partition). Same plan, different results. So requiring `bounds_type` replaces silent wrong answers with an explicit error rather than breaking anything that works today.

Four follow-ups fall out of that, which I'm happy to file:

- **substrait-go** defaults its builder to UNSPECIFIED and has tests asserting that's valid — needs a default change plus test updates.
- **duckdb-substrait-extension** should reject unset instead of falling through to RANGE (`from_substrait.cpp:1639`).
- **DataFusion** should reject unset instead of inferring from `sorts` (`consumer/expr/window_function.rs:74-83`).
- **substrait-validator** has no `bounds_type` check at all and could gain one.

Unrelated to this PR, but worth recording: Gluten never references `bounds_type` — its fork carries its own top-level `enum WindowType { ROWS = 0; RANGE = 1; }`. Separate consolidation item.

Still outstanding on metadata:

- Title needs the `!` — `feat(protos)!: support expression in window aggregate bounds`. Per `.releaserc.mjs` that bumps the minor on 0.x, not 1.0.0.
- The `BREAKING CHANGE:` footer isn't in the body yet, so this won't render under `### ⚠ BREAKING CHANGES`. Since buf reports a deprecation as clean, nothing in CI enforces it — and the `bounds_type` requirement makes it more warranted, since it does invalidate substrait-go's default output.
- `Closes #617` and `Closes #930` would be worth adding.

Nit, take or leave: the new comments write `'sorts'` where the file's convention is `` `sorts` `` — 26 backticked identifiers against 11 single-quoted, and this PR adds 9 of the latter.

pyproject.toml
yongchul: wait why is this here?!

nielspardon: there was an editorconfig GH actions glitch yesterday and I suspect your Copilot might have tried to work around that somehow and included the commit on the PR branch

yongchul: reverted the change.

proto/substrait/algebra.proto
```diff
@@ -1390,59 +1405,143 @@ message Expression {
...
+        // to CurrentRow. Because `offset` cannot represent zero, producers
+        // should emit CurrentRow rather than a zero `offset_expr` so that the
+        // bound remains readable by consumers that do not yet support
+        // `offset_expr`. Use Following for a bound after the current row.
+        //
+        // Field references resolve against the input schema of the relation
+        // containing this window function; the expression must behave
+        // as if evaluated once per input record.
+        // It must not contain window or aggregate functions.
+        Expression offset_expr = 2;
```
benbellick: This is a bit pedantic, but you technically can have multiple matching impls under different URNs, so this is not sufficient information to disambiguate. 

In general, I think explicit is better than implicit anyways. Two options I can think of:
1. add a field that is a `function_reference` to explicitly say what is being used as `add` or `substract` here, or
2. we put the complete boundary expression here e.g. `add($<reference to order column>, <2 days>)`, but I'm not sure what the right way is to reference the order column

proto/substrait/algebra.proto
```diff
@@ -1390,59 +1405,143 @@ message Expression {
...
-    // yield an equal record of values for both. When computing the window
-    // function, only the subset of records within the bounds that are also in
-    // the same partition as the current record are aggregated.
+    // yield an equal record of values for both (nulls are considered equal).
+    // When computing the window function, only the subset of records within
+    // the bounds that are also in the same partition as the current record
+    // are aggregated.
     repeated Expression partitions = 2;
 
     // Defines the bounds type: ROWS, RANGE
```
```suggestion
    // Defines how the window bounds are interpreted.
    // Required when either bound is CurrentRow, Preceding, or Following
```

We should put the fact that it is required on the field, rather than the message definition.

proto/substrait/algebra.proto
```diff
@@ -1390,59 +1405,143 @@ message Expression {
...
+    // the empty frame (e.g., SUM yields NULL, COUNT(*) yields 0).
     // Optional; defaults to the end of the partition.
     Bound upper_bound = 4;
 
     enum BoundsType {
+      // Unspecified. Producers must set one of the defined bounds types when
+      // either bound is CurrentRow, Preceding or Following. Consumers must
+      // reject such plans rather than assuming a bounds type, since the two
+      // types disagree whenever the ordering has ties, including when no
+      // ordering is declared and every record is a peer.
```
benbellick: The following is my opinion 😅 so take it or leave it. 
```suggestion
      // Valid only when the frame is unbounded in both directions and therefore
      // spans the entire partition.
```
Related to the above comment. I think this can be simpler:
1. we don't need to explicitly mark every invalid behavior as needing to be rejected by consumers. That is implicit IMO
2. we can let the field itself above own the "required" semantics.

benbellick (commented):
I haven't completed my full review yet, but I wanted to do a temperature check first on the discussion around explicitly providing a function used for the range case. What do others think? I'll try and go through the rest of the comments and things tomorrow. Thanks!

benbellick:
Ah one nit I forgot to leave in my PR. The `BREAKING CHANGE` description in the PR title doesn't seem correct to me. Deprecating a field and introducing a new field isn't a breaking change. A breaking change would be deleting the old field. What is the actual breakage here?

proto/substrait/algebra.proto
yongchul: thanks! yeah, just requiring it is simpler indeed.

proto/substrait/algebra.proto
yongchul: I'm inclined to punt this to a follow up PR as so many thing is going on this PR. 😆 

I am against option 2... or I don't know. All you need here is a lambda that T -> bool : true if it is within a bound. Do we really want that? Maybe not too bad but we do need to unwrap the lambda and analyze depending on the implementaiton -- I certainly don't want to go through the pain in my code base.

option 1 feels more approachable... but need to think in the context of extending this to `GROUPS` #902

yongchul:
> Ah one nit I forgot to leave in my PR. The `BREAKING CHANGE` description in the PR title doesn't seem correct to me. Deprecating a field and introducing a new field isn't a breaking change. A breaking change would be deleting the old field. What is the actual breakage here?

Previously valid `Unspecified` are not allowed anymore. @nielspardon found that Substrait-go needs to be fixed. deprecating a field is not technically a breaking change but removal is... Vote-wise, though, deprecation should be treated as breaking change. This is hairy 😆

nielspardon:
We have been using the BREAKING CHANGE note mechanism in the past to **announce** deprecations to downstream implementations of the Substrait specification so they can prepare their changes following the breaking change policy. How would a downstream implementation get notified about the deprecation from the current PR title which will be added to the changelog without a breaking change note?

proto/substrait/algebra.proto
nielspardon: In general, given that we are a specification that wants to be unambiguous I would tend more towards communicating more explicitly rather than implicitly.

benbellick:
I opened an issue to discuss the meaning of `BREAKING CHANGE` [here](https://github.com/substrait-io/substrait/issues/1179), as I don't want to derail the conversation about this PR. I agree though that:
> Previously valid Unspecified are not allowed anymore.

does make this PR a breaking change. 

I will continue my review today. Thanks!

proto/substrait/algebra.proto
```diff
@@ -817,6 +826,8 @@ message UpdateRel {
   // The list of transformations to apply to the columns of the named_table
   repeated TransformExpression transformations = 4;
 
+  RelCommon common = 6;
```
vbarua: This should be in it own PR with it's own release notes announcing the fix. It doesn't have anything to do with this work.

proto/substrait/algebra.proto
```diff
@@ -1390,59 +1405,141 @@ message Expression {
...
-    // wrapping around as if lower/upper were swapped? error? null?).
+    // less than the lower bound (i.e., empty frame due to contradiction),
+    // then the behavior follows the semantics of the window function over
+    // the empty frame (e.g., SUM yields NULL, COUNT(*) yields 0).
     // Optional; defaults to the end of the partition.
     Bound upper_bound = 4;
 
     enum BoundsType {
+      // Valid only when the frame is unbounded in both directions and therefore
+      // spans the entire partition.
```
vbarua: Do we need a carve-out to allow this to be unspecified, or can we just require that it be set? Wasn't that the case already?

proto/substrait/algebra.proto
yongchul: good catch. this is something picked up from recent pull. It's already merged in the main by Neil's PR.

site/docs/expressions/window_functions.md
```diff
@@ -18,8 +18,26 @@ When binding a window function, the binding must include the following additiona
...
+
+`offset_expr` is the recommended way to specify an offset. A strictly positive literal integer `offset` is still accepted for compatibility but is deprecated and will be removed in a future release. At least one of the two must be set; because `offset` has no explicit presence it is considered set when it is non-zero, and zero was never a valid offset. Consumers must use `offset_expr` when it is set and ignore `offset`. During migration a producer may set both, with `offset` carrying the int64-literal equivalent — possible only when the distance is a positive int64 literal — so that consumers which do not yet read `offset_expr` can still execute the plan.
+
+`offset_expr` may reference the input's fields and must behave as if evaluated once per input record, but it must not contain window or aggregate functions. A null or negative result is invalid; a zero result is equivalent to `CurrentRow`. Because `offset` cannot represent zero, producers should emit `CurrentRow` rather than a zero `offset_expr` so that the bound remains readable by consumers that do not yet support `offset_expr`.
```
vbarua: > Because `offset` cannot represent zero, producers should emit `CurrentRow` rather than a zero `offset_expr`

I would make a stronger case here and say the `CurrentRow` MUST be used to represent a zero-offset.

proto/substrait/algebra.proto
```diff
@@ -440,10 +440,19 @@ message ConsistentPartitionWindowRel {
...
     Expression.WindowFunction.Bound lower_bound = 5;
 
+    // Defines the upper bound of the window frame.
+    // Optional; defaults to the end of the partition.
+    // See `Expression.WindowFunction.Bound`.
     Expression.WindowFunction.Bound upper_bound = 4;
 
+    // Defines the bounds type: ROWS, RANGE.
+    // The producer obligations documented on `Expression.WindowFunction.BoundsType` apply here as well;
+    // the ordering restrictions for BOUNDS_TYPE_RANGE apply to `sorts` on the enclosing `ConsistentPartitionWindowRel`.
```
```suggestion
   // Defines how this function's window bounds are interpreted.
   // Must be ROWS or RANGE when either bound is CurrentRow, Preceding, or
   // Following.
   //
   // For RANGE with a Preceding or Following bound, `sorts` on the enclosing
   // ConsistentPartitionWindowRel must contain exactly one ordering expression
   // and must not use SORT_DIRECTION_CLUSTERED or a custom comparison function.
```
nit(opinionated): I would rather just repeat the documentation than cross reference to another relation.

site/docs/expressions/window_functions.md
```diff
@@ -18,8 +18,26 @@ When binding a window function, the binding must include the following additiona
...
+`Preceding` and `Following` define offsets relative to the current record and with respect to the declared ordering. `Preceding` moves toward records or values earlier in that ordering, while `Following` moves toward records or values later in that ordering. They therefore apply in opposite value directions for ascending and descending sorts:
+
+| Sort direction | `Preceding` boundary              | `Following` boundary            |
+| -------------- | --------------------------------- | ------------------------------- |
+| Ascending      | Toward lower ordering values      | Toward higher ordering values   |
+| Descending     | Toward higher ordering values     | Toward lower ordering values    |
+
+The interpretation of an offset depends on `BoundsType`:
+
+* `BOUNDS_TYPE_ROWS` defines physical row offsets in the declared ordering. `offset_expr` must evaluate to a non-negative number of rows and must have type `int64`. If no ordering is declared, no row order is guaranteed.
```
vbarua, line 33: meta: Not for this PR, but thinking about future work to add `BOUNDS_TYPE_GROUP` I think it would be good to rework this section to something like
```
# Bound Modes
The three bounds mode (ROWS, RANGE, GROUP) impose the following restrictions on other aspects of window functions settings
## ROWS
<All ROWS invariants>

## RANGE
<All range invariants>

## GROUP
<All group invariants> 
```

I would then suggest referencing this suggestion from the protos, and reducing the documentation on the protos directly, because the fields in the protobufs are not necessarily grouped in a way to make it possible to really highlight the invariants in one place.

proto/substrait/algebra.proto
```diff
@@ -1390,59 +1405,141 @@ message Expression {
 
     // When one or more partition expressions are specified, two records are
     // considered to be in the same partition if and only if these expressions
-    // yield an equal record of values for both. When computing the window
-    // function, only the subset of records within the bounds that are also in
-    // the same partition as the current record are aggregated.
+    // yield an equal record of values for both (nulls are considered equal).
+    // When computing the window function, only the subset of records within
+    // the bounds that are also in the same partition as the current record
```
```suggestion
    // the inclusive frame bounds that are also in the same partition as the current record
```
? Maybe good to highlight inclusive here?

vbarua (approved):
The actual proto addition of the expr fields makes sense, but I think the docs are unlikely to help with correct implementation as they are. They do capture the details that would need to be enforced, but I think the protos are the wrong place for them

My suggestion would be to do something like https://github.com/substrait-io/substrait/pull/1105/changes#r3787025749 in a follow-up.

The big difference would be that we could group the invariants clearly by mode, instead of smearing the invariants across the various fields.

proto/substrait/algebra.proto
vbarua: I think we actually need this to now show up in the diff, because I think if we merge this as is we'll get
```proto
  RelCommon common = 6;
  RelCommon common = 6;
```

site/docs/expressions/window_functions.md
benbellick: ➕

site/docs/expressions/window_functions.md
benbellick: ➕

proto/substrait/algebra.proto
benbellick: We discussed offline, and we will probably do option 1 but in a follow up PR

proto/substrait/algebra.proto
yongchul: It was not required and implementations have chosen different paths when it is not set -- substrait-go allows that and have a test case. I'm more inclined to actually requiring it but when all you need is unbounded, it's kinda pointless to settings this. but then, you need to handle unbounded case anyway even if you have a specific bounds type so.... just require it and break?

Maybe this is a necessary breaking change.

site/docs/expressions/window_functions.md
```diff
@@ -18,8 +18,26 @@ When binding a window function, the binding must include the following additiona
...
+
+`offset_expr` is the recommended way to specify an offset. A strictly positive literal integer `offset` is still accepted for compatibility but is deprecated and will be removed in a future release. At least one of the two must be set; because `offset` has no explicit presence it is considered set when it is non-zero, and zero was never a valid offset. Consumers must use `offset_expr` when it is set and ignore `offset`. During migration a producer may set both, with `offset` carrying the int64-literal equivalent — possible only when the distance is a positive int64 literal — so that consumers which do not yet read `offset_expr` can still execute the plan.
```
benbellick, line 38: nit: I don't think we should have to specify how to do the migration for every field inside of the documentation. But to be able to do that, we probably should have a more formal migration policy. That can be left as a follow up, so this is okay for now.

proto/substrait/algebra.proto
```diff
@@ -1382,6 +1393,10 @@ message Expression {
     // priority; only if a sort field determines two records to be equivalent
     // is the next field queried. This field is optional, and is only allowed
```
benbellick: I think there is a possible reconfiguration of the protos that would make some of these invariants easier to express, but it almost certainly would be a large migration and hurts my head a bit to think about, so I think we can stick with this for now, but we may want to explore something like that in the future 😅

proto/substrait/algebra.proto
```diff
@@ -1390,59 +1405,141 @@ message Expression {
...
-        // CurrentRow for offset zero and Following for negative offsets.
-        int64 offset = 1;
+        // A strictly positive integer specifying the offset of this bound.
+        // Use CurrentRow for offset zero and Following for negative offsets.
+        //
+        // At least one of `offset` and `offset_expr` must be set; setting
+        // neither is invalid. Because `offset` has no explicit presence, it is
+        // considered set when it is non-zero; zero was never a valid offset.
+        // Consumers must use `offset_expr` when it is set and ignore `offset`.
+        // Producers may set both during migration, in which case `offset` must
```
benbellick: I think some of the language we use around migrations needs some work in general, though we can leave that for the future. In my opinion, substrait org libraries _must_ do this. But people doing their own implementation can do whatever they want 🤷, then its just a recommendation.

benbellick (approved):
Thanks for your patience 🙏 

As an aside, I think this PR revealed a lot of improvements to documentation that needed to be taken care of. However, I think it would make PRs easier to review in the future if we tried to isolate them to the core semantic changes and updated the other docs in separate PRs. 

For example, adding docs like:
```proto
    // Defines the upper bound of the window frame.
    // Optional; defaults to the end of the partition.
    // See `Expression.WindowFunction.Bound`.
    Expression.WindowFunction.Bound upper_bound = 4;
```
are great, but completely unambiguous and I would easily approve it on its own. I also am guilty of this 😅 

Thanks so much this will be great!

proto/substrait/algebra.proto
benbellick: https://github.com/substrait-io/substrait/issues/1180

proto/substrait/algebra.proto
vbarua: > It was not required and implementations have chosen different paths when it is not set -- substrait-go allows that and have a test case.

The other perspective is that it was never allowed to be `UNSPECIFIED` 😓

Given that all the libraries have a means of setting the value, I think that it would be reasonable to explicitly declare that this must be set and mark it a BREAKING CHANGE. Anyone using this feature should be able to set the field explicitly, if they're not already.

site/docs/expressions/window_functions.md
benbellick: https://github.com/substrait-io/substrait/pull/1181

proto/substrait/algebra.proto
benbellick: https://github.com/substrait-io/substrait/pull/1181

proto/substrait/algebra.proto
nielspardon: I suspect that Copilot tried to work around a failing buf breaking check and instead of merging latest main into the PR branch it fixed it by adding the new field on main into this branch. I guess if you merge main into this branch we should have a cleaner PR state to reason about.

site/docs/expressions/window_functions.md
nielspardon: Agree on MUST, with one qualifier. offset_expr may reference the input's fields and is evaluated per record, so a producer cannot statically know that a non-literal expression will come out zero — a blanket MUST isn't satisfiable in that case. Splitting the static case from the runtime one:

```suggestion
`offset_expr` may reference the input's fields and must behave as if evaluated once per input record, but it must not contain window or aggregate functions. A null or negative result is invalid. A statically-known zero offset must be expressed as `CurrentRow` rather than as a zero `offset_expr`; `offset` cannot represent zero either way, so this also keeps the bound readable by consumers that do not yet support `offset_expr`. When a non-literal `offset_expr` evaluates to zero for a record, the bound is equivalent to `CurrentRow` for that record.
```

from my 🤖 review assistant

proto/substrait/algebra.proto
nielspardon: +1, and there's now a concrete reason beyond taste: the cross-reference has gone stale. It points at Expression.WindowFunction.BoundsType, but 8fa4079 moved the producer obligation off the enum and onto the bounds_type field, so the enum no longer documents what this comment claims it does.

It pulls the opposite way from @vbarua's [suggestion](https://github.com/substrait-io/substrait/pull/1105#discussion_r3787025749) about thinning the protos and giving the invariants one home in the docs. Worth picking a direction rather than landing between them.

from my 🤖 review assistant

site/docs/expressions/window_functions.md
nielspardon: +1. It would also give the cross-field invariants a single home.

This and @benbellick's [suggestion](https://github.com/substrait-io/substrait/pull/1105#discussion_r3787013974) on ConsistentPartitionWindowRel.bounds_type are alternatives rather than complements, though.

One thing worth carrying over whichever way it goes: the per-mode invariants are only half of it. The other half is what a consumer must do when the mode isn't set, which is where the implementations actually diverge today — substrait-java rejects, DuckDB assumes RANGE, DataFusion infers from sorts.

yongchul: It is a complex piece so better to clearly describe the model, invariants, and what to do and not to do in a doc. I have a strong preference to one definition or one source of truth. This, clearly, one doc is better than copy and pasting the doc around and go stale.

At least this PR prescribes the necessary components as a spec, so I'll create a follow up issue to properly document the concept, invariants, and what to do/not to.

yongchul: #1182

proto/substrait/algebra.proto
yongchul: updated the branch. no longer showing up as change.

proto/substrait/algebra.proto
yongchul: #1182

proto/substrait/algebra.proto
yongchul: Unspecified is now explicitly not allowed.

PR #1107, opened 2026-06-18
fix(grammar): give binary operators conventional precedence

### LINKED ISSUE #1106: SubstraitType.g4: binary operators lack precedence (single left-recursive alternative)
## Summary

In [`grammar/SubstraitType.g4`](https://github.com/substrait-io/substrait/blob/main/grammar/SubstraitType.g4),
all binary operators are defined in a **single left-recursive alternative**:

```antlr
expr
  : ...
  | left=expr op=(And | Or | Plus | Minus | Lt | Gt | Eq | Ne |
        Lte | Gte | Asterisk | ForwardSlash) right=expr   #BinaryExpr
  | ...
  ;
```

Because ANTLR4 derives operator precedence from the *order of alternatives* in a left-recursive rule, putting
every operator into one alternative gives them all **the same precedence** and left associativity. There is no
`*`/`/`-over-`+`/`-` precedence, no comparison-below-arithmetic precedence, etc.

As a result, a type-derivation expression such as:

```
1 + 2 * 3 - 4 / 5
```

parses as `((((1 + 2) * 3) - 4) / 5)` and evaluates to **1**, rather than the conventional `1 + (2*3) - (4/5) = 7`.

## Why this matters

Type-derivation expressions in the simple-extension YAMLs are arithmetic over type parameters (e.g. precision/
scale computations for `decimal`). Authors will reasonably expect standard mathematical precedence. With the
current grammar, an expression that mixes `*`/`/` and `+`/`-` (or arithmetic and comparison) at the same level
silently computes a different value than intended.

Today the bundled core extensions happen to avoid this by parenthesizing via function calls (`min(...)`,
`max(...)`) and not mixing multiplicative and additive operators at the same level, so the bug is latent — but
it is a correctness trap for extension authors and for any consumer that evaluates these derivations.

This surfaced while migrating
[substrait-validator](https://github.com/substrait-io/substrait-validator) to consume the
`substrait-antlr` crate generated from this grammar: the validator's previous (bespoke) grammar implemented
conventional precedence via a precedence cascade, so the same derivation evaluated to `7` there. See
[substrait-validator#526](https://github.com/substrait-io/substrait-validator/issues/526) for the broader
grammar gap analysis.

## Suggested fix

Give the operators conventional precedence by splitting `#BinaryExpr` into ordered left-recursive alternatives,
highest precedence first, e.g.:

```antlr
expr
  : ...
  | left=expr op=(Asterisk | ForwardSlash) right=expr   #MulDiv
  | left=expr op=(Plus | Minus) right=expr              #AddSub
  | left=expr op=(Lt | Gt | Lte | Gte) right=expr        #Comparison
  | left=expr op=(Eq | Ne) right=expr                    #Equality
  | left=expr op=And right=expr                          #And
  | left=expr op=Or right=expr                           #Or
  | ...
  ;
```

(Exact grouping/associativity up to the maintainers — the key point is to introduce distinct precedence levels
rather than a single flat alternative.)

### Compatibility note

This is a behavior change for any expression that mixes operator classes at the same level without explicit
grouping. A scan of the core extension YAMLs suggests none currently rely on the flat-precedence behavior, but
it is worth confirming and noting in release notes, since downstream tools generate parsers directly from this
grammar.

nielspardon:
## Summary

In `grammar/SubstraitType.g4` all binary operators (`* / + - < > = != <= >= and or`) were defined in a **single left-recursive alternative** (`#BinaryExpr`). Because ANTLR4 derives precedence from the order of left-recursive alternatives, a single flat alternative gave every operator the **same precedence** and left associativity.

As a result a type-derivation expression such as `1 + 2 * 3 - 4 / 5` parsed as `((((1 + 2) * 3) - 4) / 5)` instead of the conventional `((1 + (2 * 3)) - (4 / 5))`.

## Fix

Split the flat alternative into ordered left-recursive alternatives, highest precedence first (per the suggestion in the issue):

```antlr
| left=expr op=(Asterisk | ForwardSlash) right=expr   #MulDiv
| left=expr op=(Plus | Minus) right=expr              #AddSub
| left=expr op=(Lt | Gt | Lte | Gte) right=expr       #Comparison
| left=expr op=(Eq | Ne) right=expr                   #Equality
| left=expr op=And right=expr                         #And
| left=expr op=Or right=expr                          #Or
```

Precedence, highest→lowest: `* /` > `+ -` > comparison > equality > `and` > `or`. ANTLR left-recursive alternatives are left-associative by default, which is correct for all of these operators.

## Changes

- `grammar/SubstraitType.g4`: replace `#BinaryExpr` with the ordered precedence cascade above.
- Regenerated the checked-in ANTLR parsers under `tests/type/antlr_parser/` (`pixi run generate-antlr`). `BinaryExprContext` is replaced by `MulDivContext`, `AddSubContext`, `ComparisonContext`, `EqualityContext`, `AndContext`, `OrContext`. No hand-written code references `BinaryExpr`.
- Added a precedence regression test (`test_operator_precedence`) that asserts parse-tree structure (the grammar defines no evaluation semantics).

## Compatibility note

This is a behavior change for any expression that mixes operator classes at the same level without explicit grouping. A scan of the core extension YAMLs confirms none currently rely on the flat-precedence behavior, and the existing `test_extension_yaml_type_expressions_are_grammar_compliant` test continues to pass. Worth noting in release notes, since downstream tools generate parsers directly from this grammar.

Fixes #1106

🤖 Generated with AI

- - -
This change is [Reviewable](https://reviewable.io/reviews/substrait-io/substrait/1107)

### DIFF
--- a/grammar/SubstraitType.g4
+++ b/grammar/SubstraitType.g4
@@ -73,8 +73,12 @@ expr
   | Number                                                                                  #LiteralNumber
   | Identifier isnull=QMark?                                                                #ParameterName
   | Identifier OParen (expr (Comma expr)*)? CParen                                          #FunctionCall
-  | left=expr op=(And | Or | Plus | Minus | Lt | Gt | Eq | Ne |
-        Lte | Gte | Asterisk | ForwardSlash) right=expr                                     #BinaryExpr
+  | left=expr op=(Asterisk | ForwardSlash) right=expr                                      #MulDiv
+  | left=expr op=(Plus | Minus) right=expr                                                 #AddSub
+  | left=expr op=(Lt | Gt | Lte | Gte) right=expr                                          #Comparison
+  | left=expr op=(Eq | Ne) right=expr                                                       #Equality
+  | left=expr op=And right=expr                                                            #And
+  | left=expr op=Or right=expr                                                             #Or
   | If ifExpr=expr Then thenExpr=expr Else elseExpr=expr                                    #IfExpr
   | (Bang) expr                                                                             #NotExpr
   | ifExpr=expr QMark thenExpr=expr Colon elseExpr=expr                                     #Ternary
--- a/grammar/SubstraitType.g4
+++ b/grammar/SubstraitType.g4
@@ -73,12 +73,12 @@ expr
   | Number                                                                                  #LiteralNumber
   | Identifier isnull=QMark?                                                                #ParameterName
   | Identifier OParen (expr (Comma expr)*)? CParen                                          #FunctionCall
-  | left=expr op=(Asterisk | ForwardSlash) right=expr                                      #MulDiv
-  | left=expr op=(Plus | Minus) right=expr                                                 #AddSub
-  | left=expr op=(Lt | Gt | Lte | Gte) right=expr                                          #Comparison
+  | left=expr op=(Asterisk | ForwardSlash) right=expr                                       #MulDiv
+  | left=expr op=(Plus | Minus) right=expr                                                  #AddSub
+  | left=expr op=(Lt | Gt | Lte | Gte) right=expr                                           #Comparison
   | left=expr op=(Eq | Ne) right=expr                                                       #Equality
-  | left=expr op=And right=expr                                                            #And
-  | left=expr op=Or right=expr                                                             #Or
+  | left=expr op=And right=expr                                                             #And
+  | left=expr op=Or right=expr                                                              #Or
   | If ifExpr=expr Then thenExpr=expr Else elseExpr=expr                                    #IfExpr
   | (Bang) expr                                                                             #NotExpr
   | ifExpr=expr QMark thenExpr=expr Colon elseExpr=expr                                     #Ternary
--- a/tests/type/test_type_grammar.py
+++ b/tests/type/test_type_grammar.py
@@ -115,6 +115,39 @@ def test_parse_valid_type_expressions():
         parse_type_expression(case)
 
 
+def render_precedence(ctx):
+    """Render an expr parse tree as a fully-parenthesized string.
+
+    Binary operators all expose ``left``/``op``/``right``; parenthesized
+    expressions pass through to their inner expr so the rendered structure
+    reflects grammar-derived precedence rather than the source parentheses.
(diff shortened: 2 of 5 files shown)


grammar/SubstraitType.g4
```diff
@@ -73,8 +73,12 @@ expr
...
   | Identifier isnull=QMark?                                                                #ParameterName
   | Identifier OParen (expr (Comma expr)*)? CParen                                          #FunctionCall
-  | left=expr op=(And | Or | Plus | Minus | Lt | Gt | Eq | Ne |
-        Lte | Gte | Asterisk | ForwardSlash) right=expr                                     #BinaryExpr
+  | left=expr op=(Asterisk | ForwardSlash) right=expr                                      #MulDiv
+  | left=expr op=(Plus | Minus) right=expr                                                 #AddSub
+  | left=expr op=(Lt | Gt | Lte | Gte) right=expr                                          #Comparison
+  | left=expr op=(Eq | Ne) right=expr                                                       #Equality
+  | left=expr op=And right=expr                                                            #And
+  | left=expr op=Or right=expr                                                             #Or
```
vbarua: incredibly minor: could you align the `#` at the end.

tests/type/test_type_grammar.py
```diff
@@ -115,6 +115,39 @@ def test_parse_valid_type_expressions():
...
+def test_operator_precedence():
+    """Binary operators bind with conventional precedence."""
+    cases = [
+        ("1 + 2 * 3", "(1 + (2 * 3))"),
+        ("1 + 2 * 3 - 4 / 5", "((1 + (2 * 3)) - (4 / 5))"),
+        ("1 * 2 + 3", "((1 * 2) + 3)"),
+        ("(1 + 2) * 3", "((1 + 2) * 3)"),
+        ("1 + 2 < 3 * 4", "((1 + 2) < (3 * 4))"),
+        ("a and b or c", "((a and b) or c)"),
+    ]
```
vbarua, line 144: This is super clear. Thanks for adding this test.

vbarua (approved):
Changes make sense to me.

PR #1117, opened 2026-07-03
docs: add AGENTS.md guidance for AI agents

nielspardon:
## Proposed change

Add an `AGENTS.md` documenting the conventions AI agents (and new contributors) need when changing this specification repository, and a `CLAUDE.md` that imports it (`@AGENTS.md`) so Claude Code picks it up automatically.

The guidance is distilled from recurring patterns and captures:

- **Cross-SDK impact analysis** — proto changes ripple into substrait-java/-go/-python/-rs, so deprecated-field removals should be analyzed against all four SDKs first, with wire/source/semantic compatibility classified.
- **Pixi-based tooling** — `pixi run format` / `lint` / `test` / `generate`, so versions match CI.
- **Generated code** — `gen/` is gitignored; the ANTLR parsers under `tests/*/antlr_parser/` are committed and must be regenerated when grammars change.
- **CI gates** — the conventional-commit PR-title check and the `BREAKING CHANGE:` footer requirement when `buf breaking` trips.
- **Docs** — semantic changes usually need matching `site/docs` updates and externally-validated examples.

`AGENTS.md` is the emerging cross-tool convention; `CLAUDE.md` is kept to a one-line import so there is a single source of truth.

🤖 Generated with AI

- - -
This change is [Reviewable](https://reviewable.io/reviews/substrait-io/substrait/1117)

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -1,157 +1,80 @@
 # AGENTS.md
 
-Guidance for AI agents making changes to the Substrait specification repository.
+Entry point for AI agents working on the Substrait specification repository.
+Substrait is a **cross-language spec** for relational algebra (query plans) —
+this repo holds the specification, not an implementation. Read the shared docs
+below first, then keep the agent-specific notes in mind.
 
-## What this repository is
+## Start here
 
-Substrait is a **cross-language specification** for relational algebra (query
-plans). This repo holds the specification itself, not an implementation. The
-canonical artifacts are:
+- **[`README.md`](README.md)** — what Substrait is and how the repo is laid out.
+- **[`CONTRIBUTING.md`](CONTRIBUTING.md)** — environment ([Pixi](https://pixi.prefix.dev)),
+  the `pixi run` build/test/lint/generate tasks (use these rather than global
+  tool installs so versions match CI), committed vs. generated code, doc
+  examples, commit conventions, and breaking-change mechanics.
+- **Spec policy** — the [breaking-change policy](site/docs/spec/breaking_change_policy.md),
+  [versioning policy](site/docs/spec/versioning.md), and
+  [governance](site/docs/governance.md) (PMC votes for spec deprecations).
 
-- **Protobuf definitions** — [`proto/substrait/*.proto`](proto/substrait) (the
-  serialized plan format; the source of truth).
-- **Text/grammar** — [`grammar/*.g4`](grammar) (ANTLR grammar for the type and
-  test-case text format) and [`text/`](text) JSON schemas for extensions and
-  dialects.
-- **Extensions** — [`extensions/*.yaml`](extensions) (standard function
-  definitions).
-- **Docs site** — [`site/`](site) (MkDocs; published spec documentation).
+For GitHub work (issues, PRs, searching the SDK repos), use whatever access is
+configured — the `gh` CLI or a GitHub MCP server.
 
-Because this is a spec, changes here ripple into the downstream SDKs. Treat a
-change as an API change to an ecosystem, not a local edit.
+## What agents specifically need to get right
 
-## The downstream SDKs (critical context)
+These are the things agents tend to miss even after reading the docs above.
 
-Any change to `proto/` affects the SDKs that implement the spec. The maintained
-list is [`active_libraries.md`](site/docs/community/active_libraries.md), which
-separates **active libraries** (currently substrait-go, -java, -python, -rs) from
-inactive ones — treat that doc as the source of truth rather than hardcoding the
-set here. External projects also consume Substrait directly (e.g. Apache
-DataFusion, DuckDB's substrait extension); they are useful *usage signal* when
-gauging a change's blast radius, but they are not release-blocking — the active
-libraries are the gate.
+### A spec change is an ecosystem API change
 
-### Removing or changing a deprecated field
+A change to `proto/` ripples into every SDK that implements the spec. The active
+libraries are the release gate — the maintained list is
+[`active_libraries.md`](site/docs/community/active_libraries.md); treat it as the
+source of truth rather than hardcoding the set. External consumers (e.g. Apache
+DataFusion, DuckDB's substrait extension) are useful *usage signal* for gauging
+blast radius but are not release-blocking.
 
-This is the single most common task in this repo's history, and it is governed by
-the **[breaking-change policy](site/docs/spec/breaking_change_policy.md)** and the
-**[versioning policy](site/docs/spec/versioning.md)**: compatibility is maintained
-via deprecation, and a breaking change must ship an explicit **migration strategy
-that is implemented in all active libraries before the breaking change lands**.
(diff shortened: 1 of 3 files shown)


AGENTS.md
```diff
@@ -0,0 +1,144 @@
...
+  test-case text format) and [`text/`](text) JSON schemas for extensions and
+  dialects.
+- **Extensions** — [`extensions/*.yaml`](extensions) (standard function
+  definitions).
+- **Docs site** — [`site/`](site) (MkDocs; published spec documentation).
+
+Because this is a spec, changes here ripple into the downstream SDKs. Treat a
+change as an API change to an ecosystem, not a local edit.
+
+## The downstream SDKs (critical context)
```
yongchul: I think deprecation preference or guidance may be useful here? Also, it would be nice if we can include the usage survey -- like blast radius analysis for a change in other open source projects like... data fusion?

AGENTS.md
```diff
@@ -0,0 +1,144 @@
...
+   access is available — the `gh` CLI or a GitHub MCP server — or clone them
+   locally if the user has them checked out (they often do, as sibling dirs like
+   `../substrait-go`).
+2. Summarize impact per SDK (e.g. "read-path fallback only", "N/A — regenerates
+   from proto") and classify the change: *wire-compatible*, *source-breaking*,
+   and/or *semantically breaking*.
+3. Only then propose the change, often as a **draft PR** pending community
+   discussion, with companion PRs to the SDKs where needed.
+
+Wire-compatibility details matter and are expected in PR descriptions: reserve
```
yongchul: this... I don't know whether it is worth to encode more precisely... or just link to a page in protobuf compatibility doc in protobuf official documentation.

AGENTS.md
```diff
@@ -0,0 +1,144 @@
...
+Keep descriptions high-signal. The maintainer has repeatedly asked to remove:
+
+- **Lists of files touched** — obvious from the diff.
+- **Claims that CI-verified things pass** — e.g. "buf lint passes", "tests
+  pass". These run as PR checks; if they failed the checks would be red.
+- **Obvious process notes** — e.g. "draft pending review", "coordinating
+  companion PRs" once that's implicit in the PR being a draft.
+
+Do include: the rationale, the SDK impact analysis, and wire/source/semantic
+compatibility analysis for proto changes.
```
yongchul: where these should go? always create issue per PR?

AGENTS.md
nielspardon: Good call — rather than inventing new guidance, I've grounded this in the existing policy docs. It now links the [breaking-change policy](site/docs/spec/breaking_change_policy.md), [versioning policy](site/docs/spec/versioning.md), and [governance](site/docs/governance.md) (PMC votes for spec deprecations), and reorders the workflow to match the policy's rule that the migration must land in all active libraries before the breaking change. For the library list, it now points to [`active_libraries.md`](site/docs/community/active_libraries.md) as the source of truth instead of hardcoding it. And on blast radius: added external consumers (Apache DataFusion, DuckDB's substrait extension) as non-blocking *usage signal*, with the active libraries as the actual gate. (c241d19)

AGENTS.md
nielspardon: Agreed — no point re-deriving protobuf's rules here. Trimmed this down to the substrait-specific bit (reserve removed field numbers/names, enforced by `buf breaking`, so old plans carry unknown fields rather than failing to parse) and linked out to protobuf's [Updating A Message Type](https://protobuf.dev/programming-guides/proto3/#updating) guide for the general wire-compatibility rules. Dropped the dissolved-`oneof` note. (c241d19)

AGENTS.md
nielspardon: Clarified this in the doc: the impact/migration/compatibility analysis goes in the PR (or draft PR) body — no requirement to open a separate issue per PR. Issues are for surfacing design discussion on larger or contentious changes before the design is settled (the URI→URN cookbook is a good example of that flow). Does that match the intended process? Happy to adjust the wording if you'd frame it differently. (c241d19)

benbellick (commented):
This is perhaps a philosophical discussion, but I personally don't see that much value in `AGENTS.md`, as most things in those files I would also want a human to read or do.

For example, the information added about `pixi` already exists in `CONTRIBUTING.md`. In my view:

- Are agents not reading that file? Let's add a reference to `CONTRIBUTING.md` in `AGENTS.md` rather than repeating the information.
- Do agents find that it doesn't contain enough information? Let's improve `CONTRIBUTING.md` so that it is sufficiently clear to both humans and agents.

I personally have been using coding agents with this repo quite a bit without an `AGENTS.md` file, so adding a file like this essentially adds context to my agent regardless of whether it improves the situation, unless I now explicitly opt out by hiding or deleting the file locally.

Thoughts?

nielspardon:
> Thoughts?

They are finding some of this information by themselves but not always. You can cross-link to other files. The AGENTS.md is an entry point for agents which allows you to steer them in the right direction so they don't have to do too much grepping. Or requiring you to tell them to things differently.

I created the file by asking Claude based on the work I have been doing with it on this repo what we should put into AGENTS.md. That does not necessarily mean it needs all of that only in AGENTS.md. We can also take this information, put it CONTRIBUTING.md and reference it in AGENTS.md.

benbellick:
@nielspardon To be clear, I'm not against having the file. Its that I believe it should be thin and rely on shared docs e.g.:
```markdown
# Agents.md
Please read @README.md for details on what this project is. Please read @Contributing.md for details on how to contribute.
...
```
And then we can update those sources directly and have one source of truth. If we figure out something that we would want to tell _specifically an agent rather than a person_, then we can choose to put it in only that file (e.g. the suggestion that proto changes should come with docs changes is a great thing to keep in the `AGENTS.md` file). 

This is just my personal view, but it sounds like we are not at odds with one another :)

nielspardon:
> @nielspardon To be clear, I'm not against having the file. Its that I believe it should be thin and rely on shared docs e.g.:

All good. I put it up so we can have exactly this discussion.

nielspardon:
Thanks Ben — agreed, reworked it along these lines (18f287a).

`AGENTS.md` is now a thin index pointing at `README.md`, `CONTRIBUTING.md`, and the policy docs (157 → 80 lines). The shared content moved into `CONTRIBUTING.md` for a single source of truth — the committed-vs-gitignored generated code, and a new *Breaking Changes* section (`buf breaking`, reserving field numbers, the `!` / `BREAKING CHANGE:` footer). `pixi` etc. now lives only there and is referenced, not duplicated.

What's left in `AGENTS.md` is only genuinely agent-specific: the per-library wire/source/semantic compatibility procedure for `proto/` changes, docs-travel-with-a-semantic-change, and keeping PR descriptions high-signal. Also folded in @yongchul's notes.

🤖 The `AGENTS.md` revision was drafted with an AI agent (fittingly).

yongchul (approved):
+1 I think it is good. We can starts with this and iterates on.

Issue #214, opened 2022-05-30
Mismatch between site & protobuf in description of aggregate grouping sets


westonpace:
The site description of an aggregate relation contains:

```
| Direct Output Order  | The list of distinct columns from each grouping set (ordered by their first appearance) followed by the list of measures in declaration order, **followed by an integer describing the associated particular grouping set the value is derived from.** |
```

```
| Grouping Sets    | One or more grouping sets.                                   | Optional, required if no measures.      |
| Per Grouping Set | A list of expression grouping that the aggregation measured should be calculated for. | Optional, defaults to 0.                |
```

There does not appear to be any representation of `Per Grouping Set` in the protobuf.

Also, some questions:

 1. Does the extra integer column describing the grouping set get included in the output even if there are measures?
 2. What in the world is a grouping set?

jacques-n:
The protobuf exposes grouping sets as a list of groupings (where each grouping is a list of discrete expressions). Grouping sets come from the sql feature of the same name.

jacques-n:
The intention was to include the grouping ID in all outputs unless one remaps (whether there are measures or). 

That being said, we should consider whether we state that the column exists only in the case of two or more grouping sets. It's pretty pointless in the case of zero or one grouping set.

westonpace:
Hmm...can we rephrase `The list of distinct columns from each grouping set (ordered by their first appearance)` to something like `The output value from each grouping expression`?

The fact that values are `distinct` seems to be unrelated to the output order.

What is a `grouping ID`?  For example, if I have:

```
SELECT MEAN(cost) FROM products GROUP BY category, vendor
```

And my input data is a single row

cost | category | vendor
--- | --- | ---
20 | toys | target

and my output data is then (I think)

category | vendor | MEAN(cost) | grouping set id
--- | --- | --- | ---
toys | target | 20 | ???

jacques-n:
The "distinct columns..." could be "distinct expressions..." but it does directly influence things. Imagine a query like:

```
SELECT state, city sum(sales) as sales from t GROUP by GROUPING SETS ((state,city), (state))
```

The current definition of the grouping would give an output something like:

|State|City|Sales|Grouping Set Id|
|-|-|-|-|
|California|San Francisco|10|0|
|California|San Jose|1|0|
|California|Calistoga|5|0|
|California| |16|1|

The key is we don't want to output the same expressions that is defined in separate grouping sets multiple times. 

So if we remove grouping sets and just have something like
```
[...]

PR #1136, opened 2026-07-21
refactor(extensions)!: move example YAML files out of the extensions folder

nielspardon:
## Problem

The core `extensions/` folder mixes **official function extensions** (`functions_*.yaml`) with **example/placeholder** files:

- `type_variations.yaml` (`extension:io.substrait:type_variations`) — example type variations, no functions
- `extension_types.yaml` (`extension:io.substrait:extension_types`) — the `point`/`line` user-defined types used as the docs' UDT example

Because they live alongside the real extensions, consumers ingest them inconsistently. This surfaced in [duckdb-substrait-extension#239](https://github.com/substrait-io/duckdb-substrait-extension/pull/239) (issue #214), where the now-removed `unknown.yaml` placeholder was expanded into bogus catch-all function registrations that clobbered valid ones.

> **Note:** the third placeholder, `extensions/unknown.yaml`, was already removed in #1081, which replaced it with the built-in `unbound` type. This PR was rebased accordingly and no longer touches it.

Filtering behavior across the ecosystem for the two remaining files is inconsistent:

| Consumer | Strategy | type_variations | extension_types |
|---|---|---|---|
| DuckDB (post-#239) | allowlist `functions_*.yaml` | skip | skip |
| substrait-python | allowlist `functions*.yaml` | skip | skip |
| substrait-java | hardcoded curated list | excluded | **loaded** |
| substrait-go | load all | loaded | loaded |
| substrait-rs (+packaging) | load all; rs skips by URN | loaded (inert) | **skip** (stale) |

## Change

Move the two files to `site/examples/extensions/` — the existing, schema-validated, docs-embeddable examples folder — so the catalog is exactly the YAML files in `extensions/`.

This resolves open question 2 from the earlier revision of this description in favour of the docs tree. The alternative was shipping examples under `extensions/examples/` so they travel with the extensions tree for downstream testing, on the reasoning that `site/` is not part of the release artifact. That reasoning does not hold up: which files land in an artifact is a packaging decision, and `substrait-packaging` already assembles `substrait-extensions` from four separate spec directories (`extensions/`, `text/`, `tests/cases/`, `dialects/tests/`). Adding `site/examples/` is a fifth entry in scripts that already do exactly this, and it needs no restructuring here. Contorting the specification layout around a downstream glob would be the tail wagging the dog — and it would have cost a mkdocs snippet base-path change, rewrites of every `--8<--` include, and a semantics change to extension discovery in the test helpers, all to save one line per packaging artifact.

So `site/examples/` stays the single home for documentation examples — extension YAML, type YAML and protobuf-textformat plans alike — and nothing else moves.

### Example URNs

Every example extension URN now uses the **`extension:org.example:`** owner, replacing a mix of `extension:example:`, `extension:org.example:` and `extension:io.substrait:`. The [extension docs](https://substrait.io/extensions/) prescribe reverse-domain notation for the owner segment, `example.org` is reserved for documentation by [RFC 2606](https://datatracker.ietf.org/doc/html/rfc2606), and a uniform owner means the URN alone tells you whether something is a catalog entry. `extension:io.substrait:` is now used only by the 16 official files.

One of those renames was not cosmetic. **`lambda_function_example.yaml` declared `urn: extension:io.substrait:functions_list`** — the *same URN* as the official `extensions/functions_list.yaml`. The two files never shared a directory, so this looked harmless, but downstream artifacts bundle the examples and the catalog together, and a duplicate URN breaks any URN-keyed loader there (substrait-rs, for instance, builds a `(urn, raw source)` table). It is now `extension:org.example:lambda_functions`.

`site/examples/README.md` records the convention and states that examples are not catalog entries and carry no compatibility guarantee.

`types/type_parsing.md` also carried an inline copy of the `distance_functions.yaml` example — an abbreviated one, with a single-argument `distance` and no `urn`, so it was never schema-checked. It now includes the real file, per the guidance in `CONTRIBUTING.md` to use external example files rather than inline blocks. That duplication was not hypothetical: the URN rename above had to be applied to the inline copy and the example file separately.

### Incidental fix: five invalid type names in the examples

The examples are linted three ways — `yamllint` repo-wide, `check-jsonschema` against `simple_extensions_schema.yaml`, and `editorconfig-checker` — but all three are syntactic. The schema treats argument and return types as opaque strings, so nothing validated the Substrait content, and four example files carried type names that do not exist:

| File | Was | Now |
|---|---|---|
| `distance_functions.yaml` | `return: f64` | `fp64` |
| `any_type_function.yaml` | `return: int64` | `i64` |
| `any1_type_function.yaml` | `return: int64` | `i64` |
| `deprecation_example.yaml` | `value: int` / `return: int` | `i64` |

None of `f64`, `int64` or `int` are tokens in `SubstraitLexer.g4`. All four files are embedded in the published extension documentation, so `distance_functions.yaml` — the canonical example for referencing user-defined types across extensions — was teaching a return type that does not exist. Closing the gap properly means running the registry's type mapper over the examples in CI; that is left as a follow-up rather than bundled here.

`tests/baseline.json` drops `dependency_count` from 18 to 16, since the two relocated files each contributed a key. The baseline is a floor, so this was already stale on the branch after the #1081 rebase.

## Downstream follow-ups (separate PRs)

**substrait-packaging** — vendor `site/examples/` into the `substrait-extensions` artifact, so SDKs get the examples as parser fixtures without re-vendoring from the docs. Verified against the current scripts:

| Artifact | Change needed |
|---|---|
| **python** | `cp -r` the folder, plus one `package-data` entry. A single `"substrait_extensions" = ["examples/**/*.yaml"]` ships all 18 nested files — confirmed by building a wheel; without it they reach the source tree and are silently dropped. |
| **java** | One `from("$specDir/site/examples") { into(...) }` line. Gradle copies the tree recursively — confirmed by running `:extensions:processResources` with a probe file. |
| **rust** | One `cp -r`. Note `build.rs` walks the tree with `WalkDir`, so the examples must be vendored as data only, and kept out of the `EXTENSIONS` map and the `SIMPLE_EXTENSIONS` URN table. |
| **cpp** | One `cp -r`; `install(DIRECTORY ...)` is already recursive. |

Open question there: whether to vendor `site/examples/proto-textformat/` too. Those `.textproto` plan examples would make useful round-trip fixtures, but they are not extension YAML and sit oddly in an artifact named `substrait-extensions`.

Consumers that vendor the extensions folder themselves:

- **substrait-java**: loads `extension_types.yaml` and defines the `EXTENSION_TYPES` URN constant; loses the `point`/`line` UDTs unless it defines them locally or loads the example under its new URN.
- **substrait-go**: loses `extension_types` types when it re-embeds.
- **substrait-rs**: the URN skip for `extension_types` becomes dead (its cited bug #935 is already fixed).
- **DuckDB (#239)** and **substrait-python**: `functions*` allowlist — no impact.

Because the moved files are examples with no compatibility guarantee, the migration ordering in the breaking-change policy does not gate this on the active libraries. The practical question for each is only whether it wants the `point`/`line` types.

BREAKING CHANGE: `extensions/extension_types.yaml` and `extensions/type_variations.yaml` are no longer part of the extension catalog. They move to `site/examples/extensions/`, and their URNs change from `extension:io.substrait:extension_types` and `extension:io.substrait:type_variations` to `extension:org.example:extension_types` and `extension:org.example:type_variations`. 

Consumers that resolve `extension:io.substrait:extension_types` — the `point`/`line` user-defined types, which substrait-java and substrait-go both register today — will no longer find it in the catalog, and should either define those types locally or load the example explicitly under its new URN. 

The remaining example files keep their location but also move to the `extension:org.example:` owner, including `lambda_function_example.yaml`, which previously duplicated the official `extension:io.substrait:functions_list` URN. 

Nothing in the `extension:org.example:` namespace carries a compatibility guarantee: those files exist to illustrate the extension schema and to serve as parser fixtures, and may change without a deprecation cycle.

- - -
This change is [Reviewable](https://reviewable.io/reviews/substrait-io/substrait/1136)


.gitignore
yongchul, line 1: nit: separate PR?

extensions/extension_types.yaml
yongchul, line 1: Is this a `officially` supported type? I'd say no. why would you restrict lat and long to i32? 😆 So I would put it in an example.

We can ship these example under extensions/example in the official artifact. Or, having a separate example artifact is also fine -- seen both in wild and I really don't care. Since there are only few, we could include it in the artifact. Then we may consider have a separate example if we have enough number of examples maybe?

yongchul (approved):
+1 refactoring makes sense but the projects I involved do not use examples at all. Other PMC may know their dependencies to the examples and have stronger opinions.

benbellick (approved):
LGTM, though if I am being overly precise, I think this is a breaking change 😅 

This is a type that was previously in our catalog and is now being removed. Even if no one had support for it.

extensions/extension_types.yaml
nielspardon: I'm changing substrait-packaging to include the examples in the `substrait-extensions` artifacts so downstream implementations can use those for testing / validation.

PR #1139, opened 2026-07-21
feat(protos): add partition fields to ReadRel.LocalFiles

nielspardon:
Adds support for path-derived partition fields (commonly called partition
columns) on `ReadRel.LocalFiles`. Partitioned datasets commonly encode some
field values in the file path (for example `.../year=2024/month=03/...`) rather
than in the file contents, so a consumer needs those values to reconstruct the
full records.

## Change

A new `PartitionField` message is added to `FileOrFiles`, plus a repeated
`partition_fields` field:

```proto
message PartitionField {
  Expression.ReferenceSegment field = 1;
  Expression.Literal value = 2;
}
repeated PartitionField partition_fields = 16;
```

Each entry supplies the constant value of one field for every record produced
from that item; the remaining fields are read from the file itself.

## Design notes

- **Positional field reference.** The target field is identified by
  `Expression.ReferenceSegment` — ordinal field positions with optional child
  segments for nested structs — rather than by name. This matches how the read
  `projection` (`Expression.MaskExpression`) addresses fields and keeps names
  confined to `NamedStruct`. It also means partition values can target a
  **nested** field, not only a top-level one.
- **Terminology.** Named `PartitionField` / `partition_fields` to match
  Substrait's field-oriented vocabulary (`NamedStruct` describes "struct
  fields", references are `FieldReference` / `StructField`). "Partition columns"
  is noted as the common alias in the docs. Field-oriented naming is also more
  accurate now that a partition value can address a nested field.
- **Typed values.** The value is an `Expression.Literal`, so a partition value
  keeps its real type (e.g. an `int` year) rather than being carried as a
  string.
- **Self-contained.** The information lives entirely on `FileOrFiles`; a field
  referenced here is path-derived, everything else comes from the file. This
  does not touch the core `NamedStruct` type. (Broader field-role concepts such
  as metadata fields and row-index fields are intentionally left out of scope
  for a separate discussion.)
- **Field numbering.** This uses field `16`; field `15` is left for the
  concurrent JSON read options change (#1138) so the two can merge in either
  order without colliding.

## Motivation

Part of the same effort to upstream capabilities from Apache Gluten's vendored
Substrait fork (catalogued in Gluten's
[SubstraitModifications.md](https://github.com/apache/gluten/blob/main/docs/developers/SubstraitModifications.md)),
so that fork can converge back onto the released spec. Gluten carries a per-file
partition-column concept; this PR proposes a spec-idiomatic form (positional,
typed, nested-capable field reference). Proposed on its own merits; the Gluten
alignment is motivating context.

A cross-engine survey of how 20 engines and table formats represent path-derived partition values -- and of the gap this closes in three existing Substrait implementations -- is in #1171.

## Compatibility

Non-breaking — this only adds a new message and a new field. `buf lint`,
`buf format`, `buf breaking` (against `main`), and code generation all pass.

🤖 Generated with AI

- - -
This change is [Reviewable](https://reviewable.io/reviews/substrait-io/substrait/1139)

### DIFF
--- a/proto/substrait/algebra.proto
+++ b/proto/substrait/algebra.proto
@@ -251,6 +251,31 @@ message ReadRel {
         DwrfReadOptions dwrf = 13;
         DelimiterSeparatedTextReadOptions text = 14;
       }
+
+      // A partition field and its constant value for this item.  Partitioned
+      // datasets commonly encode some field values in the file path (e.g.
+      // `.../year=2024/month=03/...`) rather than in the file contents.  Each
+      // entry supplies the constant value of one field for every record
+      // produced from this item; the remaining fields are read from the file
+      // itself.
+      message PartitionField {
+        // Reference to the (possibly nested) field in base_schema whose value
+        // is supplied by this partition rather than stored in the file.  Field
+        // positions are ordinal, with optional child segments for nested
+        // structs -- the same addressing the read `projection`
+        // (Expression.MaskExpression) uses.
+        Expression.ReferenceSegment field = 1;
+        // The constant value of that field for this item.
+        Expression.Literal value = 2;
+      }
+
+      // Partition fields (commonly called partition columns) whose values are
+      // constant for this item and are not stored in the file contents.  This
+      // describes value-based (identity) partitioning; it does not describe
+      // hash, range, or bucket partitioning schemes (a hash/bucket
+      // redistribution is expressed with ExchangeRel).  Optional; empty for
+      // non-partitioned data.
+      repeated PartitionField partition_fields = 16;
     }
   }
 }
--- a/site/docs/relations/logical_relations.md
+++ b/site/docs/relations/logical_relations.md
@@ -78,6 +78,12 @@ Local files are references to files which are locally-accessible. This may inclu
 | --------------------------- | ----------------------------------------------------------------- | -------- |
 | Items                       | An array of Items (path or path glob) associated with the read.   | Required |
 
+##### Partition Fields
+
+Partitioned datasets commonly encode some field values in the file path (for example `.../year=2024/month=03/...`) rather than in the file contents. Each item may carry a set of partition fields (commonly called partition columns), where each pairs a reference to a (possibly nested) field in the read schema with a constant `value` (a literal). The field is addressed by ordinal position, with optional child segments for nested structs — the same addressing used by the read projection. Every record produced from that item takes the given constant value for that field, while the remaining fields are read from the file itself. This is optional and is empty for non-partitioned data.
+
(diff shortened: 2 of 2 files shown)


nielspardon:
Following up on the community sync question about how universal this concept is across engines: I surveyed 20 engines and table formats and wrote it up in #1171, reading the actual type or message that carries a per-file partition value in each system rather than the documentation.

Headline findings:

- **The concept is universal — 20/20.** No engine that reads file-based tables lacks a notion of "this field's value is constant for this file and is not stored in the file."
- **Everything that ships a scan description across a process boundary carries the values explicitly** rather than re-deriving them from the path: Impala's Thrift descriptors (`THdfsPartition.partitionKeyExprs`, commented *"These are Literal expressions"*), Comet's protobuf, DataFusion's protobuf, Dremio's protobuf, and Gluten's fork. 13 of 20 carry, 7 re-derive, and the split falls exactly along that line. That is the class Substrait is in.
- **For Iceberg and Delta the values are not recoverable from the path at all** — the manifest and the transaction log are authoritative — so a consumer handed only a file list cannot reconstruct them by parsing. This is a correctness requirement, not a convenience.
- **On typing, 12 of 20 carry typed values; all 8 string-based ones inherit the Hive metastore `list<string> values` convention**, and each has had to add a null sentinel (`__HIVE_DEFAULT_PARTITION__` / `\N` / `""` / `optional`), a normative per-type serialization spec (Delta's PROTOCOL.md, which still admits two timestamp encodings), or session-dependent casts (ClickHouse honours `cast_string_to_date_time_mode`). Comet and Gluten solve the *identical* problem — Spark JVM planner to native executor over protobuf — and Comet chose typed literals while Gluten chose `{string key, string value}`. This PR proposes the Comet answer.
- **It also fixes a live bug class in three existing Substrait implementations, independent of Gluten:** datafusion-substrait drops partition values in both its physical producer and consumer (a DataFusion → Substrait → DataFusion round trip of a Hive-partitioned Parquet scan returns rows missing their partition columns), substrait-java's Spark module hardcodes `new StructType()` as the partition schema, and Acero cannot express a partitioned read at all despite Arrow's dataset layer supporting it.

Two things the survey argues *against* overselling, which are worth having on the record here:

- The **nested** addressing that `Expression.ReferenceSegment` buys is justified by **Iceberg alone** (identity partitioning on a dotted nested source field). It comes for free from reusing `ReferenceSegment` and costs nothing, but it is not a widely demanded capability.
- The **per-item `(field, value)` pairs** shape differs from Spark/Comet/DataFusion, which declare partition columns once per scan and give each file a positional value list. The defence is that Iceberg partition-spec evolution gives files within one scan different specs (`constantsMap` is computed per task from `task.spec()`), and `FileOrFiles` is already heterogeneous per item since it carries its own `file_format`.

#1171 has the full table with source links, the four representational families, and the per-consumer implementation mapping. Happy to add any engine reviewers think is missing.

PR #1141, opened 2026-07-22
docs(extensions): use block scalar for decimal power description

kadinrabo:
The decimal `power` description was the only multiline quoted description in the YAML tree. Multiline descriptions otherwise use block scalars, so this moves it to the same style.

- - -
This change is [Reviewable](https://reviewable.io/reviews/substrait-io/substrait/1141)

### DIFF
--- a/extensions/functions_arithmetic_decimal.yaml
+++ b/extensions/functions_arithmetic_decimal.yaml
@@ -169,8 +169,9 @@ scalar_functions:
         return: "DECIMAL<38,0>"
   -
     name: "power"
-    description: "Take the power with x as the base and y as exponent.
-    Behavior for complex number result is indicated by option complex_number_result"
+    description: >
+      Take the power with x as the base and y as exponent.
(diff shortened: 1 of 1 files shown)


extensions/functions_arithmetic_decimal.yaml
```diff
@@ -169,8 +169,9 @@ scalar_functions:
         return: "DECIMAL<38,0>"
   -
     name: "power"
-    description: "Take the power with x as the base and y as exponent.
-    Behavior for complex number result is indicated by option complex_number_result"
+    description: >-
```
benbellick: nit: looks like the others use `>` instead of `>-`. 
```suggestion
    description: >
```

benbellick (approved):
thanks! 🚀

benbellick:
The following requires 1 vote per governance:
>Documentation Updates w/ no format/specification changes (i.e formatting, moves)

merging!

PR #1148, opened 2026-07-30
fix: remove ignored nullability markers from MIRROR declarations

### LINKED ISSUE #853: Disallow usage of ? for returnType of functions with MIRROR nullability
When defining the return type of a function with MIRROR nullability, include a `?` after the return type is meaningless, as the nullability is entirely determined based on the nullability of the _inputs_. For example, given a definition like:
```yaml
  - name: "foo"
    impls:
      - args:
          - name: x
            value: i8
        nullability: MIRROR
        return: string?
```
the `?` at the end of `return: string?` is meangingless.

### LINKED ISSUE #1041: Functions in `functions_boolean.yaml` should have inconsistent return type and nullability handling
https://github.com/substrait-io/substrait/pull/949 clarified that nullability handling MIRROR must not have nullable function outputs. Many of the functions in [functions_boolean.yaml](https://github.com/substrait-io/substrait/blob/main/extensions/functions_boolean.yaml) break this rule. 

E.g.
https://github.com/substrait-io/substrait/blob/ff70d18c25fe3e078d40bb3cff6074cf299d0b50/extensions/functions_boolean.yaml#L106-L115

We probably should alter these functions to have output `BOOLEAN` rather than `BOOLEAN?`

nielspardon:
Under `MIRROR` nullability the output nullability is derived from the arguments, so the spec requires that neither argument nor return types carry a `?` marker, since it is ignored.

`text/simple_extensions_schema.yaml`:

> MIRROR: Output is nullable if any argument is nullable; argument **and return** types must not include nullability markers as they will be ignored for output nullability derivation.

Seven impls in the standard catalog violated this. I enumerated them by loading every `extensions/functions_*.yaml` through a parser rather than grepping, so this is the complete set:

| file | function | markers removed |
| --- | --- | --- |
| `functions_boolean.yaml` | `or` | 1 argument, return |
| `functions_boolean.yaml` | `and` | 1 argument, return |
| `functions_boolean.yaml` | `and_not` | 2 arguments, return |
| `functions_boolean.yaml` | `xor` | 2 arguments, return |
| `functions_boolean.yaml` | `not` | 1 argument, return |
| `functions_rounding_decimal.yaml` | `ceil` | return (`decimal?<precision, 0>`) |
| `functions_rounding_decimal.yaml` | `floor` | return (`decimal?<precision, 0>`) |

`bool_and`, `bool_or` and `round` are left as-is: they declare `DECLARED_OUTPUT`, where the return marker is authoritative.

## This is not a breaking change

Derived behavior is unchanged. With the markers ignored per spec, `or(boolean, boolean) -> boolean` and `or(boolean?, boolean?) -> boolean?`, which is what Kleene logic wants and what the existing test cases in `tests/cases/boolean/` already assert. No plan that was legal before becomes illegal.

It does matter for implementations that validate a plan's declared output type against the declaration, though: a producer that reads `return: boolean?` literally will declare `boolean?` for `or(boolean, boolean)` and be rejected by a conforming validator. That inconsistency is what this fixes.

## Regression guard

Also adds `validate_nullability_markers` plus a test over the shipped catalog, so the markers cannot be reintroduced. The check:

- skips `DISCRETE`, where markers are meaningful on both sides;
- only inspects outermost nullability, so nested markers like `func<any1 -> boolean?>` and `list<i32?>` still pass;
- reads the final line of a derivation expression, so a `?` ternary in an intermediate line (`scale = init_prec > 38 ? scale_after_borrow : init_scale`) is not mistaken for a marker.

Verified it fails with a clear message when either marker is reintroduced.

Fixes #1041
Fixes #853

🤖 Generated with AI

- - -
This change is [Reviewable](https://reviewable.io/reviews/substrait-io/substrait/1148)

### DIFF
--- a/tests/coverage/extensions.py
+++ b/tests/coverage/extensions.py
@@ -21,6 +21,97 @@ def substrait_type_str(rule_num):
     return FuncTestCaseLexer.symbolicNames[rule_num].lower()
 
 
+def find_extension_files(dir_path: str):
+    """Paths of the function extension YAML files under dir_path, sorted."""
+    extensions = []
+    for root, _dirs, files in os.walk(dir_path):
+        for file in files:
+            if file.endswith(".yaml") and file.startswith("functions_"):
+                extensions.append(os.path.join(root, file))
+
+    extensions.sort()
+    return extensions
+
+
+FUNCTION_SECTIONS = ("scalar_functions", "aggregate_functions", "window_functions")
+
+
+def declared_type(type_str):
+    """The concrete type of a declaration, dropping any derivation expression.
+
+    A ``return`` may be a multi-line derivation expression whose leading lines
+    assign intermediate values.  Those lines can contain a ``?`` that is a
+    ternary operator rather than a nullability marker (e.g.
+    ``scale = init_prec > 38 ? scale_after_borrow : init_scale``), so only the
+    final line names the declared type.
+    """
+    return str(type_str).strip().split("\n")[-1].strip()
+
+
+def validate_impl_nullability_markers(impl, loc):
+    """Validate that one impl carries no meaningless nullability markers.
+
+    Under ``MIRROR`` the output nullability is derived from the arguments, so a
+    ``?`` on an argument or on the return type is ignored.  Under
+    ``DECLARED_OUTPUT`` the return marker is authoritative, but the outermost
+    argument nullability is still stripped before binding.  A marker that is
+    ignored is misleading, so it must not appear.  ``DISCRETE`` is exempt: there
+    markers are meaningful on both sides.
+
+    Only the outermost nullability is checked; nested markers are meaningful
+    (e.g. ``func<any1 -> boolean?>``, ``list<i32?>``).
+
+    Returns a list of error strings (empty if everything is valid).
+    """
+    nullability = impl.get("nullability", "MIRROR")  # MIRROR is the spec default
+    if nullability == "DISCRETE":
+        return []
+
+    errors = []
+    for arg in impl.get("args") or []:
+        value = arg.get("value")
+        if value is None:
+            continue  # option (enum) argument, not a value
+        if type_str_is_outer_nullable(declared_type(value)):
+            errors.append(
+                f"{loc}: argument '{arg.get('name', value)}' is declared '{value}' "
+                f"under {nullability} nullability, but the outermost nullability of "
+                f"an argument is stripped before binding; remove the '?'"
+            )
+
+    return_type = impl.get("return")
+    if (
+        nullability == "MIRROR"
+        and return_type is not None
+        and type_str_is_outer_nullable(declared_type(return_type))
+    ):
+        errors.append(
+            f"{loc}: return type '{declared_type(return_type)}' carries a nullability "
+            f"marker under MIRROR nullability, where output nullability is derived "
+            f"from the arguments; remove the '?'"
+        )
+    return errors
+
+
+def validate_nullability_markers(dir_path: str):
+    """Validate every function declaration under dir_path.
+
+    See :func:`validate_impl_nullability_markers` for the rules applied.
+
+    Returns a list of error strings (empty if everything is valid).
+    """
+    errors = []
+    for path in find_extension_files(dir_path):
+        with open(path, "r") as fh:
+            data = yaml.load(fh, Loader=yaml.FullLoader)
+        for section in FUNCTION_SECTIONS:
+            for func in data.get(section) or []:
+                loc = f"{os.path.basename(path)}: {func['name']}"
+                for impl in func.get("impls") or []:
+                    errors.extend(validate_impl_nullability_markers(impl, loc))
+    return errors
+
+
 def build_type_to_short_type():
     rule_map = {
         FuncTestCaseLexer.I8: FuncTestCaseLexer.I8,
@@ -160,14 +251,7 @@ class Extension:
 
     @staticmethod
     def read_substrait_extensions(dir_path: str):
-        # read files from extensions directory
-        extensions = []
-        for root, _dirs, files in os.walk(dir_path):
-            for file in files:
-                if file.endswith(".yaml") and file.startswith("functions_"):
-                    extensions.append(os.path.join(root, file))
-
-        extensions.sort()
+        extensions = find_extension_files(dir_path)
 
         scalar_functions = {}
         aggregate_functions = {}
--- a/tests/coverage/test_coverage.py
+++ b/tests/coverage/test_coverage.py
@@ -5,11 +5,7 @@ import pytest
 from antlr4 import InputStream
 from tests.coverage.case_file_parser import parse_stream, parse_one_file
 from tests.coverage.coverage import validate_nullability
-from tests.coverage.extensions import (
-    Extension,
-    validate_impl_nullability_markers,
-    validate_nullability_markers,
-)
+from tests.coverage.extensions import Extension, validate_impl_nullability_markers
 from tests.coverage.visitor import ParseError
 from tests.coverage.nodes import CaseLiteral, FuncCallArg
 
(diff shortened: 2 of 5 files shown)


nielspardon:
## Breaking-change policy: no code impact in any active library

Per the [breaking change policy](https://github.com/substrait-io/substrait/blob/main/site/docs/spec/breaking_change_policy.md), a breaking change needs a migration strategy implemented in all [active libraries](https://github.com/substrait-io/substrait/blob/main/site/docs/community/active_libraries.md) before it lands. I audited all four. **None of them reads the declared marker under `MIRROR`, so none needs a migration.**

| library | how MIRROR output nullability is computed | reads declared `?`? |
| --- | --- | --- |
| `substrait-go` | `extensions/variants.go:155` — `outType.WithNullability(Required/Nullable)` from `allNonNull`, overriding the declared type | No. Arguments use `MatchWithoutNullability` for MIRROR/DECLARED_OUTPUT (`variants.go:230`) |
| `substrait-java` | isthmus `SimpleExtensionToSqlOperator.determineNullability` → `opBinding.collectOperandTypes().anyMatch(RelDataType::isNullable)` | No. Core `SimpleExtension.validateOutputType` is a disabled no-op (commented out as "too restrictive in the case of nullability conversion"), and `resolveType` has no callers in core |
| `substrait-python` | `extension_registry/function_entry.py` — after `evaluate(...)`, overwrites output nullability to `NULLABLE`/`REQUIRED` based on the call signature | No. `check_nullability` is set true only for `DISCRETE` |
| `substrait-rs` | No derivation step; parse/validate layer only | Stores `Impl::return_type`, but nothing consumes it for derivation |

Supporting checks:

- **Compound signature keys are unchanged.** Markers are stripped when building them, so `or:bool` stays `or:bool` — asserted directly in substrait-go's `extensions/extension_mgr_test.go`.
- **No library fixture encodes the affected declarations.** The only `boolean?` occurrences across the four are type-parser and derivation-expression unit tests on literal strings (substrait-go `types/parser/parse_test.go`, substrait-python `tests/test_derivation_expression.py`).
- **It propagates as data, not code.** substrait-java vendors the spec as a submodule. The Rust path goes through the `substrait-extensions` crate, which copies the extension YAMLs verbatim and exposes them as `include_str!` constants; its generated Rust types come from the `text/*.yaml` schemas, which this PR does not touch, so there is no generated-code diff.

All four were inspected at commits level with their respective upstream `main`.

### One observable difference, for the record

`substrait-rs`'s parsed `Impl::return_type.nullable` does flip to `false` for these 7 impls. Because substrait-rs has no MIRROR derivation, a consumer reading `return_type` directly as a call's output type would move from "always nullable" (wrong when all arguments are non-nullable) to "always required" (wrong when an argument is nullable). Neither is spec-correct without implementing the derivation, so this is not a regression the policy would act on — but it is the one place where the parsed value changes, and I would rather flag it here than have it turn up later.

🤖 Generated with AI

tests/coverage/test_coverage.py
```diff
@@ -946,3 +950,142 @@ def test_mirror_options_skip_false_positive(self):
...
+            },
+            "functions_test.yaml: f",
+        )
+        assert errors == []
+
+    def test_all_standard_extensions_are_clean(self):
+        """The shipped extension catalog must have no ignored markers."""
+        script_dir = os.path.dirname(os.path.abspath(__file__))
+        extensions_path = os.path.join(script_dir, "../../extensions")
+        assert validate_nullability_markers(extensions_path) == []
```
benbellick:
```suggestion
```

I think this test duplicates this one below:
https://github.com/substrait-io/substrait/blob/f2a4982026bcca2de038ab86fa31bd34aad9e8de/tests/test_extensions.py#L61-L71

The other file is a better spot for it IMO, and it would be better to just have that one.

tests/coverage/extensions.py
```diff
@@ -21,6 +21,97 @@ def substrait_type_str(rule_num):
...
+def find_extension_files(dir_path: str):
+    """Paths of the function extension YAML files under dir_path, sorted."""
+    extensions = []
+    for root, _dirs, files in os.walk(dir_path):
+        for file in files:
+            if file.endswith(".yaml") and file.startswith("functions_"):
+                extensions.append(os.path.join(root, file))
+
+    extensions.sort()
+    return extensions
```
benbellick, line 33: Could this function be replaced by this one?
https://github.com/substrait-io/substrait/blob/f2a4982026bcca2de038ab86fa31bd34aad9e8de/tests/type/test_type_grammar.py#L70-L74

or could we centralize to only have one? I think the one I linked from `test_type_grammar.rs` is fine (bc it also checks the examples) but might benefit from moving to a more shared location.

benbellick: If it turns out unifying both makes a larger diff, we can just do it in another PR, just wanted to call it out.

tests/coverage/test_coverage.py
```diff
@@ -946,3 +950,142 @@ def test_mirror_options_skip_false_positive(self):
...
+        assert "MIRROR" in errors[0]
+        assert "boolean?" in errors[0]
+
+    def test_mirror_rejects_argument_marker(self):
+        errors = validate_impl_nullability_markers(
+            {"args": [{"value": "boolean?", "name": "a"}], "return": "boolean"},
+            "functions_boolean.yaml: not",
+        )
+        assert len(errors) == 1
+        assert "argument 'a'" in errors[0]
```
benbellick, line 979:
```suggestion
```

This test doesn't seem to test what it is named, and is basically identical to `test_nullability_defaults_to_mirror`

benbellick (approved):
Some comments but otherwise seems reasonable, thanks

tests/coverage/test_coverage.py
nielspardon: Agreed, removed in a69e403. `tests/test_extensions.py` is the better home, and its assertion also reports which declarations are offending rather than just failing, so nothing is lost.

I re-verified the remaining check still catches a regression — reintroducing the marker on `or` fails with:

```
AssertionError: 1 ignored nullability marker(s) found:
  functions_boolean.yaml: or: return type 'boolean?' carries a nullability marker under MIRROR
  nullability, where output nullability is derived from the arguments; remove the '?'
```

tests/coverage/test_coverage.py
nielspardon: Fixed the naming problem in a69e403, though I read the duplication slightly differently — flagging it in case you disagree.

The three MIRROR tests all relied on the spec default rather than declaring `nullability: MIRROR`, so their names claimed to test MIRROR when they were really testing the default. They now set it explicitly.

On the duplication: this test asserts `"argument 'a'" in errors[0]`, so it covers the argument-marker branch, whereas `test_nullability_defaults_to_mirror` covers the return-marker branch — different code paths. The genuine near-duplicate was `test_mirror_rejects_return_marker`, one above: same shape, same assertion, both leaning on the default. Making MIRROR explicit there resolves it, and `test_nullability_defaults_to_mirror` is now the only test exercising the defaulting behavior.

Happy to collapse further if you'd still prefer fewer cases here.

tests/coverage/extensions.py
nielspardon: Taking you up on the offer to defer this one — the two helpers aren't quite interchangeable, so unifying them is a bigger change than it looks:

- `extension_yaml_files()` yields every `extensions/*.yaml` **plus** `site/examples/**/*.yaml`, as `Path` objects.
- `find_extension_files()` walks a caller-supplied directory recursively, filters to `functions_*.yaml`, and returns `str` paths — because it also backs `read_substrait_extensions()`, whose contract is a directory argument.

Widening the marker check to `site/examples` is a behavioral decision I'd rather not smuggle into this fix, since a doc example may legitimately show a marker. And centralizing properly means picking a home above both `tests/coverage/` and `tests/type/` and editing `test_type_grammar.py`, which is otherwise untouched here.

For what it's worth, `find_extension_files()` isn't net-new duplication — the walk was already inline in `read_substrait_extensions()`, and I hoisted it so both callers share one copy.

Want me to open a follow-up issue for the centralization?

nielspardon: Filed as #1161, with the concrete differences between the two helpers and the three design questions that need settling first (chiefly: whether `site/examples/**` should be in scope for the marker check, since that one is a behavioral call rather than a refactor).

Tagged `good first issue` — it is self-contained and test-only. Also noted there that the traversal is duplicated too, not just the file discovery: `iter_type_expressions()` and `validate_nullability_markers()` walk the same document shape but need different things out of it.

nielspardon:
Merged `upstream/main` (through a902858) to bring this up to date. Flagging it since the approval here isn't dismissed on push, and the resolution has a small behavioral consequence worth a second look.

#953 widened the file filter in `read_substrait_extensions` from `.yaml` + `startswith("functions_")` to just `.yaml`, because the extension it adds is named `unsigned_integers.yaml` rather than `functions_*.yaml`. That is the exact line this PR had extracted into `find_extension_files()`, so it conflicted.

I kept the extraction and moved the widened filter into the shared helper. Taking either side verbatim would have been wrong: keeping the prefix filter would have silently undone #953 and dropped `unsigned_integers.yaml` from the coverage registry, and reverting the extraction would have lost the shared helper.

The consequence is that the nullability-marker check now covers all 18 extension YAMLs rather than the 15 `functions_*` ones — it picks up `extension_types.yaml`, `type_variations.yaml` and `unsigned_integers.yaml`. All three are clean, so nothing new is being asserted, but the check's scope is slightly wider than what you reviewed. `unsigned_integers.yaml` does declare `u!u64?` returns; those are under `nullability: DECLARED_OUTPUT`, where the marker is authoritative and correctly left alone.

I've also corrected the discovery-helper comparison in #1161, since the prefix filter it described is now gone.

PR #1158, opened 2026-07-31
docs: clarify dependency alias scope

### LINKED ISSUE #1152: docs: dependency aliases cannot actually prefix type variation names
The [simple extensions docs](https://substrait.io/extensions/#referencing-user-defined-types) say that a dependency alias can be used:

> This alias can then be used as a `.`-separated namespace prefix wherever a type class or type variation name is expected.

The "or type variation name" part of that sentence has no place where it can be applied.

A type variation name is only ever written in two places:

1. As a *declaration*, in `type_variations[].name` of the extension YAML that defines it. An alias prefix is meaningless there.
2. As a *reference from a plan*, via extension URN + name in `SimpleExtensionDeclaration`, which does not use YAML dependency aliases at all.

There is no production in `grammar/SubstraitType.g4` that names a type variation, so an alias-prefixed variation name cannot appear in any type expression either. The only cross-extension reference a type variation can make is its `parent`, and that is a type class reference, which the sentence already covers.

## Suggested fix

Narrow the sentence to type classes, e.g.:

> A YAML file can also reference type classes defined in another YAML file. [...] This alias can then be used as a `.`-separated namespace prefix wherever a type class name is expected, including the `parent` of a type variation.

The preceding sentence in that paragraph ("reference types and type variations defined in another YAML file") needs the same trim, as does the comment on the `dependencies` property in `text/simple_extensions_schema.yaml`, which reads "For reusing type classes and type variations from other extension files."

Alternatively, if prefixing variation names somewhere *is* intended, then the gap is in the grammar rather than the prose and this should become an enhancement instead.

## Affected files

- `site/docs/extensions/index.md`
- `text/simple_extensions_schema.yaml`

Spotted while verifying whether #496 was fully addressed by #265.

lntutor:
Clarify that dependency aliases qualify type class references, including variation parents, but not type variation names. Keep the website prose and schema comments consistent.

Fixes #1152

## Validation

- `uv run --group dev check-jsonschema --schemafile text/simple_extensions_schema.yaml site/examples/**/*.yaml`
- `uv run --group dev check-jsonschema --schemafile text/simple_extensions_schema.yaml extensions/*.yaml`
- `uv run --group dev yamllint text/simple_extensions_schema.yaml`
- `uv run --project .. --group dev mkdocs build --strict` (from `site/`)

🤖 Generated with AI

- - -
This change is [Reviewable](https://reviewable.io/reviews/substrait-io/substrait/1158)

### DIFF
--- a/site/docs/extensions/index.md
+++ b/site/docs/extensions/index.md
@@ -40,7 +40,7 @@ Within YAML extension files, [user-defined types](../types/type_classes.md#user-
 
 [Built-in types](../types/type_classes.md#built-in-types) like `i32`, `string`, or `list` (as in `list<fp64>`) do not use any prefix.
 
-A YAML file can also reference types and type variations defined in another YAML file. To do this, it must declare the extension it depends on using a key-value pair in the `dependencies` key, where the value is the extension URN, and the key is a valid identifier that can then be used as an identifier-safe alias for the extension URN. This alias can then be used as a `.`-separated namespace prefix wherever a type class or type variation name is expected. Note that user-defined types still require the `u!` prefix when referenced via namespace aliases (e.g., `ext.u!point`).
+A YAML file can also reference type classes defined in another YAML file. To do this, it must declare the extension it depends on using a key-value pair in the `dependencies` key, where the value is the extension URN, and the key is a valid identifier that can then be used as an identifier-safe alias for the extension URN. This alias can then be used as a `.`-separated namespace prefix wherever a type class name is expected, including the `parent` of a type variation. Note that user-defined types still require the `u!` prefix when referenced via namespace aliases (e.g., `ext.u!point`).
 
 !!! note "Grammar"
     The grammar for referencing [user-defined types](../types/type_classes.md#user-defined-types) is (in [ABNF](https://datatracker.ietf.org/doc/html/rfc5234)):
--- a/text/simple_extensions_schema.yaml
+++ b/text/simple_extensions_schema.yaml
@@ -8,10 +8,10 @@ properties:
   urn:
     type: string
   dependencies:
-    # For reusing type classes and type variations from other extension files.
+    # For reusing type classes from other extension files.
     # The keys are namespace identifiers that you can then use as dot-separated
-    # prefix for type class and type variation names in functions and the base
-    # type class for variations. The values must be extension URNs, following
+    # prefixes for type class names in functions and the parent type class for
+    # variations. The values must be extension URNs, following
     # the same format and conventions as those used in the proto plans.
     type: object
     patternProperties:


lntutor:
Current head `79134088768f133854e19d916f82afe7affebed9` is mergeable, the PR title checks and CLA are green, and there are no unresolved review threads on the current PR state. This looks ready for maintainer review when convenient.

site/docs/extensions/index.md
```diff
@@ -40,7 +40,7 @@ Within YAML extension files, [user-defined types](../types/type_classes.md#user-
...
 
-A YAML file can also reference types and type variations defined in another YAML file. To do this, it must declare the extension it depends on using a key-value pair in the `dependencies` key, where the value is the extension URN, and the key is a valid identifier that can then be used as an identifier-safe alias for the extension URN. This alias can then be used as a `.`-separated namespace prefix wherever a type class or type variation name is expected. Note that user-defined types still require the `u!` prefix when referenced via namespace aliases (e.g., `ext.u!point`).
+A YAML file can also reference type classes defined in another YAML file. To do this, it must declare the extension it depends on using a key-value pair in the `dependencies` key, where the value is the extension URN, and the key is a valid identifier that can then be used as an identifier-safe alias for the extension URN. This alias can then be used as a `.`-separated namespace prefix wherever a type class name is expected, including the `parent` of a type variation. Note that user-defined types still require the `u!` prefix when referenced via namespace aliases (e.g., `ext.u!point`).
```
benbellick, line 43:
```suggestion
A YAML file can also reference types defined in another YAML file. To do this, it must declare the extension it depends on using a key-value pair in the `dependencies` key, where the value is the extension URN, and the key is a valid identifier that can then be used as an identifier-safe alias for the extension URN. This alias can then be used as a `.`-separated namespace prefix wherever a type name is expected. Note that user-defined types still require the `u!` prefix when referenced via namespace aliases (e.g., `ext.u!point`).
```
[If I am reading the docs correctly](https://substrait.io/types/type_classes/), "class" means the set of values the type may assume, which is (subtly) different from the type itself. I realize this paragraph previously used "type class" but I think "type" is better here :)

text/simple_extensions_schema.yaml
```diff
@@ -8,10 +8,10 @@ properties:
   urn:
     type: string
   dependencies:
-    # For reusing type classes and type variations from other extension files.
+    # For reusing type classes from other extension files.
```
benbellick, line 11:
```suggestion
    # For reusing types from other extension files.
```

text/simple_extensions_schema.yaml
```diff
@@ -8,10 +8,10 @@ properties:
...
     # The keys are namespace identifiers that you can then use as dot-separated
-    # prefix for type class and type variation names in functions and the base
-    # type class for variations. The values must be extension URNs, following
+    # prefixes for type class names in functions and the parent type class for
+    # variations. The values must be extension URNs, following
```
benbellick, line 14:
```suggestion
    # prefixes for type names in functions. 
    # The values must be extension URNs, following
```

benbellick (approved):
Thanks for the contribution! Some minor cleanup on the usage of type class (I am curious to see if other PMCs agree with my correction), and also a suggestion to drop the mention of "parents" of type variations. 


Thanks!

yongchul (changes requested):
Looks good. Please adapt Ben's suggestions `type class` -> `type` or `type name`

PR #1159, opened 2026-07-31
docs: describe interval arithmetic implementations

### LINKED ISSUE #1157: Add impl-level descriptions to the interval arithmetic functions in functions_datetime.yaml
## Summary

`add` and `subtract` in [`extensions/functions_datetime.yaml`](https://github.com/substrait-io/substrait/blob/main/extensions/functions_datetime.yaml) each carry a single function-level `description` that is applied to six or seven impls whose semantics differ materially. Several of those differences are not recoverable from the signature alone. Since `impls[].description` is already supported by the schema, these should be documented per impl.

## The schema already allows this

`description` is a permitted property on each entry of `impls` for scalar, aggregate and window functions ([`text/simple_extensions_schema.yaml`](https://github.com/substrait-io/substrait/blob/main/text/simple_extensions_schema.yaml), in the `impls.items.properties` blocks).

There is existing precedent in `count` ([`functions_aggregate_generic.yaml#L5-L26`](https://github.com/substrait-io/substrait/blob/main/extensions/functions_aggregate_generic.yaml#L5-L26)), which uses the function-level description as an umbrella and impl-level descriptions to disambiguate:

```yaml
  - name: "count"
    description: Count the number of values or records.
    impls:
      - description: Count the non-null values in a set of values.
        ...
      - description: Count the number of records (not field-referenced).
        ...
```

## What is currently undocumented

`add`'s description is *"Add an interval to a date/time type"* plus three lines of IANA timezone-string boilerplate. Applied uniformly across the impls, this hides the following:

1. **`interval_year` and `interval_day` impls are different kinds of operation.** Per the type docs, `interval_day` is *"only the total elapsed time is significant"* — a duration, added directly to the value. `interval_year` is *"only the total number of months is significant"* — a calendar field operation, including end-of-month clamping (`2026-01-31 + 1 month` = `2026-02-28`). Nothing in the YAML says the clamping happens, or that it makes `add`/`subtract` non-inverse for month intervals.

2. **The 3-argument `precision_timestamp_tz` + `interval_year` impl performs a local-civil round trip** — convert the UTC instant to civil time in the given zone, do the month arithmetic there, convert back. That is the entire reason the argument exists, and it is not discoverable from the signature. It also means the result depends on the zone in two distinct ways (which local date the instant falls on, and whether the offset changes between source and target).

3. **The timezone boilerplate applies to exactly one impl out of six**, but sits at function level where it reads as applying to all of them. The `interval_day` impls deliberately take no timezone, because elapsed-time arithmetic needs no calendar — a point raised and settled in the #337 review, and currently recorded nowhere in the file.

4. **`add(date, interval_day) -> precision_timestamp` widens the type while `add(date, interval_year) -> date` does not.** The reason is sound (an `interval_day` carries sub-day components, so the result may not land on a date boundary; an `interval_year` cannot) but unstated, so the asymmetry reads as arbitrary. This was the substance of #576, and a per-impl note would stop it being re-litigated.

`subtract` has the same description and the same gaps.

## Illustrative shape

```yaml
      - description: >-
          Add a year/month interval to a UTC instant by converting it to civil time in the
          given timezone, adding the months in that calendar (clamping the day of month to
          the last valid day of the target month), and converting the result back to UTC.
          The timezone is not retained in the result.
        args:
          - name: x
            value: precision_timestamp_tz<P>
          - name: y
            value: interval_year
          - name: timezone
            description: Timezone string from IANA tzdb.
            value: string
        return: precision_timestamp_tz<P>
      - description: >-
          Add an elapsed-time interval to a UTC instant. No timezone is required, because
          interval_day denotes total elapsed time rather than calendar days.
        args:
          - name: x
            value: precision_timestamp_tz<P>
          - name: y
            value: interval_day<P>
        return: precision_timestamp_tz<P>
```

Wording is illustrative; the point is that each impl states the semantics its signature does not convey.

## Scope

Documentation-only and non-breaking: no signatures, return types, argument names or compound names change, so no plan or engine is affected. Reasonable candidate for a first contribution.

`add_intervals` and `multiply` are lower priority — their impls are homogeneous, so the single function-level description is adequate there.

These descriptions render on substrait.io with no additional tooling work. [`site/mkdocs.yml`](https://github.com/substrait-io/substrait/blob/main/site/mkdocs.yml) runs [`docs/extensions/generate_function_docs.py`](https://github.com/substrait-io/substrait/blob/main/site/docs/extensions/generate_function_docs.py) through the `gen-files` plugin at build time, and that script emits each impl's `description` as an italic line directly beneath the signature it belongs to. The `count` descriptions quoted above are live today at <https://substrait.io/extensions/functions_aggregate_generic/>.

Because the description renders inline as a single italic line under the signature, one or two sentences per impl reads best; longer prose is better placed in the function-level description or the spec.

## Related

- #1153 — arity gap for `add`/`subtract` of `interval_year` on `precision_timestamp_tz`; the semantics in item 2 above are the justification for that argument existing.
- #1156 — `precision_timestamp_tz` is documented as "timezone-aware" but carries no timezone.
- #576 — the return-type asymmetry in item 4, now closed.

mgalore:
docs: add per-implementation interval arithmetic descriptions

Closes #1157.

Adds concise implementation-level descriptions for every `add` and `subtract` interval signature. The descriptions distinguish calendar month arithmetic from elapsed-time arithmetic, explain timezone-local handling where applicable, and document date-to-timestamp widening for `interval_day`.

Validation:
- `pytest .` (137 passed)
- `yamllint .`
- `check-jsonschema --schemafile text/simple_extensions_schema.yaml extensions/*.yaml`
- `ruff check .` and `ruff format --diff .`
- `ec`
- `mkdocs build --strict`
- `git diff --check`

- - -
This change is [Reviewable](https://reviewable.io/reviews/substrait-io/substrait/1159)

### DIFF
--- a/extensions/functions_datetime.yaml
+++ b/extensions/functions_datetime.yaml
@@ -182,13 +182,19 @@ scalar_functions:
       Examples: "Pacific/Marquesas", "Etc/GMT+1".
       If timezone is invalid an error is thrown.
     impls:
-      - args:
+      - description: >-
+          Add a year/month interval using calendar arithmetic, clamping the day of month to
+          the last valid day of the target month when necessary.
+        args:
           - name: x
             value: precision_timestamp<P>
           - name: y
             value: interval_year
         return: precision_timestamp<P>
-      - args:
+      - description: >-
+          Add a year/month interval after converting the UTC instant to civil time in the
+          supplied timezone, then convert the result back to UTC. The timezone is not retained.
+        args:
           - name: x
             value: precision_timestamp_tz<P>
           - name: y
@@ -197,25 +203,33 @@ scalar_functions:
             description: Timezone string from IANA tzdb.
             value: string
         return: precision_timestamp_tz<P>
-      - args:
+      - description: >-
+          Add a year/month interval using calendar arithmetic, clamping the day of month to
+          the last valid day of the target month when necessary.
+        args:
           - name: x
             value: date
           - name: y
             value: interval_year
         return: date
-      - args:
+      - description: Add an elapsed-time interval.
+        args:
           - name: x
             value: precision_timestamp<P>
           - name: y
             value: interval_day<P>
         return: precision_timestamp<P>
-      - args:
+      - description: Add an elapsed-time interval. No timezone is required because interval_day denotes elapsed time.
+        args:
           - name: x
             value: precision_timestamp_tz<P>
           - name: y
             value: interval_day<P>
         return: precision_timestamp_tz<P>
-      - args:
+      - description: >-
+          Add an elapsed-time interval. The result is a timestamp because interval_day can
+          contain sub-day components.
+        args:
           - name: x
             value: date
           - name: y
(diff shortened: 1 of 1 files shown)


yongchul (approved):

nielspardon:
Hi @mgalore, if you want us to include your PR changes, then you will have to sign the contributor license agreement (CLA) by logging into CLA assistant with your Github account.

mgalore:
done

PR #1160, opened 2026-08-01
feat(extensions)!: deprecate subtract(precision_timestamp_tz, interval_year) overload

### LINKED ISSUE #1153: Inconsistent arity for add/subtract of interval_year with precision_timestamp_tz
## Summary

`add` and `subtract` disagree on the arity of the `precision_timestamp_tz` + `interval_year` variant. `subtract` offers both a 2-argument form and a 3-argument form taking a `timezone`; `add` offers only the 3-argument form.

| function | 2-arg `(precision_timestamp_tz<P>, interval_year)` | 3-arg `(precision_timestamp_tz<P>, interval_year, string)` |
|---|---|---|
| `add` | ❌ missing | ✅ [functions_datetime.yaml#L191-L199](https://github.com/substrait-io/substrait/blob/main/extensions/functions_datetime.yaml#L191-L199) |
| `subtract` | ✅ [functions_datetime.yaml#L307-L312](https://github.com/substrait-io/substrait/blob/main/extensions/functions_datetime.yaml#L307-L312) | ✅ [functions_datetime.yaml#L313-L321](https://github.com/substrait-io/substrait/blob/main/extensions/functions_datetime.yaml#L313-L321) |

In compound-name terms: `subtract:ptstz_iyear` exists but `add:ptstz_iyear` does not, while `add:ptstz_iyear_str` and `subtract:ptstz_iyear_str` both do.

This is separate from the return-type inconsistency tracked in #576, which is resolved.

## Proposed fix

Remove the 2-argument `subtract(precision_timestamp_tz<P>, interval_year)` impl, leaving the 3-argument form as the only way to add or subtract a year/month interval on a `precision_timestamp_tz`. This is a breaking change and needs PMC sign-off; a deprecation cycle first would soften it.

## Origin

The divergence looks like an oversight in #337 ("fix: correct return of temporal add and subtract and add timezone parameter"), whose commit message states:

> BREAKING CHANGE: (add/subtract)ing an interval to a timestamp_tz now requires a time zone and returns a timestamp_tz

For `add`, that PR **converted** the existing 2-argument impl in place by appending the `timezone` argument. For `subtract`, it **added** the 3-argument impl alongside the existing 2-argument one instead of converting it. So the leftover 2-argument `subtract` contradicts the stated intent of the PR that introduced the timezone argument.

## Why the timezone argument is load-bearing for `interval_year`

The spec gives the two interval types different kinds of semantics:

- `interval_day<P>` — *"only the total elapsed time is significant, i.e. `1d 0s` is considered equal to `0d 86400s`"*. A duration. Adding it to an instant is `t + Δ` on the underlying int64, with no calendar involved.
- `interval_year` — *"only the total number of months is significant"*. A count of months, and a month is not a quantity of time (31d, 28d, 30d, ...). There is no `Δ` for which "+1 month" is `t + Δ`.

Month arithmetic is therefore only definable as a *field* operation on civil components: decompose into `(Y, M, D, h, m, s)`, add to `M` with carry into `Y`, clamp `D` to the last valid day of the target month, recompose. The clamping is intrinsic — it is why months do not form a group (`2026-01-31 +1mo -> 2026-02-28 -1mo -> 2026-01-28`).

That decompose/recompose step is exactly where the timezone is required, because the mapping between an instant and a civil datetime *is* the timezone. The 3-argument form means:

1. UTC instant → local civil time in `tz` (equivalent to `local_timestamp`)
2. month field arithmetic in that civil calendar, with clamping
3. local civil result → UTC instant (equivalent to `assume_timezone`)

The zone changes the answer in two independent ways. Which civil date the instant lands on, affecting what gets clamped:

```
instant 2026-03-31T02:00:00Z
  in UTC -> civil Mar 31 -> +1mo -> Apr 31 clamped to Apr 30 -> 2026-04-30T02:00:00Z
  in NY  -> civil Mar 30 -> +1mo -> Apr 30, no clamp         -> 2026-05-01T02:00:00Z
```

And the offset at the target instant differing from the offset at the source, so identical civil arithmetic implies different elapsed time:

```
instant 2026-02-15T12:00:00Z, +1 month
  UTC calendar: elapsed 28 days
  NY  calendar: elapsed 27 days, 23:00:00   (DST started in between)
```

Note that the timezone argument never appears in the result type — `precision_timestamp_tz<P>` in, `precision_timestamp_tz<P>` out. The zone is not attached to anything; it is borrowed as a calendar for step 2 and discarded, much like a rounding mode.

## The resulting pattern

A timezone is needed exactly where an *instant* meets *calendar* semantics — one cell out of six:

| | `date` (civil) | `precision_timestamp` (civil) | `precision_timestamp_tz` (instant) |
|---|---|---|---|
| `interval_year` (months, calendar) | 2-arg | 2-arg | **needs zone** |
| `interval_day` (elapsed time, duration) | 2-arg | 2-arg | 2-arg |

`date` and `precision_timestamp` need no zone for months because they are already civil values, so there is no instant↔civil conversion to perform. `interval_day` needs no zone anywhere because it never touches civil fields — this was raised and settled in the #337 review.

The yaml matches this matrix exactly, in both functions, with the single exception of `subtract:ptstz_iyear`, which sits in the one cell that requires a zone and does not take one.

## Why "just retain the operand's timezone" is not an option

A tempting third resolution is to let the result keep the timezone of its `precision_timestamp_tz` operand, making the argument redundant. That is not representable: `precision_timestamp_tz` carries no timezone, in either the type or the value.

- `Type.PrecisionTimestampTZ` has only `precision`, `type_variation_reference`, `nullability` ([type.proto#L187-L192](https://github.com/substrait-io/substrait/blob/main/proto/substrait/type.proto#L187-L192)), and the type takes precision as its only parameter.
- The literal reuses `PrecisionTimestamp` — `precision` plus `int64 value`, documented as *"Time passed since 1970-01-01 00:00:00.000000 in UTC for PrecisionTimestampTZ"* ([algebra.proto#L1151-L1156](https://github.com/substrait-io/substrait/blob/main/proto/substrait/algebra.proto#L1151-L1156)).
- `type.proto#L42` annotates the oneof field itself with `// value is since UNIX epoch in UTC`.
- Most tellingly, `local_timestamp(precision_timestamp_tz<P>, timezone) -> precision_timestamp<P>` requires the zone to be passed back in. If the value remembered its zone, that argument would be dead weight.

The removed `timestamp_tz` type had no zone field either, so the "TZ" suffix has always denoted a UTC-normalized instant rather than a zone-carrying value. This differs from Arrow, where `timestamp[unit, tz]` does carry the zone in the type. The docs invite the wrong model here; see the separate documentation issue.

## Alternative considered

Converge on 2-argument everywhere by dropping both 3-argument variants and specifying that the month arithmetic is performed in the UTC calendar. Local-calendar semantics would remain expressible by decomposition, since all three steps already exist as functions:

```
assume_timezone(add(local_timestamp(ts_tz, tz), interval_year), tz)
```

This is coherent and has one real advantage — UTC has no DST transitions, so it sidesteps the disambiguation gap noted below. The cost is that Substrait could no longer directly express session-timezone month arithmetic as performed by Postgres, Spark and Snowflake; engines would have to pattern-match a three-function tree to recover their native operation.

## Related, separate

Step 3 above is not always a function: a local civil time can fail to exist (spring-forward gap) or occur twice (fall-back), and month arithmetic can land on both. Neither the 3-argument impls nor `assume_timezone` specifies a disambiguation rule, so two conforming engines can differ by an hour. That gap is independent of this issue and applies as long as the 3-argument forms exist.

jagjitkaur0000:
## Description

This PR deprecates the `subtract(timestamp_tz, interval_year)` overload.

## Changes

- Marked the `subtract(timestamp_tz, interval_year)` overload as deprecated.
- Added deprecation metadata with version `0.100.0`.
- Added guidance to use the three-argument overload with an explicit timezone instead.

## Related Issue

Fixes #1153

- - -
This change is [Reviewable](https://reviewable.io/reviews/substrait-io/substrait/1160)

### DIFF
--- a/extensions/functions_datetime.yaml
+++ b/extensions/functions_datetime.yaml
@@ -309,6 +309,9 @@ scalar_functions:
             value: precision_timestamp_tz<P>
           - name: y
             value: interval_year
+        deprecated:
+          since: "0.99.0"
+          reason: "Use the three-argument overload with an explicit timezone."
         return: precision_timestamp_tz<P>
       - args:
           - name: x
--- a/extensions/functions_datetime.yaml
+++ b/extensions/functions_datetime.yaml
@@ -310,7 +310,7 @@ scalar_functions:
           - name: y
             value: interval_year
         deprecated:
-          since: "0.99.0"
+          since: "0.100.0"
           reason: "Use the three-argument overload with an explicit timezone."
         return: precision_timestamp_tz<P>
       - args:


extensions/functions_datetime.yaml
```diff
@@ -309,6 +309,9 @@ scalar_functions:
             value: precision_timestamp_tz<P>
           - name: y
             value: interval_year
+        deprecated:
+          since: "0.99.0"
```
nielspardon: here we would put the release which will include the deprecation which is going to be `0.100.0`

```suggestion
          since: "0.100.0"
```

assuming that we merge it in time before next release

jagjitkaur0000:
Thanks for the review! I've updated the deprecation version to 0.100.0 as suggested.

nielspardon (approved):
LGTM

yongchul (approved):

Issue #633, opened 2024-04-18
add a CI check to lint function extensions


vbarua:
A number of PRs have introduced changes to functions extensions which don't conform to the spec. The most recent example was https://github.com/substrait-io/substrait/pull/247.

Manual review has proven insufficient to catch these issues, and honestly we shouldn't rely on manual review when we can have a lint for these issues instead.

nielspardon:
#1176 implements the first of the two lints listed here — **invalid type names**.

The gap it closes is narrower than it looks: `simple_extensions_schema.yaml` types every argument, return and structure field as an opaque string, and the type *grammar* has to accept a bare identifier in a type position, because that is how a signature names a parameter (`decimal<P, S>`) and how a derivation expression names an intermediate value. So a misspelled built-in parses cleanly — `f64` reaches the `ParameterName` alternative rather than `scalarType`. The check keys on that: numeric parameter slots reduce to a distinct `NumericParameterName` node, so a `ParameterName` means an identifier landed where a *type* was expected, and outside a derivation expression nothing legitimately introduces such a name.

It found five, all of which had shipped in example files embedded in the published documentation: `f64` for `fp64` in `distance_functions.yaml`, `int64` for `i64` in the two `any`-type examples, and a bare `int` in `deprecation_example.yaml`. Fixed in #1136. It also catches nested cases such as `list<f64>`, and runs over `extensions/` as well as the examples.

**Duplicate function definitions — the second bullet — is not covered.** Two observations from scanning the current catalog, for whoever picks that up:

- No two `impls` of the same function share an argument signature, so there are no unresolvable overloads today.
- But four functions are declared as *two separate entries* in the same section rather than one entry with several `impls`: `regexp_match_substring`, `regexp_count_substring` and `regexp_replace` in `functions_string.yaml`, and `count` in `functions_aggregate_decimal_output.yaml`. Their signatures do not overlap — e.g. `regexp_replace` has one entry with `(string, string, string, i64, i64)` and `(varchar<L1>, varchar<L2>, varchar<L3>, i64, i64)`, and a second entry with `(string, string, string)` — so they are mergeable rather than conflicting.

That means a duplicate-definition lint is not a free addition: it would flag those four, and whoever adds it has to decide whether splitting a function across entries is a defect to fix or a legal pattern to permit. Worth noting that `functions_aggregate_decimal_output.yaml` is being deprecated in #1149, which would remove one of the four.

PR #1176, opened 2026-08-14
test(extensions): reject type names that do not exist

nielspardon:
Addresses the **invalid type names** lint asked for in #633. That issue also asks for a duplicate-function-definitions lint, which this does not cover — see [my comment there](https://github.com/substrait-io/substrait/issues/633#issuecomment-5291001909) for what that one would have to decide first, so #633 stays open.

## Problem

Nothing in CI can tell a Substrait type name from an arbitrary word.

`text/simple_extensions_schema.yaml` types every argument, return and structure field as an opaque string, so `check-jsonschema` cannot help. `yamllint` and `editorconfig-checker` are about formatting. And `test_extension_yaml_type_expressions_are_grammar_compliant` — which already walks `extensions/` *and* `site/examples/**` — validates only that type expressions parse.

Parsing is not enough, and not because the grammar is weak. The type grammar *has* to accept a bare identifier in a type position, because that is how a signature names a parameter (`decimal<P, S>`) and how a derivation expression names an intermediate value. A misspelled built-in is a perfectly good identifier, so `f64` reaches the `ParameterName` alternative instead of `scalarType` and parses without complaint:

```
f64          -> ParameterNameContext     # not a Substrait type
fp64         -> TypeLiteralContext
i64          -> TypeLiteralContext
any1         -> TypeLiteralContext
u!point      -> TypeLiteralContext
```

Five such names shipped as a result — `f64` for `fp64`, and `int64` and `int` for `i64` — every one of them in an example file embedded in the published extension documentation. `distance_functions.yaml`, the canonical example for referencing user-defined types across extensions, declared `return: f64`. Readers were being taught a type that does not exist. #1136 corrects them; nothing stops the sixth.

## Change

Report identifiers that sit where the grammar expected a type.

The discriminator is already in the grammar. Numeric parameter slots — `decimal<precision, scale>`, `varchar<length>` — reduce to `NumericParameterName`, a distinct node from `ParameterName`. So a `ParameterName` node means an identifier landed in a *type* slot, and outside a derivation expression nothing legitimately introduces such a name.

`MultilineDefinition` is therefore exempt. A derivation expression binds its own identifiers by assignment and reads signature parameters freely:

```
init_scale = max(S1, S2)
DECIMAL<prec, scale>
```

Those names are resolved by the type system, not by this grammar, and `functions_arithmetic_decimal.yaml` and `functions_rounding_decimal.yaml` are full of them. Without the exemption they would be false positives; with it, the check is clean across the entire checked-in corpus.

The check runs over `extensions/` as well as the examples — the mistake is not specific to examples, it just happened to land there — and catches nested cases where only a type parameter is wrong, e.g. `list<f64>`.

## Limits

This validates *names*, not semantics. It does not check that a type is used with the right number or kind of parameters, that a `u!`-prefixed reference resolves to a type the file's `dependencies` actually declare, or that a derivation expression's identifiers are bound. Those would need the type system rather than the grammar.

It also cannot flag an identifier in a type position *inside* a derivation expression, since that whole expression is exempt. A typo in the final type of a multiline derivation goes unnoticed.

- - -
This change is [Reviewable](https://reviewable.io/reviews/substrait-io/substrait/1176)

### DIFF
--- a/tests/type/test_type_grammar.py
+++ b/tests/type/test_type_grammar.py
@@ -74,6 +74,40 @@ def extension_yaml_files():
     yield from sorted((repo_root / "site" / "examples").glob("**/*.yaml"))
 
 
+def undefined_type_names(expression: str):
+    """Identifiers in ``expression`` that sit where a type name belongs.
+
+    The grammar has to accept an arbitrary identifier in a type position,
+    because that is how a signature names a parameter (``decimal<P, S>``) or a
+    derivation expression names an intermediate value.  A misspelled built-in
+    therefore parses cleanly: ``f64`` is not a Substrait type, but it is a
+    perfectly good identifier, so it reaches the ``ParameterName`` alternative
+    instead of ``scalarType``.
+
+    Numeric parameter slots reduce to ``NumericParameterName``, a different
+    node, so a ``ParameterName`` means an identifier landed where the grammar
+    expected a *type*.  Outside a derivation expression nothing legitimately
+    introduces such a name, so it is a type that does not exist.
+
+    Derivation expressions are exempt: ``MultilineDefinition`` binds its own
+    identifiers by assignment and freely references signature parameters, and
+    those are resolved by the type system rather than by this grammar.
+    """
+    tree = parse_type_expression(expression)
+    if isinstance(tree.getChild(0), SubstraitTypeParser.MultilineDefinitionContext):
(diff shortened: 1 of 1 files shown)


yongchul (approved):

benbellick (approved):
LGTM

PR #1178, opened 2026-08-14
docs: allow an empty argument signature for zero-argument functions

### LINKED ISSUE #1162: Function signature grammar cannot express a zero-argument function
## Summary

The function signature grammar requires at least one argument type, so a zero-argument function impl has no expressible signature. Since #537 made compound names mandatory in extension references, those impls have no spec-valid reference form at all.

This is not a `count(*)` niche: seven impls in the core extension set declare no arguments.

## The gap

[`site/docs/extensions/index.md`](https://github.com/substrait-io/substrait/blob/c9f697e6a1773e9d1248e87a2bd0fabe0c0e7906/site/docs/extensions/index.md#L73-L74) (pinned to `c9f697e`, which predates the fix proposed in #1178):

```abnf
function-signature = function-name ":" argument-signature
argument-signature = short-arg-type *("_" short-arg-type)
```

`*("_" short-arg-type)` makes only the *repetition* optional; one `short-arg-type` is required.

Zero-argument impls in the core extension set:

| impl | file |
|---|---|
| `row_number`, `rank`, `dense_rank`, `percent_rank`, `cume_dist` | `extensions/functions_arithmetic.yaml:1769-1808` (`args: []`) |
| `count` — "Count the number of records (not field-referenced)" | `extensions/functions_aggregate_generic.yaml` |
| `count` (star) | `extensions/functions_aggregate_decimal_output.yaml` (deprecation proposed in #1149, not yet merged) |

## The spec already uses the empty argument signature

[`dialects/tests/functions_test.yaml:81-84`](https://github.com/substrait-io/substrait/blob/c9f697e6a1773e9d1248e87a2bd0fabe0c0e7906/dialects/tests/functions_test.yaml#L81-L84):

```yaml
- source: arithmetic
  name: rank
  supported_impls:
    - ""
```

That file is validated by the `check-dialects` lint task, and `text/dialect_schema.yaml` types `supported_impls` entries as strings with no `minLength`, so `""` is legal. Under `function-signature = function-name ":" argument-signature`, that entry denotes `rank:` — an empty argument signature the ABNF forbids.

## Producers already converged on `name:`

The key is built as `name + ":" + join(arg_types, "_")` with no zero-arity branch, implemented independently three times:

- substrait-java — `SimpleExtension.java:586`, `... + arguments.stream().map(Argument::toTypeString).collect(Collectors.joining("_"))`
- substrait-go — `extensions/variants.go:359,473,595`, `return s.name + ":" + s.impl.signatureKey()` (scalar, aggregate and window variants alike)
- substrait-python — same construction in its extension registry

Evaluated on an empty argument list this yields `count:` / `rank:`. substrait-go asserts `"name": "count:"` in a golden plan test, and substrait-python emits `count:` when exercised. `SubstraitBuilder.countStar()`'s `"count:"` literal is a catalog lookup that throws if the key is absent, so the key is load-bearing there, not merely emitted.

So the convention exists and is consistent; what is missing is permission for it in the grammar.

## How the constraint got there

It looks like a transcription artifact rather than a decision:

- #863 (*docs: clarify function signatures for variadic functions*) first formalized the informal prose as `<argument_signature> ::= <short_arg_type> { _ <short_arg_type> }*`. Its subject was variadic arguments.
- #936 (*docs: clarify UDT references must begin with `u!`*) mechanically converted that to RFC 5234 ABNF, producing the current line. Its subject was UDT references.

Neither thread discusses zero-argument functions. #537 then removed bare simple names, which is what makes the omission bite today.

## Proposal

Make the argument signature optional, ratifying what producers and the dialect fixture already do:

```abnf
function-signature = function-name ":" argument-signature
argument-signature = [short-arg-type *("_" short-arg-type)]
```

`count:` and `rank:` become valid. The Uniqueness Constraint is unaffected — a zero-argument impl is unique within its function by construction, since there is only one way to take no arguments.

If the preferred answer is instead that zero-argument functions should not be reachable by signature at all, that is worth stating explicitly, because the extension YAMLs currently declare seven of them.

## Alternatives considered

- **Treating `count(*)` as a star/wildcard argument.** Not applicable: #401 established that for `count`, `*` is represented as the absence of an argument (*"the Substrait equivalent of `COUNT(*)` is the 0-arg version `e.g. count()`"*), and the YAML impl is described as "not field-referenced". This issue takes that as given and asks only that the resulting zero-argument impl be nameable.

    To be clear, this is narrower than it may sound: the community has separately discussed a `*`-like capability for propagating columns that a plan never references through to its output ("runtime column propagation", or "dynamic star"), in the 2026-06-03 community sync. That concerns dynamic schema handling rather than signatures, is not resolved by this issue, and nothing here should be read as foreclosing it.
- **A dedicated `Expression` variant**, as #945 did for `current_date`/`current_timestamp`/`current_timezone` after #524/#548 tried the zero-argument-function route. That fits context variables, but `count(*)` and the five window functions are genuine aggregate/window functions that belong in the extension set.

## Related

- #1151 raises the zero-argument `count` as "the sharpest case", but for `AggregateFunction.arguments`/`output_type` in `algebra.proto` rather than the extension-declaration name string.
- #911 is the only place upstream where `count:` is written down.
- #1121 / #1149 / #606 / #670 concern return-type overloading — the same grammar, a different hole. Both #1121 and #1149 have to write "`count` (star)" in prose precisely because the variant has no expressible signature.
- #634 covers the URI component of function names.

## Context

Found while auditing function-signature emission in [duckdb-substrait-extension](https://github.com/substrait-io/duckdb-substrait-extension), which currently emits a bare `count` / `row_number` / `rank` with no suffix for these impls. We would like to align with `count:`, but would rather not encode a convention the spec forbids on paper.

nielspardon:
Closes #1162.

## Problem

`argument-signature = short-arg-type *("_" short-arg-type)` makes only the *repetition* optional, so one short argument type is required and an implementation that declares no arguments has no expressible signature. Since #537 made compound names mandatory in extension references, those implementations have no spec-valid reference form at all.

The core extension set declares seven of them: the record-counting `count` impl in `functions_aggregate_generic.yaml`, its decimal-output counterpart in `functions_aggregate_decimal_output.yaml`, which #1149 proposes to deprecate but which is still live on `main`, and the five window functions `row_number`, `rank`, `dense_rank`, `percent_rank` and `cume_dist` in `functions_arithmetic.yaml`.

#1162 traces the constraint to a transcription chain rather than a decision: #863 formalized the informal prose while discussing variadic arguments, and #936 mechanically converted that to ABNF while discussing UDT references. Neither thread considered zero-argument functions.

## Change

Make the argument signature optional:

```abnf
function-signature = function-name ":" argument-signature
argument-signature = [short-arg-type *("_" short-arg-type)]
```

A new *Zero-Argument Functions* subsection states that the colon separator is retained (`count:`, `rank:`) and that a bare function name with no colon is not a signature — that last point is the practical one, since it is the form some producers emit today. The Uniqueness Constraint is unaffected: a function can have at most one zero-argument implementation, because a second one would produce the same signature.

The dialect schema's `supported_impls` description now says a zero-argument implementation is identified by the empty string, so the existing `""` entry in `dialects/tests/functions_test.yaml` is documented rather than incidental.

The MIRROR row of the nullability-handling table now spells out the consequence for this case: with no arguments there is nothing nullable to propagate, so the derived return type is non-nullable. The marker check from #1148 already enforces that — its argument loop is empty and its MIRROR return check is arity-independent — but nothing exercised it at arity zero, since all seven zero-argument implementations use DECLARED_OUTPUT. Two unit tests pin it, for an explicit empty `args` list and for an omitted `args` key.

## Ecosystem alignment

This ratifies a convention already converged on independently, so no migration is implied — the change makes existing producer behavior legal rather than asking anyone to change.

| Library | Construction | Zero-arity result |
|---|---|---|
| substrait-java | `SimpleExtension.java:583` — `name + ":" + arguments.stream().map(Argument::toTypeString).collect(joining("_"))` | `count:`; `SubstraitBuilder.countStar()` looks up the literal `"count:"` and throws if absent, so the key is load-bearing |
| substrait-go | `extensions/variants.go:359,473,595` — `s.name + ":" + s.impl.signatureKey()`, identically for scalar, aggregate and window variants | `count:`, asserted in a golden plan test and in `extension_mgr_test.go` for both the aggregate and window lookups |
| substrait-python | `extension_registry/function_entry.py:49` — `f"{self.name}:{'_'.join(self.normalized_inputs)}"` | `count:` |

Downstream, duckdb-substrait-extension currently emits a bare `count` / `row_number` / `rank` with no suffix and would like to align with `count:`, which is what surfaced this.

## Drive-by

The *Examples* table headers were inverted — the left column holds declarations like `add(i8, i8) => i8` and the right column holds signatures like `add:i8_i8`, but they were labelled "Function Signature" and "Function Name". That is a leftover from the same #863 rename that introduced the grammar line, and adding the zero-argument rows made it conspicuous, so it is corrected here.

## Alternative considered

#1162 notes the other possible answer: that zero-argument functions should not be reachable by signature at all. This PR takes the opposite branch, on the grounds that the core extension set declares seven such implementations and three SDKs already reference them by key, so what stands today is a documentation gap rather than a deliberate prohibition. If the preferred answer is instead the prohibition, that is worth stating explicitly in the spec rather than leaving it implicit in the grammar — say so and I will close this in favour of that.

- - -
This change is [Reviewable](https://reviewable.io/reviews/substrait-io/substrait/1178)


site/docs/extensions/index.md
```diff
@@ -80,6 +80,12 @@ Argument types (`short_arg_type`) are encoded using the Type Short Names given b
 
 For variadic functions, the variadic argument is included *once* in the argument signature.
 
+#### Zero-Argument Functions
+
+An implementation that declares no arguments has an empty argument signature. The colon separator is still required, so such an implementation is referenced by its function name followed by a trailing colon: `count:` for the record-counting implementation of `count`, or `rank:` for `rank`. A bare function name with no colon is not a function signature and cannot be used to reference an implementation.
+
+A function can have at most one zero-argument implementation, since a second one would produce the same signature and violate the Uniqueness Constraint below.
```
yongchul: link? 😄

site/docs/extensions/index.md
```diff
@@ -80,6 +80,12 @@ Argument types (`short_arg_type`) are encoded using the Type Short Names given b
 
 For variadic functions, the variadic argument is included *once* in the argument signature.
 
+#### Zero-Argument Functions
```
yongchul, line 83: Do we need to be clear about the return type of this function (i.e., no parameterized types or any allowed as a return type)?

text/dialect_schema.yaml
```diff
@@ -244,7 +244,9 @@ definitions:
         type: array
         description: >
           List of one or more implementations supported by this function. Use
-          argument signatures to identify the implementations.
+          argument signatures to identify the implementations. An implementation
+          that takes no arguments has an empty argument signature and is
+          identified by the empty string.
```
yongchul: should we include an example for this? so `["i64", ""]` represents there are impl that takes one i64 arg and no arg. right?

site/docs/extensions/index.md
nielspardon: Done — it now links to the [Uniqueness Constraint](#uniqueness-constraint) section instead of saying "below". 5ac4675

site/docs/extensions/index.md
nielspardon: Good call — added in 5ac4675. The constraint I stated is that the return type must be *fully concrete*: with no arguments, there is nothing for `any` / `any[\d]` or for a type expression over argument type parameters to resolve against.

One nuance on the parenthetical: a *parameterized* type is still fine as long as its parameters are concrete. The record-counting `count` in `functions_aggregate_decimal_output.yaml` returns `decimal<38,0>` today, so ruling out parameterized types outright would invalidate an existing impl. The wording allows that case and rules out only unresolvable placeholders and derivations.

I deliberately left nullability alone. `MIRROR` with zero arguments is degenerate rather than invalid — vacuously all arguments are non-nullable, so the output is non-nullable — and all seven zero-argument impls in the core set already declare `DECLARED_OUTPUT`. Saying `MIRROR` is disallowed would be a new rule rather than a consequence of the existing binding rules, so I left it out; happy to add it here or as a follow-up if you think it is worth pinning down.

text/dialect_schema.yaml
nielspardon: Added in 5ac4675, and yes — your reading is exactly right. `["i64", ""]` says the dialect supports two implementations of that function: the one taking a single `i64` argument, and the zero-argument one. The description now spells that out as an example.

site/docs/extensions/index.md
yongchul: Yes, concrete type is sufficient for return type. Regarding nullability, it is worth to spell out the behavior of MIRROR somewhere -- probably MIRROR mode itself that zero arg implies not nullable.

site/docs/extensions/index.md
```diff
@@ -80,6 +80,14 @@ Argument types (`short_arg_type`) are encoded using the Type Short Names given b
...
+
+An implementation that declares no arguments has an empty argument signature. The colon separator is still required, so such an implementation is referenced by its function name followed by a trailing colon: `count:` for the record-counting implementation of `count`, or `rank:` for `rank`. A bare function name with no colon is not a function signature and cannot be used to reference an implementation.
+
+A function can have at most one zero-argument implementation, since a second one would produce the same signature and violate the [Uniqueness Constraint](#uniqueness-constraint).
+
+Because there are no arguments to bind against, the return type must be fully concrete: neither the `any` and `any[\d]` placeholders nor a type expression over argument type parameters has anything to resolve against. A parameterized type with concrete parameter values, such as `decimal<38,0>`, is unaffected.
```
yongchul: perhaps `with unresolved type parameters`?

yongchul (approved):

site/docs/extensions/index.md
nielspardon: Done in 37ed748 — spelled out in the `MIRROR` row of [Nullability Handling](https://substrait.io/expressions/scalar_functions/#nullability-handling) rather than in the zero-argument section, so it sits with the mode it describes: "An implementation that declares no arguments therefore has a non-nullable return type, since there are no nullable inputs to propagate."

Placed there it also covers aggregate and window functions, since `aggregate_functions.md` inherits all scalar-function properties. I left it single-sourced in that row rather than restating it under Zero-Argument Functions, to avoid two copies drifting apart.

site/docs/extensions/index.md
nielspardon: Applied in 37ed748 — it now reads "nor a type expression with unresolved type parameters has anything to resolve against". Better than my original, thanks: the problem is the unresolved parameters, not where they came from.

site/docs/expressions/scalar_functions.md
```diff
@@ -102,7 +102,7 @@ Both enumeration arguments and options accept values from a fixed set of strings
...
-| `MIRROR`          | This means that the function has the behavior that if at least one of the input arguments are nullable, the return type is also nullable. If all arguments are non-nullable, the return type will be non-nullable. Since the nullability of the output is determined by the nullability of the inputs, argument types and return types must not include nullability markers (`?`). The function binds regardless of argument nullability. An example of a function with `MIRROR` nullability is the `add` function. |
+| `MIRROR`          | This means that the function has the behavior that if at least one of the input arguments are nullable, the return type is also nullable. If all arguments are non-nullable, the return type will be non-nullable. An implementation that declares no arguments therefore has a non-nullable return type, since there are no nullable inputs to propagate. Since the nullability of the output is determined by the nullability of the inputs, argument types and return types must not include nullability markers (`?`). The function binds regardless of argument nullability. An example of a function with `MIRROR` nullability is the `add` function. |
```
benbellick, line 105: 🤔 ooh that is an interesting point... good update!

benbellick (approved):
LGTM

tests/coverage/test_coverage.py
```diff
@@ -1093,3 +1093,21 @@ def test_option_arguments_are_skipped(self):
...
+        assert len(errors) == 1
+        assert "i64?" in errors[0]
+
+    def test_zero_argument_without_args_key(self):
+        """An impl that declares no arguments may omit the 'args' key."""
+        errors = validate_impl_nullability_markers(
+            {"nullability": "MIRROR", "return": "i64"},
+            "functions_test.yaml: count",
+        )
+        assert errors == []
```
nielspardon, line 1113: I just thought of adding an explicit test case for the zero-arg mirror nullability behavior which is already covered by the existing validation function

site/docs/extensions/index.md
```diff
@@ -80,6 +80,14 @@ Argument types (`short_arg_type`) are encoded using the Type Short Names given b
 
 For variadic functions, the variadic argument is included *once* in the argument signature.
 
+#### Zero-Argument Functions
+
+An implementation that declares no arguments has an empty argument signature. The colon separator is still required, so such an implementation is referenced by its function name followed by a trailing colon: `count:` for the record-counting implementation of `count`, or `rank:` for `rank`. A bare function name with no colon is not a function signature and cannot be used to reference an implementation.
+
+A function can have at most one zero-argument implementation, since a second one would produce the same signature and violate the [Uniqueness Constraint](#uniqueness-constraint).
+
+Because there are no arguments to bind against, the return type must be fully concrete: neither the `any` and `any[\d]` placeholders nor a type expression with unresolved type parameters has anything to resolve against. A parameterized type with concrete parameter values, such as `decimal<38,0>`, is unaffected.
```
yongchul: nit: `fully` is redundant.

nielspardon: Applied in 46b62ab — it now reads "the return type must be concrete". Agreed: "concrete" already excludes the placeholder and unresolved-parameter cases the rest of the sentence enumerates, so the qualifier was only adding emphasis.

Issue #1179, opened 2026-08-14
Clarify the meaning of a breaking change


benbellick:
Discussion on #1105 revealed disagreement about whether a deprecation should be marked as a `BREAKING CHANGE` so that it is prominently announced to downstream implementors. We should clarify what constitutes a breaking change in our release notes and contribution guidance.

I propose that a deprecation is not considered breaking as long as the deprecated behavior remains supported. Removing it later would be breaking. [Semantic Versioning](https://semver.org/spec/v2.0.0.html) explicitly requires a minor version increment when public API functionality is deprecated and a major increment for backward-incompatible changes. [Conventional Commits](https://www.conventionalcommits.org/en/v1.0.0/) similarly defines `BREAKING CHANGE` and `!` as breaking API changes that correspond to a major release. This also aligns with [Substrait's versioning documentation](https://github.com/substrait-io/substrait/blob/main/site/docs/spec/versioning.md), which says that we maintain backward compatibility through deprecation.

For example, consider this API in version 1.2.0:

```proto
message Request {
  string old_field = 1;
}
```

We then add a replacement and deprecate the old field:

```proto
message Request {
  string old_field = 1 [deprecated = true];
  string new_field = 2;
}
```

Existing producers and consumers remain compatible. Calling this a breaking change would make the next version 2.0.0 even though nothing has been removed or made invalid.

If `BREAKING CHANGE` is being used as an announcement mechanism so downstream implementors notice deprecations, we should find a separate mechanism rather than change the meaning of `BREAKING CHANGE`.

This distinction is less relevant while Substrait remains pre-1.0, but if we use Semantic Versioning, we should still follow its conventions.

Issue #1109, opened 2026-06-24
IntervalDay.precision is the only precision field that is optional with an implicit default of 6 — consider making it consistent


nielspardon:
## Summary

`Type.IntervalDay.precision` is the only sub-second-precision field in the spec that is declared `optional` and carries an implicit "if unset, treat as 6" default. Every other precision-bearing type uses a plain required `int32 precision` where `0` means seconds and there is no unset state. This inconsistency forces every consumer to special-case `IntervalDay`, and the libraries diverge on whether they actually do.

## The inconsistency

`proto/substrait/type.proto`:

```proto
message IntervalDay {
  // Sub-second precision, 0 means the value given is in seconds, 3 is
  // milliseconds, 6 microseconds, 9 is nanoseconds, etc.
  // if unset, treat as 6.
  optional int32 precision = 3;
}

message PrecisionTimestamp   { int32 precision = 1; ... }   // required, 0 = seconds
message PrecisionTimestampTZ { int32 precision = 1; ... }   // required, 0 = seconds
message PrecisionTime        { int32 precision = 1; ... }   // required, 0 = seconds
message IntervalCompound     { int32 precision = 3; ... }   // required, 0 = seconds
```

So `IntervalDay`:
- is the only one where `precision` has **explicit presence** (`optional`), making "unset" distinguishable from `0`, and
- is the only one whose **unset state maps to 6 (microseconds)** rather than `0` (seconds).

This originates from #665 / #664, where precision was retrofitted into the pre-existing `IntervalDay` type for backwards compatibility (the literal `IntervalDayToSecond` uses a `oneof { microseconds [deprecated] | precision }` for the same reason). By contrast, when precision was added to timestamps/time, the spec introduced **new** types (`PrecisionTimestamp`, etc.) deprecating the old ones, rather than retrofitting presence. `IntervalDay` is the lone exception to that pattern.

## Why it matters — cross-library survey

We surveyed the four main implementations to see who actually depends on the `optional` + "treat as 6" behavior:

| Library | Presence check (unset vs 0)? | Implements "treat unset as 6"? | Impact if `precision` became required `int32` |
|---|---|---|---|
| **substrait-java** | Yes — `hasPrecision()` | Yes — falls back to `6` | Silent semantic shift; fallback branches become dead code |
[...]

PR #1181, opened 2026-08-14
docs: document protobuf field replacement migrations

benbellick:
Define a standard migration sequence for replacing protobuf fields. This keeps field documentation focused on field-specific behavior instead of repeating general migration guidance.

- - -
This change is [Reviewable](https://reviewable.io/reviews/substrait-io/substrait/1181)


site/docs/spec/breaking_change_policy.md
```diff
@@ -4,13 +4,24 @@ title: Breaking Change Policy
...
 ## Breaking Change Cookbooks
 
-### URI to URN Migration
+### Replacing a Protobuf Field
+
+Replacing a protobuf field requires a staged migration so that producers and consumers can be updated independently:
+
+1. Add the replacement using a new field number and deprecate the old field.
+2. Update consumers to accept both representations. When the replacement is present, consumers must use it and ignore the deprecated field.
+3. Update producers to write the replacement. Producers must also write the deprecated field whenever the replacement has an exact representation in the old field, and both fields must be semantically equivalent. When no equivalent old representation exists, producers write only the replacement. This preserves compatibility with older consumers whenever possible without limiting the replacement's new capabilities.
```
nielspardon, line 19: Do you want to cover the `oneof` case explicitly or does the following part assume that if the new field is in a `oneof` with the deprecated field there is no way to emit an `equivalent old representation` in adition to the new field even if they would be semantically the same?

> When no equivalent old representation exists, producers write only the replacement.

benbellick: Has there been a case where we migrated from `fieldA` to `fieldB` where we _needed_ to put both in a `oneof`? I'm inclined to add it as a second option if it comes up. 

I know we have put them in `oneof`s before, but I thought those may have been cases where we later realized putting them outside of a `oneof` is better. Let me investigate... 😅

nielspardon: Yes, the id-based outer reference resolution was in a oneof if you remember. https://github.com/substrait-io/substrait/pull/1031

benbellick: A few more examples here: 

- `FetchRel.offset` → `offset_expr` ([#748](https://github.com/substrait-io/substrait/pull/748))
- `FetchRel.count` → `count_expr` ([#748](https://github.com/substrait-io/substrait/pull/748))
- `IntervalDayToSecond.microseconds` → `precision` and `subseconds` ([#665](https://github.com/substrait-io/substrait/pull/665))

nielspardon: - expression support for FetchRel: https://github.com/substrait-io/substrait/pull/748
- `microseconds` and `precision` for `IntervalDayToSecond` https://github.com/substrait-io/substrait/pull/1116

ryanmaka021-creator: Ues

benbellick: @nielspardon I think those linked ones are the same as my listed examples. 

This is a bit tough because its hard to provide a prescriptive rule. My personal feeling is that we should avoid doing `oneof` migrations unless we have reason to believe that the implementation will be difficult otherwise. (Which is the case in the id-based outer reference case.)

What do you think about me adding another section for the `oneof` migration strategy along with some light guidance on preferring the non-oneof path?

nielspardon: > @nielspardon I think those linked ones are the same as my listed examples.

Yeah, I think we overlapped there.

nielspardon: > This is a bit tough because its hard to provide a prescriptive rule. My personal feeling is that we should avoid doing `oneof` migrations unless we have reason to believe that the implementation will be difficult otherwise. (Which is the case in the id-based outer reference case.)
> 
> What do you think about me adding another section for the `oneof` migration strategy along with some light guidance on preferring the non-oneof path?

Yes, I think we should add some guidance. I will run it by Claude to see if it has a good idea how to frame it or you framing is good enough to get some inspiration.

benbellick: I am changing my mind a bit and I actually think we should outright discourage the usage of `oneof` for migration purposes, unless the author can make a compelling argument for why a standard migration isn't sufficient. A standard migration allows scalar fields to retain their intended implicit-presence semantics:

```proto
// Before
int64 field_a = 1;

// During migration
int64 field_a = 1 [deprecated = true];
int64 field_b = 2;

// After migration
reserved 1;
reserved "field_a";
int64 field_b = 2;
```

By contrast, Protobuf only documents moving a single [explicit-presence field into a new `oneof`](https://protobuf.dev/programming-guides/proto3/#wire-safe-changes) as wire-safe, and safely removing the `oneof` requires the replacement scalar to retain explicit presence:

```proto
// Before
optional int64 field_a = 1;

// During migration
oneof value {
  int64 field_a = 1 [deprecated = true];
  int64 field_b = 2;
}

// After migration
reserved 1;
reserved "field_a";
optional int64 field_b = 2;
```

IMO, a migration should not determine a field’s long-term [presence semantics](https://protobuf.dev/programming-guides/field_presence/#considerations-for-change-compatibility) so IIUC the above shows that we should generally just do the migration without a `oneof`. Thoughts?

nielspardon: I think this sounds reasonable. I'm currently checking with Claude if there's anything we should consider when adding such guidance. Will share what it has to say (and will ask it to keep it succinct).

nielspardon: As promised, I ran this by Claude. Sharing its findings.

---

- **The position holds, and is stronger than stated.** Protobuf only lists *"changing a single explicit presence field into a member of a new `oneof`"* as wire-safe, and almost nothing in these protos has explicit presence — exactly three `optional` scalar fields in the whole tree (`type.proto:131`, `algebra.proto:36`, `algebra.proto:247`) against 38 oneofs. So moving a typical Substrait field into a `oneof` is not the documented-safe case at all; it is an implicit→explicit cardinality change, which the field-presence docs treat as byte-compatible but *not* semantically safe once anything depends on presence.
    - `buf breaking` against the repo's `FILE` config agrees: add-and-deprecate is clean, while the `oneof` step reports two findings.

- **Suggested framing: anchor the rule to dual-write rather than to presence.** A `oneof` makes step 3 structurally impossible — #1132 had to spell out "the migration is consumer-first rather than dual-write" for exactly that reason, and the URI→URN cookbook directly below only worked *because* there was no `oneof`. Two things that framing buys:
    - It self-scopes to migrations, so it does not read as discouraging `oneof` where mutual exclusion is the real data model (`Rel.rel_type`, `Literal.literal_type`, …).
    - It sidesteps the "is this really a migration?" argument — `steps_out` → `rel_reference` was a capability-add in #1031 and only became a migration in #1132.

- **Add a positive requirement alongside it.** If the policy pushes authors off `oneof`, it also needs to require that the replacement have *detectable presence*, or step 2 becomes unimplementable: "when the replacement is present, consumers must use it" is undecidable for a bare implicit-presence scalar where `0` is legal, and the example in this thread uses `int64 field_b = 2`. The repo has already hit this — `extension_urn_reference` is a bare `uint32`, and #900 had to add "0 is a valid anchor/reference, but prefer non-zero values for ergonomics" as a documentation workaround. So a replacement field should be one of:
    - a message type (presence comes for free),
    - `optional`, or
    - a scalar with a documented-invalid zero, as `rel_reference`'s "Must be >= 1" does.

- **Supporting evidence that is also a bug.** #1110 decided in community sync that an unset `IntervalDay.precision` must be *rejected*, keeping `optional` so that unset stays detectable. Six days later #1116 dissolved the `precision_mode` oneof, and the literal's `precision` fell out as a bare `int32 precision = 4` (`algebra.proto:1189`). Since precision `0` means seconds, an unset precision now silently reads as seconds on the literal side while the type side rejects it. That is the thesis in concrete form — a migration decided the field's long-term presence semantics — and the asymmetry looks worth its own issue.

- **Keep the wording forward-looking**, since #665, #748 and #1031 all used `oneof`.

- **Do not lean on the changelog/`BREAKING CHANGE` cost in the doc text** while #1179 is still open — though it is worth knowing that the `oneof` additive step would force a *major* bump post-1.0, once the `releaseRules` minor override goes away.

- **The "use `optional` more" convention is not written down anywhere.** It currently exists only as Niels's closing comment on #1109 plus the sync decision recorded in #1110's body — nothing under `site/docs/` mentions field presence at all. This PR may be the right place to fix that.

nielspardon: I hope this is useful. I told it about the breaking change issue you opened and the previous discussions we had on the use of `optional` in our protos.

benbellick: I am opting to only document the non-`oneof` case. I think we are in agreement that that should be the default. If a case in the future comes up where we _should_ use a `oneof`, we can discuss it then. Thanks

nielspardon:
I'm not sure whether 48f6348df13ce71f1d56d5b1c066fb972e77e2b0 was meant to be in this PR. Looks to me like this is for another issue: https://github.com/substrait-io/substrait/issues/1182

site/docs/spec/breaking_change_policy.md
nielspardon: I think it would be better to add a statement to discourage the use of `oneof` for field migrations explicitly since this is a trap we did step into not only once.

benbellick: I added it, lmk what you think! Thanks

benbellick:
> I'm not sure whether [48f6348](https://github.com/substrait-io/substrait/commit/48f6348df13ce71f1d56d5b1c066fb972e77e2b0) was meant to be in this PR. Looks to me like this is for another issue: #1182

@nielspardon yes this was intentional. [The impetus of this PR was to centralize migration docs so that it can be moved out of the proto docs.](https://github.com/substrait-io/substrait/pull/1105#discussion_r3787149083) Now that that PR has merged, I opted to include it in this PR. Would you rather I remove it?

nielspardon:
> @nielspardon yes this was intentional. [The impetus of this PR was to centralize migration docs so that it can be moved out of the proto docs.](https://github.com/substrait-io/substrait/pull/1105#discussion_r3787149083) Now that that PR has merged, I opted to include it in this PR. Would you rather I remove it?

My instinct was that it would make sense to sync with @yongchul since I had understood he was planning a much broader consolidation of the proto comments into markdown documentation via #1182 and the changes you are including here just seem to cover a relatively narrow scope.

site/docs/spec/breaking_change_policy.md
nielspardon: Thanks, the `oneof` guidance looks good to me and will hopefully serve us well 🤞 .

benbellick:
> > @nielspardon yes this was intentional. [The impetus of this PR was to centralize migration docs so that it can be moved out of the proto docs.](https://github.com/substrait-io/substrait/pull/1105#discussion_r3787149083) Now that that PR has merged, I opted to include it in this PR. Would you rather I remove it?
> 
> My instinct was that it would make sense to sync with @yongchul since I had understood he was planning a much broader consolidation of the proto comments into markdown documentation via #1182 and the changes you are including here just seem to cover a relatively narrow scope.

@nielspardon Sure, lets get @yongchul's opinion 🙂

site/docs/spec/breaking_change_policy.md
```diff
@@ -4,13 +4,24 @@ title: Breaking Change Policy
...
 ## Breaking Change Cookbooks
 
-### URI to URN Migration
+### Replacing a Protobuf Field
+
+Replacing a protobuf field requires a staged migration so that producers and consumers can be updated independently. Do not use a `oneof` solely for migration because it prevents producers from writing both representations and may change field-presence semantics. Use a `oneof` only when mutual exclusivity is part of the final schema and the proposal explains why the standard migration is insufficient.
+
+1. Add the replacement using a new field number and deprecate the old field.
+2. Update consumers to accept both representations. When the replacement is present, consumers must use it and ignore the deprecated field.
+3. Update producers to write the replacement. Producers must also write the deprecated field whenever the replacement has an exact representation in the old field, and both fields must be semantically equivalent. When no equivalent old representation exists, producers write only the replacement. This preserves compatibility with older consumers whenever possible without limiting the replacement's new capabilities.
```
vbarua: minor clarification
```suggestion
3. Update producers to write the replacement. Producers must also write the deprecated field whenever the replacement has a semantically equivalent representation in the old field. When no equivalent old representation exists, producers write only the replacement. This preserves compatibility with older consumers whenever possible without limiting the replacement's new capabilities.
```

site/docs/spec/breaking_change_policy.md
```diff
@@ -4,13 +4,24 @@ title: Breaking Change Policy
...
+Proposals for breaking changes to the specification should either reference an applicable migration strategy below or include an explicit migration strategy. This is based on the assumption that users are operating Substrait in a multi-system, multi-language environment where parallel and synchronous deployments are impossible. Breaking changes should be made so that users can update systems independently and asynchronously.
 
 The migration strategy must be implemented in all [active libraries](../community/active_libraries.md) before the breaking change is implemented.
 
 ## Breaking Change Cookbooks
 
-### URI to URN Migration
+### Replacing a Protobuf Field
+
+Replacing a protobuf field requires a staged migration so that producers and consumers can be updated independently. Do not use a `oneof` solely for migration because it prevents producers from writing both representations and may change field-presence semantics. Use a `oneof` only when mutual exclusivity is part of the final schema and the proposal explains why the standard migration is insufficient.
```
yongchul: what is standard migration?

site/docs/spec/breaking_change_policy.md
```diff
@@ -4,13 +4,24 @@ title: Breaking Change Policy
...
+
+1. Add the replacement using a new field number and deprecate the old field.
+2. Update consumers to accept both representations. When the replacement is present, consumers must use it and ignore the deprecated field.
```
yongchul, line 18: I have mixed feeling about the dual write... it is the right way to handle the migration in a defensive way at the cost of work twice. If there is a way to tell producers "target substrait version", things would be better. But it is okay.

yongchul (approved):
Thanks for the change! LGTM

site/docs/spec/breaking_change_policy.md
benbellick: By the standard migration, I meant what is being described here, `Replacing a Protobuf Field`. So basically, "use something standard as defined in this file, or give a good reason why you are not." I'll edit that to make it a bit clearer. Thanks!

site/docs/spec/breaking_change_policy.md
benbellick: Yes, it is work twice but it is the safest migration policy. This is not an absolute rule though, its just that the default is this approach. If there are cases where it is really difficult, than we can consider other migration patterns. I agree that it can be a bit annoying though.

PR #1183, opened 2026-08-19
docs: fix the Structs in Structs example in named_structs.md

andrew-coleman:
The `Structs in Structs` example is not self-consistent.

- `names` lists nine entries, `a` through `i`, while the arrows below label ten, `a` through `j`.
- The type literal has four `struct<` and five `>`, so it does not parse.

This fix adds `j` to `names` and removes one `>`.

- - -
This change is [Reviewable](https://reviewable.io/reviews/substrait-io/substrait/1183)


nielspardon (approved):
LGTM

PR #1184, opened 2026-08-19
docs: state the output nullability of each join type

### LINKED ISSUE #818: Specify output type derivation rules for SetRel
Similar to https://github.com/substrait-io/substrait/issues/558, we should provide examples of output derivations for the various types of joins as there a number of them
```proto
  enum JoinType {
    JOIN_TYPE_UNSPECIFIED = 0;
    JOIN_TYPE_INNER = 1;
    JOIN_TYPE_OUTER = 2;
    JOIN_TYPE_LEFT = 3;
    JOIN_TYPE_RIGHT = 4;
    JOIN_TYPE_LEFT_SEMI = 5;
    JOIN_TYPE_LEFT_ANTI = 6;
    JOIN_TYPE_LEFT_SINGLE = 7;
    JOIN_TYPE_RIGHT_SEMI = 8;
    JOIN_TYPE_RIGHT_ANTI = 9;
    JOIN_TYPE_RIGHT_SINGLE = 10;
    JOIN_TYPE_LEFT_MARK = 11;
    JOIN_TYPE_RIGHT_MARK = 12;
  }
```

andrew-coleman:
The `Join Types` table has no output-nullability column, while the `Set Operation` table at `logical_relations.md:342-351` has one. Nullability is not mentioned anywhere in the Join section except the Left Mark row's own "nullable boolean".

An extra sentence has been added: that output types come from the join type and the input schemas alone, and are never narrowed by analysis of the join expression or of cardinality. So, for example, a `Left Single` whose condition is provably total still has a nullable right side. I believe this is how it has been implemented, but it is not written down.

Resolves #818 

- - -
This change is [Reviewable](https://reviewable.io/reviews/substrait-io/substrait/1184)

## Summary by CodeRabbit

* **Documentation**
  * Updated join type documentation to describe output nullability for every join type.
  * Clarified that join output types and nullability depend only on the join type and input schemas.
  * Documented that nullable sides remain nullable, including for `Left Single` joins.


site/docs/relations/logical_relations.md
```diff
@@ -252,20 +252,22 @@ The join operation will combine two separate inputs into a single output, based
...
-| Right Single | Same as *Left Single* except that the right and left inputs are switched. |
-| Left Mark | Returns one record for each record from the left input.  Appends one additional "mark" column to the output of the join. The new column will be listed after all columns from left side and will be of type nullable boolean.  If there is at least one join partner in the right input where the join condition evaluates to true then the mark column will be set to true.  Otherwise, if there is at least one join partner in the right input where the join condition evaluates to NULL then the mark column will be set to NULL.  Otherwise the mark column will be set to false. |
-| Right Mark | Same as *Left Mark* except that the right and left inputs are switched. |
+A join's output field types are determined by the join type and the input schemas alone. They do not depend on the join expression, nor on anything a consumer can prove about cardinality: an output field that is nullable below remains nullable even where a consumer can show that no null will be produced.
```
yongchul, line 255: this part feels redundant as "join type and the input schemas alone" already tells this and saying more likely introduce another confusion. For instance, I don't quite understand "anything a consumer can prove about cardinality" and relate the example -- to me, whether something can be null or not is a constraint (logical fact) not an estimate (cardinality).

andrew-coleman: Sure, I agree that second sentence is confusing as written.  I think it's still worth clarifying that analysis can't narrow the output type.  Perhaps the following is clearer:

> A join's output field types are determined by the join type and the input schemas alone. 
> They are never narrowed by analysis: where the join type makes a field nullable, it stays nullable even if the join condition, the post-join filter, or the data guarantees that no null is produced. A `Left Single` whose condition matches every left record still has a nullable right side.

Happy to remove it entirely if you don't agree though.

yongchul: Thank you! The rewrite reads fine to me.

yongchul (approved):

Issue #1185, opened 2026-08-25
Standard extension impls share integer type parameters across independent arguments


nielspardon:
## Problem

`site/docs/expressions/scalar_functions.md` defines shared parameter names as a binding constraint:

> When the same parameter name is used multiple times in a function definition, the function can only bind if the exact same value is used for all parameters of that name.

Several standard extension impls use the *same* integer parameter name in two **independent** argument positions. Where the two arguments are unrelated quantities, that constraint is almost certainly unintended: it makes the impl unbindable unless the caller's arguments happen to carry identical precision/length.

This stayed invisible for years because implementations that ignore type parameters never enforced it. It surfaces once a binding implementation derives parameterized return types: calls like `strpos(varchar(100), varchar(3))` and `lte(precision_timestamp(3), precision_timestamp(6))` then fail to resolve.

The clearest proof that at least some of these are typos is `strpos`, whose two parameterized impls disagree with each other — the `varchar` impl shares `L1` while the `fixedchar` impl right below it uses `L1`/`L2` for the same two arguments.

## Sweep

This finds every impl sharing an integer parameter across two or more arguments. On `main` at `de9e632` it reports **24 impls**:

```bash
cd extensions && python3 - <<'PY'
import re, glob
pat = re.compile(r'<([^<>]*)>')
for f in sorted(glob.glob('*.yaml')):
    lines = open(f).read().split('\n')
    fn = impl = ret = None; args = []
    def flush():
        if len(args) < 2: return
        per = []
        for ln, val in args:
            ps = [t.strip() for m in pat.finditer(val) for t in m.group(1).split(',')
                  if re.fullmatch(r'[A-Z]\d*', t.strip())]
            per.append((ln, val, ps))
        flat = [p for _, _, ps in per for p in ps]
        shared = sorted(p for p in set(flat) if sum(1 for _, _, ps in per if p in ps) > 1)
        if shared:
            print(f"{f}:{impl} fn={fn} shared={shared} return={ret}")
            for ln, val, _ in per: print(f"    L{ln}: {val}")
    for i, l in enumerate(lines, 1):
        s = l.strip()
        if re.match(r'-?\s*name:\s*"?[A-Za-z0-9_]+"?$', s) and (l.startswith('    name:') or l.startswith('  - name:')):
            flush(); fn = s.split('name:')[1].strip().strip('"'); args = []; ret = None; impl = i
        elif re.match(r'^\s+- args:$', l):
[...]

PR #1186, opened 2026-08-25
fix(extensions): split shared type parameters in strpos and logb

nielspardon:
Two impls in the standard extensions declare the same integer type parameter in two independent argument positions. Per the binding rule in `scalar_functions.md` — "the function can only bind if the exact same value is used for all parameters of that name" — each is bindable only when the caller's two arguments happen to carry identical length or precision.

**`strpos`, varchar impl.** Both `input` and `substring` were `varchar<L1>`, forcing the needle's declared length to equal the haystack's. `strpos(varchar(100), varchar(3))` could not resolve. The `fixedchar` impl immediately below already used `L1`/`L2` for the same two arguments, so the catalog contradicted itself; the `fixedchar` form is the correct one.

**`logb`, decimal impl.** Both `x` and `base` were `decimal<P1,S1>`, forcing the logarithm base to carry the value's precision *and* scale. The `fp32` and `fp64` impls of the same function impose no relationship between the two arguments, and there is no reason a base should share a representation with the value.

Both impls return a type that does not mention the shared parameter — `i64` for `strpos`, `fp64` for `logb` — so nothing in output derivation depends on the two arguments agreeing.

## Compatibility

Splitting a shared parameter is a pure relaxation. Any call that bound under the old signature still binds under the new one with an identical output type; the change only admits argument combinations that were previously rejected. The compact function signature is derived from type classes rather than parameter names, so `strpos:vchar_vchar` and `logb:dec_dec` are unchanged — this is wire-compatible, source-compatible, and not semantically breaking, and no active library needs a companion migration.

## Scope

This fixes only the two cases where the shared parameter is unambiguously a typo. #1185 documents the full sweep — 24 impls on `main` — and deliberately leaves the other 22 alone:

- 20 impls (12 temporal comparisons, `add`/`subtract`/`add_intervals`, `round_temporal`) all reduce to one policy question: whether Substrait admits temporal operands of differing precision or requires producers to cast. #926 owns that question for the arithmetic subset and #927 was closed in favour of casting. #1185 adds two arguments specific to the comparisons that did not come up there, but it is a design discussion and does not belong in a YAML diff.
- 2 impls are unrelated defects that renaming a parameter would not fix: a duplicate `origin` argument in two `round_calendar` impls, and `center`'s return of `varchar<L1>`, whose correct width depends on a value argument and so cannot be derived at all.

#1185 also corrects the record on `max(P, Q)`, which #926 treats as possibly inexpressible: it is valid today, parses against the committed grammar, and is already used for decimal arithmetic. No part of the deferred group is blocked on the expression language.

Closes nothing on its own; #1185 tracks the remainder.

- - -
This change is [Reviewable](https://reviewable.io/reviews/substrait-io/substrait/1186)

### DIFF
--- a/extensions/functions_logarithmic.yaml
+++ b/extensions/functions_logarithmic.yaml
@@ -202,7 +202,7 @@ scalar_functions:
           - value: decimal<P1,S1>
             name: "x"
             description: "The number `x` to compute the logarithm of"
-          - value: decimal<P1,S1>
+          - value: decimal<P2,S2>
             name: "base"
             description: "The logarithm base `b` to use"
         options:
--- a/extensions/functions_string.yaml
+++ b/extensions/functions_string.yaml
@@ -628,7 +628,7 @@ scalar_functions:
           - value: "varchar<L1>"
             name: "input"
             description: The input string.
-          - value: "varchar<L1>"
+          - value: "varchar<L2>"
             name: "substring"
             description: The substring to search for.
         options:


benbellick (approved):
LGTM. Can we simplify the PR body so that it makes for a concise commit message? Thanks!

yongchul (approved):

PR #1191, opened 2026-08-28
fix(extensions): remove duplicate origin arguments from round_calendar

fallintoplace:
## What changed

- Remove the extra `origin` value argument from the `date` overload.
- Remove the extra `origin` value argument from the `precision_time<P>` overload.
- Add a check that catches repeated named arguments in extension signatures.

The enum `origin` argument and `multiple` argument stay unchanged. These overloads now expose the intended `round_calendar` signatures without requiring a second origin value.

Follow-up to #1186.

- - -
This change is [Reviewable](https://reviewable.io/reviews/substrait-io/substrait/1191)

### DIFF
--- a/tests/type/test_type_grammar.py
+++ b/tests/type/test_type_grammar.py
@@ -246,3 +246,29 @@ def test_extension_yaml_type_names_are_defined():
                 )
 
     assert failures == [], "\n".join(failures)
+
+
+def test_extension_yaml_argument_names_are_unique():
+    """No checked-in function implementation repeats a named argument."""
+    repo_root = Path(__file__).parents[2]
+    failures = []
+    for path in extension_yaml_files():
+        with path.open() as f:
+            extension = yaml.load(f, Loader=yaml.FullLoader)
+
+        for section in ("scalar_functions", "aggregate_functions", "window_functions"):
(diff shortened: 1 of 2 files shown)


tests/type/test_type_grammar.py
```diff
@@ -246,3 +246,29 @@ def test_extension_yaml_type_names_are_defined():
                 )
 
     assert failures == [], "\n".join(failures)
+
+
+def test_extension_yaml_argument_names_are_unique():
```
benbellick, line 251: This is a great change / test, but I don't think the docs actually spell out explicitly that (optional) names for arguments to functions need to be unique. Perhaps we should add that in this PR as well then?

nielspardon: The other angle here is that we need better documentation of function impls and their arguments (and testcases). 

I took a look at this family functions a while back and the PR that introduced them had a lot of discussions but it was unclear to me how the different arguments interact with each other. This made it difficult to understand whether `origin` was added twice for a good reason since the original author meant two different origins or whether they introduced effectively a typo and the `origin` parameter is actually redundant.

In either case requiring them to be unique might be a good starting point to make at parameter names less ambiguous.

PR #1193, opened 2026-08-29
fix(extensions): bind quantile return type

### LINKED ISSUE #1150: extensions: `quantile` return type uses a plain `any`, which cannot be bound
## Problem

`quantile` declares its return type using a **plain** `any`, which the spec says carries no identity, so the return type is not determined by anything.

`extensions/functions_arithmetic.yaml` (in the `aggregate_functions:` section):

```yaml
          - value: any            # L1740 — argument "distribution"
            name: distribution
...
        nullability: DECLARED_OUTPUT
        ordered: true
        return: LIST?<any>        # L1766
```

Per `site/docs/extensions/index.md:139`:

> The `any` type indicates that the argument can take any possible type. In the `foo` function above, arguments `a` and `b` can be of any type, **even different ones in the same function invocation**.

and L144:

> The `any[\d]` types (i.e. `any1`, `any2`, ..., `any9`) impose an additional restriction. Within a single function invocation, all `any` types with the same numeric suffix _must_ bind to the same type.

So a plain `any` binds independently at each occurrence and is deliberately *not* tied to any other occurrence. A plain `any` in a **return** position therefore has nothing to bind to: an implementation deriving the output type from the declaration cannot resolve `LIST?<any>`, and two conforming implementations could legitimately produce different answers.

## Scope

This is the only occurrence in the standard catalog. Scanning every `extensions/*.yaml` for a plain `any` (no digit) in a `return:` or `intermediate:` position yields exactly one hit:

```
extensions/functions_arithmetic.yaml:1766:        return: LIST?<any>
```

The five plain-`any` *argument* declarations are all fine, since independent binding is meaningful for inputs:

```
extensions/functions_aggregate_approx.yaml:15
extensions/functions_aggregate_generic.yaml:11
extensions/functions_aggregate_decimal_output.yaml:10
extensions/functions_aggregate_decimal_output.yaml:38
extensions/functions_arithmetic.yaml:1740
```

## Suggested fix

Change both the argument and the return to `any1`, so the return is bound by the `distribution` argument:

```yaml
          - value: any1
            name: distribution
...
        return: LIST?<any1>
```

This looks like it was simply missed when #618 clarified the `any` vs `any[\d]` distinction (2024-04) — `quantile` predates that clarification.

_Found while validating a Java implementation's function-binding type derivation against the spec: `quantile` is the only aggregate in the standard catalog whose output type cannot be derived from its declaration._

bvolpato:
Fixes #1150.

`quantile` accepts its distribution as unbound `any` and returns `LIST?<any>`. Anonymous `any` does not create a reusable type binding, so consumers cannot derive a concrete list element type from the input. This surfaced while substrait-java added aggregate output type derivation in https://github.com/substrait-io/substrait-java/pull/1017.

Bind both positions through `any1`. Accepted input types remain unchanged, while the declared return type becomes derivable from the distribution argument.

- - -
This change is [Reviewable](https://reviewable.io/reviews/substrait-io/substrait/1193)

## Summary by CodeRabbit

* **Bug Fixes**
  * Corrected the type declarations for the `quantile` aggregate function to improve type consistency for distribution inputs and list results.

### DIFF
--- a/extensions/functions_arithmetic.yaml
+++ b/extensions/functions_arithmetic.yaml
@@ -1737,7 +1737,7 @@ aggregate_functions:
             description: >
               A positive integer which defines the number of quantile
               partitions.
-          - value: any
+          - value: any1
             name: distribution
             description: >
               The data for which the quantiles should be computed.
@@ -1763,7 +1763,7 @@ aggregate_functions:
             values: [ TIE_TO_EVEN, TIE_AWAY_FROM_ZERO, TRUNCATE, CEILING, FLOOR ]
         nullability: DECLARED_OUTPUT
         ordered: true
-        return: LIST?<any>
+        return: LIST?<any1>
 
 window_functions:
   - name: "row_number"


benbellick:
@coderabbitai review

yongchul (approved):

benbellick (approved):

PR #1195, opened 2026-08-29
ci: compile rewritten protobufs without task cache

bvolpato:
Proto-prefix CI rewrites protobuf files into `output/` and points the Buf workspace there. It then invokes `generate-protobuf`, whose Pixi cache inputs only cover `proto/**/*.proto`. With a primed cache, Pixi skips that task and never compiles rewritten files.

Invoke `buf generate` directly in this job so compilation always uses the rewritten workspace. Fresh Actions jobs normally miss this cache today, but direct execution makes the check reproducible and prevents future cache restoration from turning it into a false green.

- - -
This change is [Reviewable](https://reviewable.io/reviews/substrait-io/substrait/1195)

## Summary by CodeRabbit

- **Chores**
  - Updated the protobuf validation workflow to use the current generation command.

### DIFF
--- a/.github/workflows/pr.yml
+++ b/.github/workflows/pr.yml
@@ -132,4 +132,4 @@ jobs:
           echo "directories:" >> buf.work.yaml
           echo "  - output" >> buf.work.yaml
       - name: Compile rewritten proto files
-        run: pixi run generate-protobuf
+        run: pixi run buf generate


benbellick:
@coderabbitai review

benbellick (approved):

benbellick:
This is a CI modification so only 1 PMC + 1 committer (@bvolpato) necessary.

PR #1196, opened 2026-08-29
test(extensions): reject unknown dependency URNs

bvolpato:
Function test headers declare `SUBSTRAIT_DEPENDENCY` entries for helper extensions. Coverage validation checked only `SUBSTRAIT_INCLUDE`, so a misspelled dependency remained stored but silently contributed no validation failure. This allowed malformed test metadata to stay green.

Validate every declared dependency through the existing extension registry before processing cases. Focused coverage preserves multiple valid dependencies and rejects an unknown URN with its exact value.

- - -
This change is [Reviewable](https://reviewable.io/reviews/substrait-io/substrait/1196)


benbellick (approved):

benbellick:
Merging with 1 PMC approval as this is a CI change

PR #1204, opened 2026-09-02
ci: render the pixi.lock diff into the job summary

nielspardon:
[§1 of Pixi's security guide](https://pixi.prefix.dev/latest/security/#1-make-dependency-resolution-reproducible) puts reviewing `pixi.lock` diffs first, ahead of every other measure: the lock is what pins the exact artefacts and their sha256, so a change to it is the moment a new package enters our builds.

That review is currently impossible in the GitHub UI. `.gitattributes` marks `pixi.lock` with `-diff`, so git and GitHub render it as a binary blob with no line-level diff at all — a lock change shows up as "Bin 211861 -> 273748 bytes" and nothing more. The `-diff` attribute is there for good reason (it also carries `merge=binary` to stop three-way merges producing a corrupt lock), so the fix is to render the diff another way rather than to remove it.

Renovate compounds this by design: #1075 groups all `pyproject.toml` bumps into one PR, because the conda and PyPI dependencies share a single lock and ungrouped PRs would each rewrite it and conflict. So the one file a reviewer most wants to read is both invisible and carrying several bumps at once.

## How it works

A path-filtered workflow runs only when `pixi.lock` itself changes, and appends the rendered diff to the run's job summary:

- `pixi-diff` and `pixi-diff-to-markdown` (both on conda-forge, all four of our platforms) go into a `lock-diff` pixi environment rather than being fetched at run time, so the renderer is version-pinned in the same lock it reports on. The environment sets `no-default-feature = true`: it does not need python, buf, ANTLR or Node.js, and `pixi-diff-to-markdown` requires python >=3.14, which the default feature's `>=3.13,<3.14` pin would otherwise block.
- `ci/lock_diff.sh` takes a base revision, so the same rendering is available locally as `pixi run lock-diff [<rev>]` — useful when preparing or reviewing a bump.
- `--manifest-path pyproject.toml` lets `pixi-diff` separate explicit dependencies from transitive ones, so the two or three lines a human declared are shown expanded and the hundreds that follow from them are collapsed behind a `<details>` element.

## Job summary rather than a comment

A comment needs `pull-requests: write`, which cuts directly against #1203 (least-privilege tokens for the same reason this series exists). The job summary is rendered markdown, keeps `<details>` collapsing, is attached to the run that produced it, and needs nothing beyond `contents: read`. If maintainers would rather have the diff inline in the PR conversation, that trade-off is worth making deliberately.

## Notes for review

- The lock change here is purely additive: the new environment and its packages, with no line removed from the existing content and `version: 7` unchanged.
- The workflow is new, so it is not among the repository's required checks. On a PR that does not touch `pixi.lock` it does not run at all, which would leave a required check permanently pending if it were ever added as one.

- - -
This change is [Reviewable](https://reviewable.io/reviews/substrait-io/substrait/1204)

### DIFF
--- a/.github/workflows/lock_diff.yml
+++ b/.github/workflows/lock_diff.yml
@@ -1,0 +1,30 @@
+name: Lock File Diff
+
+on:
+  pull_request:
+    paths:
+      - pixi.lock
+
+# Rendering the diff only reads the repository; the result goes to the job
+# summary rather than a comment so that no write scope is needed.
+permissions:
+  contents: read
+
+jobs:
+  lock-diff:
+    name: Render pixi.lock diff
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v7
+        with:
+          # the base revision is needed to diff against
+          fetch-depth: 0
+      - uses: prefix-dev/setup-pixi@v0.10.2
+        with:
+          pixi-version: v0.73.0
+          cache: true
+          environments: lock-diff
+      - name: Render pixi.lock diff
+        run: pixi run lock-diff "$BASE_SHA" >> "$GITHUB_STEP_SUMMARY"
+        env:
+          BASE_SHA: ${{ github.event.pull_request.base.sha }}
--- a/ci/lock_diff.sh
+++ b/ci/lock_diff.sh
@@ -1,0 +1,34 @@
+#!/usr/bin/env bash
+# shellcheck shell=bash
+# SPDX-License-Identifier: Apache-2.0
+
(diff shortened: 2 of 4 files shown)


jamietanna:
Might be worth showing what the update would look like on this PR?

nielspardon:
> Might be worth showing what the update would look like on this PR?

Good idea. I'm personally not 100% convinced how useful this specific check is going to be. Seeing an example might make it easier to decide whether this is something we want to include or not.

PR #1205, opened 2026-09-02
ci: scan the installed pixi environments weekly

nielspardon:
We have no vulnerability scanning of the conda half of the toolchain, and no dependency bot can give us one: GHSA and OSV have no conda ecosystem, so neither dependabot nor Renovate reports a conda advisory. [§5 of Pixi's security guide](https://pixi.prefix.dev/latest/security/#5-scan-the-installed-environment-directly) is explicit that scanning the installed prefix is the workaround until conda packages carry PURLs (conda/ceps#63).

This adds a weekly scheduled scan of the installed environments.

## How it works

- A `scan` pixi environment holds `syft` and `grype` (both on conda-forge, all four of our platforms), so the scanners are version-pinned in `pixi.lock` rather than fetched from Anchore's actions at run time — one toolchain, updated the same way as everything else. `no-default-feature = true` keeps ~200 MB of Go binaries out of `default` and `dev`.
- `ci/scan_environments.sh` runs over `.pixi/envs/*`, so it covers whatever environments are declared without a list to maintain in the workflow, and it is runnable locally as `pixi install --all && pixi run scan`. The catalogers are selected explicitly (`+conda-meta-cataloger`, `+cargo-auditable-binary-cataloger`) and Grype consumes Syft's native JSON rather than a CycloneDX conversion, both as the guide recommends.
- `.grype.yaml` pins `match.stock.using-cpes: true`. conda packages get CPEs but generally not PURLs, so they are matched through the stock matcher's CPE path. That is Grype's default today; pinning it follows the same reasoning as #1202 — a security-relevant default we depend on should be visible in the repository.
- The report goes to the job summary, with the highest-risk rows shown and the rest collapsed. SBOMs and full reports are uploaded as an artifact for triage.

## How failures surface, and what is left to maintainers

The workflow **reports without failing**. `FAIL_ON` is unset, so Grype's threshold check is not armed, and a run stays green regardless of findings. That is deliberate rather than timid: the first scan already finds plenty, and a job that is red from the day it lands gets ignored rather than triaged.

Baseline from a local run of this branch, by severity:

| environment | critical | high | medium | low | other | total |
|-|-|-|-|-|-|-|
| `default` | 15 | 107 | 79 | 25 | 4 | 230 |
| `dev` | 17 | 138 | 104 | 28 | 4 | 291 |
| `scan` | 0 | 4 | 2 | 0 | 1 | 7 |

Concentrated in a handful of packages: `openssl` (27 distinct CVEs at high or critical, currently 3.6.1 against fixes in 3.6.3/3.6.4), `python` (11), and the Go module graph inside the `buf` binary plus the `zulu` JRE that ANTLR pulls in. `openssl` and `python` are each reported twice, once from `conda-meta-cataloger` and once from Syft's binary classifier — the same finding seen through two catalogers, not two problems.

Three decisions are maintainers' to make, and I have not taken any of them:

1. **Where the failure threshold belongs, if anywhere.** Setting `FAIL_ON: high` in the workflow env is a one-line change once the baseline above is triaged.
2. **Whether this should ever gate a pull request.** It is scheduled-only here. Gating PRs on a scanner whose database changes daily means a PR can go red without the PR changing anything, so it is a policy call rather than a technical one.
3. **Whether findings should be filed as issues.** An auto-filed issue is far more visible than a green run whose summary nobody opens, but it needs `issues: write` on a workflow that runs third-party package code, which cuts against #1203. Left out on purpose; the reporting knobs are `FAIL_ON` and, for noise, Grype's `--only-fixed`.

## Notes

- `pixi.lock` grows by the two scanner packages only; no existing line is removed and `version: 7` is unchanged. It touches the same file as #1204, so whichever lands second needs a `pixi lock` re-run.
- The cron minute is deliberately not `:00`.

- - -
This change is [Reviewable](https://reviewable.io/reviews/substrait-io/substrait/1205)

### DIFF
--- a/.github/workflows/security_scan.yml
+++ b/.github/workflows/security_scan.yml
@@ -1,0 +1,35 @@
+name: Security Scan
+
+on:
+  schedule:
+    # weekly, Tuesday 05:17 UTC -- off the hour so it does not pile onto the
+    # top-of-hour scheduling peak
+    - cron: "17 5 * * 2"
+  workflow_dispatch:
+
+# Reads the repository and reports to the job summary; nothing is written back.
+permissions:
+  contents: read
+
+jobs:
+  scan:
+    name: Scan installed environments
+    runs-on: ubuntu-latest
+    if: ${{ github.repository == 'substrait-io/substrait' }}
+    steps:
+      - uses: actions/checkout@v7
+      - uses: prefix-dev/setup-pixi@v0.10.2
+        with:
+          pixi-version: v0.73.0
+          cache: true
+      - name: Install every environment
+        run: pixi install --all --locked
+      - name: Scan installed environments
+        run: pixi run scan >> "$GITHUB_STEP_SUMMARY"
+      - name: Upload SBOMs and reports
+        if: ${{ !cancelled() }}
+        uses: actions/upload-artifact@v7
+        with:
+          name: sbom
+          path: sbom/
+          retention-days: 30
--- a/.gitignore
+++ b/.gitignore
@@ -11,3 +11,6 @@ __pycache__
 
 # uv virtual environment
 .venv
+
+# SBOMs and vulnerability reports from ci/scan_environments.sh
+sbom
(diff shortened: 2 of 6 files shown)


nielspardon:
I feel like this should be opening issues when finding something and assign the issues to CODEOWNERS aka the PMC.

Issue #1187, opened 2026-08-25
Add support for window function testing


Slimsammylim:
We currently only have test support for [scalar and aggregate functions](https://github.com/substrait-io/substrait/tree/main/tests/cases). Let’s extend our testing framework to support window functions.

## Proposed Solution
Static Frame Tests:
* Defines one frame directly, as a table whose rows are written in the order the function observes them
* Asserts what a window function returns at each position in that frame
* The frame is fixed: no partitioning or bounds clause to compute it from
* The test supplies the window function's input exactly as the function sees it

## Test Format
* Frame: 
  * The input a window function observes, defined as a table. The function is evaluated once at each position in the frame, with the whole frame in scope.
* OVER: names the input frame the function is evaluated against.
* Input:
  * Comma-separated list of arguments to the function. The arguments can be literals or column references.
* Output:
  * The expected result of the function. Either SUBSTRAIT_ERROR or a list of literal values.

* Template for window functions:

```
DEFINE f1(<type>) = (<value>, ..., <value>)
window_func(col0) OVER f1 = (<result>::<result_type>, ..., <result>::<result_type>)
```

## Example
```
### SUBSTRAIT_WINDOW_TEST:V1
### SUBSTRAIT_INCLUDE: extension:io.substrait:functions_arithmetic

# lag tests
DEFINE f1(i32) = [1998::i32, 1999::i32, 2000::i32, 2001::i32]
lag(col0) OVER f1 = (Null::i32?, 1998::i32?, 1999::i32?, 2000::i32?)

DEFINE f1(i32) = [1998::i32, 1999::i32, 2000::i32, 2001::i32]
lag(col0, 2::i32) OVER f1 = (Null::i32?, Null::i32?, 1998::i32?, 1999::i32?)

DEFINE f1(i32) = [1998::i32, 1999::i32, 2000::i32, 2001::i32]
lag(col0, 2::i32, 5::i32) OVER f1 = (5::i32?, 5::i32?, 1998::i32?, 1999::i32?)

# row_number tests
DEFINE f1(i32) = [1998::i32, 1999::i32, 2000::i32, 2001::i32]
row_number() OVER f1 = [1, 2, 3, 4]
```

Example SQL representation:
```
# Static frame test
DEFINE f1(i32) = [1998, 1999, 2000, 2001]
lag(f1) = (Null::i32?, 1998::i32?, 1999::i32?, 2000::i32?)

# SQL version
SELECT lag(v) OVER (
[...]

PR #1209, opened 2026-09-02
feat: add support for window function testing

Slimsammylim:
Resolves issue #1187.

This PR adds support for window function tests. Specifically, it adds support for static frame tests. It also includes two new test files for `lag` and `row_number`.

**Static frame test:**
* Define a frame: the data a window function observes, as a table (same table syntax used by aggregate tests).
* The rows are written in the order the function observes them.
* The test asserts what the window function returns at each position in that frame.

**Test format examples:**
```
### SUBSTRAIT_WINDOW_TEST:V1
### SUBSTRAIT_INCLUDE: extension:io.substrait:functions_arithmetic

# lag tests
DEFINE f1(i32) = ((1998), (1999), (2000), (2001))
lag(col0) OVER f1 = (Null, 1998, 1999, 2000)::i32?

DEFINE f1(i32) = ((1998), (1999), (2000), (2001))
lag(col0, 2::i32) OVER f1 = (Null, Null, 1998, 1999)::i32?

# row_number tests
DEFINE f1(i32) = ((1998), (1999), (2000), (2001))
row_number() OVER f1 = (1, 2, 3, 4)::i64?
```

- - -
This change is [Reviewable](https://reviewable.io/reviews/substrait-io/substrait/1209)

## Summary by CodeRabbit

- **New Features**
  - Added support for defining and parsing window-function tests.
  - Supports `OVER` clauses, window arguments, typed results, and expected errors.
  - Added coverage for `lag`, `row_number`, empty inputs, offsets, default values, and multi-column windows.

- **Documentation**
  - Updated test-format documentation with window-test syntax, frame behavior, and examples.

- **Tests**
  - Added validation for frame references and result/frame row-count mismatches.


tests/README.md
Slimsammylim, line 176: This has been outdated for a while for both scalar and aggregate functions, so perhaps its better to update in a separate PR(?)

tests/README.md
```diff
@@ -123,6 +128,44 @@ A testcase with mixed arguments
     ((20), (-3), (1), (10)) LIST_AGG(col0::fp32, ','::string) = 1::fp64
 ```
 
+### Window Test Cases
+A test case consists of the following elements:
+- **frame definition**: `DEFINE <frame-name>(<datatype>(, <datatype>)*) = ((<literal>(, <literal>)*), ...)`
```
benbellick, line 133: Probably good to avoid this pseudo grammar stuff since we removed it in https://github.com/substrait-io/substrait/pull/1212. Maybe here we just list an example instead?

tests/README.md
```diff
@@ -123,6 +128,44 @@ A testcase with mixed arguments
...
+    - A table (same syntax as aggregate tables) the function observes.
+    - Rows are written in the exact order the function sees them, so there's no separate partitioning, ordering, or bounds clause.
+- **function**: The name of the function being tested. The function name must be an identifier alphanumeric string.
```
benbellick, line 136: Better to say that it is a function defined in the YAML file with URN matching the include statement, as opposed to specifying what is allowable

tests/README.md
```diff
@@ -123,6 +128,44 @@ A testcase with mixed arguments
...
+- **arguements**: Comma-separated list of arguments to the function. The arguments can be literals or a bare column reference (e.g. `col0`) into the frame.
```
benbellick, line 137:
```suggestion
- **arguments**: Comma-separated list of arguments to the function. The arguments can be literals or a bare column reference (e.g. `col0`) into the frame.
```
Does the aggregate section explain how the column naming works? We should reference that or explain it here. Just like a very simple the first col is `col0`, second is `col1`, etc.

tests/README.md
```diff
@@ -123,6 +128,44 @@ A testcase with mixed arguments
...
+- **over**: Names the input frame the function is evaluated against. Must match the frame name given in the frame definition.
+- **options**: Optional comma-separated list of options in key:value format. The options describe the behavior of the function. The test should be run only on dialects that support the options. If options are not specified, the test should be run for all permutations of the options.
+- **result**: The expected result of the function, written as a column of data using the same format as aggregate function arguments, giving one value per row of the frame in row order (e.g. `(1, 2, 3, 4)::i64?`), or `SUBSTRAIT_ERROR`.
```
benbellick, line 140: I think in scalar tests at least, the error token is `<!ERROR>`. Can we use that here too?

Slimsammylim: Yes! Let me switch that for all the test types.

tests/README.md
Slimsammylim: Yes, let me add that!

tests/coverage/visitor.py
benbellick, line 1: I'm not sure that we validate that the row shape matches the declared column shape. Something like this should throw an error on parse:
```
 DEFINE f1(i32, i32) = ((1), (2))
```
does it work in the current setup?

benbellick: FYI I opened up this pr (https://github.com/substrait-io/substrait/pull/1231) which is relevant. Basically we should separate out the parser code from the coverage code. No need to wait on that one though, its okay for now to add it here.

tests/parser/visitor.py
Slimsammylim: Hm, that does work with the current set up. A frame is just a table so that means we will need to add that validation for tables in general. I'll work on that and perhaps that's better as a separate PR.

tests/README.md
```diff
@@ -145,6 +128,45 @@ A testcase with mixed arguments
     ((20), (-3), (1), (10)) LIST_AGG(col0::fp32, ','::string) = 1::fp64
 ```
 
+### Window Test Cases
+A test case consists of the following elements:
+- **frame definition**:
```
yongchul, line 157: Just curious about the next step of this. Do you think adding a testing how we produce frames (i.e., ROWS and RANGE)?

yongchul (approved):

tests/parser/visitor.py
Slimsammylim: I opened an issue and #1234 and set up a PR to address this: [https://github.com/substrait-io/substrait/pull/1235](https://github.com/substrait-io/substrait/pull/1235)

tests/README.md
Slimsammylim: Yes, the next step would be to add validation for how we are producing frames!

benbellick (approved):
Nice! There is a merge conflict from the PR I just merged in, sorry about that! But after that we can get this in 💯

tests/parser/visitor.py
```diff
@@ -152,8 +184,34 @@ def visitMultiArgAggregateFuncCall(
...
+            group=None,
+            options={},
+            rows=rows,
+            args=args,
+            result=SubstraitError("uninitialized"),
+            comment="",
+        )
+
+    def visitWindowFuncCall(self, ctx: FuncTestCaseParser.WindowFuncCallContext):
+        table_name, column_types, rows = self.visitTableData(ctx.tableData())
```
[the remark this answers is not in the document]
benbellick: I believe this will be the followup issue here: https://github.com/substrait-io/substrait/issues/1234

PR #1215, opened 2026-09-10
fix(extensions): merge duplicate regexp function definitions

strausza:
Since #750, `regexp_match_substring`, `regexp_count_substring` and
`regexp_replace` have each had two top-level definitions in
`functions_string.yaml`, one long-arity and one short-arity. Compound names stay
unique so plans are unaffected, but anything resolving a function by `name` — a
dialect file entry, for instance — picks one definition and silently misses the
other's impls, the generated docs render each function twice, and the paired
descriptions have already diverged.

Optional trailing arguments are modelled elsewhere as impls of a single
definition (`substring`, `lead`/`lag`, `strptime_timestamp`). This does the same
here: the short-arity impls move under the existing definition and the defaults
are described once.

```yaml
- name: regexp_match_substring
  description: >-
    ... (existing text) ...
    If `position`, `occurrence` and `group` are omitted they default to `1`, `1`
    and `0` respectively, i.e. the first occurrence from the beginning of the
    input string, returning the substring matching the full regular expression.
  impls:
    - args: [varchar<L1>, varchar<L2>, i64, i64, i64]   # existing
      ...
    - args: [string, string, i64, i64, i64]             # existing
      ...
    - args: [varchar<L1>, varchar<L2>]                  # new
      ...
    - args: [string, string]                            # moved
      ...
```

`regexp_count_substring` defaults `position` to `1`; `regexp_replace` defaults
`position` and `occurrence` to `1` and `0`, its short form replacing every match
rather than the first.

The short forms also gain the `varchar` and `fixedchar` impls their long forms
already offered, and every impl of the three functions now has a test case.

Compound names and semantics are unchanged and the new impls are additive, so
this is not a breaking change. `num_scalar_functions` falls in
`tests/baseline.json` because the duplicate names were previously counted as
separate entries.

- - -
This change is [Reviewable](https://reviewable.io/reviews/substrait-io/substrait/1215)


extensions/functions_string.yaml
```diff
@@ -149,7 +149,9 @@ scalar_functions:
       expression capture group can be specified using the `group` argument. Specifying `0`
       will return the substring matching the full regular expression. Specifying `1` will
       return the substring matching only the first capture group, and so on. The `group`
-      argument should be a non-negative integer.
+      argument should be a non-negative integer. If `position`, `occurrence` and `group` are
```
yongchul, line 152: can we be explicit about which is which? is (position, occurrence) -> 1 and (group) -> 0?

Issue #1217, opened 2026-09-11
Support deeply nested plan


yongchul:
# Motivation

Substrait happens to adopt protobuf as an encoding of the protocol, which has a quarky settings called recursion depth limit. The limit is there for a good reason but for a natively nested data structure like query plan, this is just sheer PITAS to dealing with very large plan.

Substrait does offer a way to breaking large relational tree -- ReferenceRel. But there is no such thing for an expression so if you have a big expression tree hanging around leaf, you are hitting a dead end if you haven't chopped your plan proactively somewhere above or you are facing a dead end if your expression is 50 deep (substrait encoding has extra nesting overhead per expression and per rel) with the default limit 100 (in C++/Java).

# Proposal: detached expression and detached rel in PlanRel

Introducing detached expressions and detached rels with referencing mechanism work around this mechanism in a following way.

* Along the serialization, consumers can choose reasonable number to override the protobuf recursion depth limit (like 4x of the *chunking* depth of relational tree and or expression tree to be conservative), then simply serialize the rels/expressions into a new detached rels/expressions, with reference.
* Why under PlanRel? The detached rel/expressions are purely encoding artifacts (i.e., they make sense only a single point of the plan) and not intended for sharing. Thus, keep them under PlanRel would make it clear that these are needed to deserialize the wrapping PlanRel (i.e., plan rel is self-contained modulo really shared stuff).
* Stiching is easy: deserialize the proto, then just follow over the link. You only need to pass around the PlanRel message as a context to resolve the correct deserialization or anchoring.

# Alternative 1: Raise the limit!

You will face the another problem like stack overflow. Thank you.

# Alternative 2: Flatten the world!

You could go another extreme by modeling all linkage of operator as id based. That would be Substrait v2 or something-else. That's too much change in how we represent the plan from completely nested to flat. If you want to mix the two, -100 from me.

# Extra credit: streaming plan

This requires more careful coordination of ordering of the messages and tags.

# Survey and design guidance -- Powered by GH Copilot

For very deep logical nesting, the robust general solution is to keep the protobuf message nesting shallow and represent deeper relationships using IDs or indexes. The logical tree may then be arbitrarily deep without requiring the protobuf parser to recurse once per logical level.

## Representations used by existing projects

[...]

PR #1218, opened 2026-09-11
feat(proto): support detached expression subtrees

yongchul:
Allow producers to bound protobuf recursion depth by moving expression subtrees into `PlanRel.detached_expressions` and replacing them with zero-based `Expression.detached_expression_ordinal` references.

Detached ordinals are encoding-only and local to their containing `PlanRel`. Every detached expression must be reachable, referenced exactly once, and part of an acyclic reference graph. Fully detaching recursive expression children provides an effectively flat expression encoding when portable bounded nesting is required.

- - -
This change is [Reviewable](https://reviewable.io/reviews/substrait-io/substrait/1218)

## Summary by CodeRabbit

* **New Features**
  * Plans can reference detached expressions by ordinal, supporting recursive expression trees with bounded nesting depth.
  * Detached expressions retain reachability and unique-reference validation requirements.

* **Documentation**
  * Updated binary serialization guidance for bounded expression nesting.
  * Added a plan relation example demonstrating a detached expression reference.

* **Tests**
  * Expanded validation coverage for detached expression ordinals, bounds, and referenced literal values.


proto/substrait/algebra.proto
```diff
@@ -687,6 +687,9 @@ message Rel {
     ExchangeRel exchange = 15;
     ExpandRel expand = 16;
     TopNRel top_n = 23;
+    // Zero-based ordinal of an encoding-only reference into the containing
+    // PlanRel.detached_rels.
+    uint32 detached_rel_ordinal = 25;
```
benbellick, line 692: I'm not sure why we need this new field. Is using the existing `ReferenceRel` insufficient?

yongchul: Hahaha... I got exactly the same question internally 😆 

Mainly two reasons.
1. In-memory overhead reduction: using reference rel will end up creating another protobuf message in memory. If you imagine completely flat representation, this is a significant overhead in memory. See  https://github.com/substrait-io/substrait/issues/1217#issuecomment-5650127174 for more detail.
2. There is another subtlety. PlanRel is a boundary intended to be shared. I prefer to keep it in that way and this detached expressions and rels as an encoding artifact and local to within a PlanRel rather than a global level stitching (i.e., relevant fragments are local to each PlanRel thus a lot easier to **slice a large Plan for distributed execution**).

From dealing with a mega plan, (1) is primary reason but (2) is also a good separation of concern IMO. If there are strong push back from community to not introduce `detached_rel_ordinal` but use `ReferenceRel`, that is fine as long as we keep the ordinal as part of oneof so that we do not incur extra overhead for full flat representation.

Note that these deep plans in both relational and scalar that we do face in production.

jcamachor: This change is only being proposed to address concerns related to the underlying protobuf implementation used to deserialize these messages, so I do think there's value in having a separate mechanism for this split vs. existing mechanisms such as `reference_rel` which may be used by a producer to signal reuse across consumers.

yongchul: Or reserve reference rel to  semantically meaningful reference (e.g., sharing, execution boudnary) -- slightly broader meaning -- rather than oh, i need to chop it because of protobuf runtime limitation. Yes, it is arbitrary. 😄

benbellick: Thanks! I personally still don’t see that much difference between `detached_rel_ordinal` and `ReferenceRel` in practice.

Yes, the intended semantics are slightly different, but does that difference have any observable effect? A `ReferenceRel` referenced once seems equivalent to a detached relation, while one referenced multiple times represents sharing.

I would prefer to keep changes to the spec small unless we have a strong reason to do otherwise. I worry that the semantic separation between these two mechanisms is not worth the confusion it may cause when someone wonders, “Why should I use one over the other?”

Also, using `ReferenceRel` has the practical advantage that it is already supported by the libraries.

The expression case clearly requires something new because there is no equivalent reference mechanism today. Is there a reason why we can’t add detached expressions while continuing to use `ReferenceRel` for relations here? Or could we at least add detached expressions first, then see whether a separate relation mechanism is also needed in practice?

yongchul: I'll change this PR to focus on the expression as clearly we need more discussion about the relational side and we can continue to discuss the relation side in other PR/issue.

yongchul: Repurposed the PR to cover only expression.

site/docs/serialization/binary_serialization.md
```diff
@@ -13,6 +13,33 @@ The main top-level object used to communicate a Substrait plan using protobuf is
...
 
+=== "PlanRel Message"
+
+    ```proto
+%%% proto.message.PlanRel %%%
+    ```
+
+### Bounded Expression Nesting
+
+Protobuf implementations impose recursion limits while parsing nested messages. A producer can limit the physical message depth of an expression tree by moving expression subtrees into the `detached_expressions` repeated field of their containing `PlanRel` and replacing each moved subtree with its ordinal. Detached expressions may themselves contain detached expression ordinals, allowing a producer to split an arbitrarily deep expression tree into bounded-depth chunks. Relation trees can be split using [`ReferenceRel`](../relations/logical_relations.md#reference-operation).
```
benbellick:
```suggestion
[Protobuf implementations impose recursion limits while parsing nested messages](https://protobuf.dev/programming-guides/proto-limits/). A producer can limit the physical message depth of an expression tree by moving expression subtrees into the `detached_expressions` repeated field of their containing `PlanRel` and replacing each moved subtree with its ordinal. Detached expressions may themselves contain detached expression ordinals, allowing a producer to split an arbitrarily deep expression tree into bounded-depth chunks. Relation trees can be split using [`ReferenceRel`](../relations/logical_relations.md#reference-operation).
```

site/docs/serialization/binary_serialization.md
```diff
@@ -13,6 +13,33 @@ The main top-level object used to communicate a Substrait plan using protobuf is
...
+
+For reliable communication across protobuf implementations, producers of plans with unknown or unbounded expression depth should detach every recursive expression child. This yields an effectively flat expression representation connected entirely by ordinals.
```
benbellick: This recommendation feels too strong. I personally think that the "happy" path is to avoid detaching expressions unless you must due to recursion depth. Otherwise, the plans becoming extremely complex. In this case, it may be better to give no recommendation at all.

site/docs/serialization/binary_serialization.md
```diff
@@ -13,6 +13,33 @@ The main top-level object used to communicate a Substrait plan using protobuf is
...
+
+Detached expression ordinals are encoding details and do not change the meaning of a plan. Substituting every detached expression ordinal with its target must produce the logical expression trees represented by the `PlanRel`. They are not a mechanism for sharing computation.
+
+The following rules apply independently to each `PlanRel`:
+
+1. A detached expression ordinal is the zero-based index of an entry in `detached_expressions` on the same `PlanRel`.
+2. Every detached expression must be reachable from the `rel` or `root` field and referenced exactly once.
+3. Detached expression ordinals must not form a cycle.
+4. Reordering detached expressions requires updating their ordinals and does not otherwise change plan semantics.
```
benbellick:
```suggestion
```
nit: I don't see the value of explaining how to reorder. It just seems like a way of saying something that is already implicit, which is just that all plans need to be "correct".

site/docs/serialization/binary_serialization.md
benbellick:
```suggestion
```

This recommendation feels too strong. I personally think that the "happy" path is to avoid detaching expressions unless you must due to recursion depth. Otherwise, the plans becoming extremely complex. In this case, it may be better to give no recommendation at all.

site/docs/serialization/binary_serialization.md
```diff
@@ -13,6 +13,33 @@ The main top-level object used to communicate a Substrait plan using protobuf is
...
+
+Producers may inline expression subtrees when they can guarantee that the resulting protobuf nesting remains within every target implementation's limit. Logical expression depth is not a portable measure of protobuf message depth.
```
benbellick:
```suggestion
```
Also unsure if this is worth saying. You don't actually know the nesting limit of a target implementation since those are configurable in some of the languages.

benbellick: And if we leave that sentence in, it sounds like libraries are obligated to be able to do depth-analysis, which IIRC we decided against during the last community sync. I think it is okay to leave that part a bit underspecified and see if issues come up in practice :)

proto/substrait/plan.proto
```diff
@@ -20,6 +20,11 @@ message PlanRel {
     // The root of a relation tree
     RelRoot root = 2;
   }
+
+  // Expression subtrees detached to limit protobuf message nesting. Entries
+  // are local to this PlanRel and referenced by their zero-based ordinal. Every
+  // entry must be reachable from rel_type and referenced exactly once.
+  repeated Expression detached_expressions = 3;
```
benbellick, line 27: nit (and really just my opinion): I would rather put this lower after the `Plan` message definition. Putting it earlier makes it a bit less readable, as seems to "prioritize" `detached_expression` over `Plan`

benbellick (approved):

site/docs/serialization/binary_serialization.md
yongchul: we can still give some guidance to work with 100 nesting limit. if library does not want to deal with depth limit altogether, it can go complete flat model using reference rel. Starting with underspecification is fine with me.

site/docs/serialization/binary_serialization.md
yongchul: It will matter one day if Substrait is truly operating with a system that you have no control over. The most pessimistic but surely works everywhere is the full flat plan, which reference rel today will be bad.

The plan is not complex. it's just how they are laid out over the wire and memory. Also, knowing the other party's recursion depth limit value may not always possible unless you really do figure out by trial and error or do some out of band negotiation.

proto/substrait/plan.proto
yongchul: I don't understand this comment. can you be more specific? Are you suggesting moving this to Plan message?

benbellick: I mean literally moving the codeblock below message Plan. So rather than this:
```protobuf
  repeated Expression detached_expressions = 3;
  ...
  message Plan {
  ...
  };
```
you do:
```protobuf
  message Plan {
  ...
  };
  ...
  repeated Expression detached_expressions = 3;
```

though again this is just a nit, not a big deal

yongchul: I am confused. Both Plan and PlanRel are top-level messages. the detached expressions is a field of PlanRel. Plan has repeated field of PlanRel. What am I missing here?

yongchul: Or you mean swap the order of PlanRel and Plan messages?

benbellick: 🤦 you are right, I misread the proto files. Please ignore this one!

PR #1221, opened 2026-09-15
feat(proto): support plan-scoped functions

### LINKED ISSUE #1220: Support a plan-level function definition in a plan
# Motivation

* Not only Rel's but Expression can be shared a "real" common subexpression. However, in Substrait model, sharing expression is not easy due to the field references, which relies on "offset" of binding schema rather than the "identity".

* Another case is a real function defined in terms of substrait scalar expressions (i.e., UDF).

# Proposal

Both scenario can be modeled a *function* and all the field references can be replaced as a function argument. On referenced site, we can substitute the common expression as a function invocation.

Depending on implementation, one can choose either make it true function call or inline the expression with the specific parameter bindings.

We can reuse the existing lambda infrastructure for definition and call site but the definition must reside in plan scope and in a separate namespace (i.e., separate anchor).

yongchul:
Allow plans to define typed expression functions once and invoke them by anchor from multiple call sites. This supports shared expression subtrees and functions implemented directly in Substrait expressions without requiring an external extension implementation.

Plan-scoped functions reuse the existing lambda parameter and body representation while using an anchor namespace separate from extension functions. Definitions are closed over their parameters, and plan-function references must form an acyclic graph so consumers may either execute calls directly or inline their definitions.

The dialect schema and documentation describe the new invocation form, scoping rules, typing requirements, and namespace behavior. A full-plan textproto example demonstrates reuse of one definition at multiple call sites.

Closes #1220

- - -
This change is [Reviewable](https://reviewable.io/reviews/substrait-io/substrait/1221)

## Summary by CodeRabbit

- **New Features**
  - Added reusable, plan-defined functions expressed as lambda expressions.
  - Added support for invoking plan-defined functions with type-checked arguments.
  - Added dialect schema support for plan-function invocation expressions.

- **Documentation**
  - Documented function definitions, invocation rules, serialization, and usage examples.
  - Updated terminology from “plan-scoped functions” to “plan-defined functions.”

- **Examples & Tests**
  - Added an example demonstrating reuse of a plan-defined identity function.
  - Added validation coverage for plan-defined function examples.


proto/substrait/plan.proto
```diff
@@ -65,6 +65,26 @@ message Plan {
 
   // Execution behavior settings for this plan, required.
   ExecutionBehavior execution_behavior = 10;
+
+  // An optional list of plan-defined functions.
+  repeated PlanFunctionDefinition plan_functions = 11;
+}
+
+// A function defined within a plan using a lambda expression.
+message PlanFunctionDefinition {
```
benbellick, line 74: Just an idea, i'm not sure it is a good one yet 😅 

But did you consider instead making these plan-level functions instead an additional enum case of the `SimpleExtensionDeclaration`?
https://github.com/substrait-io/substrait/blob/272d8dc153ac980094d54b63660451d99eab9f89/proto/substrait/extensions/extensions.proto#L26-L31

It might not work because we would need the ability to use the contents of `algebra.proto` there, but something worth considering. The way I am thinking about it, there is really nothing special about `PlanFunction` vs. `ExtensionFunction` except for the fact that extension functions are defined in the YAML. What do you think?

yongchul: I like the idea!

yongchul: but copilot does not like it because of mixing URN based definitions and inline definition, which I kinda agree after seeing that as well as protobuf import cycle dependency problem.

Ah... now that I see it, it is not also declaration but definition (oh well, declare + definition).

So I'd stay the same way as type_aliases. Maybe we need another wrapper message to host all these definitions (e.g., type_aliases, inlined function definitions, shared schema definition)

Issue #1216, opened 2026-09-10
The Decimal Division example derives a precision `divide` does not, and has no upper bound


alexandrefimov:
The worked example for Decimal Division under [Example Type Expressions](https://substrait.io/expressions/scalar_functions/#example-type-expressions) is SQL Server's rule with the bound dropped, and it computes a precision that `divide` does not.

`site/docs/expressions/scalar_functions.md:215`, at `8ca7db0`:

> `divide(Decimal<P1,S1>, Decimal<P2,S2>) => Decimal<P1 -S1 + S2 + MAX(6, S1 + P2 + 1), MAX(6, S1 + P2 + 1)>`

That is the [T-SQL rule](https://learn.microsoft.com/en-us/sql/t-sql/data-types/precision-scale-and-length-transact-sql?view=sql-server-ver17) for `e1 / e2`, precision `p1 - s1 + s2 + max(6, s1 + p2 + 1)` and scale `max(6, s1 + p2 + 1)`, minus the sentence that follows it there: "The result precision and scale have an absolute maximum of 38."

The `divide` return expression in `extensions/functions_arithmetic_decimal.yaml` writes the same scale and a different precision, `P1 - S1 + P2 + init_scale`, with `P2` where the example has `S2`, and it does cap at 38, taking the surplus out of the scale.

| operands | `divide` | the example |
| --- | --- | --- |
| `dec(10,2) / dec(5,1)` | `dec(21,8)` | `dec(17,8)` |
| `dec(38,10) / dec(38,10)` | `dec(38,6)` | `dec(87,49)` |

The missing bound is wrong whatever anyone decides about `S2` against `P2`: `type_classes.md` gives `DECIMAL<P, S>` `P <= 38` and `S <= P`. Enumerating every valid pair of decimal operand types, 554816 of 606841 put the example's result outside those bounds.

Both precisions are in use. Deriving the first row through the implementations I have to hand, substrait-java, substrait-python, substrait-validator and substrait-go return `dec(21,8)`, and Spark, which implements the T-SQL rule, returns `dec(17,8)`.

Happy to send the documentation fix if the example is meant to track the extension. Whether `divide` itself should follow T-SQL is the larger question underneath, and that one is a released return type.

Issue #996, opened 2026-03-06
Improve documentation and consistency of options across extension files


kadinrabo:
Options like `overflow`, `on_domain_error`, and `on_division_by_zero` appear across
multiple extension files but lack description fields and are inconsistently applied.

**Problems:**
- No description fields on options explaining their behavior (e.g. what SILENT, SATURATE,
  ERROR mean for overflow). See functions_list.yaml for a good example of documented options.
- `functions_arithmetic_decimal.yaml` divide is missing `on_division_by_zero` and
  `on_domain_error` options that exist on integer divide in `functions_arithmetic.yaml`.
- As more extension files are added (e.g. unsigned_integers.yaml), these inconsistencies
  will compound.

**Proposed work:**
- Add description fields to all options across all extension files in a single pass
- Audit option sets on equivalent functions across files (e.g. divide) and reconcile
- Consider a template or generation approach to prevent future drift

Context:
- https://github.com/substrait-io/substrait/pull/953#discussion_r2866226958
- https://github.com/substrait-io/substrait/pull/953#discussion_r2841832901

Issue #993, opened 2026-03-04
Define common arithmetic options


vbarua:
In our definitions of arithmetic functions, we include a number of "common" options like:

**overflow**
https://github.com/substrait-io/substrait/blob/e4ce3f8710d91cb95cce010c6b59eb1d618bb2a6/extensions/functions_arithmetic.yaml#L14-L16

**rounding**
https://github.com/substrait-io/substrait/blob/e4ce3f8710d91cb95cce010c6b59eb1d618bb2a6/extensions/functions_arithmetic.yaml#L48-L50

The behaviour of these options is captured somewhat in tests like [add.test](https://github.com/substrait-io/substrait/blob/main/tests/cases/arithmetic/add.test), [substract.test](https://github.com/substrait-io/substrait/blob/main/tests/cases/arithmetic/subtract.test), but having to go look at the test vectors to understand what the options do is less than ideal. They are _somewhat_ based on common concepts in numeric processing, but for spec purposes we should also define them explicitly in text.

alexandrefimov:
Nothing saturates. At an integer type boundary, where clamping would be likeliest, Spark in ANSI mode and Trino raise; Hive and Spark with ANSI off wrap, `127 + 1` giving `-128`; MySQL and ClickHouse widen the result type instead, and raise or wrap where they cannot. Decimals give the same three answers, never a clamp. (Spark 4.0.1, Hive 4.0.1, Trino 483, MySQL 8.4.11, ClickHouse 26.8.6.5.)

So `ERROR` and `SILENT` can each be defined against something an engine does, and `SATURATE` cannot: defining it means finding an engine that clamps, or writing it down as a permission nobody exercises.

PR #1223, opened 2026-09-17
docs(extensions): say where the decimal precision and scale rules come from

alexandrefimov:
`functions_arithmetic_decimal` derives a precision and scale that no other extension file does, and the file says nothing about where the rule comes from. Asked for in review of #1213.

The rule is Arrow Gandiva's [`DecimalTypeUtil`](https://github.com/apache/arrow/blob/apache-arrow-7.0.0/java/gandiva/src/main/java/org/apache/arrow/gandiva/evaluator/DecimalTypeUtil.java#L38-L90), which [#151](https://github.com/substrait-io/substrait/pull/151) names as the reference these return expressions were fixed against; Hive and Spark derive the same types. Transcribing the three beside the YAML and comparing them over every operand type pair `type_classes.md` allows (779 × 779 = 606841), `add`, `subtract` and `multiply` agree with all three on every one.

`divide` is left out: its precision uses `P2` where Gandiva, Hive, Spark and Trino use `S2`, so the same sentence would not be true of it. [#1216](https://github.com/substrait-io/substrait/issues/1216) is open on that.

`modulus` is left out for a different reason: its unbounded precision never exceeds 38, so the reduction the sentence describes cannot arise there.

Two things worth a maintainer's call:

- The extension schema has no file-level `description`, so a shared paragraph has to be repeated per function, as here. The alternative is one paragraph in the prose docs — no repetition, but a reader on the generated function page never sees it.
- Engines differ on the reduction above 38: Hive and Spark take the surplus out of the scale, Trino and DataFusion keep the scale and overflow instead. Saying so here would be the spec recording its own engine-specificity. Left out of this change; happy to add it.

- - -
This change is [Reviewable](https://reviewable.io/reviews/substrait-io/substrait/1223)

## Summary by CodeRabbit

- **Documentation**
  - Clarified decimal arithmetic behavior for addition, subtraction, and multiplication.
  - Documented how result precision and scale are determined, including handling for results exceeding 38 digits.
  - Specified that oversized results reduce scale while retaining at least six fractional digits.
  - Clarified that scales of six or fewer remain unchanged.


extensions/functions_arithmetic_decimal.yaml
```diff
@@ -4,7 +4,10 @@ urn: extension:io.substrait:functions_arithmetic_decimal
 scalar_functions:
   -
     name: "add"
-    description: "Add two decimal values."
+    description: >
+      Add two decimal values. The precision and scale of the result follow Arrow Gandiva's DecimalTypeUtil, the same
+      rules Hive and Spark use: where an exact result would need more than 38 digits of precision, the result takes
+      precision 38 and the surplus comes out of the scale, down to a minimum of six fractional digits.
```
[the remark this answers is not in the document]
alexandrefimov: Fixed in 8bb6851: the three descriptions now say that a scale of six or less is left as it is.

extensions/functions_arithmetic_decimal.yaml
```diff
@@ -4,7 +4,11 @@ urn: extension:io.substrait:functions_arithmetic_decimal
 scalar_functions:
   -
     name: "add"
-    description: "Add two decimal values."
+    description: >
+      Add two decimal values. The precision and scale of the result follow Arrow Gandiva's DecimalTypeUtil, the same
```
yongchul: I generally do not like referring specific implementations in the spec. Also, this concretely defines the edge case behaviors which is previously underspecified. To that degree, this is a breaking change.

We will probably need to define the overflow or underflow behavior called out in the description as  runtime options rather than prescribing it in the description. Also, having a compatibility matrix in the spec to follow up would be nice addition.

What do you think?

extensions/functions_arithmetic_decimal.yaml
```diff
@@ -45,6 +54,11 @@ scalar_functions:
           DECIMAL<prec, scale>
   -
     name: "multiply"
+    description: >
+      Multiply two decimal values. The precision and scale of the result follow Arrow Gandiva's DecimalTypeUtil, the
```
yongchul: Would you open to contribute file-level description? 😄

extensions/functions_arithmetic_decimal.yaml
alexandrefimov: Agreed, names removed. It isn't breaking, though: the `return` expression already fixes this type, the text only says it in words. Value overflow is the `overflow` option (#993, #996), rounding is #1224. A compatibility matrix would be good as a separate change.

extensions/functions_arithmetic_decimal.yaml
alexandrefimov: Sure 🙂 I'll send a schema PR and move this paragraph there.

extensions/functions_arithmetic_decimal.yaml
benbellick: That may have come from some of the conversation here: https://github.com/substrait-io/substrait/pull/1213

I agree with @yongchul here. The point I was really trying to make is that there is no real value of having functions which are specific to one engine in this repo. We really want to codify semantics that are in some way _generic_ across many engines. 

That being said, I am fairly sure we are using these functions internally at work. So it is not completely safe to remove them 😅 


I think the doc improvements in this PR are good though.

PR #1228, opened 2026-09-21
fix(grammar)!: bind ! tighter than AND/OR and right-associate ? :

### LINKED ISSUE #721: Define order of operations in derivation expressions
This is a continuation of a discussion from slack. Extensions function return type expressions allow expressing types with math operators as defined in the [spec](https://substrait.io/expressions/scalar_functions/), but the following points aren't clear:
- Are parenthesis allowed? - spec doesn't say one way or the other but antlr [grammar](https://github.com/substrait-io/substrait-java/blob/main/core/src/main/antlr/SubstraitType.g4) in substrait-java seems to assume they are.
- Should evaluation follow the usual order of operations (pemdas)? - once again, spec is unclear, while antlr grammar doesn't follow pemdas, or to be exact it treats parenthesis with the highest precedence, but doesn't differentiate between multiplication/division and addition/subtraction.

I think we should both allow parenthesis and follow pemdas as that is what's most intuitive in the context. One way or the other, this should be made clear in the spec.

nielspardon:
## Summary

The `expr` rule orders its alternatives by precedence, but `#NotExpr` and `#Ternary` sit below `#And` and `#Or`, which gives each of them the wrong binding.

ANTLR gives a prefix alternative's operand the minimum precedence of the alternative's own position, so `!` near the bottom of the rule swallows the whole boolean expression rather than binding to its operand:

```
!a AND b          parses as   !(a AND b)        expected  (!a) AND b
a AND !b OR c     parses as   a AND !(b OR c)   expected  (a AND !b) OR c
```

`#Ternary` is left-recursive with no associativity annotation, and ANTLR defaults those to left-associative. That is right for the binary operators but not for a conditional, so a chained one groups to the left:

```
c1 ? 1 : c2 ? 2 : 3   parses as   (c1 ? 1 : c2) ? 2 : 3   expected  c1 ? 1 : (c2 ? 2 : 3)
```

The left-associated form is usually a type error rather than merely surprising, because the `else` of the outer conditional becomes the condition of the next one.

Underneath both: the prose never stated precedence at all, so the grammar has been its only statement and a reader of the docs alone could not derive it. That is what #721 asks for, so the documentation is part of this PR rather than a follow-up.

## Fix

Move `#NotExpr` above `#MulDiv`, making `!` the tightest-binding operator, and mark `#Ternary` `<assoc=right>`. Precedence, tightest to loosest: `!` > `* /` > `+ -` > comparison > equality > `AND` > `OR` > `if ... then ... else ...` > `? :`. Both conditional forms are right-associative.

This is the same class of change as #1107, which gave the binary operators their cascade, and is marked breaking on the same grounds.

## Changes

- `grammar/SubstraitType.g4`: move the `#NotExpr` alternative, annotate `#Ternary`.
- Regenerated `tests/type/antlr_parser/SubstraitTypeParser.py` (`pixi run generate-antlr`). Only the parser changes — the alternative labels are unchanged, so the listener and visitor are untouched, and `tests/coverage/antlr_parser/` is generated from a different grammar.
- Precedence regression tests asserting parse-tree structure, and `render_precedence` extended to render the unary and conditional forms. It previously matched only alternatives carrying an `op` label, so `!` and `? :` rendered as flat source text and could not be asserted on.
- `site/docs/expressions/scalar_functions.md`: a precedence and associativity table, a statement that parentheses group, and a correction to the operator list, which named three spellings the grammar does not accept and omitted three it does — the grammar defines `And : 'AND'`, `Or : 'OR'` and `Eq : '='`, not `&&`, `||` and `==`, and it also accepts `<=`, `>=` and `!=`.

## Compatibility note

No checked-in extension YAML uses `!` or a chained `? :` — the shipped return programs use only an unchained `? :` — so nothing in the standard catalog moves. Downstream tools generate parsers directly from this grammar, so they pick the change up on the next release. Note substrait-go still pins the v0.87.0 grammar (`grammar/generate.go`), so it has flat binary precedence today and will take both changes together when substrait-io/substrait-go#268 lands.

🤖 Generated with AI

BREAKING CHANGE: Fixes the precedence of `!` and the associativity of the `? :` conditional when parsing return type expressions from extension YAMLs. This can be breaking for user defined extension YAMLs which may have relied on `!` binding looser than `AND` and `OR`, or on a chained `? :` grouping to the left; both now follow conventional precedence and associativity.

Closes #721

- - -
This change is [Reviewable](https://reviewable.io/reviews/substrait-io/substrait/1228)


tests/type/test_type_grammar.py
```diff
@@ -182,6 +195,36 @@ def test_operator_precedence():
...
+    """`!` is the highest-precedence operator, so it takes only its operand."""
+    cases = [
+        ("!a and b", "((! a) and b)"),
+        ("!a or b", "((! a) or b)"),
+        ("a and !b or c", "((a and (! b)) or c)"),
+        ("!a < b", "((! a) < b)"),
+        ("!a = b", "((! a) = b)"),
+        ("!(a and b)", "(! (a and b))"),
+        ("!a ? 1 : 2", "((! a) ? 1 : 2)"),
+    ]
```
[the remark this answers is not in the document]
nielspardon: Added in 8bc6e7e: `!a * b`, `!a / b`, `!a + b`, `!a - b` and `a + !b * c`. You are right that the existing cases did not pin the position — they all pass with `!` merely above `#Comparison`, whereas the change puts it above `#MulDiv`.

benbellick:
This seems reasonable but I don't understand why this PR and #1229 are separated. If this is changing the grammar then we should update the docs at the same time.

tests/type/test_type_grammar.py
```diff
@@ -182,6 +195,41 @@ def test_operator_precedence():
...
+
+    for expression, expected in cases:
+        tree = parse_type_expression(expression)
+        assert render_precedence(tree.expr()) == expected, expression
+
+
+def test_conditional_is_right_associative():
+    """A chained `? :` nests to the right, so each `:` pairs with the nearest `?`."""
+    cases = [
+        ("c1 ? 1 : c2 ? 2 : 3", "(c1 ? 1 : (c2 ? 2 : 3))"),
```
yongchul, line 247: add a case to have ternary in the then

nielspardon:
Good call — combined. The docs commit from #1229 is now on this branch (862b6a7) and #1229 is closed; this PR closes #721.

tests/type/test_type_grammar.py
nielspardon: Added in d79f821: `c1 ? c2 ? 1 : 2 : 3`, `c1 ? c2 ? 1 : 2 : c3 ? 3 : 4` and `a > b ? c > d ? 1 : 2 : 3`. The `then` case is worth having for a different reason than the `else` ones — the inner conditional is consumed by that branch rather than left for the outer `:`, so it nests correctly even without the associativity fix, and the third case pins both sides at once.

site/docs/expressions/scalar_functions.md
```diff
@@ -180,11 +180,36 @@ These types are evaluated using a small set of operations to support common scen
...
+| 3 | `*` `/` | left |
+| 4 | `+` `-` | left |
+| 5 | `<` `>` `<=` `>=` | left |
+| 6 | `=` `!=` | left |
+| 7 | `AND` | left |
+| 8 | `OR` | left |
+| 9 | `if ... then ... else ...` | right |
+| 10 (loosest) | `... ? ... : ...` | right |
+
+So `P - S + 1` is `(P - S) + 1`, `S1 + P2 * 2` is `S1 + (P2 * 2)`, and `!a AND b` is `(!a) AND b`. Both conditional forms nest to the right, so `c1 ? 1 : c2 ? 2 : 3` is `c1 ? 1 : (c2 ? 2 : 3)`, and `if ... then ... else ...` binds tighter than `? :`.
```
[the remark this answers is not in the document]
nielspardon: You are right, and it is broader than the `else` branch. Documented in b6165bc, with tests.

Measured against the regenerated parser: the condition and the *middle* branch of both forms are delimited on both sides — by `if`/`then`, by `then`/`else`, and by `?`/`:` — so each takes a complete expression whatever its precedence, not just an unrestricted `expr(0)` in the `if` case:

```
if a ? b : c then d else e   ->  if (a ? b : c) then d else e
if a then b ? c : d else e   ->  if a then (b ? c : d) else e
a ? if b then c else d : e   ->  a ? (if b then c else d) : e
```

Only the trailing branch is governed by the table, which is what makes both forms right-associative (`if a then b else c AND d` is `if a then b else (c AND d)`, while `if a then b else c ? d : e` is `(if a then b else c) ? d : e`). And because `if` begins with a keyword rather than an operand, a whole one can be a tighter operator's operand: `x AND if a then b else c` is `x AND (if a then b else c)` — which a flat table cannot express either.

I documented the rule rather than changing the grammar: altering the branch precedences would change parse behaviour further, beyond what this PR is fixing.

yongchul (approved):

site/docs/expressions/scalar_functions.md
```diff
@@ -180,11 +180,57 @@ These types are evaluated using a small set of operations to support common scen
...
+| 5 | `<` `>` `<=` `>=` | left |
+| 6 | `=` `!=` | left |
+| 7 | `AND` | left |
+| 8 | `OR` | left |
+| 9 | `if ... then ... else ...` | right (trailing branch only, see below) |
+| 10 (loosest) | `... ? ... : ...` | right (trailing branch only, see below) |
+
+So `P - S + 1` is `(P - S) + 1`, `S1 + P2 * 2` is `S1 + (P2 * 2)`, and `!a AND b` is `(!a) AND b`. Both conditional forms nest to the right, so `c1 ? 1 : c2 ? 2 : 3` is `c1 ? 1 : (c2 ? 2 : 3)`, and `if ... then ... else ...` binds tighter than `? :`.
+
+The two conditional forms need one qualification, because only their trailing branch is governed by the table. Their condition and middle branch are delimited on both sides -- by `if` and `then`, by `then` and `else`, and by `?` and `:` -- so each accepts a complete expression whatever its precedence:
```
benbellick:
```suggestion
For `if ... then ... else ...`, the `then` keyword ends the condition and the `else` keyword ends the `then` branch. Those two parts may contain any expression without extra parentheses:
```

site/docs/expressions/scalar_functions.md
```diff
@@ -180,11 +180,57 @@ These types are evaluated using a small set of operations to support common scen
...
+
+```
+if a ? b : c then d else e   is   if (a ? b : c) then d else e
+if a then b ? c : d else e   is   if a then (b ? c : d) else e
+a ? b ? c : d : e            is   a ? (b ? c : d) : e
```
benbellick:
```suggestion
```

site/docs/expressions/scalar_functions.md
```diff
@@ -180,11 +180,57 @@ These types are evaluated using a small set of operations to support common scen
...
+| 1 (tightest) | grouping `( )`, function call `f(...)` | n/a |
+| 2 | `!` | right (unary prefix) |
+| 3 | `*` `/` | left |
+| 4 | `+` `-` | left |
+| 5 | `<` `>` `<=` `>=` | left |
+| 6 | `=` `!=` | left |
+| 7 | `AND` | left |
+| 8 | `OR` | left |
+| 9 | `if ... then ... else ...` | right (trailing branch only, see below) |
+| 10 (loosest) | `... ? ... : ...` | right (trailing branch only, see below) |
```
benbellick:
```suggestion
| 9 | `if ... then ... else ...` | right |
| 10 (loosest) | `... ? ... : ...` | right |

See [`grammar/SubstraitType.g4`](https://github.com/substrait-io/substrait/blob/main/grammar/SubstraitType.g4) for the formal grammar.
```
nit: I would simplify the below with an example section

site/docs/expressions/scalar_functions.md
```diff
@@ -180,11 +180,57 @@ These types are evaluated using a small set of operations to support common scen
...
+if a then b else c ? d : e   is   (if a then b else c) ? d : e
+```
+
+Finally, because `if ... then ... else ...` begins with a keyword rather than with an operand, a complete one may appear wherever a value may, including as the operand of a tighter operator:
+
+```
+x AND if a then b else c     is   x AND (if a then b else c)
+```
+
+The normative statement of this grammar is [`grammar/SubstraitType.g4`](https://github.com/substrait-io/substrait/blob/main/grammar/SubstraitType.g4); the table above records the precedence that grammar derives.
```
benbellick:
```suggestion
##### Examples

| Expression | Binds as |
| ---------- | -------- |
| `P - S + 1` | `(P - S) + 1` |
| `S1 + P2 * 2` | `S1 + (P2 * 2)` |
| `!a AND b` | `(!a) AND b` |
| `c1 ? 1 : c2 ? 2 : 3` | `c1 ? 1 : (c2 ? 2 : 3)` |
| `if a then b else if c then d else e` | `if a then b else (if c then d else e)` |
| `if a ? b : c then d else e` | `if (a ? b : c) then d else e` |
| `if a then b ? c : d else e` | `if a then (b ? c : d) else e` |
| `a ? b ? c : d : e` | `a ? (b ? c : d) : e` |
| `if a then b else c ? d : e` | `(if a then b else c) ? d : e` |
```
IMO, the grammar is sufficient for defining these things, examples are just nice to answer some obvious questions.

benbellick (approved):

site/docs/expressions/scalar_functions.md
nielspardon: Applied in 8e3f6d1 — both rows back to plain `right`, with the grammar link under the table.

site/docs/expressions/scalar_functions.md
nielspardon: Applied in 8e3f6d1, with two rows added to your set:

```
| `if a then b else c AND d` | `if a then b else (c AND d)` |
| `x AND if a then b else c` | `x AND (if a then b else c)` |
```

The second is the one the table alone gets wrong: `if` is row 9 and `AND` is row 7, so a reader would expect `(x AND if a then b else c)` to be impossible, but because `if` starts with a keyword rather than an operand a whole one can sit there. That was CodeRabbit's point above, and with the prose gone the examples are what carry it. I verified all eleven rows by parsing both the expression and the claimed binding and comparing the trees, and `test_conditional_branches_delimited_on_both_sides_take_a_full_expression` pins them.

